PR #2, opened 2013-01-06
Generic mount framework and mount_to_var mountspec parser function

devurandom:
These commits add functions to create fstab-like files that will be parsed in the pre-mount (and pre-pivot) phases.

I will follow this up (once it is accepted) with commits that actually make use of it to allow mounting of NFS paths and tmpfs to arbitrary locations. (My feature/nfsmounts and feature/tmpfs branches.)


vtolstov:
Does it possible to mount overlayfs on squashfs with tmpfs ?=)

devurandom:
We dismantled the system this was intended for a long time ago.

PR #3, opened 2013-01-06
Parse network related variables in cmdline options and add hostname fallback function

devurandom:
These commits add functions to parse network related variables from Kernel cmdline options (e.g. `var=%{hostname}`) and a bare-bones implementation for hostname, in case the executable is not available.

I will follow this up (once it is accepted) with commits that actually make use of it to allow mounting of NFS paths with variables in them. (My feature/nfsmounts branch.)


devurandom:
We dismantled the system this was intended for a long time ago.

PR #4, opened 2013-01-06
[95aufs] Support root=aufs:...

devurandom:
This commit adds support for having the rootfs provided by aufs.

This feature will become more useful once my nfsmounts and tmpfs branches are merged, which allow mounting of NFS paths and tmpfs to arbitrary locations. (My feature/nfsmounts and feature/tmpfs branches.)


devurandom:
The system we were using this on was dismantled a while ago.

PR #14, opened 2014-04-03
Break at switch_root only for bare rd.break

### LINKED ISSUE #13: rd.break=some_phase also causes a break before switch_root
If I boot with `rd.break=pre-mount`, for example, it breaks before mounting, and then _also_ breaks again before switching root. I'm pretty sure this is just because `init.sh` uses a plain `getarg rd.break` (with no _=value_) for the last breakpoint before switching root, and `getarg` returns `true` for that regardless of what value was on the command line.

I was going to submit a patch but figured I'd ask first whether that was intended behavior for any reason.

jcflack:
Previously, any rd.break=breakpoint would cause a break
at the desired breakpoint and also another one at switch_root.

To resolve #13.

### DIFF
--- a/modules.d/99base/init.sh
+++ b/modules.d/99base/init.sh
@@ -353,7 +353,9 @@ wait_for_loginit
 # remove helper symlink
 [ -h /dev/root ] && rm -f -- /dev/root
 
-getarg rd.break -d rdbreak && emergency_shell -n switch_root "Break before switch_root"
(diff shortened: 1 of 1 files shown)


jandob:
This change results in different behavior if the systemd module is included. 
E.g. with kernel parameters `rd.break rd.break=pre-pivot`
-> with systemd: break at pre-pivot and switch-root
-> without systemd: break at pre-pivot only

PR #27, opened 2014-08-19
WIP add mptcp support

vtolstov:
Sometimes kernel have multipath-tcp support, in this case user may want to enable it in boot time to speedup image loading.
This patch does not wait for getting configured interfaces (write for ipv6 slaac only), what is the best method for waiting to configure interfaces, but before anything else?

Signed-off-by: Vasiliy Tolstov <v.tolstov@selfip.ru>


vtolstov:
Please forget. mptcp does not included in present kernel versions =(

PR #35, opened 2014-12-19
dracut: Ajusting variables name for --include

ErwanAliasr1:
When reading the --include part of the script, we had the following
issues to make the code easy to read:
- src & tgt were extract for the original options
- i variable was a file or a directory from src
- s variable was the directory name in case $i was a directory

"s" sounds very close to "src" while "s" is on the "tgt" side. Very
confusing.

"s" was defined before the "if it's a directory" statement while it's
only used inside the "if".

"i" was commit from the "src" but wasn't really explicit.

Having some lines mixing "i" and "s" takes a little time to get read
properly.

This patch offer the following changes:
- "i" is renamed to "objectname" as we don't know if its a file or a
  directory

- "s" is renamed to "object_destdir" as the object name becomes a
  directory on the destdir

- "object_destdir" (former "s") is moved inside the "if" statement as it's
  only used here

- tgt is finally renamed to "target" to be more explicit. We are not all
  native english ;o)

My 2 (semantic) cents,

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1487,27 +1487,29 @@ if [[ $kernel_only != yes ]]; then
     done
 fi
 
-while pop include_src src && pop include_target tgt; do
-    if [[ $src && $tgt ]]; then
+while pop include_src src && pop include_target target; do
+    if [[ $src && $target ]]; then
         if [[ -f $src ]]; then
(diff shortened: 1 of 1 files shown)


ErwanAliasr1:
Any comments on that ?

PR #37, opened 2014-12-28
99base: Properly remove files with rd.hostonly=0

gonzzor:
Dracut will generate systemd units for additional devices that should be
brought up during boot, e.g. swap devices. These unit files are broken
symlinks with \ in the filename, e.g.
/etc/systemd/system/initrd.target.wants/dev-disk-by\x2duuid-e6a54f99\x2da4fd\x2d4931\x2da956\x2d1c642bcfee5e.device.

Both the backslash and the broken symlink causes problems for shell
scripts, [ -e "$file" ] isn't enough and read requires the additional -r
argument to not react on the \.


haraldh:
Thanks! Pushed

PR #38, opened 2015-02-03
90dm/dm-shutdown.sh: dmsetup remove_all -> dmsetup remove

l1k:
The function 99shutdown/shutdown.sh:_check_shutdown() assumes that shutdown scripts report success or failure via their return value. However, "dmsetup remove_all" always reports success, even if some of the device mappings could not be removed.

I submitted a [patch for dmsetup](https://www.redhat.com/archives/lvm-devel/2014-December/msg00001.html) but the lvm2 folks [rejected it](https://www.redhat.com/archives/lvm-devel/2014-December/msg00004.html), asserting that its behaviour is correct, that "remove_all" should only be used by developers and that the proper solution would be to invoke "dmsetup remove" on each device. This does report success or failure via the return value.

Apart from fixing that issue, this commit also adds the dmsetup option "--noudevsync". Without it, dmsetup would hang after removal of a device while trying to communicate with systemd-udevd, which is no longer running at this shutdown stage.

### DIFF
--- a/modules.d/90dm/dm-shutdown.sh
+++ b/modules.d/90dm/dm-shutdown.sh
@@ -1,11 +1,12 @@
 #!/bin/sh
 
 _do_dm_shutdown() {
-    local ret
+    local ret=0
     local final=$1
     info "Disassembling device-mapper devices"
-    dmsetup -v remove_all
-    ret=$?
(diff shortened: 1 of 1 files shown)


haraldh:
Thanks! Pushed

PR #42, opened 2015-02-18
dmsquash: Add rd.live.overlay.thin

fabiand:
This option changes the underlying mechanism for the overlay in the
dmsquash module.
Instead of a plain dm snapshot a dm thin snapshot is used. The advantage
of the thin snapshot is, that the TRIM command is recognized, which
means that at runtime, only the occupied blocks will be claimed from
memory, and freed blocks will really be freed in ram.

Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>


haraldh:
Thanks! Pushed

PR #39, opened 2015-02-05
Feature/crypt sshd

artem-sidorenko:
This is a module to allow remote unlocking of luks encrypted systems via ssh.
This module is optional and is a separate RPM package like dracut-network because of dropbear dependency (Btw dropbear is coming from EPEL on RHEL systems)

Original sources:
 - https://bugzilla.redhat.com/show_bug.cgi?id=524727#c7
 - https://github.com/mdcurtis/dracut-earlyssh
 - https://github.com/mk-fg/dracut-crypt-sshd

I spoke with all three authors ( @mk-fg , @mdcurtis and @zenczykowski ), its ok to release it under GPL2 as dracut module.

PR #43, opened 2015-03-08
crypt-ssh: a module for unlocking luks volumes via ssh

artem-sidorenko:
Review done in #39, its a rebased and squashed MR.

This module is optional and is a separate RPM package like dracut-network because of dropbear dependency (Btw dropbear is coming from EPEL on RHEL systems)

Original sources are:
 - https://github.com/mdcurtis/dracut-earlyssh
 - https://github.com/mk-fg/dracut-crypt-sshd
 - https://bugzilla.redhat.com/show_bug.cgi?id=524727#c7

Signed-off-by: Artem Sidorenko <artem@2realities.com>


tyll:
I created a similar module, where I use a small C helper to properly kill the password dialog:
https://github.com/tyll/dracut-cryptssh/blob/master/wait_for_file.c

Also the module patches the cryptroot-ask.sh file to write the required cryptsetup call to root's .profile, so the same code is called regardless of whether one uses the dracut prompt or the SSH login. IMHO this is cleaner, because it does not need an additional crypttab parser and only the proper processes are killed.

tyll:
Also IMHO the default SSH port 22 should be used by default instead of an arbitray chosen one.

artem-sidorenko:
@tyll Do you want to create a PR to https://github.com/artem-sidorenko/dracut/tree/rebased/crypt-sshd , so we can review and update this PR?

tyll:
On Mon, May 11, 2015 at 04:26:48AM -0700, Artem Sidorenko wrote:
> @tyll Do you want to create a PR to https://github.com/artem-sidorenko/dracut/tree/rebased/crypt-sshd , so we can review and update this PR?

ok, this sounds good.

thypon:
Hangs on boot on voidlinux. Plus the Makefile inside isn't triggered during compilation neither during `dracut --force` phase + the Makefile should specify CC=gcc to made possible to crosscompile modules or to use something like clang.

rbu:
Maintainers, what's needed to get this merged?

haraldh:
@rbu : I think this can live as a separate project and doesn't have to be upstream.

haraldh:
May I ask, why 50-udev-pty.rules is necessary?

mdcurtis:
@haraldh it has been a long time since I last looked at this, but I'm pretty sure this was for support for CentOS 6.4 which did not include suitable udev rules by default.

artem-sidorenko:
@haraldh It's a bit confusing and disappointing to read such message now and not earlier where I asked you if you are open for such PR.

Anyway, I'll go this way and create a dedicated module when I have a bit more time

haraldh:
Sorry, if I disappointed you. But because of the sheer complexity of the module and setting it up, I don't have the resources to support it upstream.

haraldh:
Also, nothing is lost in functionality, if it is a managed as a standalone module.

artem-sidorenko:
@haraldh Yes, that's right. I use it this way already for months, thx!

rbu:
I have created a project to maintain a standalone version of this at https://github.com/dracut-crypt-ssh/dracut-crypt-ssh

PR #52, opened 2015-03-26
Take into account lib64 dirs when detecting version, modules and params

haraldh:


haraldh:
obsoleted by pull #55

PR #69, opened 2015-06-08
Require all interfaces to be up before using iscsistart -b

rosslagerwall:
If multiple targets are specified in the ibft, iscsistart will log into
all of them, possibly using multiple interfaces. Since iscsistart is run
indirectly from ifup, require that all interfaces are up before actually
logging into the targets.


haraldh:
hmm, when you have a raid, you might want to start degraded if one link/connection fails

PR #72, opened 2015-06-15
crypt: don't use systemd for crypto handling

jsynacek:
There already is the code necessary to handle an encrypted root
device. The "rd.luks.key" option is not handled the same way in systemd
as it is in dracut, so some setups (like having the keyfile on a
separate device) don't work.


haraldh:
well, the goal should be to remove all dracut shell code and move functionality over to systemd, not the other way round :-)

jsynacek:
I was told once that dracut is the place to fix this: http://lists.freedesktop.org/archives/systemd-devel/2015-February/028654.html. IMHO it's also much easier to do in dracut.

haraldh:
We should create systemd generators, use unit files move functionality from dracut shell code to systemd upstream wherever possible.

PR #75, opened 2015-06-25
dmsquash-live: Add wc

fabiand:
Commit cfa365a32d introduced a dependency on wc, but wc was not added.
This patch takes care to install wc.

Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>

### DIFF
--- a/modules.d/90dmsquash-live/module-setup.sh
+++ b/modules.d/90dmsquash-live/module-setup.sh
@@ -22,7 +22,7 @@ installkernel() {
 
 # called by dracut
 install() {
-    inst_multiple umount dmsetup blkid dd losetup grep blockdev find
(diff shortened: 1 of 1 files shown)


fabiand:
$ man lvchange
```
       -k, --setactivationskip {y|n}
              Controls   whether  Logical  Volumes are persistently flagged to be skipped during activation. By default, thin snapshot volumes are flagged for activation skip.  To activate such volumes, an
              extra -K/--ignoreactivationskip option must be used.  The flag is not applied during deactivation. To see whether the flag is attached, use lvs command where the state of the flag is reported
              within lv_attr bits.

       -K, --ignoreactivationskip
              Ignore the flag to skip Logical Volumes during activation.
```

fabiand:
See also https://github.com/haraldh/dracut/pull/76

PR #79, opened 2015-07-07
skipcpio.c: Linux only accepts 070701 cpio magic, this also fixes a c…

haraldh:
…omparision of signed char to 0xc7 (type limits)


haraldh:
doesn't compile

PR #80, opened 2015-07-13
crypt-gpg: For GnuPG >= 2.1 support gpg keyfile decryption via a OpenPGP smartcard in a CCID reader

Calrama:
This PR implements support for using an OpenPGP smartcard on a CCID smartcard reader to decrypt a gpg keyfile instead of using a password if gpg >= 2.1 with scdaemon and direct CCID access is available; it can be disabled at boot time by setting the kernel option "rd.luks.smartcard=0". Tested smartcards are the YubiKey NEO and Yubikey NEO-N (both have integrated CCID reader via usb).

Steps to try out this feature on a setup already working with a password-based gpg keyfile:
1. Make a backup of your current initramfs and current gpg keyfile
2. Re-encrypt the gpg keyfile for yourself as the recipient (optionally also using a password, so the resulting keyfile can be decrypted either with password or with the recipient's private decryption key)
3. Verify that you can decrypt the re-encrypted gpg keyfile with your smartcard
4. Copy your public key to /etc/dracut.conf.d/crypt-public-key.gpg
5. Rebuild your initramfs with dracut as usual
6. Reboot with your smartcard connected.
7. If everything went well, you should be greeted by a "PIN (OpenPGP card xyz)" prompt instead of the normal "Password [...]" prompt. Enter your smartcard's PIN and press [ENTER].
8. Your system should now unlock and continue booting as usual

### DIFF
--- a/modules.d/91crypt-gpg/module-setup.sh
+++ b/modules.d/91crypt-gpg/module-setup.sh
@@ -5,6 +5,12 @@
 check() {
     require_binaries gpg || return 1
 
+    if [ -f "${initdir}/root/crypt-public-key.gpg" ]; then
+        require_binaries gpg-agent || return 1
+        require_binaries gpg-connect-agent || return 1
+        require_binaries /usr/libexec/scdaemon || return 1
+    fi
+
     return 255
 }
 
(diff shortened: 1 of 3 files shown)


haraldh:
Can you add a README in the module directory how to set this up?

Calrama:
Done in https://github.com/Calrama/dracut/commit/232ed28516b3a60415d49c0ec39e04dfdf2a4695. It explains how to move from a setup using a purely password-based gpg keyfile to a setup using a gpg keyfile that supports both password and smartcard decryption.

Calrama:
Anything still missing?

Calrama:
@haraldh Could I get feedback on this?

MarioGeckler:
@haraldh I would really like to see this implemented

Calrama:
@haraldh Since this PR has received no feedback at all for nearly two years while other PRs are being routinely merged, I must conclude that you have no interest in this contribution to dracut.

modules.d/91crypt-gpg/module-setup.sh
```diff
@@ -5,6 +5,12 @@
 check() {
     require_binaries gpg || return 1
 
+    if [ -f "${initdir}/root/crypt-public-key.gpg" ]; then
```
haraldh: check() is before install() runs, so $initdir is not yet populated

haraldh:
sorry, somehow this PR slipped under my radar

modules.d/91crypt-gpg/module-setup.sh
Calrama: Thanks, should be fixed in 0982fcb2e5ae334790851fa8ff7cf65281842ad1.

Calrama:
@haraldh Okay, I'll try again with #210 (Can't continue here since I made the mistake back then to use my master for feature development).

PR #83, opened 2015-07-16
dracut-systemd/module-setup.sh: don't re-create existing symlinks

puleglot:
These symlinks are created earlier by the systemd module.

This is also a workaroud for incorrect behaviour of
ln from coreutils-8.21 that breaks prelink:

$ mkdir -p /tmp/test/dir
$ touch /tmp/test/file
$ ln -sfnr /tmp/test/file /tmp/test/dir/file
$ stat --printf '%N\n' /tmp/test/dir/file
«/tmp/test/dir/file» -> «../file»
$ ln -sfnr /tmp/test/file /tmp/test/dir/file
$ stat --printf '%N\n' /tmp/test/dir/file
«/tmp/test/dir/file» -> «file»

prelink error:
/usr/sbin/prelink: Failed searching /usr/lib/: Too many levels of
symbolic links

### DIFF
--- a/modules.d/98dracut-systemd/module-setup.sh
+++ b/modules.d/98dracut-systemd/module-setup.sh
@@ -43,16 +43,9 @@ install() {
 
     inst_script "$moddir/rootfs-generator.sh" $systemdutildir/system-generators/dracut-rootfs-generator
 
-    for i in \
-        emergency.target \
-        rescue.service \
-        systemd-ask-password-console.service \
-        systemd-ask-password-plymouth.service \
-        ; do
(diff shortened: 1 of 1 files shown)


haraldh:
Why not just check, if it already exists? It used to be necessary.

puleglot:
If it's really necessary to create these symlinks in both modules: "systemd" and "dracut-systemd", then feel free to close this PR.
I don't really care much about old coreutils. gentoo already started stabilization of coreutils-8.23 and it is unlikely that other distros ship latest dracut with old coreutils. :)

haraldh:
fixed already in master. thanks

PR #90, opened 2015-08-25
dracut.sh: remove duplicate call of push_host_devs

ldzhong:
There is no way exit between the two places that call push_host_devs


haraldh:
true :)

PR #96, opened 2015-09-28
Conceal ftp user passwd in rdsosreport

joeuo:
When using authenticated FTP, the password contained in the kernel command line.
This causes it to end up in the generated /run/initramfs/rdsosreport.txt

Signed-off-by: Zhiguo Deng bjzgdeng@linux.vnet.ibm.com

### DIFF
--- a/modules.d/95zfcp/parse-zfcp.sh
+++ b/modules.d/95zfcp/parse-zfcp.sh
@@ -5,6 +5,8 @@
 getargbool 1 rd.zfcp.conf -d -n rd_NO_ZFCPCONF || rm /etc/zfcp.conf
 
 for zfcp_arg in $(getargs rd.zfcp -d 'rd_ZFCP='); do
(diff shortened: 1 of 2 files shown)


haraldh:
What about all the other sensitive information? iSCSI passwords? Shouldn't the file be reviewed before submitted anyway?

joeuo:
[Conceal-user-passwd-in-rdsosreport.txt](https://github.com/haraldh/dracut/files/36315/Conceal-user-passwd-in-rdsosreport.txt)
When using rd.cmdline=ask and supplying credential parameters from dracut shell, these infos will appear in /etc/cmdline.d/*.conf.  
The patch in attachment is another way of concealing passwords. However, this might be rough. 
For example, the iscsi allows multiple options, due to the greed match nature of sed, then only one
option will be showed in the rdsosreport.
The accurate way of concealing passwords could be parsing, substitution, and then reconstructing the cmdline, like sourcing 'dracut-lib',etc.   And for dmesg and journalctl, this may be a lot of work.

haraldh:
ok to test

haraldh:
@centos-ci 
ok to test

PR #97, opened 2015-10-09
dmsquash-live-root:  Persistence without overlay for uncompressed live installations

FGrose:
dmsquash-live-root: Use non-persistent metadata snapshots for transient overlays.

Temporary snapshots can take advantage of smaller, non-persistent metadata structures.
Make the --readonly option explicit rather than inferred for the readonly_overlay target.
Assure that the live-base target is on the BASE_LOOPDEV.


haraldh:
please remove the bashisms like [[ ]]

FGrose:
See 8b5ee88ff667693b2d8da85a8552ad8f5ab95127 without the bashisms.

PR #102, opened 2015-11-09
dmsquash-live-root: Request overflow support for persistent snapshot.

FGrose:
With kernel 4.3, we can take advantage of the Persistence with Overflow snapshot store type.
https://git.kernel.org/cgit/linux/kernel/git/device-mapper/linux-dm.git/commit/?h=dm-4.3&id=b0d3cc011e532d8c9db76cf717bcafa53c135595


FGrose:
This change is compatible with Fedora 22 Live (kernel 4.0.4-301.fc22.x86_64) and Fedora 23 Live (kernel 4.2.5-300.fc23.x86_64) as both ignore the extra character in the store type code.

PR #105, opened 2015-11-19
livenet: add livenet-generator

antonym:
Resolves issues where systemd attempts to boot a live URL as an NFS mount.

This patch uses systemd's generator arg[2] to generate an early sysroot.mount
that preempts systemd-fstab-generator.

See https://bugzilla.redhat.com/show_bug.cgi?id=1280103


haraldh:
Wow, thank you! :+1:

Issue #104, opened 2015-11-18
Overlayfs instead of device-mapper


probonopd:
I have the exact same issue as described on https://lists.fedoraproject.org/pipermail/users/2015-July/463627.html and https://lists.fedoraproject.org/pipermail/devel/2015-July/212956.html

> The fact is with overlayfs (or aufs from 2014 Ubuntu release) Live Linux you can apt-get install whatever (write to /) as much data as possible; till it becomes full and returns -ENOSPACE (and continues usable if clean some data), on this Fedora Live OS with device-mapper and simple sudo dnf install a few packages crashes kernel. Which is more mature?

The live system crashes when installing large packages despite having 8GB or RAM in the system. Other live systems such as casper do not have this issue. As far as I can see, there is no solution for this yet.

The difference for me as a user is that I can use casper-based live systems for my actual daily work (do not crash) while I cannot do the same with dracut-based Fedora live images (always crash).

Please make it possible to overlay-mount a large(!) amount of RAM so that the live system can handle large file system operations without crashing. 

Thanks for your consideration.

probonopd:
The following does NOT work for me:

```
#!/bin/sh

# Mount the overlayfs kernel module from inside the Live system
# since the kernel module is missing in the initrd (seems to work)
/sysroot/usr/sbin/insmod /sysroot/lib/modules/*/kernel/fs/overlayfs/overlay.ko.xz

mkdir /run/sysroot
mount --move /sysroot /run/sysroot # FIXME: "--move" fails

# Use overlayfs to mount /tmp "over" the root filesystem coming from the Live ISO
mkdir /work
mount -t overlay -o lowerdir=/run/sysroot,upperdir=/tmp,workdir=/work overlay /sysroot

# Not sure whether this is needed but dmsquash-live-root
# has symlinked /dev/mapper/live-rw to /dev/root
rm /dev/root

exit 0
```

probonopd:
Here is something that is beginning to work for me:
https://github.com/probonopd/SystemImageKit/blob/master/boot/iso/additional-initramfs/dracut-overlayfs-support/usr/lib/dracut/hooks/pre-pivot/01-liveoverlayfs.sh

crquan:
thanks @probonopd it looks promising; please continue continue push till it can become a default in next fedora live and I would like a test

haraldh:
@FGrose : what do you think?

probonopd:
[...]

PR #107, opened 2015-12-20
Enable the use of the OverlayFS for the LiveOS root filesystem.

FGrose:
Integrate the option to use an OverlayFS as the root filesystem
into the 90dmsquash-live module for testing purposes.

The rd.live.overlay.overlayfs option allows one to request an
OverlayFS overlay.  If a persistent overlay is detected at the
standard LiveOS path, the overlay & type detected will be used.

Tested primarily with transient in-RAM overlay boots on vfat-
formatted Live USB devices, with persistent overlay directories
on ext4-formatted Live USB devices, and with embedded, persistent
overlay directories on vfat-formatted devices. (Persistent overlay
directories on a vfat-formatted device must be in an embedded
filesystem that supports the creation of trusted.\* extended attributes,
and must provide valid d_type in readdir responses.)

The rd.live.overlay.readonly option, which allows a persistent
overlayfs to be mounted read only through a higher level transient
overlay directory, has been implemented through the multiple lower
layers feature of OverlayFS.

~~This implementation currently requires the enforcing=0 kernel command
line option.  Further SELinux tuning may change this limitation.~~
(SELinux is supported in OverlayFS since kernel 4.8.)

The default transient DM overlay size has been adjusted up to 32 GiB.
This change supports comparison of transient Device-mapper vs.
transient OverlayFS overlay performance.  A transient DM overlay
is a sparse file in memory, so this setting does not consume more
RAM for legacy applications.  It does permit a user to use all of
the available root filesystem storage, and fails gently when it is
consumed, as the available free root filesystem storage on a typical
LiveOS build is only a few GiB.  Thus, when booted on other-
than-small RAM systems, the transient DM overlay should not overflow.

OverlayFS offers the potential to use all of the available free RAM
or all of the available free disc storage (on non-vfat-devices)
in its overlay, even beyond the root filesystem available space,
because the OverlayFS root filesystem is a union of directories on
two different partitions.

### DIFF
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -820,10 +820,12 @@ only filesystem from the SquashFS and apply a writable device-mapper snapshot
 over the read only filesystem.  Using this method ensures a relatively fast
 boot and lower RAM usage. Users **must be careful** to avoid writing too many
 blocks to the snapshot volume.  Once the blocks of the snapshot overlay are
-exhausted, the root filesystem becomes unusable and requires a reboot.  A
-persistent overlay is marked Invalid, and requires a difficult recovery.
-Non-persistent overlays default to 512 MiB in RAM, but the size can be adjusted
-with the **rd.live.overlay.size=** kernel command line option.
+exhausted, the root filesystem becomes read only and may cause application
+failures.  The overlay file is marked 'Overflow', and a difficult recovery is
+required to repair and enlarge the overlay offline.  Non-persistent overlays
+are sparse files in RAM that only consume content space as required blocks are
+allocated.  They default to an apparent size of 32 GiB in RAM.  The size can be
+adjusted with the **rd.live.overlay.size=** kernel command line option.
 +
 The filesystem structure is expected to be:
 +
@@ -832,7 +834,7 @@ The filesystem structure is expected to be:
 squashfs.img          |  Squashfs from LiveCD .iso downloaded via network
    !(mount)
    /LiveOS
-       |- ext3fs.img  |  Filesystem image to mount read-only
+       |- rootfs.img  |  Filesystem image to mount read-only
             !(mount)
             /bin      |  Live filesystem
             /boot     |
@@ -854,7 +856,7 @@ of the root filesystem.
 Uncompressed live filesystem image:::
 When the live system was installed with the '--skipcompress' option of the
 __livecd-iso-to-disk__ installation script for Live USB devices, the root
-filesystem image, `ext3fs.img`, is expanded on installation and no SquashFS
+filesystem image, `rootfs.img`, is expanded on installation and no SquashFS
 is involved during boot.
 +
 - If **rd.live.ram=1** is used in this situation, the full, uncompressed
@@ -919,8 +921,8 @@ root=live:torrent://example.com/liveboot.img.torrent
 Enables debug output from the live boot process.
 
 **rd.live.dir=**__<path>__::
-Specifies the directory within the squashfs where the ext3fs.img or rootfs.img
-can be found.  By default, this is __LiveOS__.
+Specifies the directory within the boot device where the squashfs.img or
+rootfs.img can be found.  By default, this is __LiveOS__.
 
 **rd.live.squashimg=**__<filename of SquashFS image>__::
 Specifies the filename for a SquashFS image of the root filesystem.
@@ -938,6 +940,8 @@ used to persist the changes made to the device specified by the
 **root=live:__<url>__** option.
 - _none_ specifies no overlay when an uncompressed live root filesystem is
 available.
+If a persistent overlay is detected at the standard LiveOS path, the overlay &
+overlay type detected (whether Device-mapper or OverlayFS) will be used.
 +
 [listing]
 .Example
@@ -946,7 +950,7 @@ rd.live.overlay=/dev/sdb1:persistent-overlay.img
 --
 
 **rd.live.overlay.size=**__<size_MiB>__::
-Specifies a non-persistent overlay size in MiB.  The default is _512_.
+Specifies a non-persistent overlay size in MiB.  The default is _32768_.
 
 **rd.live.overlay.readonly=**1::
 Specifies a non-persistent, writable snapshot overlay to be stacked over a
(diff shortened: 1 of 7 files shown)


FGrose:
Mailing list post: http://article.gmane.org/gmane.linux.kernel.initramfs/4308

bcl:
This is worth exploring, but if it can't support SELinux it can't be used for live installations. We also need to make sure that /dev/mapper/live-base or something similar points to the original base filesystem so that local changes don't get copied to the target during installation.

FGrose:
This version continues to provide `/dev/mapper/live-base` even with the `rd.live.overlay.overlayfs` option in order to be compatible with the current installation structures.  Some users may eventually want to avoid any Device-mapper overhead, and the OverlayFS does depend on the read-only base image, `ext3fs.img` or `rootfs.img`, mounted at `/run/rootfsbase`.

The purpose of this version is to provide a means to optionally boot an OverlayFS version of LiveOS in Fedora Rawhide so that others can test its features and limitations, with the hope that those with expertise in SELinux and other affected components could explore adaptations to complete the integration.

haraldh:
ok to test

Conan-Kudo:
@bcl @FGrose My understanding is that this should work with SELinux now? I think Dan Walsh made it work in recent kernels so that Docker+SELinux+OverlayFS worked.

haraldh:
@centos-ci 
ok to test

haraldh:
@centos-ci
retest this please

FGrose:
I forced pushed amended commits to simplify checking the overlayfs commandline argument for those cases where only one check is needed.

FGrose:
I force pushed a couple of fixes: 1. to better deal with the readonly_overlay code path for OverlayFS, and 2. to take advantage of the $sz variable while setting up live-osimg-min.
Here was the diff:
```shell
diff --git a/modules.d/90dmsquash-live/dmsquash-live-root.sh b/modules.d/90dmsquash-live/dmsquash-live-root.sh
index 987065b..60ef6c1 100755
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -164,13 +164,15 @@ do_live_overlay() {
 
     # set up the snapshot
     sz=$(blockdev --getsz $BASE_LOOPDEV)
-    if [ -n "$readonly_overlay" ]; then
-        echo 0 $sz snapshot $BASE_LOOPDEV $OVERLAY_LOOPDEV P 8 | dmsetup create --readonly live-ro
-        base="/dev/mapper/live-ro"
-        over=$RO_OVERLAY_LOOPDEV
-    else
-        base=$BASE_LOOPDEV
-        over=$OVERLAY_LOOPDEV
+    if [ -z "$overlayfs" ]; then
+        if [ -n "$readonly_overlay" ]; then
+            echo 0 $sz snapshot $BASE_LOOPDEV $OVERLAY_LOOPDEV P 8 | dmsetup create --readonly live-ro
+            base="/dev/mapper/live-ro"
+            over=$RO_OVERLAY_LOOPDEV
+        else
+            base=$BASE_LOOPDEV
+            over=$OVERLAY_LOOPDEV
+        fi
     fi
 
     if [ -n "$thin_snapshot" ]; then
@@ -295,7 +297,7 @@ fi
 if [ -b "$OSMIN_LOOPDEV" ]; then
     # set up the devicemapper snapshot device, which will merge
     # the normal live fs image, and the delta, into a minimzied fs image
-    echo "0 $( blockdev --getsz $BASE_LOOPDEV ) snapshot $BASE_LOOPDEV $OSMIN_LOOPDEV P 8" | dmsetup create --readonly live-osimg-min
+    echo "0 $sz snapshot $BASE_LOOPDEV $OSMIN_LOOPDEV P 8" | dmsetup create --readonly live-osimg-min
 fi
 
 ROOTFLAGS="$(getarg rootflags)"
@@ -306,8 +308,8 @@ fi
 if [ -n "$overlayfs" ]; then
     mkdir -m 0755 /run/rootfsbase
     if [ -n "$reset_overlay" ] && [ -L /run/overlayfs ]; then
-        rm -r -- $(readlink /run/overlayfs)
-        mkdir -m 0755 $(readlink /run/overlayfs)
+        ovlfs=$(readlink /run/overlayfs)
+        rm -r -- ${ovlfs}/* ${ovlfs}/.*
     fi
     if [ -n "$readonly_overlay" ]; then
         mkdir -m 0755 /run/rootfsbase-r
```

haraldh:
``` rm -r -- ${ovlfs}/* ${ovlfs}/.*```

Wouldn't that lead to error messages like:
```
rm: refusing to remove '.' or '..' directory: skipping […]
rm: refusing to remove '.' or '..' directory: skipping […]
```

**Comment by haraldh** [comment]
@centos-ci 
retest please

**Comment by haraldh** [comment]
@centos-ci
retest this please

**Comment by FGrose** [comment]
@haraldh re: `rm -r -- ${ovlfs}/* ${ovlfs}/.*`  -- Yes, 2 message lines in several pages.  This version drops them into `/dev/null`.

I've force pushed a couple more fixes: 1. to deal properly with a missing OverlayFS overlay, and 2. to die if 'rd.live.overlay.overlayfs' is specified with a Device-mapper overlay file.
The diff from before is the following:
```patch
diff --git a/usr/lib/dracut/modules.d/90dmsquash-live/dmsquash-live-root.sh b/modules.d/90dmsquash-live/dmsquash-live-root.sh
index 6a5d0c2..27ef3a7 100755
--- a/usr/lib/dracut/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -119,7 +119,10 @@ do_live_overlay() {
             umount -l /run/initramfs/overlayfs || :
             oltype=$(det_img_fs $OVERLAY_LOOPDEV)
             if [ $oltype = DM_snapshot_cow ]; then
-                if [ -n "$reset_overlay" ]; then
+                if [ -n "$overlayfs" ]; then
+                    die "Found a Device-mapper overlay--not OverlayFS... " \
+                        "rd.live.overlay.overlayfs is not valid."
+                elif [ -n "$reset_overlay" ]; then
                     dd if=/dev/zero of=$OVERLAY_LOOPDEV bs=64k count=1 conv=fsync 2>/dev/null
                 fi
             else
@@ -131,7 +134,7 @@ do_live_overlay() {
             ln -s /run/initramfs/overlayfs$pathspec /run/overlayfs$opt
             ln -s /run/initramfs/overlayfs$pathspec/../ovlwork /run/ovlwork$opt
         fi
-        setup="yes"
+        blkid $OVERLAY_LOOPDEV >/dev/null 2>&1 || [ -d "/run/overlayfs$opt" ] && setup="yes"
     fi
     if [ -n "$overlayfs" ]; then
         modprobe overlay
@@ -309,7 +312,7 @@ if [ -n "$overlayfs" ]; then
     mkdir -m 0755 /run/rootfsbase
     if [ -n "$reset_overlay" ] && [ -L /run/overlayfs ]; then
         ovlfs=$(readlink /run/overlayfs)
-        rm -r -- ${ovlfs}/* ${ovlfs}/.*
+        rm -r -- ${ovlfs}/* ${ovlfs}/.* >/dev/null 2>&1
     fi
     if [ -n "$readonly_overlay" ]; then
         mkdir -m 0755 /run/rootfsbase-r
(END)
```

FGrose:
This rebased commit accomplishes the OverlayFS feature and includes the fixes from pull request #190, superseding it, if this commit is accepted.

See also some testing of `sudo dnf upgrade` on a LiveOS with OverlayFS at issue #104.

haraldh:
@centos-ci
retest this please

Conan-Kudo:
@FGrose @haraldh How does this look now?

haraldh:
@centos-ci
retest this please

FGrose:
Added documentation for the OverlayFS overlays.

Added feedback for missing overlays.   But I couldn't find a more generic test for an active plymouth splash, such as rhgb, in the DRACUT_SYSTEMD case.

I also tried
`systemd-ask-password --timeout=0 message`, but the rhgb case fails to show the message prompt (a bug, I think, in plymouth). If I could sendkey <Esc>, this would show the startup console.

I also tried
```sh
cat > /run/systemd/ask-password/ask.0000<<EOF
[Ask]
Socket=/dev/null
Message=${m[*]}
timeout=0
EOF
```
but the message is limited to a single line and the notification doesn't pause and is lost in a lot of other messages at the startup console.

Conan-Kudo:
@haraldh It looks like the current test failures are not related to this PR, and the latest test run in master appears to be broken on the same tests.

At this point, could you consider it acceptable to merge so that it could be available in Fedora 26? @FGrose has prepared changes for my project (livecd-tools) to be able to leverage overlayfs for live media, and I'd like to pull that in and make it available for livecd-tools v26.

haraldh:
tested locally and pushed manually

FGrose:
See pull request #217 for a fix needed for newer mount utilities on Fedora 26.

PR #109, opened 2016-01-05
base/dracut-lib.sh:dev_unit_name() guard against $dev beginning with "-"

puleglot:
crypt/parse-crypt.sh generate initqueue job which always call
dev_unit_name() with an argument beginning with "-". This results
in the following error:

dracut-initqueue[307]: + systemd-escape -p -cfb4aa43-2f02-4c6b-a313-60ea99288087
dracut-initqueue[307]: systemd-escape: invalid option -- 'c'


haraldh:
thanks!

PR #112, opened 2016-01-29
Netboot: don't let NetworkManager manage bonded boot interfaces

tartina:
Boot hangs when NetworkManager takes control of bonded boot interfaces.

Other complex interface configurations could need this too, further check is needed.


haraldh:
ok to test

lkundrak:
I'm pretty sure the issue with NetworkManager is not there anymore and this can be closed.

haraldh:
@tartina please re-open, if this is still an issue

PR #114, opened 2016-02-06
rootfs-generator: use infinity as default timeout value

Keruspe:
systemd recently changed the meaning of 0 for timeouts (not in a release yet)
which makes it impossible to boot with dracut as the device timeouts immediately.

Signed-off-by: Marc-Antoine Perennou <Marc-Antoine@Perennou.com>


poettering:
Note that this is a systemd bug, we will fix before the release, in order not to break compatibility here.

Keruspe:
Fixed in systemd

poettering:
btw, i think actually setting the timeout to "infinity" in dracut would indeed be nicer. But then again, that's only supportted in the upcoming 229, and the old syntax continues to be supported, hence probably not worth the effor now...

PR #138, opened 2016-06-24
permissions on dev/shm should be 777, not 755 - #1347436

AdamWill:
dracut creates (in some circumstances) /run/initramfs/state/
dev/shm/lldpad.state , which is then sync'ed into the real
root by fedora-import-state. Only dracut creates its dev/shm
directory with 0755 permissions, not 0777. /dev/shm is supposed
to be 0777, and various things break if it isn't (e.g. webkit).

Resolves: RHBZ #1347436


haraldh:
Fedora-import-state has to be fixed. I will not create world writable directories under /run/initramfs.

PR #143, opened 2016-07-07
dracut.png: Optimize with zopflipng -m

scop:


haraldh:
what does it do? reduce the size?

scop:
Yes

haraldh:
yay... 1057 bytes saved :)

scop:
:+1:

PR #150, opened 2016-08-16
qemu: include the qemu_fw_cfg kernel module

dm0-:
This adds support for /sys/firmware/qemu_fw_cfg on QEMU guests started with the -fw_cfg option.


dm0-:
Ping? @haraldh?

PR #153, opened 2016-08-25
livenet: support nfs:// urls in livenet-generator

fuhry:
This is a further improvement on PR #105 which re-adds support for nfs:// urls to root=live:nfs://... Symptoms prior to applying this patch are that sysroot.mount times out when following the nfs:// syntax for the path to the live image. An additional case is added to livenet-generator to support the nfs protocol.

### DIFF
--- a/modules.d/90livenet/livenet-generator.sh
+++ b/modules.d/90livenet/livenet-generator.sh
@@ -16,6 +16,9 @@ fi
 [ "${liveroot%%:*}" = "live" ] || exit 0
 
 case "$liveroot" in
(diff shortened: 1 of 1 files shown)


fuhry:
@haraldh 
Would appreciate getting this reviewed and merged quickly - it makes a big difference for our organization where we make the Fedora live environments available via livenet and NFS. It also means the bug is not actually fully fixed in the release version of F24.

RH bugzilla ref # [1317780](https://bugzilla.redhat.com/show_bug.cgi?id=1317780)

PR #154, opened 2016-09-01
Update nfs-lib.sh

honza801:
anaconda_nfsv6_to_var(): fix options parsing. $1 contains  : at the end, this is done by previous functions.


honza801:
cmdline contains
root=nfs:[2001:713::97e]:/mnt/data/fai/nfsroot

before this patch, dracut ends with

/lib/nfs-lib.sh@155(mount_nfs): mount -t nfs -orw,:,nolock '[2001:713::97e]:/mnt/data/fai/nfsroot' /sysroot
mount.nfs: an incorrect mount option was specified

PR #160, opened 2016-10-19
rescue: use proper path for 0-rescue.conf

yuwata:
If ESP is mounted on /boot/efi, which is Fedora's default I think, 51-dracut-rescue.install fails to create ${MACHINE_ID}-0-rescue.conf. This PR fixes the problem by using the proper mounted point of ESP.

### DIFF
--- a/51-dracut-rescue.install
+++ b/51-dracut-rescue.install
@@ -59,16 +59,15 @@ if ! [[ ${BOOT_OPTIONS[*]} ]]; then
     exit 1
 fi
 
-LOADER_ENTRY="/boot/loader/entries/${MACHINE_ID}-0-rescue.conf"
 BOOT_DIR="/${MACHINE_ID}/0-rescue"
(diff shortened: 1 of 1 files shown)


yuwata:
When install kernel, post-script fails as follows,
```
$ df /dev/sdb1
Filesystem     1K-blocks  Used Available Use% Mounted on
/dev/sdb1         204580 76320    128260  38% /boot/efi

$ sudo dnf install kernel
(snip)
/usr/lib/kernel/install.d/51-dracut-rescue.install: line 67: /boot/loader/entries/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-0-rescue.conf: No such file or directory
warning: %posttrans(kernel-core-4.7.8-200.fc24.x86_64) scriptlet failed, exit status 1
Non-fatal POSTTRANS scriptlet failure in rpm package kernel-core
Non-fatal POSTTRANS scriptlet failure in rpm package kernel-core
(snip)
```
This PR fixes the above problem.

PR #168, opened 2016-11-01
Remove `-` prefixing the param for `dev_unit_name()`

eliasp:
This causes `dev_unit_name()` to fail with an error:

```
dracut-initqueue[442]: systemd-escape: invalid option -- '7'
```


eliasp:
Nevermind, this was already properly fixed in c41df7e1db1adc51399ab2c22f251b15eb2065b1

PR #171, opened 2016-11-07
40network: separate 'mask' and 'prefix'

danimo:
The 'mask' parameter is used for both, the (IPv4) netmask and
the prefix length. As both are in different format separate them
out into 'mask' for the netmask and 'prefix' for the prefix length.
And also prefer the use of 'prefix' where possible to ease
calculation and better IPv6 support.

References: bnc#887542

Signed-off-by: Daniel Molkentin <dmolkentin@suse.de>

I had to adjust this one for master.


haraldh:
ok to test

haraldh:
modules.d/40network/net-lib.sh: line 617: syntax error near unexpected token `}'
modules.d/40network/net-lib.sh: line 617: `}'

danimo:
retest this please

PR #182, opened 2016-11-25
95ssh-client: attempt to copy UserKnownHostsFile to kdump's initramfs

jesa7955:
Bug related to this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1360131
Now dracut only attempts to copy GlobalKnownHostsFile while generating kdump's
initramfs. This method will cause kdump's failure if users set customized
UserKnownHostsFile in /etc/ssh/ssh_config. This patch simply attempts to copy
those files too while going through /etc/ssh/ssh_config. Note that we need to
make sure ~/foo will be copied as /root/foo in kdump's initramfs.

Signed-off-by: Tong Li <tonli@redhat.com>


haraldh:
@centos-ci 
ok to test

haraldh:
@jesa7955 : why did you close this?

jesa7955:
@haraldh Xunlei helped me find some issues in this patch, but I didn't find way to commit another change to this pull request which was imported from mailing list. So I planned to reopen a request with my forked repo.

PR #190, opened 2016-12-23
dmsquash-live-root: Use persistent DM targets where needed.

FGrose:
Non persistent DM targets should only be used on transient overlays.

But transient overlays in LiveOS filesystems benefit from the new PO
target setup. Persistence with Overflow support prevents snapshot
invalidation by converting the snapshot to read-only if the overlay
overflows.  This prevents hung filesystems and booted LiveOS crashes.

The change in November 2015 with commit fc14651 was faulty.
Transient/non-persistent DM target setups are not suitable for read-only
snapshots or overlays with persistent data. (Data will be ignored.)

Fortunately, the rd.live.overlay.readonly option is rarely used, and the
$OSMIN_LOOPDEV is no longer used.

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -28,7 +28,7 @@ getargbool 0 rd.live.overlay.readonly -d -y readonly_overlay && readonly_overlay
 overlay=$(getarg rd.live.overlay -d overlay)
 getargbool 0 rd.writable.fsimg -d -y writable_fsimg && writable_fsimg="yes"
 overlay_size=$(getarg rd.live.overlay.size=)
-[ -z "$overlay_size" ] && overlay_size=512
+[ -z "$overlay_size" ] && overlay_size=32768
 
 getargbool 0 rd.live.overlay.thin && thin_snapshot="yes"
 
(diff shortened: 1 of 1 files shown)


FGrose:
Furthermore, with common production LiveCDs, the root filesystem is usually sized with only a few gigabytes of free space. For example, on Fedora 25 (Fedora-Workstation-Live-x86_64-25-1.3.iso) the root filesystem is 6.3 G with 4.0 G used and 2.4 G available or 63% use%.  When I boot this with rd.live.overlay.size=32678, and then try to copy a folder of large files to a folder in the root filesystem, if I  use Files (Nautilus) I receive a warning that there is not enough room.  If I select the 'Copy Anyway' button, I get an 'Error while copying.' dialog, Error splicing file: No space left on device (a similar message is received from a command line copy when the root filesystem has filled).  The significant point here is that the system does not crash, and I can continue to delete files to free space on the root filesystem.   Having an overlay sized larger than the free space on the root filesystem shifts the failure point for over consumption of storage to filesystem software components with more robust error handling.

haraldh:
@centos-ci  ok to test

haraldh:
merged your other patch. care to rebase?

Conan-Kudo:
@haraldh I believe #107 already included this fix, per https://github.com/dracutdevs/dracut/pull/107#issuecomment-274622822

This is okay to close.

PR #191, opened 2016-12-29
Add NTFS support to 90dmsquash-live module

robert-scheck:
Support booting from USB media with NTFS filesystem (optionally), which removes the FAT32 related 4 GB file size limit for LiveOS/squashfs.img (and any other file on the same USB media)

Note: I am aware that my patch is not great, but it would be awesome if you could help to make it proper.

### DIFF
--- a/modules.d/90dmsquash-live-ntfs/module-setup.sh
+++ b/modules.d/90dmsquash-live-ntfs/module-setup.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+command -v
+
+check() {
+    require_binaries ntfs-3g || return 1
+    return 255
+}
+
+depends() {
+    echo 90dmsquash-live
+    return 0
+}
+
+install() {
+    inst_multiple fusermount ulockmgr_server mount.fuse ntfs-3g
+    dracut_need_initqueue
+}
+
+installkernel() {
+    hostonly='' instmods fuse
+}
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -78,7 +78,20 @@ if [ -f $livedev ]; then
     esac
     [ -e /sys/fs/$fstype ] || modprobe $fstype
 else
-    mount -n -t $fstype -o ${liverw:-ro} $livedev /run/initramfs/live
+    if [ "$(blkid -o value -s TYPE $livedev)" != "ntfs" ]; then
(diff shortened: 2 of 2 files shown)


modules.d/90dmsquash-live-ntfs/module-setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+command -v
+
+check() {
+    require_binaries ntfs-3g || return 1
+    return 0
```
haraldh: This should return 255, otherwise this module gets pulled by default, and it will pull in dmsquash-live, which is probably not what you want on a normal system.

haraldh:
Yeah, it's an ugly hack... but I don't have a better solution

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -77,7 +77,20 @@ if [ -f $livedev ]; then
...
 else
-    mount -n -t $fstype -o ${liverw:-ro} $livedev /run/initramfs/live
+    if [ "$(blkid -o value -s TYPE $livedev)" != "ntfs" ]; then
+        mount -n -t $fstype -o ${liverw:-ro} $livedev /run/initramfs/live
+    else
+        # Symlinking /usr/bin/ntfs-3g as /sbin/mount.ntfs seems to boot
+        # at the first glance, but ends with lots and lots of squashfs
+        # errors, because systemd attempts to kill the ntfs-3g process?!
+        if [ -x "$(find_binary "ntfs-3g")" ]; then
+            ( exec -a @ntfs-3g ntfs-3g -o ${liverw:-ro} $livedev /run/initramfs/live ) | (while read l; do warn $l; done)
```
haraldh: I think you want vwarn here:

``` ( exec -a @ntfs-3g ntfs-3g -o ${liverw:-ro} $livedev /run/initramfs/live ) | vwarn```

modules.d/90dmsquash-live-ntfs/module-setup.sh
robert-scheck: Do you mean `require_binaries ntfs-3g || return 255`?

haraldh: The final ```return 0``` replaced with ```return 255```. Of course that would make the module optional, but the LiveCD tools can pull it in, as it does with the dmsquash-live module

robert-scheck:
@haraldh, any further comments or suggestions for improvements?

haraldh:
nope! many thanks!

PR #193, opened 2017-01-09
Support multiple default gateways from DHCP server (rhbz#1408249)

dsilakov:
Currently dracut can fail to setup dhclient in a network where DHCP server provides multiple default gateways. See https://bugzilla.redhat.com/show_bug.cgi?id=1408249 for details


haraldh:
ok to test

haraldh:
Because dhclient-script.sh should be POSIX shell, we cannot use the bash ```<<<```
maybe something like:
```
echo "$gw" | { 
    IFS=' ' read -r main_gw other_gw 
    [rest of gw handling]
}
```

So, please update the PR with some POSIX shell code

haraldh:
@centos-ci 
retest this please

PR #203, opened 2017-02-13
Remove obsolete test dependencies

danimo:
net-tools are not required anymore.

Tested on F25 with package removed.

### DIFF
--- a/README.testsuite
+++ b/README.testsuite
@@ -1,6 +1,5 @@
 For the testsuite to work, you will have to install at least the following software packages:
 dash \
-bridge-utils \
 asciidoc \
 mdadm \
 lvm2 \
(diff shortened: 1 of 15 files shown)


danimo:
ok to test

danimo:
@centos-ci ok to test

danimo:
This patch also drops the dependency on bridge-utils, by replacing them with calls to "ip". Allthough the "bridge" binary is not strictly needed, I added it instead. I can also drop it as it currently serves no purpose.

danimo:
This patch also drops the dependency on bridge-utils, by replacing them with calls to "ip".

haraldh:
@centos-ci ok to test

danimo:
@centos-ci retest this please

danimo:
Not sure what to make of that. Tests 30 and 50 succeed locally.

PR #206, opened 2017-03-09
dracut-init.sh: fix typo

mwilck:
The variable is called "_silent", not "silent". Just stumbled over this by accident.


haraldh:
thanks!

PR #210, opened 2017-03-30
crypt-gpg: For GnuPG >= 2.1 support gpg keyfile decryption via a OpenPGP smartcard in a CCID reader

Calrama:
Continuation of #80.

### DIFF
--- a/modules.d/91crypt-gpg/module-setup.sh
+++ b/modules.d/91crypt-gpg/module-setup.sh
@@ -5,10 +5,11 @@
 check() {
     require_binaries gpg || return 1
 
-    if [ -f "${initdir}/root/crypt-public-key.gpg" ]; then
-        require_binaries gpg-agent || return 1
-        require_binaries gpg-connect-agent || return 1
-        require_binaries /usr/libexec/scdaemon || return 1
+    if sc_requested; then
+        if ! sc_supported; then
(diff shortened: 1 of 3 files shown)


modules.d/91crypt-gpg/module-setup.sh
```diff
@@ -5,6 +5,13 @@
 check() {
     require_binaries gpg || return 1
 
+    if sc_requested; then
+        if ! sc_supported; then
+            dwarning "crypt-gpg: GnuPG >= 2.1 with scdaemon and libusb required for ccid smartcard support"
+            return 1
+        fi
```
haraldh: you might want to "return 0" here

"return 255" means, that the module is only included, if it is specified manually or required by other modules.

But sc_reqeusted() checks for a user configured file, so it would be save to include the module regardless.

What do you think?

haraldh: I mean inbetween line 12 and 13. The part github shows is misleading

Calrama: Thanks for the info. I agree and it's done in 0e1b686b8367925b21c76b9d9e36bf7d2f6b89ac

Calrama:
@haraldh Anything else needs fixing?

PR #213, opened 2017-04-05
Make binutils optional when elfutils are available

danimo:
This is intended for minimum host OSes where 36 MB used by binutils
are deemed too expensive. We only need "strip", which exists as eu-strip
in elfutils, which in turn is < 1 MB in size.
    
 Note that the tests (TEST-04-FULL-SYSTEMD/test.sh) still depend on
 strip from binutils. It could use sstrip in the future.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1605,7 +1605,11 @@ fi
 
 # strip binaries
 if [[ $do_strip = yes ]] ; then
-    for p in strip xargs find; do
+    # Prefer strip from elfutils for package size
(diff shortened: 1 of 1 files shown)


tpgxyz:
Looks interesting, i'll make some test on real hw.

haraldh:
@centos-ci 
ok to test

haraldh:
@centos-ci 
retest this

haraldh:
@centos-ci 
 retest this please

PR #214, opened 2017-04-05
Look for IPv6 RA in wait_for_ipv6_dad, wait_for_ipv6_dad_link and wait_for_ipv6_auto

pallotron:
As per title, you need to wait for default route to be set, this is needed before downloading Anaconda's `squashfs.img` when installing from the network.

Here I am adding the same logic we have in wait_for_ipv6_auto into the other two functions (`wait_for_ipv6_dad`, `wait_for_ipv6_dad_link`).
Also enabling RA acceptance in the `do_static` function...

### DIFF
--- a/modules.d/40network/ifup.sh
+++ b/modules.d/40network/ifup.sh
@@ -108,6 +108,9 @@ do_static() {
     if strglobin $ip '*:*:*'; then
         # note no ip addr flush for ipv6
         ip addr add $ip/$mask ${srv:+peer $srv} dev $netif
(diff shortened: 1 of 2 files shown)


pallotron:
ping? :)
We are using this in production now and works better :)

haraldh:
Are you sure every of these need ra checking?

pallotron:
pretty sure, things were still failing occasionally until I put those in... you must wait for RA to default your default gateway otherwise you won't be able to fetch things outside of your local network.

PR #215, opened 2017-04-05
use more portable /usr/bin/env in shebangs

Mic92:
This commit changes all shebangs to use `#!/usr/bin/env bash` instead of
expecting bash to be located in `/bin`. This is especially important
on Nixos, which trades LFS for a more safer pure package manager. Also
the trend in some distributions is to give up `/bin/` for putting all
binaries into `/usr/bin/`.

### DIFF
--- a/50-dracut.install
+++ b/50-dracut.install
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 COMMAND="$1"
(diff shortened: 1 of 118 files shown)


Mic92:
mhm, this does not work for `bash --norc`. But I do not see a solution to have both `/usr/bin/env` and `--norc`

PR #221, opened 2017-04-25
change the default value of rd.timeout from 0 to 200

xlpang:
Currently rd.timeout default value is 0, this will result in the
systemd job waiting for the target forever, and cause the boot
process hang forever, it's better to wait some default time other
than forever.

We met an issue under kdump, the dump target broke for some reason.
Because wait_for_dev() is invoked due to "--hostonly-cmdline" is
specified when making kdump initramfs in the 1st kernel, we have to
add "rd.timeout=X" to the 1st kernel using grub manually, which is
unacceptable for customers.

This patch sets the default value of rd.timeout to 200s, which is a
little longer than the default value(180s) of rd.try.

Signed-off-by: Xunlei Pang <xlpang@redhat.com>

### DIFF
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -147,14 +147,14 @@ Misc
 
 **rd.retry=**__<seconds>__::
     specify how long dracut should retry the initqueue to configure devices.
-    The default is 30 seconds. After 2/3 of the time, degraded raids are force
+    The default is 180 seconds. After 2/3 of the time, degraded raids are force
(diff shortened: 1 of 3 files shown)


xlpang:
I can't see why there are three check failures, is there any pointer?

haraldh:
Sorry, TEST-50-MULTINIC had a bug. Please rebase your patch to current master and repush to your PR branch.

xlpang:
close this old one, will submit a new rebased version.

Issue #152, opened 2016-08-21
prelink missing libelf.so.1


gagern:
As reported in [Gentoo bug 585106](https://bugs.gentoo.org/show_bug.cgi?id=585106), dracut 044 has problems on Gentoo when prelinking:

```
dracut: *** Resolving executable dependencies ***
dracut: *** Resolving executable dependencies done***
dracut: *** Pre-linking files ***
/usr/sbin/prelink: error while loading shared libraries: libelf.so.1: cannot open shared object file: No such file or directory
dracut: *** Pre-linking files done ***
dracut: *** Stripping files ***
dracut: *** Stripping files done ***
```

The [corresponding dracut code](https://github.com/dracutdevs/dracut/blob/044/dracut.sh#L1559-L1564) reads thus:

``` sh
        dinfo "*** Pre-linking files ***"
        inst_multiple -o prelink /etc/prelink.conf /etc/prelink.conf.d/*.conf
        chroot "$initdir" "$PRELINK_BIN" -a
        rm -f -- "$initdir/$PRELINK_BIN"
        rm -fr -- "$initdir"/etc/prelink.*
        dinfo "*** Pre-linking files done ***"
```

So apparently it lists the files `prelink` will probably need, does the prelink step, then uninstalls said files again. But it misses the `libelf` reported by `ldd`:

```
# ldd /usr/sbin/prelink 
    linux-vdso.so.1 (…)
    libelf.so.1 => /usr/lib64/libelf.so.1 (…)
    libc.so.6 => /lib64/libc.so.6 (…)
    libz.so.1 => /lib64/libz.so.1 (…)
    /lib64/ld-linux-x86-64.so.2 (…)
```

I think `dracut-install --ldd` (or some shell function invoking this, e.g. `DRACUT_RESOLVE_DEPS=1 inst_multiple`) should be able to install the `prelink` binary with all accompanying files. But cleaning that up again might be tricky.

tpgxyz:
Interesting why prelink support is still available in dracut, as it has been proven that prelink does more bad than good. (https://pagure.io/fesco/issue/1183)

Btw, here is the patch that fixes that prelink
https://github.com/OpenMandrivaAssociation/dracut/blob/master/dracut-034-fix-prelink.patch

haraldh:
@tpgxyz care to create a pull request with that patch?

tpgxyz:
@haraldh done, please merge. Thanks

Massimo-B:
Any news here?

[...]

PR #225, opened 2017-04-29
fix prelink missing libelf.so.1

tpgxyz:
this fixes issue #152


gagern:
I'd be slightly concerned that this would remove libelf even in situations where some other binary may depend on it. Is this a valid concern? If so, it might make sense to install the libs to some different directory and include that on `$LD_LIBRARY_PATH`.

haraldh:
right.. we need to check, which libelf* were already there and not remove them afterwards.

Massimo-B:
Please reconsider this patch as it solves the issue with prelink on Gentoo: https://bugs.gentoo.org/585106#c2

PR #226, opened 2017-05-03
Handle curl using libnssckbi for TLS (RHBZ #1447777)

AdamWill:
curl in Fedora recently changed its default CA trust store. The
Fedora package no longer specifies an OpenSSL-format bundle file
during build, and curl itself has been patched to use an NSS
plugin called libnssckbi when no bundle file or directory is
specified. This appears to use a trust bundle in p11-kit format
stored in /usr/share/pki/ca-trust-source .

Unfortunately neither libcurl.so nor libnssckbi.so includes a
string specifying the exact trust store filename, as is the case
for the old approach (which dracut takes advantage of to find
the file). libnssckbi.so contains a string that seems to list the
*directories* it will look for a trust bundle in, but not the
filenames it will look for (I think it will actually use any file
of an appropriate format in the specified dirs, but not sure).

So for an initial proposal this just hardcodes the location of
the bundle file as it is on Fedora currently. I'm not sure if
it may appear elsewhere or under a different name on other
platforms, or if this location / name is standard.

This fixes TLS transactions in the initramfs environment when
using a curl that's built this new way; it's significant for
use of kickstarts and update images with the Fedora / RHEL
installer, as these are retrieved in the initramfs environment,
and are frequently retrieved via HTTPS.

### DIFF
--- a/modules.d/45url-lib/module-setup.sh
+++ b/modules.d/45url-lib/module-setup.sh
@@ -15,7 +15,7 @@ depends() {
 
 # called by dracut
 install() {
-    local _dir _crt _found _lib
+    local _dir _crt _found _lib _nssckbi _p11roots _p11root _p11item
     inst_simple "$moddir/url-lib.sh" "/lib/url-lib.sh"
     inst_multiple -o ctorrent
     inst_multiple curl
@@ -29,6 +29,7 @@ install() {
 	[[ -d $_dir ]] || continue
         for _lib in $_dir/libcurl.so.*; do
 	    [[ -e $_lib ]] || continue
+            [[ $_nssckbi ]] || _nssckbi=$(grep -F --binary-files=text -z libnssckbi $_lib)
             _crt=$(grep -F --binary-files=text -z .crt $_lib)
             [[ $_crt ]] || continue
             [[ $_crt == /*/* ]] || continue
@@ -39,6 +40,39 @@ install() {
             _found=1
         done
     done
-    [[ $_found ]] || dwarn "Couldn't find SSL CA cert bundle; HTTPS won't work."
+    # If we found no cert bundle files referenced in libcurl but we
+    # *did* find a mention of libnssckbi (checked above), install it.
+    # If its truly NSS libnssckbi, it includes its own trust bundle,
+    # but if it's really p11-kit-trust.so, we need to find the dirs
+    # where it will look for a trust bundle and install them too.
+    if ! [[ $_found ]] && [[ $_nssckbi ]] ; then
+        _found=1
+        inst_libdir_file "libnssckbi.so*" || _found=
+        for _dir in $libdirs; do
+            [[ -e $_dir/libnssckbi.so ]] || continue
+            # this looks for directory-ish strings in the file
+            for _p11roots in $(grep -o --binary-files=text "/[[:alpha:]][[:print:]]*" $_dir/libnssckbi.so) ; do
+                # the string can be a :-separated list of dirs
+                for _p11root in $(echo "$_p11roots" | tr ':' '\n') ; do
+                    # check if it's actually a directory (there are
+                    # several false positives in the results)
+                    [[ -d "$_p11root" ]] || continue
+                    # check if it has some specific subdirs that all
+                    # p11-kit trust dirs have
+                    [[ -d "${_p11root}/anchors" ]] || continue
+                    [[ -d "${_p11root}/blacklist" ]] || continue
+                    # so now we know it's really a p11-kit trust dir;
+                    # install everything in it
+                    for _p11item in $(find "$_p11root") ; do
+                        if ! inst "$_p11item" ; then
+                            dwarn "Couldn't install '$_p11item' from p11-kit trust dir '$_p11root'; HTTPS might not work."
+                            continue
+                        fi
+                    done
+                done
+            done
+        done
+    fi
+    [[ $_found ]] || dwarn "Couldn't find SSL CA cert bundle or libnssckbi.so; HTTPS won't work."
 }
 


AdamWill:
I tested this, and it works (a current Fedora 26 image built with this dracut can perform HTTPS transactions using curl from the initramfs environment, like using kickstarts from HTTPS servers; one built without this dracut cannot). But it'd probably be good to talk to @kdudka (who made the change to curl) and @kaie (who deals with CA trust stuff, including the trust store that actually gets used in this instance) about the portability of this, and whether there's a better way to do it than mine.

kdudka:
I only know that f26 libcurl uses ```libnssckbi.so``` as the default source of trust.  It can still be overridden by the ```--cacert``` option of curl to get back the previous behavior.  I have, however, no idea which files ```libnssckbi.so``` uses at run-time under which circumstances.  I believe @kaie could answer this.

kaie:
When talking about "libnssckbi.so" it's important to be clear to which one you're referring.

Name libnssckbi.so is the classic name of a module provided by the NSS library.

Fedora wanted to provided a compatible drop-in replacement with extended functionality, but didn't want to break applications that already accessed this module using this filename.

The classic module libnssckbi.so is still shipped as part of the NSS library, but it is installed into a secondary directory.

A symbolic link is used that allows applications to find the library. In addition, the update-alternatives(8) mechanism is used to cause applications to use the new p11-kit-trust.so module, if installed.

Please make it clear in discussions when you're talking about "the NSS module libnssckbi.so" and when you're referring to "the symbolic link named libnssckbi.so".

The classic NSS module libnssckbi.so doesn't read any external files. Corresponding to the NSS library version from which it was produced, it contains the static set of trust and distrust that was shipped with that version of NSS.

So, if you have a system, where the NSS module libnssckbi.so is present, and have the nspr.rpm library dependency installed, you'll get the default set of trust and distrust. This should be sufficient, if curl/libcurl loads libnssckbi.so

You won't get all the configuration mechanisms that are possible with the ca-certificates and p11-kit-trust.so modules, but it seems that isn't required by your minimal bootstrap environment.


If you wanted to avoid the NSS module and its dependencies, and if you rather wanted to install the p11-kit-trust.so module, you would have to install the full set of files that are shipped as part of the ca-certificates.rpm package, because the filenames inside that package aren't guaranteed to remain stable (except the symbolic links the package installs for backwards compatibility).

If you further require that the static trust bundle files in the various file formats are present, so that openssl and java can read the PEM and keystore files, you must execute the scriptlets that are contained in the ca-certificates package, which execute the update-ca-trust script after package installation, which tells p11-kit-trust.so to read all the multiple source trust files, combine them, and export/extract them into the external file formats.

AdamWill:
@kaie Thanks a lot, that's a lot of the information we needed. I didn't expect this patch was going to be the complete answer by itself, it was just a baseline that at least works in the immediate case.

To answer the question you asked: dracut will copy in whatever the symlink `libnssckbi.so` in the library directory ultimately points to. In a current default Fedora install this appears to be `p11-kit-trust.so`, but see below about cross-distro compatibility. If it's possible that some other distro will:

i) *not* specify a CA bundle file for curl at build time
ii) provide a *different* `libnssckbi.so` implementation that uses a different trust bundle

then we'll have to make dracut cope with that too, in order to support that distro.

 it's important to note that dracut is a *cross-distro* tool. I don't have a completely up-to-date list, but wikipedia claims that it's available on and/or used by default in Fedora, RHEL, Gentoo, Debian, Mandriva/Mageia, openSUSE and something called void. (Arch uses something called 'mkinitcpio', in case you were wondering).

dracut is basically a bunch of bash scripts implementing a modular system to create an initramfs image, mainly by copying files from the host system into the image. Here, we're dealing with the module that provides the ability to use curl from the initramfs environment.

So the basic requirement for the module is "from a bunch of different distros (several of which don't use RPM at all), figure out which files we need to copy into the initramfs so that curl will work".

I suppose dracut could implement some kind of mechanism to handle multiple packaging systems and the modules could do stuff like "on Fedora, copy all files from package ca-certificates into the environment; on Debian, copy all files from package (whateverthedebianpackageiscalled) into the environment..." and so on. But AFAIK it doesn't actually have anything along those lines at present. It basically gives you the ability to just specify file names to copy in - and obviously you can use anything that works in a bash script to *determine* the names - and some functions for doing stuff like "copy in this executable plus all the libraries it's linked against". When the dependency is something complex like this, though, dracut winds up trying to put together some kind of cross-distro-portable logic on the fly.

Like the code it's been using for this job up till now: the code right above the bit this patch adds, where it looks for strings in the curl library file itself that look like a certificate bundle file name, then copies any of the strings that actually exist on the filesystem into the initramfs. It's pretty hairy stuff (and perhaps doesn't work for distros that use a directory not a bundle file...), but then it's hard to think of less hairy ways to do it given the constraints dracut is operating under.

So practically speaking, I guess the questions are:

i) Are the paths `/etc/pki/ca-trust/source` and `/usr/share/pki/ca-trust-source` quite canonical and likely to be the same on all distros, or is there a realistic possibility some distros define *different* or *additional* paths for p11-kit to look in for trusted CAs?

ii) Is it at all likely that any other distro (e.g. RHEL) will build curl such that a *different* `libnssckbi.so` provider - perhaps with different behaviour - will be used?

Thanks!

AdamWill:
So I talked about this quite a lot to @kaie yesterday, and revised the pull request. The basic logic is this:

1) If the existing code 'works' - finds one or more `.crt` files referenced in `libcurl` and successfully finds and installs them - nothing changes, the added code does not run. Otherwise:
2) We try to `inst_libdir_file "libnssckbi.so*"`
3) We find all `libnssckbi.so` files in libdirs and grep them for strings that look like directory names (or colon-separated lists of directory names)
4) We check whether each of those directories i) exists and ii) has `/anchors` and `/blacklist` subdirectories
5) For each one that meets all the conditions, we install its entire contents (using `find`) into the initramfs

Note that we don't necessarily *expect* `libnssckbi.so` to reference any trust directories, because the actual NSS implementation doesn't. So we don't error or warn if we don't find any, we just figure this is fine.

This should, we believe, cover all known scenarios. I've tested it with current Fedora 26 and it works OK.

We may need a bit more variable quoting, I'll try and get that right. I hate shell script.

AdamWill:
Revised again with minor tweaks to messags, and more variable quoting.

AdamWill:
btw, if there's a better way to recursively install an entire directory into the initramfs than using `find`, please let me know. I couldn't see one in `dracut-init.sh`.

kdudka:
Looks good to me (although I did not test it).

dustymabe:
mergable?

haraldh:
Maybe we should "abuse" pkg-config to tell us the relevant directories to copy in the initramfs.

PR #232, opened 2017-06-02
change the default value of rd.timeout from 0 to 300

xlpang:
Currently rd.timeout default value is 0, this will result in the
systemd job waiting for the target forever, and cause the boot
process hang forever, it's better to wait some default time other
than forever.

We met an issue under kdump, the dump target broke for some reason.
Because wait_for_dev() is invoked due to "--hostonly-cmdline" is
specified when making kdump initramfs in the 1st kernel, we have to
add "rd.timeout=X" to the 1st kernel using grub manually, which is
unacceptable for customers.

This patch sets the default value of rd.timeout to 300s, which is
longer than the default value of rd.retry(180s).

Signed-off-by: Xunlei Pang <xlpang@redhat.com>

### DIFF
--- a/install/dracut-install.c
+++ b/install/dracut-install.c
@@ -1238,16 +1238,6 @@ static bool check_module_path(const char *path)
         return true;
 }
 
-static bool check_module_hostonly(struct kmod_module *mod)
-{
-        const char *name = kmod_module_get_name(mod);
-
-        if (check_hashmap(modules_loaded, name))
-                return true;
-
-        return false;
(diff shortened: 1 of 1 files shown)


xlpang:
Please trigger the CI test

daveyoung:
300 should be enough and better than forever, looks good to me

haraldh:
@centos-ci retest this please

haraldh:
@centos-ci retest this please

PR #233, opened 2017-06-02
kernel-install: Skip to create initrd if /etc/machine-id is missing or empty

### LINKED ISSUE #231: Need to adapt for systemd/systemd#5975
Dracut files `/usr/lib/kernel/install.d/50-dracut.install`, `/usr/lib/kernel/install.d/51-dracut-rescue.install` should be updated to exit cleanly when KERNEL_INSTALL_MACHINE_ID is empty.

Also, I'm not sure if `51-dracut-rescue.install` should call `systemd-machine-id-setup` on its own. That seems a strange thing to do it here.

Filing this so we don't forget.

systemd/systemd#5975

yuwata:
From systemd-234, kernel-install plugins are called even if /etc/machine-id
is missing or empty, and in that case BOOT_DIR_ABS is a fake directory.
So, let's skip to create initrd in that case.

C.f. https://github.com/systemd/systemd/commit/9d8813b3b4f69d9428e7759ed8a2956d8846b0fb.
Fixes #231.

### DIFF
--- a/50-dracut.install
+++ b/50-dracut.install
@@ -5,6 +5,12 @@ KERNEL_VERSION="$2"
 BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
 
+# If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
+# So, let's skip to create initrd.
+if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
+    exit 0
+fi
+
 ret=0
 case "$COMMAND" in
     add)
--- a/51-dracut-rescue.install
+++ b/51-dracut-rescue.install
@@ -33,11 +33,15 @@ dropindirs_sort()
 
 [[ -f /etc/os-release ]] && . /etc/os-release
 
-if [[ ! -f /etc/machine-id ]] || [[ ! -s /etc/machine-id ]]; then
-    systemd-machine-id-setup
+if [[ ${KERNEL_INSTALL_MACHINE_ID+x} ]]; then
+    MACHINE_ID=$KERNEL_INSTALL_MACHINE_ID
+elif [[ -f /etc/machine-id ]] ; then
+    read MACHINE_ID < /etc/machine-id
 fi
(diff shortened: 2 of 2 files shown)


50-dracut.install
```diff
@@ -5,6 +5,12 @@ KERNEL_VERSION="$2"
 BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
 
+# If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
+# So, let's skip to create initrd.
+if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
```
haraldh: call me blunt, but why is "-x" needed here?

51-dracut-rescue.install
```diff
@@ -33,11 +33,15 @@ dropindirs_sort()
 
 [[ -f /etc/os-release ]] && . /etc/os-release
 
-if [[ ! -f /etc/machine-id ]] || [[ ! -s /etc/machine-id ]]; then
-    systemd-machine-id-setup
+if [[ ${KERNEL_INSTALL_MACHINE_ID+x} ]]; then
```
haraldh: same here with "+x" ??

I thought [[ ]] already escapes everything

50-dracut.install
yuwata: ${FOO-x} equals to x if FOO is not defined, and equivalent to ${FOO} if defined. So the condition hits only if KERNEL_INSTALL_MACHINE_ID is defined but empty, as explained in the comment.

51-dracut-rescue.install
yuwata: ${FOO+x} equls to x if FOO is defined, and is blank if not defined. So, the above condition hits only if KERNEL_INSTALL_MACHINE_ID is defined.

Actually, the code can be altered without +x. But in this code, I intend to switch the case that systemd is 234+ or not.

yuwata:
systemd-234 was released a week ago. Are there any problem to merge this PR?

PR #236, opened 2017-06-12
udev-rules: pull in 40-redhat-cpu-hotplug.rules (#1460707)

msekletar:
Cpu hotplug policy was split off to separate file. Let's pull in the new
rules file by default.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1266322

Resolves: #1460707

### DIFF
--- a/modules.d/95udev-rules/module-setup.sh
+++ b/modules.d/95udev-rules/module-setup.sh
@@ -27,6 +27,7 @@ install() {
     fi
 
     inst_rules \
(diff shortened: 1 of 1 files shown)


msekletar:
Note that this change was only compile tested.

msekletar:
> Note that this change was only compile tested.

I mean I've build RHEL7 dracut rpm package containing the fix.

msekletar:
Now, I've also regenerated initrd with patched dracut and new rules file is pulled in correctly.

PR #242, opened 2017-06-26
network/dhclient-script: also parse dot separated $new_classless_static_routes

lnykryn:
Previously our dhclient-script expected that $new_classless_static_routes
will have all values separated by a whitespace. But at least on F25
dhclient will put there the destination descriptor in the same format
as it is used by ISC dhcp-server.
For example:
new_classless_static_routes=32.10.198.122.47 192.168.78.4
while our current code expects
new_classless_static_routes=32 10 198 122 47 192 168 78 4

So let's just accept both of these formats by adding "." to IFS.

For details plesse see https://tools.ietf.org/html/rfc3442
"Classless Route Option Format"

### DIFF
--- a/modules.d/40network/dhclient-script.sh
+++ b/modules.d/40network/dhclient-script.sh
@@ -191,7 +191,10 @@ case $reason in
             echo '. /lib/net-lib.sh'
             echo "setup_net $netif"
             if [ -n "$new_classless_static_routes" ]; then
+                OLDIFS="$IFS"
+                IFS=".$IFS"
                 parse_option_121 $new_classless_static_routes
+                IFS="$OLDIFS"
             fi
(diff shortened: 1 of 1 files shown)


lnykryn:
Just for reference here is a config for dhcpd I've used for testing:

```
option domain-name "example.org";
option domain-name-servers ns1.example.org, ns2.example.org;
default-lease-time 600;
max-lease-time 7200;
log-facility local7;

subnet 10.152.187.0 netmask 255.255.255.0 {
}

subnet 192.168.50.0 netmask 255.255.255.0 {
  range 192.168.50.10 192.168.50.30;
  option routers rtr-239-0-1.example.org, rtr-239-0-2.example.org;
  option classless-static-routes 32.10.198.122.47 192.168.78.4;
}
```

PR #243, opened 2017-06-26
Initial support of rootfs on ceph rbd

bitphage:
I did not created a tests yet because I had used separate testing environment deployed via vagrant + ansible including ceph cluster. If tests are mandatory I'll postpone PR.

### DIFF
--- a/modules.d/95rbd/module-setup.sh
+++ b/modules.d/95rbd/module-setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash
 
 # called by dracut
 check() {
--- a/modules.d/95rbd/module-setup.sh
+++ b/modules.d/95rbd/module-setup.sh
@@ -1,0 +1,48 @@
+#!/bin/bash -x
+
+# called by dracut
+check() {
+    local _rootdev
+
+    # if an rbd device is not somewhere in the chain of devices root is
+    # mounted on, fail the hostonly check.
+    [[ $hostonly ]] || [[ $mount_needs ]] && {
+
+        is_rbd() {
+            local _dev=$1 d=
+
+            [[ -L "/sys/dev/block/$_dev" ]] || return
+            d="$(readlink -ev /sys/dev/block/$_dev)"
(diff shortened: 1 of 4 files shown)


modules.d/95rbd/rbdroot.sh
```diff
@@ -0,0 +1,60 @@
...
+    rbd_bus=/sys/bus/rbd/add
+else
+    echo "ERROR: /sys/bus/rbd/add does not exist"
+    return 1
+fi
+
+# tell the kernel rbd client to map the block device
+echo "$mons name=$user,secret=$key $pool $image $snap" > $rbd_bus
+# figure out where the block device appeared
+dev=$(ls /dev/rbd* | grep '/dev/rbd[0-9]*$' | tail -n 1)
```
haraldh: With:
```
dev=$(ls -1art /dev/rbd[0-9]* | (while read l; do dev=$l ;done; echo $dev))
```
```grep``` and ```tail``` are not needed.

File: modules.d/95rbd/rbdroot.sh [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+# add partition if set
+if [ $partition ]; then
+    dev=${dev}p$partition
+fi
+
+mount -t $fstype $dev $NEWROOT -o $opts
```
haraldh: Manual mounting should only be necessary, if ```root=rbd:...```. In the case of ```netroot=rbd:...```, which should be the preferred way, ```root=UUID=...``` or ```root=LABEL=...``` should let dracut or systemd mount the disk just fine on ```$NEWROOT```.

modules.d/95rbd/module-setup.sh
```diff
@@ -0,0 +1,48 @@
...
+}
+
+# called by dracut
+installkernel() {
+    instmods rbd aes cbc
+}
+
+# called by dracut
+install() {
+    inst_multiple grep tail
```
haraldh: Might not be necessary, as stated below.

haraldh:
ping?

bitfag:
Sorry it was a one-time project for me, maybe someone else will do required changes & necessary testing.

PR #244, opened 2017-06-27
90crypt: Fixed crypttab_contains() to also work with device path in /etc/crypttab

danimo:
blkid is not available when this function is called, so block_uuid.map is put into
the initrd, mapping block devices from /etc/crypttab to UUIDs.

This fixes a bug where udev rules were created by mistake as crypttab_contains()
returned false for devices specified by path in /etc/crypttab which resulted in
error messages during boot.

Signed-off-by: Julian Wolf <juwolf@suse.de>

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -75,6 +75,8 @@ install() {
             [[ $_dev == UUID=* ]] && \
                 _dev="/dev/disk/by-uuid/${_dev#UUID=}"
 
+            echo "$_dev $(blkid $_dev -s UUID -o value)" > /usr/lib/dracut/modules.d/90crypt/block_uuid.map
+
             for _hdev in "${!host_fs_types[@]}"; do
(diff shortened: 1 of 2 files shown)


modules.d/90crypt/module-setup.sh
```diff
@@ -66,6 +66,8 @@ install() {
         inst_hook cleanup 30 "$moddir/crypt-cleanup.sh"
     fi
 
+    # Have to use exit here, return value gets ignored in dracut.sh...
+    uuid_map_file=$(mktemp /tmp/dracut.XXXXXXXXXX) || exit 1
```
haraldh: Why not write directly to "$initdir/usr/lib/dracut/modules.d/90crypt/block_uuid.map" ??

danimo: Good question. I guess the original author wanted to install the file "properly". Want me to make the change?

haraldh: yes, please :)

danimo:
Done!

PR #245, opened 2017-06-28
95debug: add strace

danimo:


haraldh:
It's already in there!! :-)

PR #249, opened 2017-07-04
dracut.sh: improve udevdir and systemdutildir fallback logic

floppym:
Check for a common binary in systemdutildir. This resolves an issue on
split-usr systems, where it is common to have both /lib/systemd[/system]
and /usr/lib/systemd[/user] present.

Check for systemd-udevd specifically, since some distros (Gentoo) allow
udev to be installed without the rest of the systemd stack.

Similar logic is applied to udevdir simply for consistency.


haraldh:
@centos-ci
ok to test

floppym:
I'm not sure what failed here, and the detailed logs seem to have been removed from the build server.

haraldh:
@centos-ci retest this please

PR #257, opened 2017-07-27
Support crypto loop keys in initrd.

### LINKED ISSUE #255: Can't use loopback crypto keys when the key is in initrd
According to https://github.com/dracutdevs/dracut/blob/master/modules.d/90crypt/parse-keydev.sh#L20, there's an undocumented feature where `rd.luks.key=keypath:/` should use keypath from the initrd itself.  However this doesn't work with crypto-loop keys.

I tried `rd.luks.key=key.img:/`, and unlocking the key failed with `Command requires device and mapped name as arguments.`, apparently from trying something like `cryptsetup -d - luksOpen /dev/loop0  ` without a name.

https://github.com/dracutdevs/dracut/blob/master/modules.d/91crypt-loop/crypt-loop-lib.sh#L19 should be updated for the case where `$mntp` is `/`.

For me this would be helpful to work around #254.

sbrudenell:
Fixes #255

### DIFF
--- a/modules.d/91crypt-loop/crypt-loop-lib.sh
+++ b/modules.d/91crypt-loop/crypt-loop-lib.sh
@@ -16,7 +16,7 @@ loop_decrypt() {
     local keydev="$3"
     local device="$4"
 
-    local key="/dev/mapper/$(str_replace "loop-keydev-$keydev-$keypath" '/' '-')"
+    local key="/dev/mapper/$(str_replace "loop-$keydev-$mntp-$keypath" '/' '-')"
 
     if [ ! -b $key ]; then
         local loopdev=$(losetup -f "${mntp}/${keypath}" --show)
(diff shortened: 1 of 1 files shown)


haraldh:
@centos-ci 
ok to test

modules.d/91crypt-loop/crypt-loop-lib.sh
```diff
@@ -16,7 +16,7 @@ loop_decrypt() {
     local keydev="$3"
     local device="$4"
 
-    local key="/dev/mapper/$(str_replace "loop-keydev-$mntp-$keypath" '/' '-')"
+    local key="/dev/mapper/$(str_replace "loop-keydev-$keydev-$keypath" '/' '-')"
```
haraldh: err.. I meant "loop-$keydev-$mntp-$keypath"

sbrudenell:
Made your suggested change. Also rebased.

PR #258, opened 2017-07-27
Don't block boot on redundant crypto cleanup.

### LINKED ISSUE #256: Can't use loopback crypto keys to unlock multiple devices
If I use a commandline like

`rd.luks.uuid=abcd rd.luks.uuid=1234 rd.luks.key=key.img`

where the intent is to unlock a keyfile once and use it to `luksOpen` multiple devices with a single password entry. The `luksOpen` chain seems to happen correctly but cleanup is broken.

After unlocking 5 devices with the key, I get:
```Device /dev/mapper/keydev--dev-sdf1-key.img is not active.
Device /dev/mapper/keydev--dev-sdf1-key.img is not active.
Device /dev/mapper/keydev--dev-sdf1-key.img is not active.
Device /dev/mapper/keydev--dev-sdf1-key.img is not active.
```

spammed hundreds of times to the console, then dropped to a shell.

It looks like the problem is:
```
dracut:/# cat /lib/dracut/hooks/initqueue/finished/crypt-loop-cleanup-10-keydev--dev-sdf1-key.img.sh
[ -e "$job" ] && rm -f -- "$job"
/sbin/cryptsetup luksClose /dev/mapper/keydev--dev-sdf1-key.img
```

this is a "finished" hook trying to close the key multiple times, which fails. Other `crypt-look-cleanup-*` hooks are there but similarly trying to cleanup multiple times. By the time I get dropped to the dracut shell, I can just

```
rm /lib/dracut/hooks/initqueue/finished/crypt-loop-cleanup-*
```

and my boot succeeds.

[rdsosreport.txt](https://github.com/dracutdevs/dracut/files/1181136/rdsosreport.txt)

sbrudenell:
Fixes #256

### DIFF
--- a/modules.d/90crypt/crypt-lib.sh
+++ b/modules.d/90crypt/crypt-lib.sh
@@ -219,7 +219,7 @@ readkey() {
                 . /lib/dracut-crypt-loop-lib.sh
                 loop_decrypt "$mntp" "$keypath" "$keydev" "$device"
                 initqueue --onetime --finished --unique --name "crypt-loop-cleanup-99-${mntp##*/}" \
-                    $(command -v umount) "$mntp; " $(command -v rmdir) "$mntp"
+                    $(command -v umount) "$mntp; " $(command -v rmdir) "$mntp; exit 0"
                 return 0
             else
                 die "No loop file support to decrypt '$keypath' on '$keydev'."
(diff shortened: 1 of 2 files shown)


haraldh:
If https://github.com/dracutdevs/dracut/pull/260 works for you, I would favor it.

Your patch does incorrect things.

Let's move the discussion over to https://github.com/dracutdevs/dracut/pull/260

PR #268, opened 2017-08-15
ssh-client: Include nss_ libraries

danimo:
SSH uses passwd database and thus need various NSS plugin libraries,
depending upon setting in nsswitch.conf.
    
SSH binary fails within the dracut environment without the libraries:
    
#:/ ssh
No user exist for uid 0


danimo:
Sorry, initial description was incorrect due to a bug in my github command line client.

PR #269, opened 2017-08-18
introduce new "--no-hostonly-default-device"  argument

xlpang:
In some cases, we do not want the default host devices like root, swap, fstab, etc.
We can use this new argument to skip those devices, and use "--mount" or "--add-device" to explicitly add the wanted devices as needed.

This is useful for kdump which only cares about its dump target.


xlpang:
modified, repushed to retrigger the CI test.

xlpang:
The previous name is root_dev(I played with RHEL7, it is root_dev), seems it needs export, and there is also an old "root_dev" export left in latest upstream dracut.sh

PR #281, opened 2017-09-06
Do not try to autodetect the output name on non-uefi

keszybz:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1475565.
When 20-grubby.install from systemd created an empty
/boot/<machine-id> directory, dracut would unexpectedly try to create
the initramfs image there. Let's use this output directory only for uefi.
This makes the behaviour match the documentation.

Also update the documentation to say that for uefi, the output directory
is different.

### DIFF
--- a/dracut.8.asc
+++ b/dracut.8.asc
@@ -17,8 +17,10 @@ DESCRIPTION
 
 Create an initramfs <image> for the kernel with the version <kernel version>.
 If <kernel version> is omitted, then the version of the actual running
-kernel is used. If <image> is omitted or empty, then the default location
-/boot/initramfs-<kernel version>.img is used.
+kernel is used. If <image> is omitted or empty, then the default location is
+/boot/initramfs-<kernel version>.img for non-UEFI installations, and
+<efidir>/Linux/linux-<kernel version>-<machine id>-<build id>.efi for UEFI
+installations.
 
 dracut creates an initial image used by the kernel for preloading the block
 device modules (such as IDE, SCSI or RAID) which are needed to access the root
--- a/dracut.sh
+++ b/dracut.sh
@@ -773,13 +773,7 @@ if ! [[ $outfile ]]; then
         mkdir -p "$efidir/Linux"
         outfile="$efidir/Linux/linux-$kernel${MACHINE_ID:+-${MACHINE_ID}}${BUILD_ID:+-${BUILD_ID}}.efi"
     else
-        if [[ -e "/boot/vmlinuz-$kernel" ]]; then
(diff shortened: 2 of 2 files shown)


msekletar:
I guess this will fix Fedora default setup which uses GRUB. However, there are Fedora users (including me) who use systemd-boot, but use traditional kernel+initrd instead of single efi binary. AFAICT, proposed fix will cause exactly the same bug on such setups. IOW, initrd will get generated to the location not used by ```systemd-boot```.  Honestly I didn't try it, so maybe I am wrong and initrd will be used in the end. In such case bug is less severe, because I will be able to boot with newly installed kernel. However initrd won't be stored in the location mandated by Boot Loader Specification.

kparal:
Can we please move this patch forward (either include it and test the corner cases, or rework it)? This is an important issue to be resolved before F27 is released. Thanks!

haraldh:
commit 791d80168994f4784569c8ea47b7bff554102f5e should fix the grubby case

keszybz:
Sorry for letting this languish.

I fixed the PR title, it was backwards (said "uefi" instead of "non-uefi").

>  there are Fedora users (including me) who use systemd-boot

That is actually easy: if they are using sd-boot, it means they are on UEFI, since sd-boot doesn't support anything else, so they don't care about this patch at all, since it only matters for non-UEFI boots.

> commit 791d801 should fix the grubby case

Dunno. Harald's patch fixes the immediate issue, so we could just leave things as they are. But I still think it'd be cleaner to apply my patch too: it's more general and provides a clean and easy split: uefi -> one location, non-uefi -> another location, in both cases independent on what is on disk.

The only reason to not apply this would be if there are non-uefi installations which want to use the bootdir-spec layout, but I don't see why there would be. On non-uefi everybody uses grub as set up by anaconda, which uses the old layout.

Changing the directory based on the existence of other files seems prone to errors and races ;)

keszybz:
Updated to resolve the merge conflict.

msekletar:
> That is actually easy: if they are using sd-boot, it means they are on UEFI, since sd-boot doesn't support anything else, so they don't care about this patch at all, since it only matters for non-UEFI boots.

AFAICT, ```$uefi``` is set only when you pass in ```--uefi``` flag and this flag isn't enabled by default on Fedora. IOW, when you are on Fedora and use systemd-boot you don't get single EFI binary which bundles initrd and kernel (unless you change your local config). Hence output file is set in else branch even on UEFI enabled systems.

keszybz:
This change would cause more problems that it solves, so let's close this for now.

PR #300, opened 2017-11-01
10i18n: Load all keymaps for a given locale

danimo:
Previously, dracut would only copy the first one found. However,
with legacy maps for some locales around, there is a chance we
pick the wrong one. Pick all matching keymaps instead

Reference: boo#1065058


haraldh:
@centos-ci 
ok to test

haraldh:
@centos-ci 
retest this

haraldh:
@centos-ci 
retest this please

PR #304, opened 2017-11-08
Unbreak configuration of a VLAN from iBFT

lkundrak:


lkundrak:
Ping?

modules.d/40network/net-lib.sh
```diff
@@ -263,10 +263,12 @@ ibft_to_cmdline() {
                     case "$vlan" in
                         [0-9]*)
                             echo "vlan=$dev.$vlan:$dev"
+                            dev="$dev.$vlan"
                             echo $mac > /tmp/net.${dev}.${vlan}.has_ibft_config
```
haraldh: Wouldn't that change:
 echo $mac > /tmp/net.${dev}.${vlan}.has_ibft_config
to
 echo $mac > /tmp/net.${dev}.${vlan}.${vlan}.has_ibft_config

effectively?

danimo:
Ping?

haraldh:
ping @lkundrak

haraldh:
ping @lkundrak

modules.d/40network/net-lib.sh
tblume: Is a command line  like "ip=ibft0.666:dhcp" allowed at all?
That would mean reconfiguring the interface that was initially set up via firmware.
If the original ip setup from firmware gets overwritten by dhcp, is the interface still to be called ibft?

PR #308, opened 2017-11-28
ifup.sh: don't wait linkup when set static ip explicitly

akihiro:
when set the kernel parameter ```ip=<client-IP>:[<peer>]:<gateway-IP>:<netmask>:<client_hostname>:<interface>:{none|off|dhcp|on|any|dhcp6|auto6|ibft}[:[<mtu>][:<macaddr>]]```, ```iface_has_carrier``` is skipped. And then fail curl fetch for ```root=live:<url>```.

fix check interface linkup (and delayed) by ```iface_has_carrier``` always.

### DIFF
--- a/modules.d/40network/ifup.sh
+++ b/modules.d/40network/ifup.sh
@@ -86,7 +86,7 @@ do_ipv6auto() {
 do_static() {
     strglobin $ip '*:*:*' && load_ipv6
 
-    if [ -z "$dev" ] && ! iface_has_carrier "$netif"; then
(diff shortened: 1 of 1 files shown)


haraldh:
So, you set ```ip=...``` but the network is not connected and you want it to fail faster?

akihiro:
No, I want ```ip=```  wait link up. and I want ```root=live:<url>```  successed.
Now, ifup.sh no wait link-up, and curl is failded by ```Network is unreachable```.
I want ifup.sh wait 3-6 seconds until ethX link up.

PR #309, opened 2017-11-30
Fix warning about multipath if not needed

danimo:
Although no device uses multipath, the module checks
for presence of the multipath binary first, printing a
warning if not present. This patch fixes the wrong ordering.


haraldh:
Thanks!

PR #310, opened 2017-12-04
Remove 00systemd-bootchart

danimo:
This was removed from systemd almost two years ago in
[c550f7a9b89d017215af084288bc44f736f774fe](https://github.com/systemd/systemd/commit/c550f7a9b89d017215af084288bc44f736f774fe), so dracut should drop support
as well.


danimo:
Any objections or can we merge this?

PR #311, opened 2017-12-05
dracut.sh: for_each_host_xx() function should indicate the empty case

daveyoung:
The caller of "for_each_host_xx func" needs to tell three cases:
func success/ fail / not be called.
E.g, in kdump case, host_devs can be empty, and we want to know it.

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -468,7 +468,8 @@ for_each_host_dev_fs()
     local _dev
     local _ret=1
 
-    [[ "${#host_fs_types[@]}" ]] || return 0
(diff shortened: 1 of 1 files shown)


daveyoung:
Some more backgrounds about this, we saw this when testing ssh kdump, for ssh dump the 
host_devs[@] is empty, we do not need any host_devs, probably this is the first time we have an empty array, and previously nobody run into this case.

haraldh:
@centos-ci  please test this

haraldh:
@centos-ci  retest this please

PR #321, opened 2017-12-15
SUSE fixes for 90multipath

danimo:
Amongst others, this adds support for legacy commandline
options multipath=off. I'd like to maintain this upstream.

If not acceptable, I'll drop this one off the stack and keep it a distro patch.

### DIFF
--- a/modules.d/90multipath/multipath-shutdown.sh
+++ b/modules.d/90multipath/multipath-shutdown.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+for i in $(multipath -l -v1); do
+    if ! $(dmsetup table $i | sed -n '/.*queue_if_no_path.*/q1') ; then
+        dmsetup message $i 0 fail_if_no_path
+    fi
+done
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -89,6 +89,7 @@ install() {
     fi
 
     inst_hook cleanup   80 "$moddir/multipathd-needshutdown.sh"
+    inst_hook shutdown  20 "$moddir/multipath-shutdown.sh"
 
     inst_rules 40-multipath.rules 56-multipath.rules \
 	62-multipath.rules 65-multipath.rules \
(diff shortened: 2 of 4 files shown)


haraldh:
Why is 
```
ConditionPathExists=/etc/multipath.conf
```
removed?

haraldh:
@centos-ci  ok to test

danimo:
According to @hreinecke:

  If rootfs is on multipath, but platform does not have an /etc/multipath.conf
  file which is not urgently needed, they system will not boot

danimo:
@centos-ci ok to test

haraldh:
the CI sucks sometimes.. no clue why it is failing here

danimo:
@centos-ci retest this please

danimo:
@haraldh new run worked!

modules.d/90multipath/multipath-shutdown.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+for i in $(multipath -l -v1); do
+    if ! $(dmsetup table $i | sed -n '/.*queue_if_no_path.*/q1') ; then
```
haraldh: This returns on a machine with no multipath:
```
# multipath -l -v1
Jan 17 11:09:57 | /etc/multipath.conf does not exist, blacklisting all devices.
Jan 17 11:09:57 | A default multipath.conf file is located at
Jan 17 11:09:57 | /usr/share/doc/device-mapper-multipath-0.7.3/multipath.conf
Jan 17 11:09:57 | You can run /sbin/mpathconf to create or modify /etc/multipath.conf
Jan 17 11:09:57 | DM multipath kernel driver not loaded
```

danimo: `$ multipath -l -v1`
is empty for me.

Seems what you are seeing is the effect of `0009-RH-don-t-start-without-a-config-file.patch` and `0011-RH-add-mpathconf.patch` in the Fedora RPMs.

It also seems that these messages go to stdout, rather than stderr, so simply ignoring stderr won't cut it. Can you check?

danimo: Maybe prepend `[ -e /etc/multipath.conf  ] || exit 1` ?

haraldh: or better fix the fedora patches 😞

haraldh: Filed https://bugzilla.redhat.com/show_bug.cgi?id=1535868

PR #324, opened 2017-12-15
90dmraid: do not delete partitions

danimo:
There is no point trying to delete partitions; dmraid works
happily even with them. On the contrary trying to delete partitions
can even be harmful when eg dmraid should _not_ be started.

References: bsc#998860


haraldh:
@centos-ci ok to test

PR #325, opened 2017-12-15
95dasd_mod: do not set module parameters if dasd_cio_free is not present

danimo:
We're now parsing the 'rd.dasd' parameter from 95dasd_rules, so
setting the 'dasd_mod' module parameter should be dropped here.


haraldh:
@hreinecke can you elaborate more on this patch please?

hreinecke:
Well, the dasd_mod module will only work if dasd_cio_free is present.
If it isn't we can't include this module.

PR #329, opened 2017-12-15
95resume: Do not resume on iSCSI

danimo:
The iSCSI configuration is started after dracut checks for resume,
so we run into a timeout here. Additionally it's questionable if
resume on iSCSI makes sense (or is even supported on the platform),
so disable it for now.

References: bsc#999663

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -842,3 +842,47 @@ ip_params_for_remote_addr() {
     fi
 
 }
+
+# block_is_nbd <maj:min>
+# Check whether $1 is an nbd device
+block_is_nbd() {
+    [[ -b /dev/block/$1 && $1 == 43:* ]]
+}
+
+# block_is_iscsi <maj:min>
+# Check whether $1 is an nbd device
+block_is_iscsi() {
+    local _dir
+    local _dev=$1
+    [[ -L "/sys/dev/block/$_dev" ]] || return
+    _dir="$(readlink -f "/sys/dev/block/$_dev")" || return
+    until [[ -d "$_dir/sys" || -d "$_dir/iscsi_session" ]]; do
+        _dir="$_dir/.."
+    done
+    [[ -d "$_dir/iscsi_session" ]]
+}
+
+# block_is_fcoe <maj:min>
+# Check whether $1 is an FCoE device
+# Will not work for HBAs that hide the ethernet aspect
+# completely and present a pure FC device
+block_is_fcoe() {
+    local _dir
+    local _dev=$1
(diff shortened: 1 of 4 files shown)


danimo:
@centos-ci ok to test

danimo:
@haraldh Can you give me the magic power to trigger CI runs?

haraldh:
@danimo you now have the power :) May the force be with you :)

haraldh:
@centos-ci ok to test

haraldh:
What, if the swap partition is local?

danimo:
Back from vacation.

Fair point. Is there a reliable way to tell if all swap storage is truly local?

haraldh:
Wouldn't it be better to just tell the user not to place "resume=<path to iSCSI partition>" on the kernel command line?

haraldh:
Ah, Suse is compiling that in the initramfs. So the resume dracut module should skip network block devices.

danimo:
Does dracut already have a function to drill down && check for the type of the underlying block device?

haraldh:
Nothing globally yet. 

```modules.d/95iscsi/module-setup.sh``` has for example:

```
    is_iscsi() {
        local _dev=$1

        [[ -L "/sys/dev/block/$_dev" ]] || return
        cd "$(readlink -f "/sys/dev/block/$_dev")"
        until [[ -d sys || -d iscsi_session ]]; do
            cd ..
        done
        [[ -d iscsi_session ]]
    }
```
```modules.d/95nbd/module-setup.sh``` has:
```
   is_nbd() { [[ -b /dev/block/$1 && $1 == 43:* ]] ;}
```

danimo:
Any suggestion on how to avoid code dupes? {iscsi,nbd}-lib.sh will not work, as they cannot be sourced from the module-setup.sh that installs them.

haraldh:
well, put those functions in dracut-functions.sh, create a function ```block_is_netdevice()```, which calls all tests for network devices and use that.

haraldh:
will you or should I?

dustymabe:
this looks related to #480 - can someone confirm?

haraldh:
@danimo ping

haraldh:
please rebase

danimo:
Done. But the `swap_on_netdevice` check should probably also be hostonly, right?

haraldh:
> Done. But the `swap_on_netdevice` check should probably also be hostonly, right?

correct

danimo:
Done.

Conan-Kudo:
@danimo It looks good, but can you rebase on current master?

danimo:
@Conan-Kudo done.

PR #331, opened 2017-12-16
call fsfreeze(8) on /boot to flush initramfs data & metadata to media

fcami:
If /boot is a journaling file system, the only guarantee after sync(1) is that data has been written to the disk media, but not metadata.
It is necessary to either umount the filesystem or call fsfreeze(8) for metadata to be written to stable disk media (elsewhere than the journal).
As grub2 does not know how to handle a journaled filesystem, e316ae0e4309726b2c067a70ac41f7b22011c063 alone does not ensure consistency.

Note that I explicitly avoided using fsfreeze(8) if /boot and / belong to the same file system - I think this would be wrong for many reasons.

This is related to https://github.com/rhboot/grubby/pull/28 and part of an on-going effort to fix https://bugzilla.redhat.com/show_bug.cgi?id=1227736 .

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1828,8 +1828,13 @@ fi
 command -v restorecon &>/dev/null && restorecon -- "$outfile"
 
 if ! sync "$outfile" 2> /dev/null; then
-    dinfo "dracut: sync operartion on newly created initramfs $outfile failed"
+    dinfo "dracut: sync operation on newly created initramfs $outfile failed"
     exit 1
+# use fsfreeze only if we're not writing to /
+elif ! [ "$(stat -c %m -- "$outfile")" == "/" ]; then
+    if ! $(fsfreeze -f $(dirname "$outfile") 2>/dev/null && fsfreeze -u $(dirname "$outfile") 2>/dev/null); then
+        dinfo "dracut: warning: could not fsfreeze $(dirname "$outfile")"
+    fi
 fi
 
 exit 0


haraldh:
Hmm, better find out about the mountpoint with:
```
$ stat -c %m -- "$outfile"
/boot
```
as the ```$outfile``` can also be e.g. ```/boot/9e3d4b6532ff41e1ab40d6b4e5609907/4.15.0-0.rc1.git1.1.fc28.x86_64/initrd```

fcami:
Thanks for the hint - I've switched to using stat.
If this one looks good I'll cherry-pick to the RHEL7 branch, but maybe you'd prefer to update it first?

dracut.sh
```diff
@@ -1828,8 +1828,13 @@ fi
 command -v restorecon &>/dev/null && restorecon -- "$outfile"
 
 if ! sync "$outfile" 2> /dev/null; then
-    dinfo "dracut: sync operartion on newly created initramfs $outfile failed"
+    dinfo "dracut: sync operation on newly created initramfs $outfile failed"
     exit 1
+# use fsfreeze only if we're not writing to /
+elif ! [ "stat -c %m -- $outfile" == "/" ]; then
```
haraldh: Well I guess that should be "$(stat...)"

haraldh: and you should put $outfile in quotes also

fcami:
fixed, thank you... does this trace look OK to you?

+ sync /boot/initramfs-4.14.3-300.fc27.x86_64.img
++ stat -c %m -- /boot/initramfs-4.14.3-300.fc27.x86_64.img
+ '[' /boot == / ']'
+++ dirname /boot/initramfs-4.14.3-300.fc27.x86_64.img
++ fsfreeze -f /boot
+++ dirname /boot/initramfs-4.14.3-300.fc27.x86_64.img
++ fsfreeze -u /boot
+ exit 0

haraldh:
Looks good... what only worries me, that if the unfreeze fails or somebody pressed ctrl-c.

PR #338, opened 2017-12-19
dracut-emergency: optionally print fs help

danimo:
Allow filesystem modules to install a fs-specific text file with instructions
on what to do when mount fails. This is printed when we go into an emergency
shell.

Signed-off-by: Mark Fasheh <mfasheh@suse.de>

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -261,6 +261,12 @@ inst_script() {
     (($? != 0)) && derror FAILED: $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${loginstall:+-L "$loginstall"} ${DRACUT_RESOLVE_DEPS:+-l}  ${DRACUT_FIPS_MODE:+-f} "$@" || :
 }
 
+inst_fsck_help() {
+    local _helper="/run/dracut/fsck/fsck_help_$1.txt"
+    $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${loginstall:+-L "$loginstall"} ${DRACUT_RESOLVE_DEPS:+-l} ${DRACUT_FIPS_MODE:+-f} "$2" $_helper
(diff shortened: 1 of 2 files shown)


danimo:
@centos-ci retest this please

dracut-init.sh
```diff
@@ -261,6 +261,12 @@ inst_script() {
     (($? != 0)) && derror FAILED: $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${loginstall:+-L "$loginstall"} ${DRACUT_RESOLVE_DEPS:+-l}  ${DRACUT_FIPS_MODE:+-f} "$@" || :
 }
 
+inst_fsck_help() {
+    local _helper="/usr/share/fsck/fsck_help_$1.txt"
+    $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${loginstall:+-L "$loginstall"} ${DRACUT_RESOLVE_DEPS:+-l} ${DRACUT_FIPS_MODE:+-f} "$2" $_helper
+    (($? != 0)) && derror $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${loginstall:+-L "$loginstall"} ${DRACUT_RESOLVE_DEPS:+-l}  ${DRACUT_FIPS_MODE:+-f} "$2" $_helper || :
+}
+
```
haraldh: ```/usr/share/fsck```
hmm... why not in ```/run/dracut``` or ```/tmp``` ???

danimo: Probably because the author was not aware of the policies in dracut's initramfs. I'll change that.

danimo:
Changed install location accordingly.

modules.d/98dracut-systemd/dracut-emergency.sh
```diff
@@ -25,6 +26,7 @@ if getargbool 1 rd.shell -d -y rdshell || getarg rd.break -d rdbreak; then
     echo 'after mounting them and attach it to a bug report.'
     echo
     echo
+    [ -f $FSTXT ] && cat $FSTXT
```
haraldh: nitpicking:
[ -f "$FSTXT" ] && cat "$FSTXT"

danimo: Fair enough, fixed.

PR #339, opened 2017-12-19
Fix parsing of "-i" and "--include"

danimo:
- dracut replaced every instance of "-i" in the cmdline,
 even if it was part of a kernel image name, e.g. "vmlinuz-i"
- Fixes boo#908452

Signed-off-by: Fabian Vogt <fvogt@suse.com>

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -286,11 +286,15 @@ dropindirs_sort()
 rearrange_params()
 {
     # Workaround -i, --include taking 2 arguments
-    set -- "${@/--include/++include}"
-
-    # This prevents any long argument ending with "-i"
-    # -i, like --opt-i but I think we can just prevent that
-    set -- "${@/%-i/++include}"
+    newat=()
+    for i in "$@"; do
+      if [[ $i == "-i" ]] || [[ $i == "--include" ]]; then
+            newat+=("++include") # Replace --include by ++include
+        else
+            newat+=("$i")
+        fi
+    done
+    set -- "${newat[@]}" # Set new $@
 
     TEMP=$(unset POSIXLY_CORRECT; getopt \
         -o "a:m:o:d:I:k:c:L:fvqlHhMN" \


haraldh:
```+ make V=1 check
make[1]: Entering directory '/root/dracut/test/TEST-01-BASIC'
TEST: root filesystem on a ext3 filesystem 0+0 records in
0+0 records out
0 bytes copied, 7.7591e-05 s, 0.0 kB/s

Unknown arguments: 4.8.6-300.fc25.x86_64

Usage: /root/dracut/dracut.sh [OPTION]... [<initramfs> [<kernel-version>]]

Version: 044-77.fc25

Creates initial ramdisk images for preloading modules

  -h, --help  Display all options

If a [LIST] has multiple arguments, then you have to put these in quotes.

For example:

    # dracut --add-drivers "module1 module2"  ...

[FAILED]
```

dracut.sh
```diff
@@ -286,11 +286,17 @@ dropindirs_sort()
...
+    newat=()
+    for i in "$@"; do
+        if [[ $i =~ ^-i(.*) ]]; then
+            newat+=("++include" "${BASH_REMATCH[1]}") # Replace -i by ++include
+        elif [[ $i == "--include" ]]; then
+            newat+=("++include") # Replace --include by ++include
+        else
+            newat+=("$i")
+        fi
+    done
```
haraldh:
```    for i in "$@"; do
        if [[ $i == "-i" ]] || [[ $i == "--include" ]]; then
            newat+=("++include") # Replace "--include" or "-i" by "++include"
        else
            newat+=("$i")
        fi
    done
```
should be good enough. 
I don't think we should support ```dracut -ifoo bar […]```

danimo: Fair enough. You or me?

haraldh:
Easier for you to repush the PR.

danimo:
Done, let's see what CI says.

danimo:

PR #340, opened 2017-12-19
mkinitrd-suse.sh: Implement functionality of -A option

danimo:
- Without this patch, -A sets host_only=0,
   but host_only wasn't used
- Translates into --no-host-only now


danimo:

PR #345, opened 2018-01-09
00warpclock: Set correct timezone

danimo:
Add module for setting correct timezone.

References: bnc#830060

For now, this module will not be included automatically due to different
expectations (see e.g. https://bugzilla.redhat.com/show_bug.cgi?id=981617).

Signed-off-by: Daniel Molkentin <daniel.molkentin@suse.com>

### DIFF
--- a/modules.d/00warpclock/module-setup.sh
+++ b/modules.d/00warpclock/module-setup.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# called by dracut
+check() {
+    # hwclock does not exist on S390(x), bail out silently then
+    local _arch=$(uname -m)
+    [ "$_arch" = "s390" -o "$_arch" = "s390x" ] && return 1
+
+    [ -e /etc/localtime -a -e /etc/adjtime ] || return 1
+    require_binaries /sbin/hwclock || return 1
+
+    return 255
+}
+
+# called by dracut
+depends() {
+    return 0
+}
+
+# called by dracut
+install() {
+    inst /usr/share/zoneinfo/UTC
+    inst /etc/localtime
+    inst /etc/adjtime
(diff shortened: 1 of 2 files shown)


danimo:
nb.: This patch is already included in openSUSE and SLE, but with `return 0` rather than `return 255`.

Conan-Kudo:
Could you please put the correct bug reference link in here? The URL doesn't go to a specific bug report...

danimo:
@Conan-Kudo Whoops, fixed!

haraldh:
We could omit the while module if  /etc/adjtime != LOCAL ... basically the warpclock.sh check could be done in module-setup.sh in check()

modules.d/00warpclock/module-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+    local _arch=$(uname -m)
+    [ "$_arch" = "s390" -o "$_arch" = "s390x" ] && return 1
+
+    [ -e /etc/localtime -a -e /etc/adjtime ] || return 1
+    require_binaries /sbin/hwclock || return 1
+}
+
+# called by dracut
+depends() {
+    return 255
```
haraldh: shouldn't this be
```
return 0
```

modules.d/00warpclock/module-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# called by dracut
+check() {
+    # hwclock does not exist on S390(x), bail out silently then
+    local _arch=$(uname -m)
+    [ "$_arch" = "s390" -o "$_arch" = "s390x" ] && return 1
+
+    [ -e /etc/localtime -a -e /etc/adjtime ] || return 1
+    require_binaries /sbin/hwclock || return 1
```
haraldh: to make it "optional" do
```
return 255
```
here and not in depends()

danimo: Done

modules.d/00warpclock/module-setup.sh
danimo: Done.

PR #351, opened 2018-01-11
Add 91zipl, which adds support for indirect booting on s390.

danimo:
The s390 zipl bootloader is not capable of reading from certain
more advanced filesystems such as btrfs. We therefore need an indirect
chaining process to boot systems like SLES. This works by letting ZIPL
boot a kernel from an ext partition, which loads a dracut-generated initrd
containing an ELF-based GRUB. This grub then loads the effective kernel
and userland via kexec.

### DIFF
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -1001,6 +1001,20 @@ be mounted read only through a higher level transient overlay directory, has
 been implemented through the multiple lower layers feature of OverlayFS.
 
 
+ZIPL
+~~~~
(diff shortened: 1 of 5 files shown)


danimo:
@hreinecke Please check my modification of https://github.com/dracutdevs/dracut/pull/351/commits/ea7ffef13e00caa80cf37dbc53938ead25fb0d72, which is supposed to not disrupt distros with direct boot.

danimo:
@centos-ci retest this please

PR #358, opened 2018-01-16
40network: Fix race condition when wait for networks

danimo:
If no network related params are specific, but rd.neednet=1 is set,
the default initqueue action is to wait until one of the network
interfaces is marked as setup properly.

This also help with initqueue's race condition when the network interface
shows up late

References: bnc#866771

### DIFF
--- a/modules.d/40network/net-genrules.sh
+++ b/modules.d/40network/net-genrules.sh
@@ -105,6 +105,9 @@ command -v fix_bootif >/dev/null || . /lib/net-lib.sh
         cond='ACTION=="add", SUBSYSTEM=="net", ENV{DEVTYPE}!="wlan|wwan"'
         # if you change the name of "91-default-net.rules", also change modules.d/80cms/cmssetup.sh
         echo "$cond, $runcmd" > /etc/udev/rules.d/91-default-net.rules
+        if [ "$NEEDNET" = "1" ]; then
+            echo 'for i in /tmp/net.*.did-setup; do [ -f "$i" ]  && exit 0; done; exit 1' >$hookdir/initqueue/finished/wait-network.sh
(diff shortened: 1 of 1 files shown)


modules.d/40network/net-genrules.sh
```diff
@@ -105,6 +105,9 @@ command -v fix_bootif >/dev/null || . /lib/net-lib.sh
         cond='ACTION=="add", SUBSYSTEM=="net", ENV{DEVTYPE}!="wlan|wwan"'
         # if you change the name of "91-default-net.rules", also change modules.d/80cms/cmssetup.sh
         echo "$cond, $runcmd" > /etc/udev/rules.d/91-default-net.rules
+        if [ "$NEEDNET" = "1" ]; then
+            echo "[ -f /tmp/net.*.did-setup ]" >$hookdir/initqueue/finished/wait-network.sh
```
haraldh:
```
$ dash -c '[ -f /tmp/net.*.did-setup ] && echo OK'
$ touch /tmp/net.foo.did-setup
$ dash -c '[ -f /tmp/net.*.did-setup ] && echo OK'
OK
$ touch /tmp/net.bar.did-setup
$ dash -c '[ -f /tmp/net.*.did-setup ] && echo OK'
dash: 1: [: /tmp/net.bar.did-setup: unexpected operator
$
$
$ dash -c '( for i in /tmp/net.*.did-setup; do [ -f "$i" ] && exit 0; done ) && echo OK'
OK
```

Better use something like:
```
echo 'for i in /tmp/net.*.did-setup; do [ -f "$i" ] && exit 0; done; exit 1' >$hookdir/initqueue/finished/wait-network.sh
```

PR #361, opened 2018-01-16
90multipath: load dm_multipath module during startup

danimo:
As the 'multipath' program will be triggered directly from
udev events it will be called before the multipath service
unit has started up. Which means we cannot rely on the
service unit to load the module for us, but we rather
have to do it early before udev is started.

References: bsc#986734


danimo:
@centos-ci retest this please

PR #365, opened 2018-01-18
Enable 'early_microcode' by default

Conan-Kudo:
This PR contains a patch from Thomas Backlund from Mageia (rebased to v046) to change the `early_microcode` setting to default to `yes`.

It was originally patched in Mageia for [mga#16657](https://bugs.mageia.org/show_bug.cgi?id=16657), but it makes sense for it to be an upstream default, even more so now with Meltdown and Spectre...


danimo:
SUSE does

```
%ifarch %ix86 x86_64
echo 'early_microcode="yes"' > %{buildroot}%{_sysconfdir}/dracut.conf.d/02-early-microcode.conf
%endif
```
Enabling it globally changes the assumption that the parameter is opt-in. When used on non x86, it will throw the warning: `"Disabling early microcode, because kernel does not support it. CONFIG_MICROCODE_[AMD|INTEL]!=y`. This is confusing.

Can you refine the code in dracut.sh:1082ff accordingly? I suggest checking for x86 and x86_64 there as long as no other microcode updates are supported.

Conan-Kudo:
@danimo My understanding is that most architectures *do* support microcode updates. Are you saying the kernel doesn't support early loading of that microcode?

danimo:
I misread the part, sorry. it only checks the kernel config for specific CPU vendors. I'd have to check how it works with non-intel, but at least there is no warning.

However, I do not see support for anything but AMD and Intel, see dracut.sh:

```
    dinfo "*** Generating early-microcode cpio image ***"
    ucode_dir=(amd-ucode intel-ucode)
    ucode_dest=(AuthenticAMD.bin GenuineIntel.bin)
    _dest_dir="$early_cpio_dir/d/kernel/x86/microcode" 
    _dest_idx="0 1"
    mkdir -p $_dest_dir
    if [[ $hostonly ]]; then
        [[ $(get_cpu_vendor) == "AMD" ]] && _dest_idx="0"
        [[ $(get_cpu_vendor) == "Intel" ]] && _dest_idx="1"
    fi
   ...
```

Am I missing something?

Conan-Kudo:
Maybe, but I'd be missing the same thing... I would expect that every major architecture would have some process for microcode loading. I guess it's just not in dracut yet...

danimo:
No objection to the patch then from my side.

PR #368, opened 2018-01-18
50plymouth: improve distro compatibility

danimo:
The existence of dpkg-achitecture is not indicative of a debian
installation. It may well be installed on systems of people who
package for both distros. The previous code path did not take
that into account.

We now traverse all known plymouth directories, locking on the first
existing one, and try to work with it.

At the same time, we do not include the module if the plymouth directory
could not be found.

This obsoletes 0300-dracut_dont_use_dpkg_defaults_on_SUSE.patch.

Also: 90plymouth: logo file is optional and may not exist

### DIFF
--- a/modules.d/50plymouth/module-setup.sh
+++ b/modules.d/50plymouth/module-setup.sh
@@ -1,8 +1,23 @@
 #!/bin/bash
 
+pkglib_dir() {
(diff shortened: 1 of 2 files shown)


danimo:
@centos-ci retest this please

danimo:
@centos-ci retest this please

PR #371, opened 2018-01-27
support kernels with built-in modules

prometheanfire:
Don't fail if modprobe fails to load a module, the kernel could have it
statically compiled in.

Signed-off-by: Matthew Thode <mthode@mthode.org>


prometheanfire:
this was the only modprobe in a service file I could find.

PR #375, opened 2018-02-09
95qeth_rules: Add new module to copy qeth rules

danimo:
Only pick rules for interfaces which have a carrier in the running
system. Those interfaces will be assembled by udev to allow booting
from those devices (i.e. iSCSI).

Reference: FATE#323440

### DIFF
--- a/modules.d/95qeth_rules/module-setup.sh
+++ b/modules.d/95qeth_rules/module-setup.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    local _arch=$(uname -m)
+    local _online=0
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@danimo Could you please put a URL reference for this? I don't know what `FATE` is or where to look for it.

haraldh:
@centos-ci retest this please

danimo:
@Conan-Kudo FATE is our feature DB, where customers can request new features. This is just to keep track of them.

Essentially this follows the other *_rules modules, which are also purely for s390, where its customary to assemble devices udev rules on demand only. So this copies the rules for qeth devices (s390 ethernet) into initrd so iscsi can be used in initramfs.

PR #384, opened 2018-03-22
Merge fips

lnykryn:


yuwata:
Is it not necessary to add Obsoletes line for the dropped packages?

lnykryn:
@yuwata yep you are right; I will create a new pull request to fix that.

lnykryn:
https://github.com/dracutdevs/dracut/pull/385

PR #386, opened 2018-03-23
plymouth: fix detection of plymouth directory

puleglot:
Some distros have both /usr/lib/plymouth and /usr/libexec/plymouth
directorirs, so we should check the existance of plymouth-populate-initrd
script.

Fixes: 421b46f8ae89cfe2b62e880a8a5079ee8c1b3aae


puleglot:
Downstream bug report: https://bugs.gentoo.org/651132

PR #387, opened 2018-03-30
make failing installation of drm modules nonfatal

Gottox:
During dracut runs users of VoidLinux reported ERROR messages. After some investigation I concluded that those errors are non-fatal and can be ignored. Nontheless, to prevent further confusion, let's make the installation of those drm modules optional so no error is occuring.


Gottox:
ref: https://github.com/voidlinux/void-packages/issues/12970

PR #392, opened 2018-04-19
Initial Stratis support

tasleson:
Ref. https://github.com/stratis-storage

Signed-off-by: Tony Asleson <tasleson@redhat.com>


haraldh:
why don't you use systemd units for the daemon?

tasleson:
@haraldh Thanks, will look into using that instead

haraldh:
now squash your commits and I will merge them

tasleson:
> now squash your commits and I will merge them

Done, thanks!

PR #399, opened 2018-05-02
Suppress sync/fsfreeze if not running on a live system

cgwalters:
It's possible for e.g. `kernel` to be installed as an RPM BuildRequires or equivalent,
and there's no reason to sync, and *definitely* no reason to fsfreeze.

Another case where this happens is rpm-ostree, which performs its own sync/fsfreeze
globally.  See e.g. https://github.com/ostreedev/ostree/commit/8642ef5ab3fec3ac8eb8f193054852f83a8bc4d0

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1832,15 +1832,22 @@ fi
 
 command -v restorecon &>/dev/null && restorecon -- "$outfile"
 
-if ! sync "$outfile" 2> /dev/null; then
-    dinfo "dracut: sync operation on newly created initramfs $outfile failed"
-    exit 1
-fi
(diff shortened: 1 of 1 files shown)


cgwalters:
For more detail see [the rpm-ostree dracut code](https://github.com/projectatomic/rpm-ostree/blob/506910d9300120f57ce710206ddc6fc0f2c9a628/src/libpriv/rpmostree-kernel.c#L440) which runs `dracut` inside a `bwrap` container, where we won't have `/run/systemd/system`.

cgwalters:
(Also on a totally unrelated topic, it'd be nice if I could pass dracut an `O_TMPFILE` fd and it would write the initramfs to that)

PR #400, opened 2018-05-03
10i18n: Fix possible infinite recursion

danimo:


danimo:
@centos-ci retest this please

PR #411, opened 2018-06-07
Harden dracut against BASH_ENV environment variable

rmetrich:
When dracut silently produces a broken initramfs, then the system will
likely not boot and this can be very problematic. Typical use case is
after the kernel has been updated.

It appears that dracut is not protected against the **BASH_ENV** variable,
causing various scripts called by dracut to possibly fail or provide
wrong output (e.g. `ldd` is one of these).
Having a broken output for `ldd` makes the generated initramfs be not
usable, typically because vital binaries will be missing (e.g.
`awk`, `udevadm`, ...).

Note: because the shebang line cannot contain more than one argument,
the `--norc` option had to be removed. IMHO, it was useless anyway.

Signed-off-by: Renaud Métrich <rmetrich@redhat.com>

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1,4 +1,4 @@
-#!/bin/bash --norc
+#!/bin/bash -p
 #
 # Generator script for a dracut initramfs
 # Tries to retain some degree of compatibility with the command line
@@ -22,6 +22,8 @@
 #
 
 # store for logging
+
+unset BASH_ENV
 
 # Verify bash version, current minimum is 4
 if (( BASH_VERSINFO[0] < 4 )); then


rmetrich:
Example:

1. Create initramfs

~~~
# dracut -f /tmp/initramfs_noenv.img $(uname -r)
# lsinitrd /tmp/initramfs_noenv.img > /tmp/initramfs_noenv.img.lsinitrd
~~~

2. Create initramfs after setting BASH_ENV

~~~
# export BASH_ENV=/root/banner
# echo "echo Evil Banner /dracut/killer" > /root/banner
# dracut -f /tmp/initramfs.img $(uname -r)
# lsinitrd /tmp/initramfs.img > /tmp/initramfs.img.lsinitrd
~~~

3. Compare

~~~
# diff /tmp/initramfs_noenv.img.lsinitrd /tmp/initramfs.img.lsinitrd | grep udevadm
< lrwxrwxrwx   1 root     root           22 Jan  5 10:18 usr/sbin/udevadm -> ../..//usr/bin/udevadm
~~~

**No udevadm in broken initramfs.**

haraldh:
why the "-p" option?

rmetrich:
The `-p` will disable executing BASH_ENV at the time `dracut` is called (through shebang).

haraldh:
@centos-ci ok to test

haraldh:
@centos-ci retest this please

PR #367, opened 2018-01-18
￼ Remove obsolete prelinking functionality

danimo:

PR #414, opened 2018-06-14
fips: Drop obsolete dependencies

WOnder93:
Explicitly require libkcapi-hmaccalc and stop installing NSS/prelink
dependencies that are no longer needed.

Cc: @t8m


WOnder93:
Note: this overlaps with #367 (I removed prelink, too, since it is also useless nad obsolete), but that should be easy to resolve when rebasing (I suggest merging #367 first).

haraldh:
merged manually

PR #418, opened 2018-06-22
Include virtio DRM drivers in hostonly initramfs

AdamWill:
The code in 50drm which tries to include all DRM drivers for
hardware attached to the system did not look for virtio devices.
So if the system is a VM using the 'virtio' graphics adapter,
the 'virtio-gpu' module which should be included is not. This
extends the code to also look for virtio devices.

https://bugzilla.redhat.com/show_bug.cgi?id=1593028

### DIFF
--- a/modules.d/50drm/module-setup.sh
+++ b/modules.d/50drm/module-setup.sh
@@ -31,7 +31,7 @@ installkernel() {
     # as we could e.g. be in the installer; nokmsboot boot parameter will disable
     # loading of the driver if needed
     if [[ $hostonly ]]; then
-        for i in /sys/bus/{pci/devices,soc/devices/soc?}/*/modalias; do
+        for i in /sys/bus/{pci/devices,virtio/devices,soc/devices/soc?}/*/modalias; do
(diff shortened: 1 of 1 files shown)


AdamWill:
You can see the effect of this patch in an [openQA test](https://openqa.stg.fedoraproject.org/tests/319530). [This screenshot](https://openqa.stg.fedoraproject.org/tests/319530#step/_do_install_and_reboot/27) shows the output of `lsinitrd /boot/initramfs-4*.img | grep virtio` (in the installed system chroot) after install of a recent Rawhide compose, where the initramfs was generated with dracut-047-32.git20180515.fc29, *without* this patch. The test then updates to a scratch build of dracut *with* this patch and then runs `dracut -f --hostonly` (again in the installed system chroot). [This screenshot](https://openqa.stg.fedoraproject.org/tests/319530#step/_do_install_and_reboot/32) shows the output of the same `lsinitrd` *after* that. As you can see, `virtio-gpu.ko.xz` is in the initramfs built with the patched dracut, but *not* in the initramfs built with the stock dracut.

PR #422, opened 2018-06-27
Introduce tri-state hostonly mode

ryncsn:
Now --hostonly accept a optional <mode> parameter, so we have
a tri-state hostonly mode:

    * generic (non-hostonly): by passing "--no-hostonly" or not
                              passing anything
    * sloppy: by passing "--hostonly sloppy" or just "--hostonly"
    * strict: by passing "--hostonly strict"

Sloppy mode is the original hostonly mode, the new introduced strict
mode will allow modules to ignore more drivers or do some extra job to
save memory and disk space, while making the image less portable.

Also introduced a helper function "optional_hostonly" to make it
easier for modules to leverage new hostonly mode.

To force install modules only in sloppy hostonly mode, use the form:

hostonly="$(optional_hostonly)" instmods <modules>

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -267,6 +267,18 @@ inst_fsck_help() {
     (($? != 0)) && derror $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${loginstall:+-L "$loginstall"} ${DRACUT_RESOLVE_DEPS:+-l}  ${DRACUT_FIPS_MODE:+-f} "$2" $_helper || :
 }
 
+# Use with form hostonly="$(optional_hostonly)" inst_xxxx <args>
+# If hosotnly mode is set to "strict", hostonly restrictions will still
+# be applied, else will ignore hostonly mode and try to install all
+# given modules.
+optional_hostonly() {
+    if [[ $hostonly_mode = "strict" ]]; then
+        printf -- "$hostonly"
+    else
+        printf ""
+    fi
+}
+
 mark_hostonly() {
     for i in "$@"; do
         echo "$i" >> "$initdir/lib/dracut/hostonly-files"
--- a/dracut.sh
+++ b/dracut.sh
@@ -147,6 +147,21 @@ Creates initial ramdisk images for preloading modules
   -H, --hostonly        Host-Only mode: Install only what is needed for
                         booting the local host instead of a generic host.
   -N, --no-hostonly     Disables Host-Only mode
+  --hostonly-mode <mode>
+                        Specify the hostonly mode to use. <mode> could be
+                        one of "sloppy" or "strict". "sloppy" mode is used
+                        by default.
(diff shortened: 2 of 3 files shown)


haraldh:
@centos-ci please test this

haraldh:
@centos-ci ok to test

daveyoung:
On 06/27/18 at 10:04am, Kairui Song wrote:
> Now --hostonly accept a optional <mode> parameter, so we have
> a tri-state hostonly mode:
> 
> Generic mode (non-hostonly): by passing "--no-hostonly" or not
>                              passing anything
> Sloppy hostonly mode: by passing "--hostonly sloppy" or just
>                       "--hostonly"
> Strict hostonly mode: by passing "--hostonly strict"

Don't know how to comment inline in github pull request thus replying
here.

It would be better to descripbe the modes like below:

hostonly mode:
*  disabled  (no hostonly)
*  generic   (sloppy in original way)
*  strict

dracut-init.sh
```diff
@@ -267,6 +267,26 @@ inst_fsck_help() {
...
+sloppy_install() {
+    local inst_func=$1; shift
+    is_func $inst_func || return 1
+    if [[ $hostonly = "strict" ]]; then
+        $inst_func $@
+    else
+        hostonly='' $inst_func $@
+    fi
+}
+
```
daveyoung: Having said in another comment, these two functions are not necessary.

haraldh:
@centos-ci retest this please

ryncsn:
@daveyoung OK, it's a good idea to split into two patches, and I'll remove the two helper function, then the patch will be simpler.

And for the hostonly mode:
If we have disabled / generic / strict hostonly mode, I think the word "generic" have opposite meaning of word "hostonly", and might be confusing, how do you think?

ryncsn:
I just noticed a minor backward compatibility problem: double dash is now mandatory between options and positional parameters, if --hostonly is the last option, because --hostonly accept an optional argument now.
eg. "dracut --hostonly image-name" will work just fine before, now will give an error as --hostonly will treat image-name as an argument. Now have to use "dracut --hostonly -- image-name"
Comnands like "dracut --hostonly --hostonly-cmdline image-name" will work just fine as before, as long as --hostonly is not the last option argument.
Will not effect --rebuild, the build-parameter.txt inside an image always have "--" in the command line.

haraldh:
why not add ```--hostonly-mode strict|sloppy|none``` as a backwards compatible option then?

ryncsn:
@haraldh Ok, good idea, I though overload --hostonly option could keep the command line simple, and then noticed there could be some compatible problem. I'll update the patch.
But I think we don't need `--hostonly-mode none` right? That seems equivalent to `--no-hostonly`.

haraldh:
yeah, maybe no ```--hostonly-mode none```

ryncsn:
Hi, I've updated the patch, it looks good now.

PR #427, opened 2018-07-10
Add a module to support building a space saving initramfs leveraging squashfs

ryncsn:
Add support for building a squashed initramfs

With all files stored in ramfs, and most of them are not compressed,
the initramfs will take up a lot of memory. This becomes much worse
with larger page size, as page cache will allocate a minimum of one
page to cache the content of one file however small it is.

One approach to reducing the memory usage and shrink initramfs size
is to put most files into a read-only squash image and keep a minimum
set of executable and libraries outside as the loader for the squash
image. Once the squash image is mounted, the real init inside will be
executed and then everything behaves as usual.

This patch will introduce a 99squash module which will never be included
by default. User can force add it, and if it is included, dracut
will perform some extra steps before creating the final image.

For now, /etc and /usr will be moved into the squashfs image. /init
will be renamed to /init.orig and replaced by /init.squash.
Files and folders need to be accessible before mounting the image will
be kept at their original place, not moving into the squashfs. A symlink
will be created inside the squashfs instead, pointing to the file
location at boot time, so no redundant binary or libraries are added.

/init.squash is now the init loader of the initramfs. It will do some
mounting and bind-mounting job, and call the original /init.orig,
then everything works as usual.

This is very helpful when mem resource is very limited, like Kdump.
According to my tests, with initramfs built with mkdumprd (which is
already a smaller image than usual), this can save about 40MB of
memory with 64K page size, or about 20MB with 4K page size.

Won't change any behavior if 99squash module is not enabled.

This pull request also contains some related commits for this new module to work properly.

### DIFF
--- a/README.modules
+++ b/README.modules
@@ -70,43 +70,42 @@ HOOKS
 
 init has the following hook points to inject scripts:
 
-/lib/dracut/hooks/cmdline/*.sh
+/var/lib/dracut/hooks/cmdline/*.sh
    scripts for command line parsing
 
-/lib/dracut/hooks/pre-udev/*.sh
+/var/lib/dracut/hooks/pre-udev/*.sh
    scripts to run before udev is started
 
-/lib/dracut/hooks/pre-trigger/*.sh
+/var/lib/dracut/hooks/pre-trigger/*.sh
    scripts to run before the main udev trigger is pulled
 
-/lib/dracut/hooks/initqueue/*.sh
+/var/lib/dracut/hooks/initqueue/*.sh
    runs in parallel to the udev trigger
    Udev events can add scripts here with /sbin/initqueue.
    If /sbin/initqueue is called with the "--onetime" option, the script
    will be removed after it was run.
-   If /lib/dracut/hooks/initqueue/work is created and udev >= 143 then
+   If /var/lib/dracut/hooks/initqueue/work is created and udev >= 143 then
    this loop can process the jobs in parallel to the udevtrigger.
    If the udev queue is empty and no root device is found or no root
    filesystem was mounted, the user will be dropped to a shell after
    a timeout.
    Scripts can remove themselves from the initqueue by "rm $job".
 
-/lib/dracut/hooks/pre-mount/*.sh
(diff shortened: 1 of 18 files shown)


daveyoung:
There are something need consider to have it in kernel,  for example need overlay the writable part to tmpfs/ramfs,  I thought about this before, but I did not get a idea which is good enough yet :)

haraldh:
one thing before I merge, please change
```/var/dracut/lib``` to ```/var/lib/dracut```
and force push.

Thanks!

ryncsn:
@haraldh Fixed, thanks for pointing out the problem

haraldh:
One more thing: ```required_libs``` and ```resolve_binary()``` seems unnecessary, if ```dracut-install``` is used, because it already does the same thing.

ryncsn:
Hi, new introduced `required_libs` and `resolve_binary()` are not used to install things, they are used to identify which file need to be out of the squash image so that initramfs can boot.

The process of the image build with squash module is:
1. every module and dracut do their installation job as usual
2. Before the final stage (early microcode, generate cpio), squash module will move everything in a squash-workspace folder
3. squash module finds out basic binaries and modules required to boot, then move them out of the squash-workspace folder. And symlinks will be created inside the workspace folder take place of moved out files, to even further reduce the image size.
4. squash module builds the squash image which contains the squash-workspace folder's content, and replace origin init script with init.squash, which combined with the basic binary and modules moved out of the squash-workspace before, they will perform the squash mounting and other init jobs when booting.
5. dracut pack everything into initramfs as usual.

The two new introduced `required_libs` and `resolve_binary()` are used in step 3 to find out the libraries required by binaries needed to boot.

ryncsn:
@haraldh Any update or comments?

After second thought, I think some logic of this patch is overcomplicated, will send a v2 which should be much simpler and using overlayfs, having better compatibility and maintainability, with a slight higher memory usage during initrd boot.

Will close this PR later if v2 goes successfully.

ryncsn:
Duplicated by #463

PR #430, opened 2018-07-24
dracut.spec: reintroduce 51-dracut-rescue-postinst.sh script

martinezjavier:
Commit 0bb9a683d4d ("spec: drop support for legacy distributions") dropped
the 51-dracut-rescue-postinst.sh script, but this is used by Fedora during
installation to create the kernel and initramfs images and on a new kernel
install if these images don't exist, when using grubby (which seems to be
the case on a Fedora system upgrade).

So this commit partially reverts 0bb9a683d4d, to bring the script again.

Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

### DIFF
--- a/51-dracut-rescue-postinst.sh
+++ b/51-dracut-rescue-postinst.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+export LANG=C
+
+KERNEL_VERSION="$1"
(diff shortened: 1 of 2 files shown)


martinezjavier:
closing it since @haraldh already pushed the same change.

PR #438, opened 2018-08-03
Add gpio and pinctrl drivers for arm*/aarch64

kwizart:
This is needed since few gpio/pinctrl can be built as modules and are
useful on early boot.

One example is jetson-tx1 where sata and external mmc can work only
after loading pinctrl-max77620 and gpio-max77620 modules.

Having theses kind of drivers bundled into the initramfs will also
avoid some deferred probes.


kwizart:
V2: add pinctrl for all arches

Issue #436, opened 2018-07-31
IMA policy fails to load if policy includes SELinux labels


acs3ss:
IMA policies can be customized to include SELinux labels for a more fine-tuned policy. According to the [IMA wiki](https://sourceforge.net/p/linux-ima/wiki/Home/#defining-an-lsm-specific-policy), this means that SELinux needs to be loaded prior to IMA in the boot process. Except that dracut's SELinux module has a big fat `FIXME` on top;

```bash
# FIXME: load selinux policy.  this should really be done after we switchroot
```

Basically, I'd like to be able to disable dracut's attempts at loading the policy, while still having it load the keys and enable EVM. I propose adding a new variable to the IMA config file which specifies whether the policy should be loaded (the example below arbitrarily uses IMAPOLICYLOAD). Default behavior would be loading as we do now.

## Current script

```bash
IMASECDIR="${SECURITYFSDIR}/ima"
IMACONFIG="${NEWROOT}/etc/sysconfig/ima"
IMAPOLICY="/etc/sysconfig/ima-policy"

load_ima_policy()
{
    # check kernel support for IMA
    if [ ! -e "${IMASECDIR}" ]; then
        if [ "${RD_DEBUG}" = "yes" ]; then
            info "integrity: IMA kernel support is disabled"
        fi
        return 0
    fi

    # override the default configuration
    [ -f "${IMACONFIG}" ] && \
        . ${IMACONFIG}

    # set the IMA policy path name
    IMAPOLICYPATH="${NEWROOT}${IMAPOLICY}"

    # check the existence of the IMA policy file
    [ -f "${IMAPOLICYPATH}" ] && {
        info "Loading the provided IMA custom policy";
        printf '%s' "${IMAPOLICYPATH}" > ${IMASECDIR}/policy || \
            cat "${IMAPOLICYPATH}" > ${IMASECDIR}/policy
    }

    return 0
}

load_ima_policy
```

## Proposed changes

```bash
IMASECDIR="${SECURITYFSDIR}/ima"
IMACONFIG="${NEWROOT}/etc/sysconfig/ima"
IMAPOLICY="/etc/sysconfig/ima-policy"

load_ima_policy()
{
    # check kernel support for IMA
    if [ ! -e "${IMASECDIR}" ]; then
        if [ "${RD_DEBUG}" = "yes" ]; then
            info "integrity: IMA kernel support is disabled"
        fi
        return 0
    fi

    # override the default configuration
[...]

Issue #435, opened 2018-07-31
Failure to load .ima keyring if trusted key is not embedded into kernel


acs3ss:
There's a bug I've noticed with how dracut loads IMA keys in certain kernels.

IMA keys can be loaded onto either the _ima or .ima keyring, depending on how the kernel is configured. The _ima keyring can be created by a root user (or by dracut) and keys can be uploaded without any special credentials. Meanwhile, if the kernel option CONFIG_IMA_TRUSTED_KEYRING is enabled, the system automatically creates the trusted .ima keyring. A key needs to be signed by a key on the .buitlin_trusted_keys keyring or, on older kernels, to be signed by a key on the system keyring, in order to be imported onto the .ima keyring. This requires first recompiling the kernel with an embedded local IMA CA X509 certificate, which is used to verify other keys added to the .ima trusted keyring. 

The way we currently determine which keyring to load is pasted below, from the [IMA policy loading script of the integrity module](https://github.com/dracutdevs/dracut/blob/master/modules.d/98integrity/ima-policy-load.sh). 

```bash
# get the IMA keyring id
line=$(keyctl describe %keyring:.ima)
if [ $? -eq 0 ]; then
    _ima_id=${line%%:*}
else
    _ima_id=`keyctl search @u keyring _ima`
    if [ -z "${_ima_id}" ]; then
        _ima_id=`keyctl newring _ima @u`
    fi
fi
```

The problem here is that IMA key loading will fail on kernels with CONFIG_IMA_TRUSTED_KEYRING enabled that haven't rebuilt their kernel to embed an X509 cert. The .ima keyring will exist, as specified by the kernel option, so `keyctl describe %keyring:.ima` will succeed and successfully snag the ID using `_ima_id=${line%%:*}`. But the subsequent attempts to load keys onto this keyring will fail, since the .ima keyring won't allow any of these keys to be loaded onto it without the embedded X509 cert. 

This means that, in order to enable IMA on a system with the CONFIG_IMA_TRUSTED_KEYRING option, you are REQUIRED to rebuild your kernel, either to 1) disable this option, or 2) embed an X509 cert. I argue that there are numerous use cases in which someone may want to implement IMA on a system without recompiling their kernel, which is currently not possible using dracut.

I propose an edit similar to the following:

```bash
load_x509_keys() {

...

    # listing the keyring prints out 'keyring is empty' if no keys have been loaded
[...]

PR #439, opened 2018-08-06
IMA policy and key loading improvements

acs3ss:
This pull request changes two things about how dracut loads IMA policies and keys. See #435 and #436 

- Attempt to load to the _ima keyring when loading to the .ima keying fails. I added a return value to the function which actually loads keys onto an IMA keyring. If you try and fail to load the keys to the keyring (almost certainly due to a permission error arising from not rebuilding your kernel with an IMA certificate), try loading to a different keyring
- Allows for disabling of IMA policy using the IMAPOLICYLOAD keyword, which can be specified in the IMA config file that dracut sources. There is a need for the ability to disable policy loading, since if the IMA policy contains SELinux labels, the policy can't be loaded by IMA.

### DIFF
--- a/modules.d/98integrity/ima-keys-load.sh
+++ b/modules.d/98integrity/ima-keys-load.sh
@@ -17,7 +17,7 @@ load_x509_keys()
         IMAKEYSDIR="/etc/keys/ima"
     fi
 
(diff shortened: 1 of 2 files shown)


haraldh:
obsolete?

PR #441, opened 2018-08-09
dracut-functions: fix the word splitting

lnykryn:
I have no idea why, but on Fedora the original code works just fine, but on rhel7 the loop is only executed for the first line. The double apostrophe seems to fix it.

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -683,7 +683,7 @@ get_loaded_kernel_modules ()
     local modules=( )
     while read _module _size _used _used_by; do
         modules+=( "$_module" )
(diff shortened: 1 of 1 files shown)


ryncsn:
Hi, Lukáš, I just created another pull request #440, and the problem is bash not handling "here string" properly in before version 4.4
I'll close my pr as this commit seems have been merged downstream, thanks for the fast downstream fix :+1:

lnykryn:
Just for reference, this is a bug in old version of bash
https://bugzilla.redhat.com/show_bug.cgi?id=1514300

PR #442, opened 2018-08-15
Add module for persistent memory support

ryncsn:
pmem devices require some time to start up, and need to make sure nvdimm
related drivers are included. So introduce a new module to load the
drivers and pull in initqueue to wait for pmem devices.

### DIFF
--- a/modules.d/90pmem/module-setup.sh
+++ b/modules.d/90pmem/module-setup.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    is_pmem() {
+        local _dev=$1
+
+        [[ -L "/sys/dev/block/$_dev" ]] || return
+        cd "$(readlink -f "/sys/dev/block/$_dev")"
+        until [[ -d sys || -f devtype ]]; do
+            cd ..
+        done
+        [[ -f devtype ]] && [[ "$(cat devtype)" == nd_* ]]
+    }
+
+    [[ $hostonly ]] || [[ $mount_needs ]] && {
+        pushd . >/dev/null
+        for_each_host_dev_and_slaves is_pmem || return 255
+        popd >/dev/null
+    }
+
+    return 0
+}
+
+# called by dracut
+depends() {
+    return 0
+}
+
+# called by dracut
+installkernel() {
+    instmods nfit =drivers/nvdimm =drivers/dax
+}
+
+# called by dracut
+install() {
+    dracut_need_initqueue
+}


modules.d/90pmem/module-setup.sh
```diff
@@ -0,0 +1,38 @@
...
+}
+
+# called by dracut
+installkernel() {
+    instmods nfit =drivers/nvdimm =drivers/dax
+}
+
+# called by dracut
+install() {
+    dracut_need_initqueue
```
daveyoung: Kdump use --device option to add devices dependencies.   In theory wait_for_dev should ensure initqueue been used, no?

daveyoung: This should be similar with the problem Lianbo was trying to fix in below request:
https://github.com/dracutdevs/dracut/pull/443

haraldh:
Shouldn't this be added to 90kernel-modules??
```
instmods nfit =drivers/nvdimm =drivers/dax
```

Why is the ```dracut_need_initqueue``` needed?

Normally ```systemd``` waits for the device to appear via ```udev``` and module autoloading.

Shouldn't they just behave like all the other devices?

ryncsn:
> Shouldn't this be added to 90kernel-modules??
> 
> instmods nfit =drivers/nvdimm =drivers/dax

Yes indeed, will move to 90kernel-modules.

> Why is the dracut_need_initqueue needed?
> 
> Normally systemd waits for the device to appear via udev and module autoloading.
> 
> Shouldn't they just behave like all the other devices?

While testing pmem devices with kdump, I found they are not working properly.
pmem modules are loaded in multiple stages, udev have to be settled before the later services start, else some required pmem device may still not showed up after module loaded. To make things clear I just added dracut_need_initqueue to make sure all devices are present, and it works very well.

haraldh:
```While testing pmem devices with kdump```
Shouldn't kdump then ensure  ```dracut_need_initqueue``` ?

**Comment by ryncsn** [comment]
@haraldh, good suggestion. This may also help some other kdump issue. Could there be any unexpected side effect if we always ensure `dracut_need_initqueue`? Slower boot progress?

**Comment by haraldh** [comment]
Well, ```dracut_need_initqueue``` basically means, "wait for udev to be settled" and/or also "I have jobs scheduled in the initqueue". In your kdump case, since you don't boot and really want your kdump actions to happen in the initqueue, it has to be set.

**Comment by haraldh** [comment]
In the normal boot process, as soon as the real root device appears, there is normally no need to wait any longer. So, if your real root is on a pmem device, the boot should automatically wait for the pmem device... initqueue or not.

So, can you elaborate again, which problems you see without ```dracut_need_initqueue```? Is it for a normal boot, or is it in case the kernel crashed and you want to kdump?

**Comment by haraldh** [comment]
```Would it be good to include initqueue when some extra devices are being included?```
Absolutely!

**Comment by haraldh** [comment]
But I thought the whole kdump action is happening in the initqueue, so it should be enabled anyway???

**Comment by haraldh** [comment]
maybe the kdump systemd service should have
```
Requires=dracut-initqueue.service
After=dracut-initqueue.service
```
then?

ryncsn:
> maybe the kdump systemd service should have
> 
> Requires=dracut-initqueue.service
> After=dracut-initqueue.service
> then?

That requires is not needed, dracut-initqueue.service states `ConditionPathExists=|/lib/dracut/need-initqueue`, so dracut_need_initquque is required to be called at first place, and if not, services depending on it won't start it, and won't fail either.

Thanks for the review, really helpful, closing this PR and will open a new one.

PR #445, opened 2018-08-17
enable journal logging to kernel ring buffer

dustymabe:
This will enable early log messages from systemd service to go
to the kernel buffer which means they should be seen on various
terminals. Even if a user is not looking at the correct primary
console of the machine they should still get the error messages
because now they would be going to all console= device and not
just the primary one.

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -228,12 +228,22 @@ install() {
 
     mkdir -p "$initdir/etc/systemd"
     # We must use a volatile journal, and we don't want rate-limiting
+    # and want early startup messages to go everywhere kernel buffer does
     {
         echo "[Journal]"
         echo "Storage=volatile"
         echo "RateLimitInterval=0"
         echo "RateLimitBurst=0"
+        echo "ForwardToConsole=yes"
+        echo "ForwardToKMsg=yes"
+        echo "MaxLevelKMsg=info"
     } >> "$initdir/etc/systemd/journald.conf"
+
+    mkdir -p "$initdir/etc/sysctl.d"
+    # Also disable rate-limiting at kernel ring buffer level
+    {
+        echo "kernel.printk_devkmsg = on"
+    } >> "$initdir/etc/sysctl.d/10-dont-ratelimit-kmsg.conf"
 
     ln_r "${systemdsystemunitdir}/multi-user.target" "${systemdsystemunitdir}/default.target"
 }


modules.d/00systemd/module-setup.sh
```diff
@@ -228,13 +228,23 @@ install() {
...
+        echo "ForwardToConsole=yes"
+        echo "ForwardToKMsg=yes"
+        echo "MaxLevelKMsg=info"
     } >> "$initdir/etc/systemd/journald.conf"
 
+    mkdir -p "$initdir/etc/sysctl.d"
+    # Also disable rate-limiting at kernel ring buffer level
+    {
+        echo "kernel.printk_devkmsg = on"
+    } >> "$initdir/etc/sysctl.d/10-dont-ratelimit-kmsg.conf"
```
dustymabe: the biggest problem here is that the default for this setting (`ratelimit`) will not get restored after this has run. I don't know how to resolve that. Really I only want to change it to `on` during initramfs and after that drop it back to whatever it was set to before (either the default  or the value from the kernel command line).

dustymabe:
related discussion: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/OI7BOWZSXDJM5W2KVDK3BBEN6S7ASLZC/

haraldh:
have you tried: ```loglevel=7 rd.systemd.log_level=debug rd.systemd.log_target= kmsg ``` ??

dustymabe:
bump. would like to at least have a discussion about this.

haraldh:
ping  @poettering

poettering:
I am not sure I follow. So far we shied away from logging to kmsg except when we have no other place to log to (because the kernel folks are pretty nasty about it, see LWN). Changing the default in Fedora to log to kmsg automatically certainly has the power to piss of those people again, and quite frankly I am a bit tired of dealing with those folks. 

I am not entirely sure why this is needed though. AFAICS the main change from this would be that people could type "dmesg" rather than "journalctl" to get all system logs during the initrd, but I am not sure that is such a great benefit, no?

What am I missing?

modules.d/00systemd/module-setup.sh
ferdnyc: If `printk.devkmsg=` is set from the kernel command line, this won't even _work_, at least according to [the docs](https://www.kernel.org/doc/Documentation/sysctl/kernel.txt): 

```
printk_devkmsg:

Control the logging to /dev/kmsg from userspace:

ratelimit: default, ratelimited
on: unlimited logging to /dev/kmsg from userspace
off: logging to /dev/kmsg disabled

The kernel command line parameter printk.devkmsg= overrides this and is
a one-time setting until next reboot: once set, it cannot be changed by
this sysctl interface anymore.
```

If you set it via `sysctl`, though, it shouldn't be any problem setting it back to `ratelimit` again once journald is online. The lockout only seems to apply when it's set from the kernel command line. (At least, I just flipped it between `on` and `ratelimit` three times on my F28 box using `sysctl`, with no apparent problems.)

dustymabe: ahh, i think that's actually good news. IOW if a user sets it from the CLI then we don't have to worry about stepping on them. so if we set it and then set it back to the default after switch root then we'd be fine

poettering:
> understood. does it help that this would only have effect for the journal during the initrd. After switch root the settings from the jourald.conf from the root filesystem would take over.

well, but that's still a lot of logging in the usual case. i mean the main part of its logging systemd does during boot…

> that depends. a lot of times people might not have actual access to the correct console device that is configured in the image so they aren't able to "type" `dmesg` or `journalctl`. All they know is their system didn't boot up and they don't know why. They might have access to one console output and it could be the wrong console device (i.e. not the last `console=` entry on the kernel command line) or it may be read-only (like grabbing the console logs from AWS). with this change they would get that critical startup information on every `console=` entry from the kernel command line.

Wouldn't a better approach be to fix /dev/console in the kernel to multiplex to all console= exactly like kmsg does it?

Quite frankly, if this is just about wanting multiplexing to multiple consoles then kmsg sounds like an awfully big hammer to me.

I am pretty sure that in the generic versions of Fedora this really shouldn't be turned on. Maybe turning this on as proposed in the cloud edition of Fedora might be OK, but it still appears to be a hack to me, and a proper fix would be to add multiplexing to /dev/console inside the kernel as suggested...

dustymabe:
> Wouldn't a better approach be to fix /dev/console in the kernel to multiplex to all console= exactly like kmsg does it?

yes. That is exactly what I want, but I don't know how to make that happen or even if it would be accepted.

dustymabe:
bump. I hit this again today. One example is where you are trying to build a disk using the same kickstart for more than one architecture so you add a bunch of `console=` lines to the kernel command line: https://pagure.io/fedora-kickstarts/blob/4665f812c6f72709b83e961defab18e3d8da7ab4/f/fedora-atomic.ks#_36

any idea how we could make the *fix /dev/console in the kernel to multiplex to all console= exactly like kmsg does it* approach happen?

poettering:
> understood. does it help that this would only have effect for the journal during the initrd. After switch root the settings from the jourald.conf from the root filesystem would take over.

not really, it's the (early) boot-time message those kernel folks had issues with, it's anyway where 95% of our log messages are generated, hence saying "only during early boot" is not that great...

dustymabe:
> > any idea how we could make the _fix /dev/console in the kernel to multiplex to all console= exactly like kmsg does it_ approach happen?
> 
> prep a kernel patch for that? of course this would need to be opt-in (probably compile time), for compat reasons, but we'd turn this on in fedora kernels by default.

FYI opened [discussion on kernel mailing list](https://lkml.org/lkml/2018/11/5/80) to see if this is a patch that would be accepted.

haraldh:
@dustymabe any progress?

PR #446, opened 2018-08-20
Rename dm_* to dm-* as dm_* has apparently been renamed.

TheReverend403:


haraldh:
```_``` and ```-``` do not matter.

```bash
$ modprobe kvm-intel && echo OK
OK
$ modprobe kvm_intel && echo OK
OK
```

Issue #433, opened 2018-07-28
Live root filesystem on squashfs


GigabyteProductions:
Before support for overlayfs existed in dracut, it made sense to require that squashfs live images simply contain a image of a writeable root file system (for use with block device overlays).

However, now that read-only root file systems can be made writeable by overlaying a tmpfs using overlayfs, it would make more sense to allow for the root file system to be provided directly by the downloaded squashfs image.

As a workaround, I am currently booting with a squashfs-in-squashfs image to satisfy the inner image search requirement.

To clarity, specifically what I am referring to is the following currently required image layout (from [dracut.cmdline(7)](https://github.com/dracutdevs/dracut/blob/bc01f822de732d85d4bfc77bae44c6bd4abe43d7/dracut.cmdline.7.asc#booting-live-images)):
```
               squashfs.img          |  Squashfs from LiveCD .iso downloaded via network
                  !(mount)
                  /LiveOS
                      |- rootfs.img  |  Filesystem image to mount read-only
                           !(mount)
                           /bin      |  Live filesystem
                           /boot     |
                           /dev      |
                           ...       |
```
I would instead like to use:
```
               squashfs.img          |  Squashfs from LiveCD .iso downloaded via network
                  !(mount)
                  /bin               |  Live filesystem
                  /boot              |
                  /dev               |
                  ...                |
```

alkisg:
[Linux Terminal Server Project](http://www.ltsp.org) would appreciate this as well.

A squashfs image is served over NBD, and by using root=live:nbd:server:export, we would be able to netboot it without having to add custom code in the initramfs.

dtardon:
#447 claims to have fixed this -> close?

johannbg:
Closing as this issue seems resolved. Thank you for your contributions.

GigabyteProductions:
Haven't utilized the feature yet. Will test, soon.

GigabyteProductions:
Yep, works. Note: Make sure /proc exists in the image, so this test passes: https://github.com/FGrose/dracut/blob/6734a42405a22d550224002045c69852f1966207/modules.d/90dmsquash-live/dmsquash-live-root.sh#L298

That bit me in running the following command:
```
sudo mksquashfs "$root" "$image" -e dev -e proc -e sys
```

instead of this one:
```
sudo mksquashfs "$root" "$image" -wildcards -e 'dev/*' -e 'proc/*' -e 'sys/*'
```

Issue #307, opened 2017-11-24
dmsquash-live-root only squashfs.img root?


kyknos123:
hello a question, how is be played change parameter someone dmsquash-live-root script without /LiveOS/rootfs.img, I want to make my custom fedora-centos but using compressed only squashfs.img, something like that filesystem.squashfs for ubuntu, it is possible?

FGrose:
@kyknos123:   Does pull request #447 address your request?

johannbg:
This issue is being marked as stale because it has not had any recent activity. It will be closed if no further activity occurs. If this is still an issue in the latest release of Dracut and you would like to keep it open please comment on this issue within the next 7 days. Thank you for your contributions.

PR #447, opened 2018-08-21
dmsquash-live: Support a flattened squashfs.img 

FGrose:
~~~~txt
A simplified root filesystem structure may be provided for OverlayFS
overlays by squashing the root filesystem directly instead of squashing
an embedded image file at /LiveOS/rootfs.img.  Detect and configure
such a squashed root filesystem for live booting.

For OverlayFS boots, avoid the read-only Device-mapper linear device
  at /dev/mapper/live-base.
Create a consistent device link at /dev/live-base for the read-only
  base loop device for all overlayed live root filesystems.
Consistently provide a link at /dev/root for wait_for_dev.

Update documentation.

Also,
dmsquash-live, livenet: Simplify OverlayFS read-only overlay setup.

Use multiple lower layer directories in a single OverlayFS mount with
  a transient overlay directory.
Tolerate a command line with rd.live.overlay.readonly and NO persistent
  overlay by reconfiguring the OverlayFS mount options.
Use more compatible shell syntax for testing symlinks, and use printf
  instead of echo -e.
~~~~
This satisfies the request in issue #433 and possibly the question in issue #307.

### DIFF
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -838,23 +838,24 @@ Booting live images
 Dracut offers multiple options for live booted images:
 
 =====================
-SquashFS with read-only filesystem image::: The system will boot with a read
-only filesystem from the SquashFS and apply a writable device-mapper snapshot
-over the read only filesystem.  Using this method ensures a relatively fast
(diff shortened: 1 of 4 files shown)


haraldh:
@centos-ci ok to test

PR #448, opened 2018-08-21
Make dracut work well with pmem device

ryncsn:
Include "nfit" module so that pmem device could work.


haraldh:
the second patch should be a separate PR.

ryncsn:
Done, move that commit to pr #449

PR #451, opened 2018-08-23
dracut-install: fix kernel module lookup path in "updates" "extra" dir

loyfan:
Kernel modules in “kernel” directory are stored in subdirectories like "drivers/block";
but in "updates" and "extra" directory, it's stored directly in the root directory.

The original implementation will only include kernel modules in “kernel” directory to initramfs, and won't include kernel modules in  "updates" and "extra".

### DIFF
--- a/install/dracut-install.c
+++ b/install/dracut-install.c
@@ -1532,13 +1532,13 @@ static int install_modules(int argc, char **argv)
                                 exit(EXIT_FAILURE);
                         }
 
-                        r = asprintf(&path1, "%s/extra/%s", kerneldir, &argv[i][1]);
+                        r = asprintf(&path1, "%s/extra", kerneldir);
                         if (r < 0) {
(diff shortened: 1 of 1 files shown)


haraldh:
@centos-ci  ok to test

haraldh:
@centos-ci retest this please

esyr-rh:
I presume that commit 290df2e1dee3724f0a7564a825be774f36584b94 should fix the problem (in case depmod is configured to use kernel modules from "extra"/"updates" as well).

loyfan:
@esyr-rh thanks
tested 290df2e with master branch, problem resolved.

And by the way, it's dracut-044-78.fc25 where I found this bug.
similar with https://bugzilla.redhat.com/show_bug.cgi?id=1413574, my raid server can't boot after an iso installation.

PR #454, opened 2018-09-03
RFC: Add NetworkManager support

lkundrak:
Let's see/discuss how would this work.

Depends on a branch of NetworkManager that is not yet merged:
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commits/lr/initrd

COPR with patched NM here: https://copr.fedorainfracloud.org/coprs/lkundrak/NetworkManager-initramfs/


haraldh:
merged manually

haraldh:
build for rawhide https://koji.fedoraproject.org/koji/taskinfo?taskID=29793164

PR #458, opened 2018-09-07
modules.d: Add a module for handling additional depmod kernel module directories

esyr-rh:
It parses depmod configuration and scans modules.dep for kernel modules
present in directories supplied in "overrides", "external", and "search"
depmod configuration options.  The resulting list of (absolute) kernel
module paths is then supplied to instmods.

* modules.d/90kernel-modules-extra/module-setup.sh: New file.

Signed-off-by: Eugene Syromiatnikov <esyr@redhat.com>

### DIFF
--- a/dracut.spec
+++ b/dracut.spec
@@ -340,6 +340,7 @@ install -m 0755 51-dracut-rescue-postinst.sh $RPM_BUILD_ROOT%{_sysconfdir}/kerne
 %{dracutlibdir}/modules.d/90dm
 %{dracutlibdir}/modules.d/90dmraid
 %{dracutlibdir}/modules.d/90kernel-modules
+%{dracutlibdir}/modules.d/90kernel-modules-extra
 %{dracutlibdir}/modules.d/90lvm
 %{dracutlibdir}/modules.d/90mdraid
 %{dracutlibdir}/modules.d/90multipath
--- a/modules.d/90kernel-modules-extra/module-setup.sh
+++ b/modules.d/90kernel-modules-extra/module-setup.sh
@@ -1,0 +1,194 @@
+#!/bin/bash
+
+# called by dracut
+#
+# Parses depmod configuration and calls instmods for out-of-tree kernel
+# modules found.  Specifically, kernel modules inside directories that
+# come from the following places are included (if these kernel modules
+# are present in modules.dep):
+#   - "search" configuration option;
+#   - "override" configuration option (matching an exact file name constructed
+#      by concatenating the provided directory and the kernel module name);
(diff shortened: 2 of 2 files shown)


haraldh:
Isn't that already happening with dracut-install using libkmod?

esyr-rh:
Well, it indeed uses it, but not depmod, that has a separate configuration and logic related to module sources processing, see [1][2].

[1] https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/tools/depmod.c#n812
[2] https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/tools/depmod.c#n599

haraldh:
Why not extend kmod? You would want it for module loading also... right?

esyr-rh:
Uh, depmod is a part of kmod project already. And making depmod part of libkmod is pretty significant change, not that it wouldn't be useful, however.

haraldh:
I mean,
```dracut-install``` using libkmod should have the same behavior as ```modprobe```.
If ```dracut-install``` does not behave like it should, maybe libkmod should be fixed then.

But maybe I don't understand the initial problem.

esyr-rh:
Well, if that's the idea, then the current behaviour (together with its usage in modules) differs pretty noticeable, as significant part of kernel modules is installed into initramfs via "=dir" syntax, and that one just ignores all the sources and configurations that are checked by kmod_module_new_from_lookup[1] (primarily, modules.dep) and uses kmod_module_new_from_path instead. As a result, not the kernel modules are used by the running system are installed, but rather those that found under specific paths in kernel/updates/extra directories. Those modules that are installed by kmod name, will be picked up in accordance with the configuration, most of the device drivers — not[2][3][4].

[1] https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/libkmod/libkmod-module.c#n558
[2] https://github.com/dracutdevs/dracut/blob/master/modules.d/90kernel-modules/module-setup.sh#L75
[3] https://github.com/dracutdevs/dracut/blob/master/modules.d/90multipath/module-setup.sh#L63
[4] https://github.com/dracutdevs/dracut/blob/master/modules.d/90kernel-network-modules/module-setup.sh#L24

esyr-rh:
One possibility is to check with kmod_module_new_from_lookup each module name discovered with directory traversal, but, again, out-of-tree modules that do not have inbox counterparts will not be covered, for example. Probably, the proper way is to specify aliases and used symbols, but that may be too fragile and requires significant effort for converting all the existing dracut modules. That's why I settled with just adding all the out-of-tree modules in the end.

PR #463, opened 2018-09-12
Add support for building a squashed initramfs

ryncsn:
With all files stored in ramfs, and most of them are not compressed,
the initramfs will take up a lot of memory. Besides, if the file number
is large, each file will waste some memory due to page fragmetation.
This is due to ramfs' design, at least one page will be allocated for
one file however small the file is. On machine with large page size,
this will become worse and waste too many memory.

One approach to reducing the memory usage is to reduce the number of
files that got directly loaded into the root ramfs, and compress files
by put most files will into a read-only squash image and keep a minimum
set of executable and libraries outside as the loader for the squash
image. After the squash image is mounted, the real 'init' will be
executed and then everything behaves as usual.

This patch will introduce a '99squash' module which will never be
included by default. User can force add it, and if it is included,
dracut will perform some extra steps before creating the final image:

For now, "/etc" and "/usr" will be moved into the squashfs image.
"/init" will be renamed to "/init.stock" and replaced by "/init.squash".
Files and folders need to be accessible before mounting the image will
be still avaliable at their original place. And due to squashfs is
readonly, an overlayfs layer will be created on top of squashfs mount
point, as many dracut module require readwrite access to "/etc" and
"/usr", "init.squash" will ultimately call "/init.stock".

An extra systemd service will be installed. This service will umount all
squashfs related mount points right before switch-root to release
resources properly. This service will not actually do anything if
switch-root is not used.

This is very helpful when mem resource is very limited, like Kdump.
According to my tests, this squash module can help save about 35MB of
memory with 64K page size, or about 15MB with 4K page size on an
ordinary kdump capture routine. This module could also help reduce
memory usage for normal boot up process.

Won't change any behavior if squash module is not enabled.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1747,6 +1747,120 @@ fi
 
 dinfo "*** Creating image file '$outfile' ***"
 
+if dracut_module_included "squash"; then
+    if ! check_kernel_config CONFIG_SQUASHFS; then
+        dfatal "CONFIG_SQUASHFS have to be enabled for dracut squash module to work"
+        exit 1
+    fi
+    if ! check_kernel_config CONFIG_OVERLAY_FS; then
+        dfatal "CONFIG_OVERLAY_FS have to be enabled for dracut squash module to work"
+        exit 1
+    fi
+    if ! check_kernel_config CONFIG_DEVTMPFS; then
+        dfatal "CONFIG_DEVTMPFS have to be enabled for dracut squash module to work"
+        exit 1
+    fi
+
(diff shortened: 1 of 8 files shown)


ryncsn:
I see the only CI failure is caused by newly introduced file not being packaged. As the squash module requires squash-tools to be installed, should we introduce a new subpackage like "dracut-squash"?

haraldh:
> I see the only CI failure is caused by newly introduced file not being packaged. As the squash module requires squash-tools to be installed, should we introduce a new subpackage like "dracut-squash"?

yes, makes sense... please add it to dracut.spec

PR #466, opened 2018-09-18
network: configure NetworkManager to use dhclient

lkundrak:
If the network module obtained a lease using dhclient, NetworkManager
must be configured to use it too, otherwise it would obtain a different
lease (and could potentially break a connection to the network volume).

### DIFF
--- a/modules.d/40network/ifup.sh
+++ b/modules.d/40network/ifup.sh
@@ -41,6 +41,12 @@ do_dhcp() {
         return 1
     fi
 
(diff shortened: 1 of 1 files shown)


rvykydal:
I think we could solve this by setting NM configuration for installer image in lorax, by the same mechanism as with eg https://github.com/weldr/lorax/blob/master/share/templates.d/99-generic/config_files/common/91-anaconda-autoconnect-slaves.conf

rvykydal:
Thinking of it, configuring only installer to use dhclient wouldn't address cases where network is activated (on installed system) in initramfs for boot from SAN or root on network storage in general.

PR #467, opened 2018-09-19
retry image download forever

naterini:
When we are using livenetroot image, we want dracut to try until it succeeds. Current setup will have the node fail to boot if there are transient network issues on the first failure. This patch will make it try forever which is usually long enough for transient issues to get worked out.

### DIFF
--- a/modules.d/90livenet/livenetroot.sh
+++ b/modules.d/90livenet/livenetroot.sh
@@ -30,7 +30,7 @@ do
     if [ ! -z "$imgfile" -a -s "$imgfile" ]; then
         break
     else
-        if [ $i -eq $RETRIES ]; then
+        if [ $i -ge $RETRIES ]; then
             warn "failed to download live image after $i attempts."
             exit 1
         fi
--- a/modules.d/90livenet/livenetroot.sh
+++ b/modules.d/90livenet/livenetroot.sh
@@ -6,8 +6,8 @@ type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 . /lib/url-lib.sh
 
 PATH=/usr/sbin:/usr/bin:/sbin:/bin
-RETRIES={$RETRIES:-100}
-SLEEP={$SLEEP:-5}
+RETRIES=${RETRIES:-100}
+SLEEP=${SLEEP:-5}
 
 [ -e /tmp/livenet.downloaded ] && exit 0
 
--- a/modules.d/90livenet/livenetroot.sh
+++ b/modules.d/90livenet/livenetroot.sh
@@ -18,7 +18,8 @@ info "fetching $liveurl"
 
 imgfile=
 #retry until the imgfile is populated with data or the max retries
-for (( i=1; i<=$RETRIES; i++))
+i=1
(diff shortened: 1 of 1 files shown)


haraldh:
I would try 100 times with a sleep 1 in between to save energy :smile:

naterini:
> I would try 100 times with a sleep 1 in between to save energy

Added suggested changes in dfc413f4.

modules.d/90livenet/livenetroot.sh
```diff
@@ -6,21 +6,33 @@ type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 . /lib/url-lib.sh
 
 PATH=/usr/sbin:/usr/bin:/sbin:/bin
+RETRIES={RETRIES:-100}
+SLEEP={SLEEP:-5}
```
ivy: Without a `$` won't these variables expand to a literal `{RETRIES:-100}` and `{SLEEP:-5}` and not defaults of `100` and `5`?

modules.d/90livenet/livenetroot.sh
```diff
@@ -6,21 +6,33 @@ type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
...
-
-> /tmp/livenet.downloaded
+imgfile=
+#retry until the imgfile is populated with data or the max retries
+for (( i=1; i<=$RETRIES; i++))
+do
+    imgfile=$(fetch_url "$liveurl")
+
+    if [ $? != 0 ]; then
+            warn "failed to download live image: error $?"
```
ivy: Should we `exit 1` like [L20](https://github.com/dracutdevs/dracut/pull/467/files#diff-7088b30306c7d75f033518ea411e591cL20), when we've exceeded `$RETRIES`?

ivy:
I left a couple comments. This is otherwise a change I'd love to see merged! :+1: (I currently catch everything with `rd.emergency=reboot` and could see this saving our PXE clients a lot of time/traffic).

naterini:
Committed a few fixes and added an exit 1 if all attempts fail in 9571b3a

haraldh:
now bonus points to make it POSIX shell compatible :)

```
$ dash
$ RETRIES=${RETRIES:-100}
$ echo $RETRIES 
100
$ for (( i=1; i<=$RETRIES; i++)); do echo $i;done
dash: 3: Syntax error: Bad for loop variable
```

naterini:
> now bonus points to make it POSIX shell compatible :)
> 
> ```
> $ dash
> $ RETRIES=${RETRIES:-100}
> $ echo $RETRIES 
> 100
> $ for (( i=1; i<=$RETRIES; i++)); do echo $i;done
> dash: 3: Syntax error: Bad for loop variable
> ```

Fixed in 28da5d65ded1bbba2fd7f7cbaea1440f1b8063bc

>$ dash -n modules.d/90livenet/livenetroot.sh; echo $?
>0

ivy:
@haraldh Is there anything else blocking this PR from being merged?

danimo:
What else is missing here?

PR #479, opened 2018-10-19
Two fixes related to host_only behavior

danimo:
Triggered on 390x, but the backslash read fix
is useful on all platforms.


sriemer:
Tested-by: Sebastian Parschauer (SUSE L3 agent)

PR #492, opened 2018-11-27
dracut-initramfs-restore: set selinux labels

lnykryn:


doverride:
where are these things being discussed? Why is this needed? Is "set selinux labels" enough justification to merge this?

PR #494, opened 2018-11-27
Ensure the AWS ENA driver is in the initrd for cloud images

nullr0ute:
Amazon ENA driver should be in initrd for various init bits. Ensure that's
the case for the supported x86_64 and aarch64 architectures.


nullr0ute:
cancelling for a better/fixed version

PR #495, opened 2018-11-28
Add virtio network and ena drivers for cloud images

nullr0ute:
There's some cloud use cases where we need network connectivity in cloud images
so ensure we have the two common ones, virtio_net for KVM/OpenStack/Cloud
providers and ena for Amazon AWS.

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -43,6 +43,14 @@ installkernel() {
             virtio virtio_blk virtio_ring virtio_pci virtio_scsi \
             "=drivers/pcmcia" =ide nvme vmd nfit
 
+        if [[ "$(uname -m)" == x86_64 || "$(uname -m)" == aarch64 ]]; then
+            # virtio_net driver for kvm, ENA driver for AWS
+            instmods \
+                "=drivers/net/ethernet/amazon/ena" \
+                virtio_net \
+                ${NULL}
+        fi
+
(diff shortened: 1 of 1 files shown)


haraldh:
Why not make a separate module for amazon?

nullr0ute:
So the virtio_net isn't getting pulled in then for the cloud images which is why I missed that there was even a module, any thoughts on why that would be?.

I can certainly look at doing it as a separate module. I suppose the query I would have is how it would be detected, because the cloud images aren't built on amazon so using detection via HW for first boot is difficult or maybe I've missed something.

haraldh:
@nullr0ute still need this PR?

danimo:
This PR seems to be abandoned. Closing for now.

nullr0ute:
It got lost in my inbox

nullr0ute:
> Well cloud images should really be built with a special tailored dracut setup/options and not take the Workstation/Desktop default. This will even reduce the initramfs size much more.

How do you specify that in a kickstart? It's also useful for workstation when running in VMs and it already deals with other virt modules here such as: "virtio virtio_ring virtio_pci virtio_scsi pci_hyperv"

johannbg:
@nullr0ute  Not sure what relevance kickstart has to with dracut but what I would do is create a cloud specific dracut override something like /etc/dracut.conf.d/99_aws.conf and include that as a file or use  
cat >> /etc/dracut.conf.d/99_aws.conf  << EOF
My amazon dracut cloud configuration
EOF

shebang ( which should work I think in ks )

Issue #476, opened 2018-10-17
Spawn emergency console on all console ttys


danimo:
Currently, when there is an emergency in dracut, it will only spawn the emergency console (e.g. on rd.break) on the primary tty. However, it should present a console on all  ttys.

mkinitrd uses sulogin for this. For systemd systems there is systemd-sulogin, which we would need to patch to issue a custom output.

Any comment on this before I get to it?

herrold:
It is usually useful to leave an error message on the console at which it was expressed.  As I understand it, your suggestion would lose that information when a console is spawned

This seems like this needs more consideration

danimo:
The information about the error has to be preserved as well as possible, of course.

danimo:
https://github.com/dracutdevs/dracut/pull/499 is the suggested PR, it does not yet address your concern, but it is an improvement over not being able to do anything on the wrong console.

danimo:
With https://github.com/dracutdevs/dracut/pull/499/commits/9a5cb9c5f2128ee5635a48ceb22e0d77e49d708b, we now have the emergency information on all ttys. All other relevant errors _should_ show up in the journal.

PR #499, opened 2018-12-05
emergency mode: use sulogin

danimo:
- allow emergency login on every console
  specified in the kernel cmdline
- require password for hostonly images
- emergency mode: Manually multiplex emergency infos

This will bring all vital information to all ttys specified
as console devices, regardless of wether they hold the C flag.

Reference: FATE#325386
Reference: #449


danimo:
Please consider https://github.com/dracutdevs/dracut/issues/476 before merging.

danimo:
I think this is good now. Please review & merge.

PR #500, opened 2018-12-07
net-lib: remove calls to ip with -oneline option

nolange:
busybox seems to not properly support this option,
and it can be be safely omitted or replaced
with piping the output through head.

### DIFF
--- a/modules.d/40network/net-lib.sh
+++ b/modules.d/40network/net-lib.sh
@@ -14,20 +14,20 @@ is_ip() {
 
 get_ip() {
     local iface="$1" ip=""
-    ip=$(ip -o -f inet addr show $iface)
+    ip=$(ip -f inet addr show $iface)
     ip=${ip%%/*}
     ip=${ip##* }
     echo $ip
 }
 
 iface_for_remote_addr() {
-    set -- $(ip -o route get to $1)
-    echo $5
+    set -- $(ip route get to $1 | sed  's/.*\bdev\b//p;q')
+    echo $1
 }
 
 iface_for_ip() {
-    set -- $(ip -o addr show to $1)
-    echo $2
+    set -- $(ip addr show to $1)
+    echo ${2%:}
 }
 
 iface_for_mac() {
@@ -593,7 +593,7 @@ wait_for_if_link() {
     timeout=$(($timeout*10))
 
     while [ $cnt -lt $timeout ]; do
-        li=$(ip -o link show dev $1 2>/dev/null)
+        li=$(ip link show dev $1 2>/dev/null)
         [ -n "$li" ] && return 0
         sleep 0.1
         cnt=$(($cnt+1))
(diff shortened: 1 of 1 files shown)


modules.d/40network/net-lib.sh
```diff
@@ -14,20 +14,20 @@ is_ip() {
...
     ip=${ip##* }
     echo $ip
 }
 
 iface_for_remote_addr() {
-    set -- $(ip -o route get to $1)
-    echo $5
+    set -- $(ip route get to $1 | sed -n 1p)
+    echo ${5}
 }
```
haraldh: Hmm... 
```
$ set -- $(ip -o route get to 127.0.0.1)
$ echo $5
src
```
Should be $4 ?? right?

haraldh: And I think `$()` eliminates all `\n` anyway...

modules.d/40network/net-lib.sh
```diff
@@ -14,20 +14,20 @@ is_ip() {
...
 
 iface_for_ip() {
-    set -- $(ip -o addr show to $1)
-    echo $2
+    set -- $(ip addr show to $1 | sed -n 1p)
+    echo ${2%:}
 }
```
haraldh: I think `$()` eliminates all `\n` anyway...

modules.d/40network/net-lib.sh
```diff
@@ -609,7 +609,7 @@ wait_for_if_up() {
     timeout=$(($timeout*10))
 
     while [ $cnt -lt $timeout ]; do
-        li=$(ip -o link show up dev $1)
+        li=$(ip link show up dev $1 | sed -n 1p)
         if [ -n "$li" ]; then
```
haraldh: And I think `$()` eliminates all `\n` anyway...

modules.d/40network/net-lib.sh
```diff
@@ -721,12 +721,12 @@ iface_has_carrier() {
 
     linkup "$1"
 
-    li=$(ip -o link show up dev $1)
+    li=$(ip link show up dev $1 | sed -n 1p)
```
haraldh: And I think `$()` eliminates all `\n` anyway...

modules.d/40network/net-lib.sh
```diff
@@ -721,12 +721,12 @@ iface_has_carrier() {
...
     strstr "$li" "NO-CARRIER" && _no_carrier_flag=1
 
     while [ $cnt -lt $timeout ]; do
         if [ -n "$_no_carrier_flag" ]; then
-            li=$(ip -o link show up dev $1)
+            li=$(ip link show up dev $1 | sed -n 1p)
```
haraldh: And I think `$()` eliminates all `\n` anyway...

modules.d/40network/net-lib.sh
nolange: It differs, if you resolve a remote ip it's $5, if the ip is from a local link it's $4. The function is not called anywhere(?), but I could use some sed to handle both cases.

Otherwise, I agree with the whitespace issues. I thought some shell might make problems which required the -oneline switch in the first place, thus I kept it conservative.

Typing this from a mobile now, will do a revised patch in a few days

nolange: Solved in v2 by cutting away everything before and including 'dev'.

nolange:
err, is that fault caused by my changes? have no clue why it fails

PR #502, opened 2018-12-12
remove superfluous quotes in parameter expansion

nolange:
as this breaks busybox hush shell.
offending functions can be found with
grep -r '${[^}]*"[^}]*}'


diabonas:
Quotes in patterns are actually required (according to POSIX) if the variable contains special pattern characters like `*` or `\`, try e.g.
```
i='a\c'; j='\'; echo "${i%$j*}"
```
vs.
```
i='a\c'; j='\'; echo "${i%"$j"*}"
```
Furthermore I couldn't find any issues with using quotes in parameter expansions with a recent busybox hush build. Since this actually breaks unlocking encrypted partitions with shells like bash, dash and ash, I suggest reintroducing variable quoting at least in patterns, see #794.

Issue #464, opened 2018-09-18
Dependence on Bash


GravisZro:
Despite there even being a module for Dash, Bash is still required to actually execute Dracut.  It wouldn't be a herculean effort to simply replace the Bashisms with POSIX conformant shell code.

GravisZro:
Bashism are non-standard and therefore will not work on shells that are not Bash.  Not everyone uses Bash.

danimo:
I think Harald's question was quite to the point: Is there a specific situation that you are in where bash is not an option for other reasons than some weird feeling that bash should somehow not be required on Linux? Dracut is quite complex for a shell project and foregoing bash specific would make it even more complex.

GravisZro:
It seems bizarre that I have to argue in favor of standards compliance.  Have we learned nothing from the days of "only works with Internet Explorer" websites?

haraldh:
I accept patches :smiling_imp:

awilfox:
Will you?  That emoji, and the "laugh" response, suggest that you think this is a joke.

It's not a joke, and to me that reflects very poorly on supporting Dracut as our primary initramfs generation system for our distribution.

johannbg:
This issue is being marked as stale because it has not had any recent activity. It will be closed if no further activity occurs. If this is still an issue in the latest release of Dracut and you would like to keep it open please comment on this issue within the next 7 days. Thank you for your contributions.

GravisZro:
Dependence on Bash is still a problem.

johannbg:
Feel free to submit a pull request that addresses that problem which will get reviewed and merged like issue #507

GravisZro:
Why did you close the issue?  It's a still a problem.

johannbg:
It's a problem that will be addressed via pull requests which no one has care to do for the past 2 years including yourself and there is no interest or apparent need for upstream to do that work and it's serves no point keeping issues open that no one is working on just for the sake of keeping them open. 

If and then when someone *decides* to rewrite dracut to be based upon "POSIX conformant shell code" that work will be reviewed just like any other PR.

GravisZro:
Just because nobody has addressed the issue yet doesn't mean it's not an issue.  Closing this only hides the issue, it does not resolve it.

johannbg:
That perspective is entirely dependent on the report being viewed as being an issue to begin with...

GravisZro:
[...]

PR #507, opened 2018-12-13
add mksh support v2

tpgxyz:
Add support for MirBSD™ Korn Shell, an actively developed free implementation of the Korn Shell programming language and a successor to the Public Domain Korn Shell (pdksh).
http://www.mirbsd.org/mksh.htm

@haraldh please review


tpgxyz:
@haraldh  mksh is a nice 100% POSIX compliat shell. Adding support for it may help to get rid of dependency on bash and sounds like little step for #464

PR #511, opened 2018-12-26
lsinitrd: list squash content as well

ryncsn:
When the initramfs is built with squash module enabled, the content
inside squash image will be invisible from the lsinitrd tool. This
commit make lsinitrd detect and list the content of the squash image as
well to avoid possible confusion.


daveyoung:
lsinitrd also support --unpack, it is good to also expand to support this option

PR #516, opened 2019-01-06
remove bashism in various boot scripts

keszybz:
When using dash as initramfs shell, the boot process is broken.

Signed-off-by: Đoàn Trần Công Danh <congdanhqx@gmail.com>

### DIFF
--- a/modules.d/90crypt/crypt-lib.sh
+++ b/modules.d/90crypt/crypt-lib.sh
@@ -191,7 +191,7 @@ readkey() {
     local device="$3"
 
     # No mounting needed if the keyfile resides inside the initrd
-    if [ "/" == "$keydev" ]; then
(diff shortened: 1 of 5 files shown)


congdanhqx:
For your reference, I x-posted my steps to reproduce here:

- Full disk encryption,
- Luks key is encrypted with gpg
- Luks key is embedded in initramfs (install_items+=" /etc/luks/key.gpg ")
- cmdline is: rd.luks.key=/etc/luks/key.gpg:/
- add_dracutmodules+=" dash "
- omit_dracutmodues+=" bash "
- build dracut

Expectation:
- the system should boot with new initramfs

Actual:
- The system failed to pick the key in initramfs
- replace dash by bash and system can pick up the correct key

PR #525, opened 2019-01-21
95zfcp_rules/parse-zfcp.sh: remove rule existence check

danimo:
Reference: bsc#1008352

Original-Patch-By: Michal Suchanek <msuchanek@suse.com>

### DIFF
--- a/modules.d/95zfcp_rules/parse-zfcp.sh
+++ b/modules.d/95zfcp_rules/parse-zfcp.sh
@@ -22,8 +22,6 @@ create_udev_rule() {
         return 0;
     fi
 
-    [ -e ${_rule} ] && return 0
-
     if [ ! -f "$_rule" ] ; then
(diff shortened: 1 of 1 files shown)


haraldh:
Is that the suse bugzilla you are referencing to?
`You are not authorized to access bug #1008352.`

A little bit more details in the commit log would be nice!

danimo:
This is a partner bug, and as such, it's not possible to open it to the public. I agree there should be more context.

By terminating the code early with this statement, only one `rd.zfcp=` statement would be processed, contrary to the man page. Without the check, several of the

```
ACTION=="add", KERNEL=="rport-*", ATTR{port_name}=="$wwpn", SUBSYSTEMS=="ccw", KERNELS=="$ccw", ATTR{[ccw/$ccw]$wwpn/unit_add}="$lun"
```
rules can be added now, fixing the issue.

PR #526, opened 2019-01-22
Add a sanity check against executable ld-linux.

shlomif:
See https://bugs.mageia.org/show_bug.cgi?id=24203 . I spent several days diagnosing and fixing this problem and would like to prevent it for others in the future.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -407,6 +407,23 @@ rearrange_params()
     fi
 }
 
+for fn in /lib/ld-linux* ; do
+    # Make sure we do not test unexpanded globs
+    if test -f "$fn"; then
+        if test -x "$fn"; then
+            cat << EOF 2>&1
+$fn which is one of the /lib/ld-linux* files is executable
+which causes an ldd warning and may make the generated initrd unbootable.
+
+Please run "chmod -x /lib/ld-linux*" as root before running dracut.
+
+Aborting.
+EOF
+            exit 1
+        fi
+    fi
+done
+
 verbosity_mod_l=0
 unset kernel
 unset outfile


haraldh:
Huh? On my machine:
```
$ ls -l  /lib*/ld-linux*
lrwxrwxrwx 10 root 14 Dez  2018 /lib64/ld-linux-x86-64.so.2 -> ld-2.28.so
$ ls -l  /lib*/ld-*.so
.rwxr-xr-x 228k root 14 Dez  2018 /lib64/ld-2.28.so
```

haraldh:
May you

haraldh:
What is your `ldd` warning?

shlomif:
```
[shlomif@telaviv1 ~]$ ldd /bin/echo
/usr/bin/ldd: line 160: /lib/ld-linux.so.2: cannot execute binary file: Exec format error
        linux-vdso.so.1 (0x00007ffd93d9f000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f02fdf84000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f02fe180000)
[shlomif@telaviv1 ~]$ uname -a
Linux telaviv1.shlomifish.org 5.0.0-rc2-00145-g7fbfee7c80de #1 SMP PREEMPT Thu Jan 17 13:42:20 IST 2019 x86_64 x86_64 x86_64 GNU/Linux
[shlomif@telaviv1 ~]$ 

```

This error does not happen on a stock mageia kernel so it may have to with me misconfiguring the custom kernel from source. See https://github.com/shlomif/shlomif-linux-kernel-configurations/blob/master/config-5.0.0-v1 . Thanks for looking into it!

shlomif:
Perhaps we can test instead if `ldd /bin/echo` emits anything to stderr.

haraldh:
Code in question is https://github.com/dracutdevs/dracut/blob/master/install/dracut-install.c#L450-L484

PR #527, opened 2019-01-23
Add a sanity check against mal functioning ldd.

shlomif:
See https://bugs.mageia.org/show_bug.cgi?id=24203 and https://github.com/dracutdevs/dracut/pull/526 .

It prevents dracut-install from working later, and we should fail early
instead of generating a broken initrd .

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -407,6 +407,28 @@ rearrange_params()
     fi
 }
 
+# Sanity check for https://github.com/dracutdevs/dracut/pull/526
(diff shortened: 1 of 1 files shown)


shlomif:
Hi! Any comments?

shlomif:
Ping!

haraldh:
dracut cannot guard against malfunctioning basic tools. We have to trust `mv` `cp` `ldd` `cpio` etc..

PR #534, opened 2019-02-10
network-manager: let the kernel generate a UUID for /etc/machine-id

floppym:
systemd normally manages /etc/machine-id, but dbus-uuidgen is more
likely to exist on systems that don't have systemd installed.

Bug: https://bugs.gentoo.org/677554

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -41,7 +41,9 @@ install() {
     elif ! [[ -e "$initdir/etc/machine-id" ]]; then
         # The internal DHCP client silently fails if we
         # have no machine-id
-        systemd-machine-id-setup --root="$initdir"
+        local UUID=$(< /proc/sys/kernel/random/uuid)
+        UUID=${UUID//-}
+        echo "$UUID" > "$initdir/etc/machine-id"
     fi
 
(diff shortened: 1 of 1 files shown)


floppym:
cc @lkundrak

haraldh:
What about:

```bash
$ UUID=$(< /proc/sys/kernel/random/uuid); UUID=${UUID//-}; echo $UUID
139bb634dacf46f2af37aa87e03e9ee7
```

floppym:
I'm not familiar with that kernel interface, and don't know how compatible that would be. Another alternative would be to use the uuidgen binary from util-linux. 

Just let me know what method you prefer.

lkundrak:
> I'm not familiar with that kernel interface, and don't know how compatible that would be.

The file seems to be there as far as the Linux git history goes (2.6.12 as of 2005). I'd say it's okay to rely on it.

> Another alternative would be to use the uuidgen binary from util-linux.
> 
> Just let me know what method you prefer.

@floppym thanks for the patch. I'd say @haraldh 's suggestion is superior to (dbus-)uuidgen and also systemd-machine-id-setup, because it has no external dependencies.

haraldh:
ping? want to update your PR?

floppym:
> ping? want to update your PR?

Done.

PR #535, opened 2019-02-12
Allow separate rootfs

zboszor:
This patchset allows running dracut on Yocto from packages' postinstall scriptlet during rootfs creation.

Only simple block devices would work (e.g. by using built-in drivers or adding them explicitly to the initramfs), as a lot of dracut modules, even during their check() and install() phases check for /proc, /sys or /dev state which is not applicable to the cross-compiled sysroot rootfs on the build host.
Therefore, some dracut modules are just not included in this initial initramfs.

These changes allow the complete initramfs to be generated during the image creation time for the cross-compiled OS. No need for re-running dracut on the target hardware at the first boot.

The validation was done on these setups:
* i686 cross-compiled Yocto 2.5 OS image on an x86_64 CentOS 7 build host (legacy target hardware)
* x86_64 cross-compiled Yocto 2.5 OS image on an x86_64 CentOS 7 build host (EFI target hardware)


zboszor:
Superceded by https://github.com/dracutdevs/dracut/pull/649

PR #541, opened 2019-02-14
install: string_hash_func should not be fed with NULL

lnykryn:
If kmod_module_get_path returns NULL, we should skip that entry,
the hash function does not like NULL pointers.


AdamWill:
Tested with a scratch build, it resolves the crash and the fix looks to make sense. Thanks. Will backport to Rawhide to avoid this causing any major issues there.

PR #543, opened 2019-02-20
99base: enable the initqueue in both 'dracut --add-device' and 'dracut --mount' cases

LianboJ:
…n time

Recently, we have encountered some kdump failures of virtual machines.
Although the device driver has been added to kdump initrd, the driver
is not loaded because the device isn't present yet, it's not brought up
in time. Which causes the kdump failure.

Therefore, the initqueue mechanism is still necessary for all devices
that need to be brought up in time.

Signed-off-by: Lianbo Jiang <lijiang@redhat.com>

### DIFF
--- a/modules.d/99base/module-setup.sh
+++ b/modules.d/99base/module-setup.sh
@@ -97,7 +97,7 @@ install() {
 
     ## save host_devs which we need bring up
     if [[ $hostonly_cmdline == "yes" ]]; then
-        if [[ -n $add_device ]]; then
+        if [[ -n "${host_devs[@]}" ]]; then
             dracut_need_initqueue
         fi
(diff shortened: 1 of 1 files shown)


LianboJ:
> ```
>     if [[ -n $add_device ]]; then
>         dracut_need_initqueue
>     fi
> ```
> 
> Looking at above code, it only covers dracut --devices case, but not covered --mount, kdump is using --mount, so instead of unconditionally use dracut_need_initqueue, can you try to check host_devs[] array, if it is not blank then add dracut_need_initqueue, in this way we can cover both cases

Thanks for your comment, daveyoung. I checked it and posted again, the host_devs[] can work well.

daveyoung:
Hi Harald,

This change looks good to me,  can you help to review as well?

Thanks
Dave

Issue #270, opened 2017-08-21
Add bluetooth module


bl33pbl0p:
RFE: Add a module to pull in bluetooth related systemd units, config files, and state from /var to be able to pair and connect to bluetooth devices like keyboards in early boot which would allow the user to enter the LUKS password. bluetoothd could power on the controller automatically if the config has AutoEnable=true on the host generating the initramfs. This would work well with kernels which have their bluetooth driver built into the kernel image, or it could be added as a module in the initramfs.

adamoa:
Hi. bluetoothd requires DBus. I prepared one module for bluetoothd and another for an adjusted version of DBus service that can start before sysinit.target. I used sed to change the service file for this. Would it be worth to submit a PR?

andrey82k:
> I prepared one module for bluetoothd
Thank you!

> Would it be worth to submit a PR?
Sure!

irreleph4nt:
I have commented on the PR as well. Can someone please review so this can get merged? Bluetooth connectivity in initramfs is a very useful thing to have. Thank you.

irreleph4nt:
Not stale.

bam80:
And how to reopen it now?

Neurognostic:
I think someone has to reopen now, because the thread was not poked soon enough.... thanks stale bot.

johannbg:
@Neurognostic I will be implementing this since I'm currently working on improving the systemd implementation in dracut. That work should hopefully land in the next dracut release, which should be released ca 2 weeks from now.

bgmilne:
With the current revision of https://github.com/dracutdevs/dracut/pull/1139, I am able to use 2 different Bluetooth keyboards in dracut, e.g. when booting with `rd.break` I can enter my root password using a Bluetooth keyboard at the `Enter root password to access an emergency shell` prompt.

However, my current setup has only `/home` on LUKS, so if someone who has their rootfs on LUKS can test, that would probably be better.

Besides using e.g. https://github.com/bgmilne/dracut/commit/9133705a616d87c97c220aad98fd43ae3fc48ab4 you also need to ensure that the `systemd` module isn't omitted (e.g. by files in /etc/dracut.conf*), and regenerate your initrd/initramfs with dracut (you may want to keep a backup). Check with `lsinitrd <initramfs file>|grep bluetooth` that at least some Bluetooth-related files are included, and then reboot.

bgmilne:
[...]

PR #544, opened 2019-02-22
Implemented bluetooth support in initrd.

adamoa:
- Included a bluetooth module that install modules, firmwares, udev rules and bluetoothd into initrd.
- Included a dbus module as bluetoothd requires it.
- This module depends on systemd.

This is a proposed solution for the issue #270 .

### DIFF
--- a/modules.d/62bluetooth/module-setup.sh
+++ b/modules.d/62bluetooth/module-setup.sh
@@ -3,7 +3,7 @@
 # called by dracut
 check() {
   require_binaries /usr/libexec/bluetooth/bluetoothd || return 1
-  
+
   return 255
 }
 
@@ -49,7 +49,8 @@ install() {
   sed -e \
 '/^\[Unit\]/aDefaultDependencies=no\
 Conflicts=shutdown.target\
-Before=shutdown.target' \
+Before=shutdown.target\
+After=dbus.service' \
     /usr/lib/systemd/system/bluetooth.service > \
     "$initdir"/usr/lib/systemd/system/bluetooth.service
 
--- a/modules.d/62bluetooth/module-setup.sh
+++ b/modules.d/62bluetooth/module-setup.sh
@@ -46,12 +46,11 @@ install() {
 
   inst_rules 69-btattach-bcm.rules 60-persistent-input.rules
 
-  sed -e \
+  sed -i -e \
 '/^\[Unit\]/aDefaultDependencies=no\
 Conflicts=shutdown.target\
 Before=shutdown.target\
 After=dbus.service' \
-    /usr/lib/systemd/system/bluetooth.service > \
     "$initdir"/usr/lib/systemd/system/bluetooth.service
 
   systemctl --root "$initdir" enable bluetooth.service > /dev/null 2>&1
--- a/modules.d/62bluetooth/module-setup.sh
+++ b/modules.d/62bluetooth/module-setup.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+  require_binaries /usr/libexec/bluetooth/bluetoothd || return 1
+  
+  return 255
+}
+
+depends() {
+  echo systemd dbus
+  return 0
+}
+
+installkernel() {
+  instmods bluetooth btrtl btintel btbcm bnep ath3k btusb rfcomm
+  inst_multiple -o \
+    /usr/lib/firmware/ar3k/AthrBT* \
+    /usr/lib/firmware/ar3k/ramps* \
+    /usr/lib/firmware/ath3k-1.fw \
+    /usr/lib/firmware/BCM2033-MD.hex \
+    /usr/lib/firmware/bfubase.frm \
+    /usr/lib/firmware/BT3CPCC.bin \
+    /usr/lib/firmware/brcm/*.hcd \
+    /usr/lib/firmware/mediatek/mt7622pr2h.bin \
+    /usr/lib/firmware/qca/nvm* \
+    /usr/lib/firmware/qca/crnv* \
+    /usr/lib/firmware/qca/rampatch* \
+    /usr/lib/firmware/qca/crbtfw* \
+    /usr/lib/firmware/rtl_bt/* \
+    /usr/lib/firmware/intel/ibt* \
+    /usr/lib/firmware/ti-connectivity/TIInit_* \
+    /usr/lib/firmware/nokia/bcmfw.bin \
+    /usr/lib/firmware/nokia/ti1273.bin
+}
+
+install() {
+  inst_multiple \
+    /usr/libexec/bluetooth/bluetoothd \
+    /usr/lib/systemd/system/bluetooth.target \
+    /usr/lib/systemd/system/bluetooth.service \
+    /etc/bluetooth/main.conf \
+    bluetoothctl
+
+  inst_multiple $(find /var/lib/bluetooth)
+
+  inst_rules 69-btattach-bcm.rules 60-persistent-input.rules
+
+  sed -e \
+'/^\[Unit\]/aDefaultDependencies=no\
+Conflicts=shutdown.target\
+Before=shutdown.target' \
+    /usr/lib/systemd/system/bluetooth.service > \
+    "$initdir"/usr/lib/systemd/system/bluetooth.service
+
+  systemctl --root "$initdir" enable bluetooth.service > /dev/null 2>&1
+}
--- a/dracut.spec
+++ b/dracut.spec
@@ -345,7 +345,7 @@ install -m 0755 51-dracut-rescue-postinst.sh $RPM_BUILD_ROOT%{_sysconfdir}/kerne
 %{dracutlibdir}/modules.d/45url-lib
 %{dracutlibdir}/modules.d/50drm
 %{dracutlibdir}/modules.d/50plymouth
-%{dracutlibdir}/modules.d/62bluetooh
+%{dracutlibdir}/modules.d/62bluetooth
 %{dracutlibdir}/modules.d/80lvmmerge
 %{dracutlibdir}/modules.d/90btrfs
 %{dracutlibdir}/modules.d/90crypt
--- a/dracut.spec
+++ b/dracut.spec
@@ -339,11 +339,13 @@ install -m 0755 51-dracut-rescue-postinst.sh $RPM_BUILD_ROOT%{_sysconfdir}/kerne
 %{dracutlibdir}/modules.d/03rescue
 %{dracutlibdir}/modules.d/04watchdog
 %{dracutlibdir}/modules.d/05busybox
+%{dracutlibdir}/modules.d/06dbus
 %{dracutlibdir}/modules.d/10i18n
 %{dracutlibdir}/modules.d/30convertfs
 %{dracutlibdir}/modules.d/45url-lib
 %{dracutlibdir}/modules.d/50drm
 %{dracutlibdir}/modules.d/50plymouth
+%{dracutlibdir}/modules.d/62bluetooh
 %{dracutlibdir}/modules.d/80lvmmerge
 %{dracutlibdir}/modules.d/90btrfs
 %{dracutlibdir}/modules.d/90crypt
--- a/modules.d/06dbus/dbus-cleanup.sh
+++ b/modules.d/06dbus/dbus-cleanup.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+systemctl stop dbus.service dbus.socket
+rm -rf /run/dbus
--- a/modules.d/06dbus/module-setup.sh
+++ b/modules.d/06dbus/module-setup.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+  require_binaries dbus-daemon || return 1
+  
+  return 255
+}
+
+depends() {
+  echo systemd
+  return 0
+}
+
+install() {
+  inst_multiple \
+    /usr/lib/systemd/system/dbus.service \
+    /usr/lib/systemd/system/dbus.socket \
+    /usr/bin/dbus-daemon \
+    /usr/bin/dbus-send
+
+  inst_multiple $(find /usr/share/dbus-1)
+  inst_multiple $(find /etc/dbus-1)
+  inst_multiple $(find /var/lib/dbus)
+
+  grep '^dbus:' /etc/passwd >> "$initdir/etc/passwd"
+  grep '^dbus:' /etc/group >> "$initdir/etc/group"
+
+  sed -e \
+'/^\[Unit\]/aDefaultDependencies=no\
+Conflicts=shutdown.target\
+Before=shutdown.target' \
+    /usr/lib/systemd/system/dbus.service > \
+    "$initdir"/usr/lib/systemd/system/dbus.service
+
+  sed -e \
+'/^\[Unit\]/aDefaultDependencies=no\
+Conflicts=shutdown.target\
+Before=shutdown.target
+/^\[Socket\]/aRemoveOnStop=yes' \
+    /usr/lib/systemd/system/dbus.socket > \
+    "$initdir"/usr/lib/systemd/system/dbus.socket
+}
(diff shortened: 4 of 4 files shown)


irreleph4nt:
Can someone please have a look at this so it can be merged? Bluetooth in initrd is very much a useful feature. Thank you.

sigboe:
Please merge this if possible. Most bluetooth keyboard doesn't even have a USB chord option.

hadess:
> Please merge this if possible

There are conflicts in one file, and the last 2 commits failed the test suite. It might also be good to squash the "typo fixes" commits together.

irreleph4nt:
> > Please merge this if possible
> 
> There are conflicts in one file, and the last 2 commits failed the test suite. It might also be good to squash the "typo fixes" commits together.

Hi. What needs to happen to resolve the dracut.spec conflict?

irreleph4nt:
I'll start tagging people now:

@lnykryn Can you please let people know what's needed to move this feature forward?

lnykryn:
This will not work on the fedora since now we are using dbus-broker, while this commit seems to use dbus-daemon. Also could you please rebase your branch a bit, so it contains two commits? One for adding the dbus module and the second one adding the Bluetooth module.

irreleph4nt:
> Also another thing I don't like is that after you pair a new keyboard you have to regenerate initramdisk to get there the /var/lib/bluetooth

I have created this functionality for Arch Linux's mkinitcpio, so have a bit of background. So, if you would not want to rebuild the initramfs, please let us know that how from your perspective we are meant to get the relevant data into the initramfs to actually work with a BT keyboard. Certainly you would not want people to pair a new device during boot, right? Because these changes AFAIK do not survive a reboot and would require an additional keyboard connected at the same time.

lnykryn:
Maybe it could be stored in efivars?

hadess:
The "right way" to do this, and the way it's implemented in devices like the Mac desktops, Lenovo all-in-ones, and plenty of other devices, is to have the linkkeys be stored on the HCI device itself. 

During boot, including EFI setup, grub, etc., the device is in HID mode, and uses the linkkeys stored on the device to talk to the mouse and keyboard. Later during boot, once storage is available, the device is switched to HCI mode, and acts as a "normal" Bluetooth adapter.

This isn't implemented in Linux because we don't have the specs for it, and haven't reverse engineered it either. The binary used to switch between the 2 modes is unsurprisingly called `hid2hci`.

haraldh:
> Can we agree on 1) here, please?

agreed.

One more thing though... can we dynamically detect the bluetooth keyboard and only then include the bluetooth module, which then pulls in dbus?

irreleph4nt:
> > Can we agree on 1) here, please?
> 
> agreed.
> 
> One more thing though... can we dynamically detect the bluetooth keyboard and only then include the bluetooth module, which then pulls in dbus?

You mean only pull in any Bluetooth device profile that has an identifier for keyboard and also only pull in any other data in case such a profile exists? I'd see two issues with that:
1. What if the profile is lacking that identifier?
2. Can you connect to more than one BT keyboard at once during init? I.e. what if I have paired two BT keyboards to my PC at the same time?

On question two, the way I solved this for Arch Linux was to have the user specify one BT address they would connect to during boot, because otherwise the controller might get enabled but no connection was established. I am not sure whether this would be different when using dracut?

haraldh:
I mean, only install the dracut bluetooth + dbus modules, if there is the need for it.
The [check()](https://github.com/dracutdevs/dracut/pull/544/files#diff-816d9e86ad011566e3cd10d6d3ae25cfR4) function in the dracut bluetooth module should check, in case of hostonly, if there is an actual bluetooth connection to a keyboard.

irreleph4nt:
> I mean, only install the dracut bluetooth + dbus modules, if there is the need for it.
> The [check()](https://github.com/dracutdevs/dracut/pull/544/files#diff-816d9e86ad011566e3cd10d6d3ae25cfR4) function in the dracut bluetooth module should check, in case of hostonly, if there is an actual bluetooth connection to a keyboard.

That particular thing I'd be unable to do. So will require someone else to look at it.

bam80:
The feature is still wanted. Please.

bgmilne:
> The check() function in the dracut bluetooth module should check, in case of hostonly, if there is an actual bluetooth connection to a keyboard.

As far as I know, something like this should be sufficient:
```
grep -iE 'Class=0x[0-9a-f]{3}5[4c]0' /var/lib/bluetooth/*/*/info
```

I have a Keychron K2 Bluetooth keyboard, which has:

```
# cat  /var/lib/bluetooth/*/DC\:*/info |grep -v ^Key
[General]
Name=Keychron K2
Class=0x000540
SupportedTechnologies=BR/EDR;
Trusted=true
Blocked=false
Services=00001000-0000-1000-8000-00805f9b34fb;00001124-0000-1000-8000-00805f9b34fb;00001200-0000-1000-8000-00805f9b34fb;

[LinkKey]
Type=4
PINLength=0

[DeviceID]
Source=2
Vendor=1452
Product=592
Version=283
```

Using an online [class calculator](http://bluetooth-pentest.narod.ru/software/bluetooth_class_of_device-service_generator.html), 
* a device with major class of 'peripheral', minor class of 'keyboard' should have a (hex) class of 0x540, which matches what I see, 
* toggling all of the Major Service Class checkboxes changes the leading 3 hex characters
* there is an additional minor device class of 'combo keyboard/pointing device' which uses lower 3 hex digits 0x5C0

With these conditions, a simple regex can exactly match all of the possible combinations, and we don't need to convert to binary and bitmask the values to find a device that is a keyboard.

There is the case of a computer that has previously had a bluetooth keyboard connected, but no longer has it, but I don't know how to reason about that vs. an initial ram-disk that is generated while the (only) bluetooth keyboard is not currently connected (e.g. user AFK during security updates and keyboard entered power saving).

If no-one else can implement this, I can take a look in a few more days (and also test with a different model Bluetooth keyboard).

bgmilne:
> I mean, only install the dracut bluetooth + dbus modules, if there is the need for it.
> The [check()](https://github.com/dracutdevs/dracut/pull/544/files#diff-816d9e86ad011566e3cd10d6d3ae25cfR4) function in the dracut bluetooth module should check, in case of hostonly, if there is an actual bluetooth connection to a keyboard.

The PR as it is has check() return 255, so as far as I understand, this should only include the bluetooth support files in the ramdisk if bluetooth was requested (e.g. by `add_dracutmodules+=bluetooth`).

So I am going to make it return 0 in the case of a keyboard device having been previously used/paired, but still return 255 in other cases.

Is this what you intended ?

modules.d/62bluetooth/module-setup.sh
```diff
@@ -0,0 +1,57 @@
...
+    /usr/lib/firmware/nokia/bcmfw.bin \
+    /usr/lib/firmware/nokia/ti1273.bin
+}
+
+install() {
+  inst_multiple \
+    /usr/libexec/bluetooth/bluetoothd \
+    /usr/lib/systemd/system/bluetooth.target \
+    /usr/lib/systemd/system/bluetooth.service \
+    /etc/bluetooth/main.conf \
```
haraldh: What about other config files:
```console
$ strings /usr/libexec/bluetooth/bluetoothd|fgrep etc/bluetooth
/etc/bluetooth/network.conf
/etc/bluetooth/input.conf
/etc/bluetooth/main.conf
```

haraldh:
> > I mean, only install the dracut bluetooth + dbus modules, if there is the need for it.
> > The [check()](https://github.com/dracutdevs/dracut/pull/544/files#diff-816d9e86ad011566e3cd10d6d3ae25cfR4) function in the dracut bluetooth module should check, in case of hostonly, if there is an actual bluetooth connection to a keyboard.
> 
> The PR as it is has check() return 255, so as far as I understand, this should only include the bluetooth support files in the ramdisk if bluetooth was requested (e.g. by `add_dracutmodules+=bluetooth`).
> 
> So I am going to make it return 0 in the case of a keyboard device having been previously used/paired, but still return 255 in other cases.
> 
> Is this what you intended ?

yes

haraldh:
and please squash your commits into two. one for dbus and one for bluetooth

irreleph4nt:
@bgmilne & @haraldh 
Are we making progress on getting bluetooth added to dracut? If this stalled again, then what is the current holdup, please?

bgmilne:
> @bgmilne & @haraldh
> Are we making progress on getting bluetooth added to dracut? If this stalled again, then what is the current holdup, please?

I have tested the changes I proposed above, and they seem to work correctly from the perspective of only including the additional files when a Bluetooth keyboard has been paired/used on a system before.

However, the distribution I use on my physical desktop machine had configured dracut to not enable systemd in the initrd, and I have not been able to get my system to enable my keyboard during boot yet (but, due to work demands, I wasn't able to spend too much time on that).

I will send/update a PR over the weekend with what I have, and maybe someone else who was previously able to get a keyboard working can test that the final changes still work.

irreleph4nt:
@bgmilne Have you had a chance to revisit your code yet?

modules.d/62bluetooth/module-setup.sh
irreleph4nt: @haraldh Those files are not required for bluetooth to work. What's needed is 3 things:
 i) main.conf from bluez, 
ii) relevant files (like firmware, software, modules) to enable BT generally, which includes bluez and possibly d-bus, 
iii) the device profile so bluez actually connects to something during boot
All 3 are already present in this PR.

irreleph4nt:
@bgmilne Kind reminder. Have you had a chance to look at this again? If not, what's the particular issue you are facing? Sounds like that with systemd enabled, no BT connectivity is established at boot?

danimo:
I have landed the dbus module. Feel free to submit improvements along with the bluetooth module.

irreleph4nt:
> I have landed the dbus module. Feel free to submit improvements along with the bluetooth module.

https://github.com/dracutdevs/dracut/blob/6fee16eab0430a38ca8b8845d9198f9dc86ad119/modules.d/06dbus/module-setup.sh#L64

Am I understanding the above code line correctly that this would get us out of the problem that bluetooth for us always started right **_after_** you had to enter the password for your encrypted drives? That has been the only real pain point with BT in dracut, as due to it BT was essentially useless.

danimo:
> Am I understanding the above code line correctly that this would get us out of the problem that bluetooth for us always started right **_after_** you had to enter the password for your encrypted drives? That has been the only real pain point with BT in dracut, as due to it BT was essentially useless.

It should do that. If not, it's unexpected and we need to fix it.

sigboe:
I don't want this merge request to go stale

haraldh:
still needs to be rebased

jo5ef:
I'm also hoping this will be merged.

sigboe:
I ended up using clevis to store my luks key in the tmp chip, then you will have to trust your display manager, but this may be good enough depending on what work you do. Ask the CSI at your work if *you* can trust your display manager for login. Assuming he is competent of course 😄 

https://aboutcher.co.uk/2020/06/fedora-linux-luks-encryption-with-tpm-unlock/

NaruFGT:
I'm interested in the functionality this provides.

johannbg:
Closing this stale PR since BT support has been added via #1139

tunix:
Hi! I've Fedora Silverblue 35 running on my computer and have the latest updates installed. I tried to enable bluetooth module by adding a file like this:

```
# cat /etc/dracut.conf.d/ble.conf
add_dracutmodules+=" bluetooth "
```

And then tried to regenerate initramfs with the following command:

```
dracut --regenerate-all --force
```

I can see the bluetooth related files inside initramfs including the files that are probably needed for my keyboard:

```
# lsinitrd initramfs-5.16.12-200.fc35.x86_64.img |grep bluetooth
bluetooth
drwxr-xr-x   1 root     root            0 Jan  1  1970 etc/bluetooth
-rw-r--r--   1 root     root        10841 Jan  1  1970 etc/bluetooth/main.conf
-rw-r--r--   1 root     root         1182 Jan  1  1970 etc/dbus-1/system.d/bluetooth.conf
drwxr-xr-x   1 root     root            0 Jan  1  1970 etc/systemd/system/bluetooth.target.wants
lrwxrwxrwx   1 root     root           41 Jan  1  1970 etc/systemd/system/bluetooth.target.wants/bluetooth.service -> /usr/lib/systemd/system/bluetooth.service
lrwxrwxrwx   1 root     root           41 Jan  1  1970 etc/systemd/system/dbus-org.bluez.service -> /usr/lib/systemd/system/bluetooth.service
-rwxr-xr-x   1 root     root       288200 Jan  1  1970 usr/bin/bluetoothctl
drwxr-xr-x   1 root     root            0 Jan  1  1970 usr/libexec/bluetooth
-rwxr-xr-x   1 root     root      1241576 Jan  1  1970 usr/libexec/bluetooth/bluetoothd
drwxr-xr-x   1 root     root            0 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/drivers/bluetooth
-rw-r--r--   1 root     root        26680 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/drivers/bluetooth/btbcm.ko.xz
-rw-r--r--   1 root     root        36504 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/drivers/bluetooth/btintel.ko.xz
-rw-r--r--   1 root     root        29044 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/drivers/bluetooth/btrtl.ko.xz
-rw-r--r--   1 root     root        46788 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/drivers/bluetooth/btusb.ko.xz
drwxr-xr-x   1 root     root            0 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth
-rw-r--r--   1 root     root       281012 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth/bluetooth.ko.xz
drwxr-xr-x   1 root     root            0 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth/bnep
-rw-r--r--   1 root     root        33300 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth/bnep/bnep.ko.xz
drwxr-xr-x   1 root     root            0 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth/hidp
-rw-r--r--   1 root     root        37252 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth/hidp/hidp.ko.xz
drwxr-xr-x   1 root     root            0 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth/rfcomm
-rw-r--r--   1 root     root        61760 Jan  1  1970 usr/lib/modules/5.16.12-200.fc35.x86_64/kernel/net/bluetooth/rfcomm/rfcomm.ko.xz
-rw-r--r--   1 root     root          822 Jan  1  1970 usr/lib/systemd/system/bluetooth.service
-rw-r--r--   1 root     root          435 Jan  1  1970 usr/lib/systemd/system/bluetooth.target
drwxr-xr-x   1 root     root            0 Jan  1  1970 var/lib/bluetooth
drwxr-xr-x   1 root     root            0 Jan  1  1970 var/lib/bluetooth/58:96:1D:2C:14:4B
drwxr-xr-x   1 root     root            0 Jan  1  1970 var/lib/bluetooth/58:96:1D:2C:14:4B/34:88:5D:A3:D1:A7
-rw-------   1 root     root          367 Jan  1  1970 var/lib/bluetooth/58:96:1D:2C:14:4B/34:88:5D:A3:D1:A7/info
....
drwxr-xr-x   1 root     root            0 Jan  1  1970 var/lib/bluetooth/mesh
```

But when I reboot my computer, it doesn't get paired with the keyboard at LUKS prompt so I can't type my encryption password in. Am I doing something wrong? Do I need to do additional stuff?

PR #549, opened 2019-03-01
Print an error if included item ('-i' option) doesn't exist or is neither a directory nor a regular file

rmetrich:
This PR proposes to make **dracut** fail when using the `-i` or `--include` option and the specified argument doesn't exist or is not either a directory or a file.


mrc0mmand:
@centos-ci retest this please

mrc0mmand:
@rmetrich excellent, thanks for the update!

PR #552, opened 2019-03-13
squash: keep less unneeded things outside the squash image

ryncsn:
Hi, This pull request contains several changes of squash module that will help reduce the memory usage even more, clean up the code a bit, and make it possible to do more optimize for kdump.


ryncsn:
Ping? Any comment?

PR #555, opened 2019-03-26
doc: Fix integrity key name

pevik:
Default value of EVMKEYDESC (in evm-enable.sh) is "evm-key" and it's
also specified previously in this README file.

Signed-off-by: Petr Vorel <pvorel@suse.cz>


danimo:
@centos-ci retest this please

PR #559, opened 2019-04-04
Fix boot with `fips` without a value

jkonecny12:
If you boot system with `fips` and not `fips=1` then you will get unary operator expected error. This will fix the problem.

Although it won't boot even after the patch but that is the same behavior as with `fips=1`, the error is fixed by this. It looks like fips on Fedora 30 and Rawhide is broken because of loop devices?

Tested also on Fedora 29 and it works correctly there.

Thanks @lnykryn for your help with this.

### DIFF
--- a/modules.d/01fips/fips-boot.sh
+++ b/modules.d/01fips/fips-boot.sh
@@ -2,6 +2,8 @@
 
 if ! fipsmode=$(getarg fips) || [ "$fipsmode" = "0" ]; then
     rm -f -- /etc/modprobe.d/fips.conf >/dev/null 2>&1
+elif [ -z "$fipsmode" ]; then
+    die "FIPS mode have to be enabled by 'fips=1' not just 'fips'"
 elif getarg boot= >/dev/null; then
     . /sbin/fips.sh
     if mount_boot; then
--- a/modules.d/01fips/fips-boot.sh
+++ b/modules.d/01fips/fips-boot.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if ! fipsmode=$(getarg fips) || [ $fipsmode = "0" ]; then
+if ! fipsmode=$(getarg fips) || [ "$fipsmode" = "0" ]; then
     rm -f -- /etc/modprobe.d/fips.conf >/dev/null 2>&1
 elif getarg boot= >/dev/null; then
     . /sbin/fips.sh
--- a/modules.d/01fips/fips-noboot.sh
+++ b/modules.d/01fips/fips-noboot.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if ! fipsmode=$(getarg fips) || [ $fipsmode = "0" ]; then
+if ! fipsmode=$(getarg fips) || [ "$fipsmode" = "0" ]; then
     rm -f -- /etc/modprobe.d/fips.conf >/dev/null 2>&1
 elif ! [ -f /tmp/fipsdone ]; then
     . /sbin/fips.sh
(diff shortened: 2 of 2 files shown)


jkonecny12:
OK I had a problem with the original patch. I was testing it with a wrong version so the `fips` only doesn't work not even for F29. So I will kill the boot with message that user should use `fips=1` instead.

The F30 & Rawhide looks like it has a problem to boot when using overlay with fips mode enabled. Even when I won't change anything by the overlay image. The boot works correctly with fips=1 without an overlay. The error looks like this:
``cannot find an unused loop device: No such device``

modules.d/01fips/fips-boot.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if ! fipsmode=$(getarg fips) || [ $fipsmode = "0" ]; then
+if ! fipsmode=$(getarg fips) || [ "$fipsmode" = "0" ]; then
```
danimo: This looks like it really should be using `getargbool 0 fips`, i.e. something like:

```
if ! getargbool 0 fips; then
```
See https://github.com/dracutdevs/dracut/blob/master/modules.d/99base/dracut-lib.sh#L245

(and for cherries on top maybe invert logic and swap if/else block to make reading easier)

modules.d/01fips/fips-noboot.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if ! fipsmode=$(getarg fips) || [ $fipsmode = "0" ]; then
+if ! fipsmode=$(getarg fips) || [ "$fipsmode" = "0" ]; then
```
danimo: dito

danimo:
Does anything outside dracut evaluate the fips cmdline parameter in FC? Because otherwise I would not know why simply specifying "fips" does not work. It would certainly be nice (and very dracutish) it it would work.

modules.d/01fips/fips-boot.sh
jkonecny12: I'm not sure this will really work. 
If you don't have fips on command line at all. It will behave the same way as if you have `fips=0` which is not what you want. That is the reason why the original conditional is so complicated AFAIK.

What you want here is to run the first condition only if you have `fips=0` the second in case `fips` and the third when there is `fips=1`. This is not easily doable by using `getargbool` thanks to the default value.

jkonecny12:
> Does anything outside dracut evaluate the fips cmdline parameter in FC? Because otherwise I would not know why simply specifying "fips" does not work. It would certainly be nice (and very dracutish) it it would work.

I agree but unfortunately wasn't able to find out.

modules.d/01fips/fips-boot.sh
danimo: I don't quite follow.  the general behavior would be the same, wouldn't it, as it only depends on those two checks that would decided the same in all cases (no param, `fips`, `fips=0` and `fips=1`)?

There is only these two `getargs` that check for the `fips` parameter. `DRACUT_FIPS_MODE` depends on fips module being included at build time.

jkonecny12: OK, after some more thinking it should behave the same as the original code.. but it doesn't.

After your proposed change missing `fips`, `fips=0`, `fips` and `fips=1` behaves the same, they all break bootup process. Unfortunately, I really don't know why, it just froze because `install.img` can't be downloaded by curl.

I guess I have to do more testing. The bad thing is, that I'm not able to get into console because it just freeze :(. And it looks like it behaves correct in the rd.break...

PR #561, opened 2019-04-12
51-dracut-rescue: do not require non-empty kernel cmdline

keszybz:
When booting with Fedora-Server-dvd-x86_64-30-20190411.n.0.iso,
/proc/cmdline is empty (libvirt, qemu host with bios, not sure if that
matters), after installation to disk, anaconda would "crash" in kernel-core
%posttrans, after calling kernel-install, because dracut would fail
with
> Could not determine the kernel command line parameters.
> Please specify the kernel command line in /etc/kernel/cmdline!

I guess it's legitimate, even if unusual, to have no cmdline parameters.
Two changes are done in this patch:
1. do not fail if the cmdline is empty.
2. if /usr/lib/kernel/cmdline or /etc/kernel/cmdline are present, but
   empty, ignore /proc/cmdline. If there's explicit configuration to
   have empty cmdline, don't ignore it.

### DIFF
--- a/50-dracut.install
+++ b/50-dracut.install
@@ -35,9 +35,7 @@ case "$COMMAND" in
             read -r -d '' -a BOOT_OPTIONS < /etc/kernel/cmdline
         elif [[ -f /usr/lib/kernel/cmdline ]]; then
             read -r -d '' -a BOOT_OPTIONS < /usr/lib/kernel/cmdline
-        fi
-
-        if ! [[ ${BOOT_OPTIONS[*]} ]]; then
+        else
             read -r -d '' -a BOOT_OPTIONS < /proc/cmdline
         fi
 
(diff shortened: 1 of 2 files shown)


keszybz:
@lnykryn could you also have a look at this one?

lnykryn:
Well, it sounds weird to only change it there, the same code is also here:
https://github.com/dracutdevs/dracut/blob/master/50-dracut.install#L34
and here:
https://github.com/systemd/systemd/blob/master/src/kernel-install/90-loaderentry.install#L50

keszybz:
Indeed. Updated.

keszybz:
One more change.

lnykryn:
https://github.com/systemd/systemd/pull/13391 got merged, so let's do the same here.

PR #562, opened 2019-04-14
dracut fixes for iscsiroot

danimo:
This rebases https://github.com/openSUSE/dracut/pull/7 on top of master.

https://github.com/openSUSE/dracut/pull/7/commits/354296ed1ae4c52191bb7955e5260756f7f6f2a6 is missing, since the node case has been removed in master by 7d0848957c0cb9f9796b7f4ec4941000bfaaf6fe. Unsure what to do for this one, @gonzoleeman.

### DIFF
--- a/modules.d/35network-legacy/net-genrules.sh
+++ b/modules.d/35network-legacy/net-genrules.sh
@@ -96,7 +96,11 @@ command -v fix_bootif >/dev/null || . /lib/net-lib.sh
 
         for iface in $IFACES; do
             if [ "$bootdev" = "$iface" ] || [ "$NEEDNET" = "1" ]; then
-                echo "${DRACUT_SYSTEMD+systemctl is-active initrd-root-device.target || }[ -f /tmp/net.${iface}.did-setup ]" >$hookdir/initqueue/finished/wait-$iface.sh
+		if [ -n "$netroot" ] && [ -n "$DRACUT_SYSTEMD" ]; then
+                    echo "systemctl is-active initrd-root-device.target || [ -f /tmp/net.${iface}.did-setup ]"
+		else
+                    echo "[ -f /tmp/net.${iface}.did-setup ]"
+		fi >$hookdir/initqueue/finished/wait-$iface.sh
             fi
         done
     # Default: We don't know the interface to use, handle all
--- a/modules.d/35network-legacy/net-genrules.sh
+++ b/modules.d/35network-legacy/net-genrules.sh
@@ -96,7 +96,7 @@ command -v fix_bootif >/dev/null || . /lib/net-lib.sh
 
         for iface in $IFACES; do
             if [ "$bootdev" = "$iface" ] || [ "$NEEDNET" = "1" ]; then
-                echo "[ -f /tmp/net.${iface}.did-setup ]" >$hookdir/initqueue/finished/wait-$iface.sh
+                echo "${DRACUT_SYSTEMD+systemctl is-active initrd-root-device.target || }[ -f /tmp/net.${iface}.did-setup ]" >$hookdir/initqueue/finished/wait-$iface.sh
             fi
         done
     # Default: We don't know the interface to use, handle all
--- a/modules.d/95iscsi/iscsiroot.sh
+++ b/modules.d/95iscsi/iscsiroot.sh
@@ -54,11 +54,11 @@ handle_firmware()
     if ! iscsiadm -m fw; then
         warn "iscsiadm: Could not get list of targets from firmware."
     else
-        ifaces=( $(echo /sys/firmware/ibft/ethernet*) )
+        ifaces=$(set -- /sys/firmware/ibft/ethernet*; echo $#)
         retry=$(cat /tmp/session-retry)
 
-        if [ $retry -lt ${#ifaces[*]} ]; then
-            let retry++
+        if [ $retry -lt $ifaces ]; then
+            retry=$((retry+1))
             echo $retry > /tmp/session-retry
             return 1
         else
@@ -88,6 +88,7 @@ handle_netroot()
     local iscsi_iface_name iscsi_netdev_name
     local iscsi_param param
     local p found
+    local login_retry_max_seen=
 
     # override conf settings by command line options
     arg=$(getarg rd.iscsi.initiator -d iscsi_initiator=)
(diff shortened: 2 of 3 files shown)


gonzoleeman:
@danimo you did the right thing. that "node" change is no longer needed.

danimo:
@haraldh ping? can you merge those?

modules.d/95iscsi/iscsiroot.sh
```diff
@@ -238,14 +233,22 @@ handle_netroot()
...
                      $EXTRA \
-                     $NULL
-        ;;
-        *)
-        ;;
-        esac
+                     $NULL"
+            $CMD
+            if [ "$netif" != "timeout" ]; then
+                $CMD --login
```
haraldh:
```
[   11.121564] dracut-initqueue[571]: iscsiadm: Invalid parameters. Login/logout and op passed in
```

danimo:
@gonzoleeman Can you check?

modules.d/95iscsi/iscsiroot.sh
gonzoleeman: You cannot use "--login" and "--op=update ..." in the same command line

gonzoleeman:
I looked over the changes, but I'd like @mwilck to check, since I believe he still has a test configuration set up.

modules.d/95iscsi/iscsiroot.sh
mwilck: Fixed with the update 403256a (pushed Oct 1st).

modules.d/35network-legacy/net-genrules.sh
```diff
@@ -96,7 +96,7 @@ command -v fix_bootif >/dev/null || . /lib/net-lib.sh
 
         for iface in $IFACES; do
             if [ "$bootdev" = "$iface" ] || [ "$NEEDNET" = "1" ]; then
-                echo "[ -f /tmp/net.${iface}.did-setup ]" >$hookdir/initqueue/finished/wait-$iface.sh
+                echo "${DRACUT_SYSTEMD+systemctl is-active initrd-root-device.target || }[ -f /tmp/net.${iface}.did-setup ]" >$hookdir/initqueue/finished/wait-$iface.sh
```
haraldh: Hmm, `NEEDNET=1` means, that somebody requested to wait until all network interfaces are up, before continuing.
This patch will change this behavior!

modules.d/35network-legacy/net-genrules.sh
```diff
@@ -96,7 +96,11 @@ command -v fix_bootif >/dev/null || . /lib/net-lib.sh
 
         for iface in $IFACES; do
             if [ "$bootdev" = "$iface" ] || [ "$NEEDNET" = "1" ]; then
-                echo "${DRACUT_SYSTEMD+systemctl is-active initrd-root-device.target || }[ -f /tmp/net.${iface}.did-setup ]" >$hookdir/initqueue/finished/wait-$iface.sh
+		if [ -n "$netroot" ] && [ -n "$DRACUT_SYSTEMD" ]; then
+                    echo "systemctl is-active initrd-root-device.target || [ -f /tmp/net.${iface}.did-setup ]"
```
haraldh: Hmm, NEEDNET=1 means, that somebody requested to wait until all network interfaces are up, before continuing.
This patch will change this behavior!

mwilck: `95iscsi/module-setup.sh` sets "`rd.neednet`" unconditionally (which is what's causing us trouble here). So does `35network-manager/nm-config.sh`. `35network-legacy/parse-ip-opts.sh` sets it if more then one `ip=` parameter is found and `rd.neednet` isn't explicitly disabled.

So the actual meaning of "`rd.neednet`" seems to be "**some** network interface needs to be up", rather than "**all** network interfaces must be up". 

The rationale of this patch is "*if it's about booting a computer* (indicated by `$netroot`), availability of the root device is sufficient to continue".  Can you give examples where this assumption is false, and where this patch would thus cause a regression?

If it's some other use of dracut (other than booting a computer), `$netroot` will not be set and dracut will wait for all interfaces as usual.

The root of the issue is that AFAICS, dracut does not have a concept of multipathing. It happens to work with Fibre Channel, because the connections don't have to be brought up actively. But with (software) iSCSI, it seems to be impossible to tell dracut "any one of these interfaces / iSCSI connections would do". Rather, dracut tries to bring up every interface and every iSCSI connection configured, and waits for each one, which contradicts the spirit of multipathing. This is what I'm trying to fix.

In the long run, a more generalized new approach to iSCSI multipath boot in dracut would be desirable, but that would be a larger effort.

mwilck: If I'm overlooking some simple way to express the "or" logic for network interface bringup logic in dracut, I'd be eager to learn about it.

mwilck:
To be honest: there is one issue with this change, although it's not directly related to dracut. 

With these patches, it can happen that not all iSCSI connections with "onboot" startup mode are brought up in the initrd, even if they're healthy. However they aren't usually brought up later either, because "iscsi.service" only starts those connections with "automatic" startup mode. This requires a change in the startup logic of open-iscsi; it's now wrong to assume that all "onboot" connections will be up as soon as the root device is mounted. "onboot" connections need to be triggered again at the same time when the "automatic" ones are being started. It's not a big deal, but it needs to be covered.

haraldh:
> To be honest: there is one issue with this change, although it's not directly related to dracut.
> 
> With these patches, it can happen that not all iSCSI connections with "onboot" startup mode are brought up in the initrd, even if they're healthy. However they aren't usually brought up later either, because "iscsi.service" only starts those connections with "automatic" startup mode. This requires a change in the startup logic of open-iscsi; it's now wrong to assume that all "onboot" connections will be up as soon as the root device is mounted. "onboot" connections need to be triggered again at the same time when the "automatic" ones are being started. It's not a big deal, but it needs to be covered.

good point

mwilck:
I was trying to figure out if the changes requested here are causing the TESTS=16 failure. Not being familiar with the dracut CI, I think the answer is "NO". Right?

haraldh:
> I was trying to figure out if the changes requested here are causing the TESTS=16 failure. Not being familiar with the dracut CI, I think the answer is "NO". Right?

yeah, TEST-16-DMSQUASH is an unrelated failure

A rebase to current master would trigger the use of the travis CI, instead of the CentOS jenkins CI.

modules.d/35network-legacy/net-genrules.sh
mwilck: Note: the man page says

     rd.neednet=1
            boolean, bring up network even without netroot set

So even there, it doesn't say "bring up all network interfaces"; rather, it hints that it's a setting that's (only?) relevant without netroot (where my patch doesn't change behavior).

Anyway, I'm aware that @haraldh, your interpretation counts more than anyone else's.
Please clarify if this is a killer argument, and if possible, give us a hint how to fix it.

One option would be to fix the cases where `rd.neednet` is set without compellent reason (see above), and modify my patch such that it wouldn't do anything with `$NEEDNET = 1`. Would that be ok for you?

haraldh: Oh, totally misread that patch somehow... sorry.. seems to be fine with me.

mwilck:
Thanks a lot!

PR #563, opened 2019-04-24
35network-legacy: Allow to use wicked as drop-in replacement tool

danimo:
- for dhclient
- (where applicable) arping


modules.d/35network-legacy/ifup.sh
```diff
@@ -18,11 +18,162 @@ netif=$1
...
+
     ip link set lo up
     ip addr add 127.0.0.1/8 dev lo
     exit 0
 fi
 
+dhcp_backend() {
+    type wicked >/dev/null 2>&1 && \
+    echo "wicked" || \
+    echo "dhclient"
```
Conan-Kudo: I'd rather this either be flipped or set up via configuration. This would be an unexpected nasty surprise for any openSUSE user who is trying to avoid wicked (such as myself).

Conan-Kudo:
@danimo I'd rather actually see a specific module for wicked rather than having it be chainsawed into the `network-legacy` module. That way, people who are using wicked can specifically pick up the wicked dracut integration.

danimo:
I explicitly tried to avoid creating a new module for this. If you want to make sure not to use wicked, don't install it. If you want a basic discussion on whether to have wicked for network management in the first place on openSUSE, that's not what this patch is trying to do. It tries to port a SUSE proprietary patch and make it useful everywhere.

Conan-Kudo:
> If you want a basic discussion on whether to have wicked for network management in the first place on openSUSE, that's not what this patch is trying to do

That's not what I'm saying either. Generally, for supporting network management systems, there's a separate dracut module for it. `35network-legacy` is for the legacy sysconfig/netconfig based network bootup process, and hybridizing it with wicked stuff could lead to unexpected behavior.

Is it really so bad to have a wicked module?

danimo:
Oh we are working on a wicked module, but that will use wicked entirely. Unless that is there, I would very much like to keep this patch (and upstream it, since wicked is mostly, but not exclusivly used on openSUSE.

Conan-Kudo:
Okay, then.

modules.d/35network-legacy/ifup.sh
haraldh: yes please

modules.d/35network-legacy/module-setup.sh
```diff
@@ -4,8 +4,9 @@
 check() {
     local _program
 
-    require_binaries ip dhclient sed awk grep || return 1
-    require_any_binary arping arping2 || return 1
+    require_binaries ip sed awk grep hostname || return 1
+    require_any_binary arping arping2 wicked || return 1
+    require_any_binary dhclient wicked || return 1
```
haraldh: probably install wicked in a separate module, so people can switch it on by purpose.
and maybe create one for dhclient also, so people can omit dhclient.

danimo: A separate wicked module is currently being developed, disjointed from the legacy module. 
Also, wicked also replaces arping2, not only dhclient in SUSE distros. How about options that can be set in `dracut.conf.d` for a more usercentric choice?

danimo:
Will still revise this one.

johannbg:
@danimo I personally nack this change since we should be aiming at deprecating it not keeping it around on some kind of life support, let alone for modules we ship ( makes absolutely no sense ) + we have iwd, NetworkManager,systemd-networkd and Wicked which should be plenty enough for the ecosystem. 

If the intend is to keep this archaeology stuff around we need to make it distribution agnostic and split the dhclient out of it not add wicked to it. 

That said you need rebase and fix your commit msg so it passes the commissary check :rofl:

PR #566, opened 2019-05-08
35network-manager Add basename as a requirement

### LINKED ISSUE #537: modules.d/35network-manager/nm-run.sh uses basename but does not require it
modules.d/35network-manager/nm-run.sh uses `basename`, but the accompanying module-setup.sh does not require it.

How to reproduce: On my system, sometimes bcache activation hangs forever, which will make `99-nm-run.sh` start before the new rootfs is online.  The console will then log lots of lines similar to: basename - no such file or directory.

Attached patch should fix this.  (Applies cleanly, not yet tested)

[dracut-049-network-manager-dependencies.patch.txt](https://github.com/dracutdevs/dracut/files/2856009/dracut-049-network-manager-dependencies.patch.txt)

See-Also: https://bugs.gentoo.org/677788

danimo:
Fixes #537

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -26,7 +26,7 @@ install() {
 
     _nm_version=$(NetworkManager --version)
 
-    inst_multiple sed grep
+    inst_multiple sed grep basename
 
     inst NetworkManager
     inst /usr/libexec/nm-initrd-generator


lkundrak:
https://github.com/dracutdevs/dracut/pull/578/commits/4c216b1db6a86373549e13b60250a7fcf94417b9 of PR #578 removes the basename requirement. I guess that is even better than dragging it in.

haraldh:
can be closed now?

PR #572, opened 2019-05-21
net-lib: strstr is not for globs

lnykryn:


danimo:

PR #575, opened 2019-05-27
move setting the "systemdutildir" variable before it's used

gdamjan:
on line 1086 it's used to check for the uefi_stub:
"${systemdutildir}/boot/efi/linux${EFI_MACHINE_TYPE_NAME}.efi.stub"

so it needs to be defined before that

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -999,6 +999,16 @@ esac
 
 abs_outfile=$(readlink -f "$outfile") && outfile="$abs_outfile"
 
+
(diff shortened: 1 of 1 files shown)


gdamjan:
this is just the single change to make `--uefi` work for me on ArchLinux without manually specifying `--uefi-stub /usr/lib/systemd/boot/efi/linuxx64.efi.stub`.

But it seems to me that all of the `systemd*` variables should be moved up too. Comments?

PR #578, opened 2019-05-28
network-manager fixes

lkundrak:
These fix the test suite with recent enough NetworkManager.

Tested with `make TESTS="01 02 03 04 11 15 17 20 30 31 50 70" check`, with and without NetworkManager.


lkundrak:
> The Travis CI build failed https://travis-ci.org/dracutdevs/dracut/jobs/538103915

It's the `bridging` subtest of `TEST-50-MULTINIC` that timed out. It works locally and a similar `bonding` test succeeded. Perhaps it was just Travis being sluggish?

danimo:
> > The Travis CI build failed https://travis-ci.org/dracutdevs/dracut/jobs/538103915
> 
> It's the `bridging` subtest of `TEST-50-MULTINIC` that timed out. It works locally and a similar `bonding` test succeeded. Perhaps it was just Travis being sluggish?

It often is. I retriggered the particular test. Let's see. In the long term, we should check if there is enough randomness, and maybe make sure we have havaged in our test images.

test/TEST-70-BONDBRIDGETEAMVLAN/test.sh
```diff
@@ -150,7 +150,7 @@ ip=192.168.57.104::192.168.57.1:24:test:ens5.0004:none
...
+    'ens3 inet 192\.168\.50\.[0-9]*/24 brd 192\.168\.50\.255 ens5\.0004 inet 192\.168\.57\.104/24 brd 192\.168\.57\.255 ens5\.3 inet 192\.168\.56\.103/24 brd 192\.168\.56\.255 vlan0001 inet 192\.168\.54\.101/24 brd 192\.168\.54\.255 vlan2 inet 192\.168\.55\.102/24 brd 192\.168\.55\.255 EOF ' \
```
haraldh: Hmm, can we test the legacy and NM output in different tests?

test/TEST-50-MULTINIC/test.sh
```diff
@@ -243,7 +243,7 @@ test_setup() {
             done
         )
         inst_multiple sh shutdown poweroff stty cat ps ln ip \
-                      mount dmesg mkdir cp ping grep ls
+                      mount dmesg mkdir cp ping grep ls basename
```
haraldh: Huh? what needs basename? It's so easy to do that in bash.

test/TEST-70-BONDBRIDGETEAMVLAN/test.sh
lkundrak: Probably. I'm not sure what would be the best way though; opened #580 that solves it by creating a copy of the test.

test/TEST-50-MULTINIC/test.sh
lkundrak: I did not know about that. Fixed now (also in #580).

lkundrak:
Ok, no idea how that many tests failed; some are clearly not affected by this patch set. I've seen the test 12 failure locally; opened #581 with my patch. Will be looking into the rest shortly.

lkundrak:
There's more ENOSPC errors, one Docker Hub rate limiting problem and some unclear ones. Things work locally; I'm likely not to blame for these.

lkundrak:
Folded in a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1719057

PR #580, opened 2019-06-10
network-manager test fixes

lkundrak:
Split  from #578 because the BONDBRIDGETEAMVLAN change may need more discussion.


lkundrak:
As far as the test failures go, they probably are not my fault, see #578.

PR #583, opened 2019-06-13
keep network device naming scheme on upgrade

tblume:
Include 70-persistent-net.rules (if present) in order to keep the persistent network device naming scheme on system upgrade.


danimo:
@tblume Harald is right. Can you adjust the patch once you are back?

tblume:
I've updated the patch accordingly.

PR #584, opened 2019-06-14
dracut-install: Add support for compressed kernel modules

tiwai:
When a module is compressed, uncompress it before packing into initrd.
Since initrd is already compressed, it makes no sense to put the
compressed module files.

The patch contains a slight refactoring and adds a helper function to
get the command for uncompressing a file per extension.

This issue came up in opensuse bugzilla:
  https://bugzilla.suse.com/show_bug.cgi?id=1135854

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -533,6 +533,14 @@ inst_libdir_file() {
     [[ $_files ]] && inst_multiple $_files
 }
 
+# get a command to decompress the given file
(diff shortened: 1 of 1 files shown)


danimo:
Restarting failed CI tests. False negatives again :(

danimo:
It's no use, the disk images are currently too small to contain binaries from the latest fedora release. This is currently being addressed elsewhere.

PR #585, opened 2019-06-15
35network-legacy: fix classless static route parsing

frdeng:
The current dhclient-script doesn't handle the case when classless static route with mask length=0, the patch fixes the problem, sample dhcp code 121:

option rfc2442-classless-static-routes 0,10,0,0,1,16,169,254,0,0,0,0


lkundrak:
Looks all right to me.

PR #590, opened 2019-06-24
95iscsi: Fix network setup

ryncsn:
There are multiple problems with network setup code in 95iscsi:

- The network script and config is in "/etc/sysconfig/network-scripts".
  So fix the look up path.

- The regex used for sed is not working, so fix it too.

### DIFF
--- a/modules.d/95iscsi/module-setup.sh
+++ b/modules.d/95iscsi/module-setup.sh
@@ -74,6 +74,7 @@ install_iscsiroot() {
     local devpath=$1
     local scsi_path iscsi_lun session c d conn host flash
     local iscsi_session iscsi_address iscsi_port iscsi_targetname iscsi_tpgt
+    local bootproto
 
     scsi_path=${devpath%%/block*}
     [ "$scsi_path" = "$devpath" ] && return 1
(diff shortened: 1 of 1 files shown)


danimo:
Hi, why did you remove
```
-        printf 'ip=%s:static ' ${ifname}
+       printf 'ip=%s:none ' ${ifname}
```

The former made no sense indeed, but what is none supposed to do?

modules.d/95iscsi/module-setup.sh
```diff
@@ -122,7 +122,7 @@ install_iscsiroot() {
     ifname=$(ip -o route get to $iscsi_address | sed -n 's/.*dev \([^ ]*\).*/\1/p')
 
     #follow ifcfg settings for boot protocol
-    bootproto=$(sed -n "/BOOTPROTO/s/BOOTPROTO='\([[:alpha:]]*6\?\)4\?'/\1/p" /etc/sysconfig/network/ifcfg-$ifname)
```
danimo: This is the path used with SUSE. Please change the code so it works with both.

ryncsn: Thanks very much, I'll add detect code for both.

ryncsn:
> Hi, why did you remove
> 
> ```
> -       printf 'ip=%s:static ' ${ifname}
> +       printf 'ip=%s:none ' ${ifname}
> ```
> 
> The former made no sense indeed, but what is none supposed to do?

Yes, "static" make no sense, but if "none" is used, it should pass the static IP address by cmdline too. I think I can fix that later, as that will require more works.

PR #610, opened 2019-07-22
modules.d: add a module for early fan control on PowerPC Macs

q66:
The goal of this module is to enable automatic loading of the thermal/fan control modules on PowerPC based Macs (usually G5, which use the `windfarm` infrastructure), as on some modular kernel configurations this will not happen automatically which will result in the fans spinning up to 100% until they are manually loaded.

This is especially a problem in live systems where it takes some time to boot up and the spin-up happens before the system even leaves initramfs, but even on normal systems this is generally beneficial, as it means the modules are correctly loaded automatically.

This is inspired by a similar `thermal` hook in Debian's initramfs-tools, though it has been written from scratch for `dracut` and today's kernels (effectively 2012+, the Debian hook still loads `therm_p72` for PowerMac7,2/7,3/RackMac3,1, this module loads the appropriate `windfarm_pm72` and `windfarm_r31` present in Linux since 2012).

### DIFF
--- a/modules.d/90ppcmac/module-setup.sh
+++ b/modules.d/90ppcmac/module-setup.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+#
+# This module attempts to properly deal with thermal behavior on PowerPC
+# based Mac systems, by installing the model-appropriate (when hostonly)
+# or all (when not) fan control/thermal kernel modules and loading them
+# in a hook.
+#
+# While this is not strictly necessary for all kernels, particularly
+# modular kernels will not autoload those drivers, even once the full
+# system is up, which results in the fans spinning up to 100%; this is
+# particularly annoying on live systems, where the system takes a while
+# to load, so it's best to load the drivers early in initramfs stage.
+#
+# The behavior of this is inspired by the thermal hook in Debian's
+# initramfs-tools, but written for dracut specifically and updated
+# for modern kernels (2012+).
+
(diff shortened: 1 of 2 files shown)


modules.d/90ppcmac/module-setup.sh
```diff
@@ -0,0 +1,82 @@
...
+# The behavior of this is inspired by the thermal hook in Debian's
+# initramfs-tools, but written for dracut specifically and updated
+# for modern kernels (2012+).
+
+# called by dracut
+check() {
+    local _arch="$(uname -m)"
+    # only for PowerPC Macs
+    [[ "$_arch" == ppc* && "$_arch" != ppc*le ]] || return 1
+    return 0
```
haraldh: Shouldn't this also check for the pm_model and only return 0 if it matches?

q66: Well, no, because the idea is that non-hostonly initramfs images will have it regardless of the machine they're created on (which may be any `ppc64` or `ppc` system). For example, I use this in bootable live images, which are meant to be run on a variety of hardware. And even on a regular system, unless you set your dracut to hostonly, moving it to another hardware should result in everything still working. Which may be between different PowerMacs or even between a non-Apple and Apple system.

haraldh: ok

PR #612, opened 2019-07-24
[RHEL-8] network-manager updates

lkundrak:
Pull in the network-manager fixes (along with the test fixes) and and extra change that will make Dracut not switch to network-manager by default when nm-initrd-generator gets updated.


mrc0mmand:
@lkundrak hey, could you please rebase this PR so it uses the *fixed* Travis config?

PR #613, opened 2019-08-01
Documentation: add systemd warning for rd.luks.key, clean up related text

jbash:
I've managed to get myself bitten *twice* in the last couple of years by the fact that rd.luks.key doesn't work with the Dracut systemd module (which of course it is on Fedora and RHEL). Apparently this has caused trouble for other people as well.

The basic problem is that if you add the option when systemd is there, it has no detectable effect; you're still asked for a password and you sit around wondering why and digging through journals. See also https://bugzilla.redhat.com/show_bug.cgi?id=905683 .

... so here's a suggested fix to the dracut.cmdline manual page.

I also cleaned up some other text in the same section. I hope I made it clearer. I definitely fixed some grammar, for whatever that's worth.

### DIFF
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -318,22 +318,37 @@ crypto LUKS
 
 crypto LUKS - key on removable device support
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


haraldh:
pushed a fixed version with 7f0b48627a8cfdaba7263a35c827ff73e956a65c

PR #614, opened 2019-08-02
90kernel-modules-extra: don't resolve symlinks before instmod

lnykryn:
When you install a third-party driver, you will probably end in a
situation, where the module will be in a different directory and
in $depmod_module_dir you will only have symlink. If we resolve the
symlink before we pass the module path to instmod, the dracut-install
will only include the module with its original path, but not the
symlink. Hence the module can't be automatically loaded.

Dracut-install is clever enough to handle symlinks and will include both
the symlink and the module to the initrd.


haraldh:
Looks good to me :) Thanks!

PR #622, opened 2019-08-20
man: fix several typos

mrc0mmand:


dracut.usage.asc
```diff
@@ -151,7 +151,7 @@ If your root partition is on the network see <<NetworkBoot>>.
 If you have to input passwords for encrypted disk volumes, you might want to set
 the keyboard layout and specify a display font.
 
-A typical german kernel command would contain:
+A typical German kernel command would contain:
```
haraldh: isn't this an adjective and therefore lowercase? .. a real fix would be: s/kernel command/kernel command line/

mrc0mmand: ah, thanks, that's indeed a mistake on my side (and I completely overlooked the missing 'line')

PR #623, opened 2019-08-20
travis: fetch only git tags

mrc0mmand:
As Travis CI builds the merge commit[0] of given PR, doing `git pull` during
testing is somewhat useless and breaks testing of PRs against non-master
branches.

One of these *broken* PRs is, for example, https://github.com/dracutdevs/dracut/pull/612.

[0] i.e. `refs/pull/XXX/merge`, see https://docs.travis-ci.com/user/pull-requests/#how-pull-requests-are-built
>  Rather than build the commits that have been pushed to the branch the pull request is from, we build the merge between the source branch and the upstream branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_script:
     sudo modprobe kvm-intel nested=1 || :
     sudo modprobe kvm-amd nested=1 || :
     dmesg | tail || :
-  - git pull --depth=100
+  - git fetch --tags --unshallow
   - |
     git describe --abbrev=0 --tags || :
     git describe --tags || :
--- a/git2spec.pl
+++ b/git2spec.pl
@@ -37,7 +37,7 @@ $tag=`git describe --abbrev=0 --tags` if not defined $tag;
 chomp($tag);
 my @patches=&create_patches($tag, $pdir);
 my $num=$#patches + 2;
-$tag=~s/[^0-9]+?([0-9]+)/$1/;
+$tag=~s/[^0-9]+?([0-9]+)/$1/ if $tag !~ /\b[0-9a-f]{5,40}\b/;
 my $release="$num.git$datestr";
 $release="1" if $num == 1;
 


mrc0mmand:
@haraldh also, it looks there's a somewhat nasty bug in the `git2spec.pl` script. This line:

https://github.com/dracutdevs/dracut/blob/52dda1f3b2603b4912aa2470974a4dbb38158cc1/git2spec.pl#L40

will malform the SHA-1 hash of the current commit in certain cases, which causes CI's to fail (both failures in this PR's run):

```
$ git rev-parse HEAD
6e8cd92261577230daa1098f7e05ec198c3c4281
$ perl -e 'use POSIX; my $tag="6e8cd92261577230daa1098f7e05ec198c3c4281"; $tag=~s/[^0-9]+?([0-9]+)/$1/; print("$tag\n")'
68cd92261577230daa1098f7e05ec198c3c4281
```

Notice the missing 'e' after the first character of the hash.

mrc0mmand:
I see, the git SHA-1 related issues were discovered because of the first commit, which drops `git pull --tags`, in which case the `git describe --abbrev=0 --tags --always` from Makefile returns a SHA-1 of the current revision instead of the latest tag (as the git repo in Travis doesn't have tags pulled in).

@haraldh do you prefer the former way of building the RPM (in Travis) from a version tag + patches or is the *current* way okay-ish? Basically, Travis now builds the dracut RPM from a current revision's source tarball, instead of using the *base* tarball (e.g. [049](https://github.com/dracutdevs/dracut/archive/049.tar.gz)) and then applying bunch of patches on top of it. This makes it compatible with non-master branches.

mrc0mmand:
tl;dr of the changes:

1cd958b - fixes a potential usecase of building dracut RPMs using a revision's SHA-1 as a version tag

1c3580b - we need tags to build dracut RPMs for certain tests, however, using `git pull` will pull changes from the currently tracking branch as well, which is an unwanted sideefect. What it originally did:
   1) Travis cloned the repo (shallow clone, without tags) and called `git fetch origin +refs/pull/623/merge: && git checkout -qf FETCH_HEAD`. This results in *detached head* state, however, the `FETCH_HEAD` would point to the merge commit of the PR branch and the respective source branch
   2) We would call `git pull --depth=100` to fetch tags. However, the active tracking branch would still be `master`, because we're just in the *detached head* state. This would pull changes from the `master` branch to our current *detached head* revision, which **might not** be the `master` branch
   3) That results in `fatal: refusing to merge unrelated histories` error and job abort for PR's based on non-master branches

The *replacement* is `git fetch --tags --unshallow` which fetches tags and metadata to link them to the respective commits.

cc @lnykryn

PR #626, opened 2019-08-22
Fix for compressed firmware support

tiwai:
I thought it was already implemented, but it was bogus.  This adds a very simple support for the compressed firmware file in dracut-install binary, as well as a cleanup of the dead code.


danimo:
LGTM. We should definitely follow up with a patch to uncompress the files. @haraldh Do you have a preference between linking to liblzma or calling xz?

PR #633, opened 2019-08-31
Suppress spurious output when using crypt module without systemd-cryptsetup-generator

diabonas:
When using the `crypt` module without systemd-cryptsetup-generator (with `dracut -o systemd`), there is some unwanted output:

- When decrypting LUKS2 partitions, a message
    ```
    WARNING: Locking directory /run/cryptsetup is missing!
    ```
    is printed.
- If `rootflags` are specified on the kernel command line, these are erroneously printed by the `rootfs-block` module.

Creating the required directory beforehand and redirecting the output of `getarg` removes these artefacts.


danimo:
LGTM

PR #634, opened 2019-09-03
modules.d: fix udev rules detection of multipath devices

bmarzins:
Starting with the 0.7.7 release of the multipath tools, the multipath
udev rules always set a value in ENV{DM_MULTIPATH_DEVICE_PATH} for any
device that multipath scans. A value of 0 means that the device is not
claimed by multipath, and a value of 1 means that it is. Because of
this, udev rules that check ENV{DM_MULTIPATH_DEVICE_PATH}=="?*" will
always return True, and act as if every scanned device is claimed by
multipath. Checking ENV{DM_MULTIPATH_DEVICE_PATH}=="1" will work
correctly for both the old and new versions of the multipath tools.


mwilck:
I'm not an admin, but I can confirm that this patch is correct.

danimo:
Merging as per @mwilck's advise.

PR #636, opened 2019-09-05
10i18n: enable rd.vconsole.* and rd.locale.* handling when systemd module is included

mmaxs:
`parse-i18n.sh`  correctly handles `rd.vconsole.*` and `rd.locale.*` command line parameters and adds respective systemd's keys into /etc/vconsole.conf and /etc/locale.conf files, so this fix makes them being available according to dracut.cmdline(7) when `systemd` dracut  module is included


mmaxs:
After some tests, I found it more supportable to use the systemd-specific kernel command line options `vconsole.*` and `locale.LANG` instead of `rd.vconsole.*` and `rd.locale.*` to override the i18n settings for the initramfs environment.

PR #640, opened 2019-09-10
51-dracut-rescue-postinst.sh: Create hmac file for rescue kernel

ryan-blakley:
When fips=1 is enabled on a server, and the server is booted into the generated rescue kernel grub menu entry. It will fail to boot since there isn't an hmac file for the rescue vmlinuz file. This patch checks if the sha512hmac binary exist, if it does it generates the hmac file for the rescue kernel. Also it checks if a rescue kernel entry exist in the grub.cfg, if it does it doesn't add a new entry. This is so when ran on existing rescue kernels, it doesn't create duplicate entries when generating the hmac. 

https://bugzilla.redhat.com/show_bug.cgi?id=923439

### DIFF
--- a/51-dracut-rescue-postinst.sh
+++ b/51-dracut-rescue-postinst.sh
@@ -16,10 +16,18 @@ fi
 [[ $MACHINE_ID ]] || exit 1
 [[ -f $KERNEL_IMAGE ]] || exit 1
 
+KERNEL_HMAC="${KERNEL_IMAGE%/*}/.vmlinuz-${KERNEL_VERSION}.hmac"
 INITRDFILE="/boot/initramfs-0-rescue-${MACHINE_ID}.img"
 NEW_KERNEL_IMAGE="${KERNEL_IMAGE%/*}/vmlinuz-0-rescue-${MACHINE_ID}"
+NEW_KERNEL_HMAC="${KERNEL_IMAGE%/*}/.vmlinuz-0-rescue-${MACHINE_ID}.hmac"
 
-[[ -f $INITRDFILE ]] && [[ -f $NEW_KERNEL_IMAGE ]] && exit 0
+if [[ -f $INITRDFILE ]] && [[ -f $NEW_KERNEL_IMAGE ]]; then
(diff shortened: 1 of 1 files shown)


haraldh:
I would try to copy a possible existing .hmac here.

https://github.com/dracutdevs/dracut/blob/c296d29d3f931fb7ce320a88a150ba7bf9ab8272/51-dracut-rescue-postinst.sh#L66-L69

```
    cp --reflink=auto "$KERNEL_IMAGE" "$NEW_KERNEL_IMAGE"
    [[ -f "${KERNEL_IMAGE}.hmac"  ]] && cp --reflink=auto "${KERNEL_IMAGE}.hmac" "${NEW_KERNEL_IMAGE}.hmac"
```

ryan-blakley:
I apologize for the delay, I never received an email notification about your comment. 

So I choose not to copy, because the hmac file for the rescue vmlinuz more than likely doesn't exist on existing installs. So I figured it was best to generate a new hmac file based on the rescue vmlinuz file, to be absolutely sure.

haraldh:
Well, best to generate a new one only if you can't copy it.

ryan-blakley:
That's true, I tweaked the cp snippet you provided, and tested that it works for both scenarios.

51-dracut-rescue-postinst.sh
```diff
@@ -58,11 +65,18 @@ fi
 
 if [[ ! -f $NEW_KERNEL_IMAGE ]]; then
     cp --reflink=auto "$KERNEL_IMAGE" "$NEW_KERNEL_IMAGE"
+    [[ ! -f $NEW_KERNEL_HMAC  ]] && cp --reflink=auto "${KERNEL_IMAGE%/*}/.vmlinuz-${KERNEL_VERSION}.hmac" "$NEW_KERNEL_HMAC"
```
haraldh: hmm, check for existence of the source or discard the cp error messages

ryan-blakley: Sure thing, I added a check for existence of the source.

PR #654, opened 2019-10-18
Move README to README.md

cgwalters:
The `README.md` was nearly empty.  Move the travis bits into
`README`, then rename `README` to `README.md`.

This matches the Github standard.  A major compelling feature
of Github is how prominently it displays a project's `README.md`,
so let's ensure ours has content.


danimo:
This makes a lot of sense from a GH perspective, but from a packagers perspective, you now end up with packaging the travis status include as well. That feels rather silly.

cgwalters:
> you now end up with packaging the travis status include as well. That feels rather silly.

We're talking about the single line of Markdown in the file?  I can't see that it matters.

(I suspect the ratio of people viewing Github to reading packaged files locally is something like 1000:1 at least too)

danimo:
Fair enough.

PR #655, opened 2019-10-18
99base: Rework `/etc/initrd-release` to derive from real os-release

cgwalters:
I'd like to rework CoreOS Ignition (which runs in the initramfs)
to include some values from the *real* `/etc/os-release` in
HTTP headers.

Looking at this, it turns out dracut eats almost all of the useful
information from it.  I don't think `dracut` should be the `ID`
here...dracut's not an OS itself, it's a way to *build* little
operating systems.  It'd be kind of like if Fedora's Koji
injected itself into `/etc/os-release`.

This code dates back a long time; not sure of all the rationale
behind it.

I changed it so that we keep extending the VERSION/PRETTY_NAME
with the dracut version, but otherwise "pass through" the
rest of the real `/etc/os-release` we were built from unchanged.

### DIFF
--- a/modules.d/99base/module-setup.sh
+++ b/modules.d/99base/module-setup.sh
@@ -69,17 +69,19 @@ install() {
 
     local VERSION=""
     local PRETTY_NAME=""
+    # default values
+    ANSI_COLOR="0;34"
     if [ -e /etc/os-release ]; then
         . /etc/os-release
         [[ -n ${VERSION} ]] && VERSION+=" "
         [[ -n ${PRETTY_NAME} ]] && PRETTY_NAME+=" "
     fi
+    # force-override values
     NAME=dracut
     ID=dracut
     VERSION+="dracut-$DRACUT_VERSION"
     PRETTY_NAME+="dracut-$DRACUT_VERSION (Initramfs)"
     VERSION_ID=$DRACUT_VERSION
-    ANSI_COLOR="0;34"
 
     [ -e "${initdir}/usr/lib" ] || mkdir -m 0755 -p ${initdir}/usr/lib
     {
--- a/modules.d/99base/module-setup.sh
+++ b/modules.d/99base/module-setup.sh
@@ -67,31 +67,36 @@ install() {
         echo ro >> "${initdir}/etc/cmdline.d/base.conf"
     fi
 
+    [ -e "${initdir}/usr/lib" ] || mkdir -m 0755 -p ${initdir}/usr/lib
+
     local VERSION=""
     local PRETTY_NAME=""
-    # default values
-    ANSI_COLOR="0;34"
+    # Derive an os-release file from the host, if it exists
     if [ -e /etc/os-release ]; then
         . /etc/os-release
+        grep -hE -ve '^VERSION=' -ve '^PRETTY_NAME' /etc/os-release >${initdir}/usr/lib/initrd-release
         [[ -n ${VERSION} ]] && VERSION+=" "
         [[ -n ${PRETTY_NAME} ]] && PRETTY_NAME+=" "
+    else
+        # Fall back to synthesizing one, since dracut is presently used
+        # on non-systemd systems as well.
+        {
+            echo NAME=dracut
+            echo ID=dracut
+            echo VERSION_ID=\"$DRACUT_VERSION\"
+            echo ANSI_COLOR='"0;34"'
+        } >${initdir}/usr/lib/initrd-release
     fi
-    # force-override values
-    NAME=dracut
-    ID=dracut
     VERSION+="dracut-$DRACUT_VERSION"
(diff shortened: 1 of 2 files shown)


cgwalters:
(Also I just noticed before submitting this PR that this code is duplicated in `01systemd-initrd`?  Which is probably being overwritten with `99base`?)

danimo:
> (Also I just noticed before submitting this PR that this code is duplicated in `01systemd-initrd`? Which is probably being overwritten with `99base`?)

Funny you should mention this... Check out https://github.com/danimo/dracut/commit/1b567c09ed71a84337348e4c76179bb205c8a55b :)

danimo:
Also this: https://github.com/danimo/dracut/commit/af072ff53dd0711421191e5212f93ea0a1525005. Would you be willing to integrate those two patches in your patchset? They are currently part of an experimental patchset that tries to satisfy shellcheck, which they do not really belong to.

As for why the ID is dracut: I don't know. But it is possible that some 3rd party modules rely on the information now that it's there. Maybe @haraldh can shed some light here. Anyway, if we change it, we need to mention in NEWS.

haraldh:
> Also this: [danimo@af072ff](https://github.com/danimo/dracut/commit/af072ff53dd0711421191e5212f93ea0a1525005). Would you be willing to integrate those two patches in your patchset? They are currently part of an experimental patchset that tries to satisfy shellcheck, which they do not really belong to.
> 

sure, make a PR

> As for why the ID is dracut: I don't know. But it is possible that some 3rd party modules rely on the information now that it's there. Maybe @haraldh can shed some light here. Anyway, if we change it, we need to mention in NEWS.

I doubt that anybody is relying on it. But yeah we should include that in the NEWS.

modules.d/99base/module-setup.sh
```diff
@@ -67,29 +67,24 @@ install() {
...
     local VERSION=""
     local PRETTY_NAME=""
     if [ -e /etc/os-release ]; then
         . /etc/os-release
+        grep -hE -ve '^VERSION=' -ve '^PRETTY_NAME' /etc/os-release >${initdir}/usr/lib/os-release
         [[ -n ${VERSION} ]] && VERSION+=" "
         [[ -n ${PRETTY_NAME} ]] && PRETTY_NAME+=" "
     fi
-    NAME=dracut
-    ID=dracut
```
haraldh: Better move that up before the `. /etc/os-release` in case it does not exist, so have a default value.

modules.d/99base/module-setup.sh
```diff
@@ -67,29 +67,24 @@ install() {
...
     VERSION+="dracut-$DRACUT_VERSION"
     PRETTY_NAME+="dracut-$DRACUT_VERSION (Initramfs)"
     VERSION_ID=$DRACUT_VERSION
-    ANSI_COLOR="0;34"
```
haraldh: Move that up, too.

modules.d/99base/module-setup.sh
```diff
@@ -67,29 +67,24 @@ install() {
...
-    [ -e "${initdir}/usr/lib" ] || mkdir -m 0755 -p ${initdir}/usr/lib
     {
-        echo NAME=\"$NAME\"
         echo VERSION=\"$VERSION\"
-        echo ID=$ID
-        echo VERSION_ID=$VERSION_ID
         echo PRETTY_NAME=\"$PRETTY_NAME\"
-        echo ANSI_COLOR=\"$ANSI_COLOR\"
-    } > $initdir/usr/lib/initrd-release
+        echo 'ANSI_COLOR="0;34"'
```
haraldh: and remove this

modules.d/99base/module-setup.sh
cgwalters: Or, just hard fail if the system doesn't have an `/etc/os-release`?

haraldh: > Or, just hard fail if the system doesn't have an `/usr/lib/os-release` (or `/etc/os-release`)?

Nah, as we learned, there are quite a few distributions using dracut without systemd. *BSD even.

cgwalters: :scream: :cry: 
Feels like it'd make sense for them to fork and have one dracut focused solely on systemd.  But OK not for this PR...

cgwalters:
I merged in https://github.com/danimo/dracut/commit/1b567c09ed71a84337348e4c76179bb205c8a55b
(but forgot to do https://github.com/danimo/dracut/commit/af072ff53dd0711421191e5212f93ea0a1525005 - I can rebase if you care)
Also addressed @haraldh 's comments.

danimo:
@cgwalters Please go ahead.

cgwalters:
> @cgwalters Please go ahead (and rebase).

OK done.

Issue #635, opened 2019-09-04
failed to install module because of missing whitespace in add_drivers


mapausch:
Please understand this more as a feature request than a issue.

Today I found me in the same situation like this: http://teridon73.blogspot.com/2019/08/rhel77-dracut-failed-to-install-module.html

I know, that `man dracut.conf`
says:

> Warning
Space-separated lists have to have a leading and trailing space!

But as it seems even RHEL package maintainers are not fully aware of this Warning, because this gets shipped by default with RHEL.

In my situation I also have 2 relevant files in /etc/dracut.conf.d:
```
[user@host dracut.conf.d]$ cat nvdimm-security.conf
# Make sure libnvdimm is loaded and ndctl is available in the initramfs
install_items+="/bin/ndctl"
add_drivers+=" libnvdimm "

[user@host dracut.conf.d]$ cat vmware-tools.conf
add_drivers+=" vmxnet3 vmw_pvscsi "
```

(by default both files didn't contain the leading and trailing spaces in the add_drivers lines).

So my feature request is:
Please enhance dracut in a way that it doesn't just concatenate the lists from multiple conf files when parsing them.

PR #657, opened 2019-10-23
dracut.sh: add check for invalid configuration files

haraldh:
Emit a warning about possible misconfigured configuration files, where
the spaces around values are missing for +=""

Better report a possible source of problems. We can fix annoying false
positives later.

Kind of fixes https://github.com/dracutdevs/dracut/issues/635


danimo:
I think this is a good approach. If we could catch `foo+=bar`s (i.e. forgotten quotes) as well, that would be even better, since they are accepted by the shell perfectly.

PR #659, opened 2019-10-29
Do not call cryptsetup in a systemd environment

### LINKED ISSUE #602: crypt-run-generator.sh depends on cryptsetup binary when not strictly required
By relying on the line:

https://github.com/dracutdevs/dracut/blob/8238f41b34777c3c75b03a9b598d4b9e1c7bf911/modules.d/90crypt/crypt-run-generator.sh#L14

The entire ``if`` block is skipped if the command fails e.g. if ``cryptsetup`` is not present. Under systemd this command is not required, so the failure incorrectly skips all discard flag processing.

danimo:
systemd provides its own cryptsetup facilities, and the
cryptsetup binary might not even exist, failing
to execute the discard flag processing.

Fixes #602

### DIFF
--- a/modules.d/90crypt/crypt-run-generator.sh
+++ b/modules.d/90crypt/crypt-run-generator.sh
@@ -11,7 +11,7 @@ crypttab_contains "$luks" "$dev" && exit 0
 allowdiscards="-"
 
 # parse for allow-discards
-if strstr "$(cryptsetup --help)" "allow-discards"; then
+if [ -n "$DRACUT_SYSTEMD" ] || strstr "$(cryptsetup --help)" "allow-discards"; then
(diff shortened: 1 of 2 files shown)


haraldh:
Hmm... `crypt-run-generator.sh` only ever runs in systemd mode... so  [ -n "$DRACUT_SYSTEMD" ] is always true.

haraldh:
> Hmm... `crypt-run-generator.sh` only ever runs in systemd mode... so [ -n "$DRACUT_SYSTEMD" ] is always true.

oh.. `module-setup.sh` even might be wrong... and we should move that out of the systemd block

danimo:
Moved the grypt-run-generator script outside the systemd module condition. Made the commit naming more consistent. Ok now?

danimo:
ping?

PR #660, opened 2019-10-29
Documentation: dynamically set current version

### LINKED ISSUE #369: Bump version in dracut.asc
I'm not familiar with the dracut release process, but if there is a checklist, bumping the release version and date should be added. 

In theory, this could be done through variables (dynamic replacement), but especially the date would make the builds non-reproducible. Also, the release number is not indicated anywhere except for the NEWS file (and we surely wouldn't want to parse it from there).

Either way, the current version statement `v3.0, October 2013` is most certainly wrong :)

danimo:
Fixes #369

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,13 +92,16 @@ endif
 
 %.xml: %.asc
 	@rm -f -- "$@"
-	asciidoc -d manpage -b docbook -o "$@" $<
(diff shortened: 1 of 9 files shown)


haraldh:
excellent.. thank you!

PR #662, opened 2019-10-29
Shutdown umount loop early-terminates on failure, not on success

### LINKED ISSUE #483: Shutdown umount loop early-terminates on failure, not on success
From modules.d/99shutdown/shutdown.sh
````
while [ $_cnt -le 40 ]; do
    umount_a 2>/dev/null || break
    _cnt=$(($_cnt+1))
done
````
umount_a returns 0 on success, so the second line should be
````
    umount_a 2>/dev/null && break
````

danimo:
umount_a returns 0 on success, fix the logic accordingly.

Fixes #483

Patch-by: Martin Whitaker <foss@martin-whitaker.me.uk>

### DIFF
--- a/modules.d/99shutdown/shutdown.sh
+++ b/modules.d/99shutdown/shutdown.sh
@@ -44,7 +44,7 @@ source_hook pre-shutdown
 
 warn "Killing all remaining processes"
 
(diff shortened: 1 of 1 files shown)


danimo:
Note: this needs reworking. See discussion on the bug report.

PR #663, opened 2019-10-30
99base: Remove duplicate nfsroot_to_var from dracut-lib.sh

### LINKED ISSUE #17: where should nfsroot_to_var() be defined?
It is defined in dracut-lib.sh and in 95nfs/nfs-lib.sh. The definitions are identical.

It's _used_ in 95nfs/nfs-lib.sh and in 95nfs/parse-nfsroot.sh, but the latter only sources dracut-lib, not nfs-lib.

Commit 9fcfa04c updated 95nfs/nfsroot.sh to source and use nfs-lib.sh, but it doesn't seem that was ever done to 95nfs/parse-nfsroot.sh. Was it supposed to be? That would allow the copy of nfsroot_to_var() in dracut-lib to be removed, I assume, in favor of the one in nfs-lib where it seems most at home.

danimo:
It already lives in nfs-lib.sh, which is the more correct library scope.

Fixes #17

### DIFF
--- a/modules.d/95nfs/parse-nfsroot.sh
+++ b/modules.d/95nfs/parse-nfsroot.sh
@@ -24,6 +24,7 @@
 #
 
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
+ . /lib/nfs-lib.sh
 
 # This script is sourced, so root should be set. But let's be paranoid
 [ -z "$root" ] && root=$(getarg root=)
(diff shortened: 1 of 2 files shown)


haraldh:
you will have to source `/lib/nfs-lib.sh` then also in `parse-nfsroot.sh`

danimo:
Done.

PR #664, opened 2019-10-30
90kernel-modules: don't install any block driver if not needed

ryncsn:
If hostonly mode is in use, and there isn't any block device included
as hostonly device, then the initramfs won't need any block driver.

Eg. for kdump built image, there could be only one nfs mount point,
or the initramfs may only execute some network operation then reboot.

In such case, skip block driver installation to save space and time.

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -2,6 +2,8 @@
 
 # called by dracut
 installkernel() {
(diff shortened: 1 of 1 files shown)


ryncsn:
Hi, @haraldh any thought on this? This help reduces the memory usage for initramfs that will only access the network, which solve some kdump OOM issue.

ryncsn:
Hi, I have tested it multiple times and so far it worked well, and it can help reduce memory usage in kdump, can any one help give some comments?

PR #665, opened 2019-10-30
TEST-30-ISCSI: test iBFT boot as well

lkundrak:
Setting up the machinery to boot with the aid of real networked boot loader
(such as iPXE) would involve much hassle, including possibly serving the kernel
and initrd via TFTP, etc.

Let us generate the iBFT table ourselves, with a Perl script. Include the
pregenerated table as well so that the test run won't depend on Perl. In the
end it's just reproducibly built static data, totally independent of the host
system.


lkundrak:
Note the test is going to fail due to https://github.com/dracutdevs/dracut/pull/666

mrc0mmand:
@centos-ci retest this please

PR #666, opened 2019-10-30
network-manager: fix getting of ifname from the sysfs path

lkundrak:
commit 5e0f8c8a4ced ('network-manager: remove useless use of basename')
somewhat carelessly didn't take into account that $_i has a slash at
the end which made the result of the ## substitution be just an empty
string.

The slash was put to the end of /sys/class/net/*/ to make sure we're only
iterating directories, but it's not strictly necessary. In an unlikely case
something else than a directory appears in /sys/class/net/, we'll already deal
with it gracefully. Remove it.

This fixes the TEST-30-ISCSI test.

### DIFF
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -6,7 +6,7 @@ else
     /usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
 fi
 
-for _i in /sys/class/net/*/
+for _i in /sys/class/net/*
 do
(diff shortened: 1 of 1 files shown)


danimo:
That test still fails, but travis does not reveal the cause. Can you check?

lkundrak:
> That test still fails, but travis does not reveal the cause. Can you check?

More tests are failing and need investigation, I don't know what went wrong there. I'll try to take a look as time permits.

I'm pretty sure I didn't break this though, it passes on my machine.

Issue #668, opened 2019-11-02
$UDEVVERSION may be empty


mikhailnov:
`prepare_udev_rules()` https://github.com/dracutdevs/dracut/blob/master/dracut-init.sh#L462 does not verify that $UDEVVERSION is not empty after `UDEVVERSION=$(udevadm --version)`

When udevadm can't be run for some reasons, e.g. as bellow, dracut scripts fails into a shell error.

![image](https://user-images.githubusercontent.com/15802528/68075290-bab7e580-fdb6-11e9-941b-a54d6eb95ab3.png)

PR #669, opened 2019-11-02
Fix dealing with udevadm --version reporting sth unexpected

mikhailnov:
Fixes https://github.com/dracutdevs/dracut/issues/668

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -465,6 +465,10 @@ prepare_udev_rules() {
         derror "Failed to detect udev version!"
         return 1
     fi
+    if [ -z "${UDEVVERSION##*[!0-9]*}" ]; then
+        derror "udevadm --version did not report an integer, udev version cannot be determined!"
+        return 1
+    fi
 
     for f in "$@"; do
         f="${initdir}/etc/udev/rules.d/$f"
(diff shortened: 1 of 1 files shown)


dracut-init.sh
```diff
@@ -461,6 +461,14 @@ inst_rules_wildcard() {
 
 prepare_udev_rules() {
     [ -z "$UDEVVERSION" ] && export UDEVVERSION=$(udevadm --version)
+    if [ -z "$UDEVVERSION" ]; then
+        derror "Failed to detect udev version"
```
danimo: Punctuation. Maybe Add '!' (see below)?

dracut-init.sh
```diff
@@ -461,6 +461,14 @@ inst_rules_wildcard() {
...
+        return 1
+    fi
+    if [ -z "${UDEVVERSION##*[!0-9]*}" ]; then
+        derror "udevadm --version reported not an integer, can't handle it"
```
danimo: Spelling: "...did not report an integer, ...".  can't -> cannot.

And "can't handle it is a bit implicit with derror, isn't it? ;-) " Maybe better:

"udevadm --version did not report an integer, udev version cannot be determined!"

(unfortunately, dracut is not consistent when it comes to punctuation here)

mikhailnov: Done, force-pushed.

mikhailnov:
@danimo , applied what you suggested and force-pushed

PR #671, opened 2019-11-08
Configure and requirement

zdtcd:
The configure script change is used for VoidLinux. It may be too local. I can drop it if it's undesirable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ install/util.o: install/util.c install/util.h install/macro.h install/log.h
 install/strv.o: install/strv.c install/strv.h install/util.h install/macro.h install/log.h
 
 install/dracut-install: $(DRACUT_INSTALL_OBJECTS)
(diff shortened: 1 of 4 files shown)


zdtcd:
On 2019-11-08 02:03:29 -0800, Daniel Molkentin wrote:
> danimo commented on this pull request.
> 
> Thanks for your contribution. The kpartx changes are trivial enough,
> as for the fts interface of some libc implementations, I think the
> check doesn't hurt.
> 
> Btw: I read musls FAQ and they mention using an alternative
> interface, which one would that be? FWIW, glibc now supports
> `_FILE_OFFSET_BITS=64` in recent versions.

musl libc doesn't ship anything for fts.

I think they meant every software vendor should include their own
version of fts (let's say: taken from BSD).

VoidLinux chose to ship a copy of NetBSD fts to reduce the workload.

-- 
Danh

danimo:
Ok then. Thanks for the clarification.

PR #676, opened 2019-11-11
Revert "travis: run TEST-14-IMSM on Fedora 29"

lkundrak:
We're on Fedora 31 that should be good enough for this test.

This reverts commit 7a2503ab8cec9473ad6dacf4f48b17344219f3c0.


danimo:
@lkundrak can you fix the conflict? What about test 13?

lkundrak:
> can you fix the conflict?

Done.

> What about test 13?

Fixed already.

PR #681, opened 2019-11-12
Replace ln with systemctl

jamacku:
Replace manual creating of symbolic links via ``mkdir`` and ``ln`` with ``systemctl`` calls.

This patch also unify how symbolic links are created in ``module-setup.sh`` files. It might help new contributors with creating new dracut modules.

### DIFF
--- a/modules.d/95iscsi/module-setup.sh
+++ b/modules.d/95iscsi/module-setup.sh
@@ -253,20 +253,18 @@ install() {
                       $systemdsystemunitdir/iscsiuio.socket \
                       iscsiadm iscsid
 
-        mkdir -p "${initdir}/$systemdsystemunitdir/sockets.target.wants"
         for i in \
                 iscsid.socket \
                 iscsiuio.socket \
             ; do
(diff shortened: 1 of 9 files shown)


modules.d/95iscsi/module-setup.sh
```diff
@@ -253,20 +253,13 @@ install() {
...
-            ln_r "$systemdsystemunitdir/${i}" "$systemdsystemunitdir/sockets.target.wants/${i}"
-        done
-
-        mkdir -p "${initdir}/$systemdsystemunitdir/basic.target.wants"
-        for i in \
                 iscsid.service \
                 iscsiuio.service \
             ; do
-            ln_r "$systemdsystemunitdir/${i}" "$systemdsystemunitdir/basic.target.wants/${i}"
+            systemctl --root "$initdir" enable "$i"
```
haraldh: iscsid.service has:
```
[Install]
WantedBy=multi-user.target
```
Which is not the `basic.target`. This is why I didn't use systemctl in the first place. There is no `multi-user.target` in the initramfs.

jamacku: Thank you. I didn't realize it.

PR #685, opened 2019-11-14
95ssh-client: improve nsswitch.conf parser

ptesarik:
The actual syntax of nsswitch.conf is slightly different from the
current regular expression:

- A comment can appear anywhere, not just at the beginning of a
  line.
- Action items take the general form [STATUS=ACTION] or
  [!STATUS=ACTION] (with some optional whitespace).
- A service name cannot include a colon. This is a nitpick,
  because there should never be more than a single colon per
  line.

Signed-off-by: Petr Tesarik <ptesarik@suse.com>

### DIFF
--- a/modules.d/95ssh-client/module-setup.sh
+++ b/modules.d/95ssh-client/module-setup.sh
@@ -68,7 +68,9 @@ install() {
     inst_multiple ssh scp
     inst_sshenv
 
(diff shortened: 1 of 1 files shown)


ptesarik:
Ugh. I don't see how my commit caused such breakage. Did the checks already fail before?

danimo:
No worries, the tests are currently broken when run in public CI places, due to several issues involving timing and the like. We are currently fixing them.

danimo:
Thanks!

PR #687, opened 2019-11-19
dmsquash-live: add rd.live.join

udeved:
Support a union join on rootfs wiih another img.

This makes throw away layers possible, or starting different setups depending on rootfs and the join.img.

Example:
* the live bootloader features a base system and a desktop system entry, where the desktop fs resides in the join.img, while rootfs would be just a console base system.
* throw away live layer, the installer extracts the rootfs, while live session fs resides in the join.img, including the installer itself

Usage:

`rd.live.join=$name.img`

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -26,6 +26,7 @@ getargbool 0 rd.live.ram -d -y live_ram && live_ram="yes"
 getargbool 0 rd.live.overlay.reset -d -y reset_overlay && reset_overlay="yes"
 getargbool 0 rd.live.overlay.readonly -d -y readonly_overlay && readonly_overlay="--readonly" || readonly_overlay=""
 overlay=$(getarg rd.live.overlay -d overlay)
+join=$(getarg rd.live.join -d join)
 getargbool 0 rd.writable.fsimg -d -y writable_fsimg && writable_fsimg="yes"
 overlay_size=$(getarg rd.live.overlay.size=)
 [ -z "$overlay_size" ] && overlay_size=32768
@@ -286,6 +287,9 @@ fi
 if [ -e /run/initramfs/live/${live_dir}/${squash_image} ]; then
     SQUASHED="/run/initramfs/live/${live_dir}/${squash_image}"
 fi
+if [ -e /run/initramfs/live/${live_dir}/${join} ]; then
+    SQUASHED_JOIN="/run/initramfs/live/${live_dir}/${join}"
+fi
 if [ -e "$SQUASHED" ]; then
     if [ -n "$live_ram" ]; then
         echo 'Copying live image to RAM...' > /dev/kmsg
@@ -293,6 +297,13 @@ if [ -e "$SQUASHED" ]; then
         dd if=$SQUASHED of=/run/initramfs/squashed.img bs=512 2> /dev/null
         echo 'Done copying live image to RAM.' > /dev/kmsg
         SQUASHED="/run/initramfs/squashed.img"
+        if [ -n ${join} ];then
+            echo 'Copying live image to RAM...' > /dev/kmsg
+            echo ' (this may take a minute)' > /dev/kmsg
+            dd if=$SQUASHED_JOIN of=/run/initramfs/${join} bs=512 2> /dev/null
+            echo 'Done copying live image to RAM.' > /dev/kmsg
+            SQUASHED_JOIN="/run/initramfs/${join}"
+        fi
     fi
 
     SQUASHED_LOOPDEV=$( losetup -f )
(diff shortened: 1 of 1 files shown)


haraldh:
@FGrose could you please review?

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -286,28 +287,49 @@ fi
...
+    SQUASHED_JOIN="/run/initramfs/live/${live_dir}/${join}"
+fi
 if [ -e "$SQUASHED" ]; then
     if [ -n "$live_ram" ]; then
         echo 'Copying live image to RAM...' > /dev/kmsg
         echo ' (this may take a minute)' > /dev/kmsg
         dd if=$SQUASHED of=/run/initramfs/squashed.img bs=512 2> /dev/null
         echo 'Done copying live image to RAM.' > /dev/kmsg
         SQUASHED="/run/initramfs/squashed.img"
+        if [ -n ${join} ];then
```
FGrose:
```suggestion
        if [ -n ${join} ]; then
```
Did you intend to support the live_ram option?  If so, please verify that this new code path works.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -286,28 +287,49 @@ fi
...
     SQUASHED_LOOPDEV=$( losetup -f )
     losetup -r $SQUASHED_LOOPDEV $SQUASHED
     mkdir -m 0755 -p /run/initramfs/squashfs
     mount -n -t squashfs -o ro $SQUASHED_LOOPDEV /run/initramfs/squashfs
 
+    if [ -n ${join} ]; then
+        SQUASHED_JOIN_LOOPDEV=$( losetup -f )
+        losetup -r $SQUASHED_JOIN_LOOPDEV $SQUASHED_JOIN
+        mkdir -m 0755 -p /run/initramfs/joinfs
+        mount -n -t squashfs -o ro $SQUASHED_LOOPDEV /run/initramfs/joinfs
```
FGrose:
```suggestion
        mount -n -t squashfs -o ro $SQUASHED_JOIN_LOOPDEV /run/initramfs/joinfs
```
Did you intend to support an unflat join.img?  I would recommend not to as you require an OverlayFS for the union join.  (You don't use the /run/initramfs/joinfs mountpoint anyway.  This section is superfluous.)

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -286,28 +287,49 @@ fi
...
+    fi
+
     if [ -d /run/initramfs/squashfs/LiveOS ]; then
         if [ -f /run/initramfs/squashfs/LiveOS/rootfs.img ]; then
             FSIMG="/run/initramfs/squashfs/LiveOS/rootfs.img"
         elif [ -f /run/initramfs/squashfs/LiveOS/ext3fs.img ]; then
             FSIMG="/run/initramfs/squashfs/LiveOS/ext3fs.img"
         fi
+        if [ -f /run/initramfs/squashfs/LiveOS/${join} ]; then
```
FGrose: This is looking for the join.img inside the squashfs.img.  Not correct.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -323,12 +345,22 @@ else
     elif [ -e /run/initramfs/live/${live_dir}/ext3fs.img ]; then
         FSIMG="/run/initramfs/live/${live_dir}/ext3fs.img"
     fi
+    if [ -e /run/initramfs/live/${live_dir}/${join} ]; then
```
FGrose: Did you intend to support an uncompressed join.img?  If so, please verify that this new code path works.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -343,6 +375,15 @@ if [ -n "$FSIMG" ]; then
             unpack_archive $FSIMG /run/initramfs/fsimg/
         fi
         FSIMG=/run/initramfs/fsimg/rootfs.img
+
+        if [ -n ${join} ]; then
```
FGrose: Did you intend to support the 'writeable_fsimg' option?  If so, please verify that this new code path works.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -358,12 +399,23 @@ if [ -n "$FSIMG" ]; then
     fi
     if [ "$FSIMG" = "$SQUASHED" ]; then
         BASE_LOOPDEV=$SQUASHED_LOOPDEV
+        if [ -n ${join} ]; then
```
FGrose: Device-mapper based configurations will not support a union join of rootfs content with joinfs.  This code is superfluous, no?

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -390,12 +442,26 @@ if [ -n "$overlayfs" ]; then
...
+            ovlfs=lowerdir=/run/rootfsjoin:/run/rootfsbase
+        else
+            ovlfs=lowerdir=/run/rootfsbase
+        fi
     fi
     mount -r $FSIMG /run/rootfsbase
+    if [ -n ${join} ]; then
+        mount -r $FSIMG_JOIN /run/rootfsjoin
+    fi
     if [ -z "$DRACUT_SYSTEMD" ]; then
```
FGrose: In order to support DRACUT_SYSTEMD systems, the ovlfs variable in dmsquash-generator.sh must be similarly modified.

udeved:
@FGrose 

I am sorry for the late reply, as soon as I have enough time at hand, I will look more closely at it.
We still haven't switched to dracut, and this rough patch did the job to test our install media boot with dracut.
Tested was a flat squash image and a squashed ext image.

danimo:
@udeved any update?

PR #688, opened 2019-11-21
Add support for dbus-broker

jamacku:


danimo:
I would like to consolidate this into `06dbus` (just landed in master). This should already work with dbus-broker (but untested). Can you test?

PR #690, opened 2019-11-26
35network-legacy: using 'replace' instead of 'add' to add route

### LINKED ISSUE #683: 35network-legacy: DHCP client MUST ignore the Router option if a Classless Static Routes option exists
According to RFC 3442:

> If the DHCP server returns both a Classless Static Routes option and
> a Router option, the DHCP client MUST ignore the Router option.

According to the following configuration file, the client requests both the
router option and classless static route option in the DHCP Parameter Request
List.
https://github.com/dracutdevs/dracut/blob/9a752007b3d33c97a0337d6e2dfa268378fe14b4/modules.d/35network-legacy/dhclient.conf#L6-L8

So the DHCP server sends back a response with both options.

Expected:
  The client should not handle the router option.
Actual:
  It adds both routes and the second one fails because "RTNETLINK answers: File exists".

wenjianhn:
This is a simple trick to honor RFC 3442:
If the DHCP server returns both a Classless Static Routes option and
a Router option, the DHCP client MUST ignore the Router option.

Fixes #683 

Signed-off-by: Jacob Wen <jian.w.wen@oracle.com>

### DIFF
--- a/modules.d/35network-legacy/dhclient-script.sh
+++ b/modules.d/35network-legacy/dhclient-script.sh
@@ -146,7 +146,7 @@ parse_option_121() {
             temp_result="$destination via $gateway dev $interface"
         fi
 
-        echo "/sbin/ip route add $temp_result"
+        echo "/sbin/ip route replace $temp_result"
     done
 }
 


wenjianhn:
@centos-ci Yes. The patch comes from one of our admins(Oracle cloud).

PR #693, opened 2019-12-02
network-manager: don't run NetworkManager when there are no connections

lkundrak:
NetworkManager would unnecessarily bring up the devices, colliding with
further attempts to rename the devices.

This is arguably a NetworkManager bug and should eventually be fixed there.
Running NetworkManager without the connection is unnecessary regardless.


cgwalters:
This can only work in hostonly mode, right?  For FCOS we want to use NM-initrd for generic (rpm-ostree server side) initramfs images.

cgwalters:
I guess we could create a generic ethernet connection in `/usr` in our initramfs or so?  Feels...odd though.

PR #696, opened 2019-12-13
Check .hmac of boot.iso in fips mode

bcl:
Instead of carrying the kernel and hmac in the install.img (which takes
up more space) use the vmlinuz installed under /images/pxeboot/ and the
.hmac in the install.img /boot.

This check is triggered by the presence of /run/install/repo/images/pxeboot/vmlinuz

Related: rhbz#1782737


jstodola:
This will work for x86_64 and aarch64, but not for ppc64le and s390x, since the paths to the kernel image are different there:
ppc64le: /run/install/repo/ppc/ppc64/vmlinuz
s390x: /run/install/repo/images/kernel.img

haraldh:
> This will work for x86_64 and aarch64, but not for ppc64le and s390x, since the paths to the kernel image are different there:
> ppc64le: /run/install/repo/ppc/ppc64/vmlinuz
> s390x: /run/install/repo/images/kernel.img

cool!

PR #697, opened 2019-12-15
iscsi: Fix ibft boot with network-manager module

ryncsn:
Currently with 35network-manager iscsi ibft won't work because parse-ibft hook is missing and the iscsi cmdline could be duplicated and flawed. This PR should fix it. Tested on Fedora.


lkundrak:
No, this is completely wrong.
nm-initrd-generator is doing the ibft parsing.

lkundrak:
If it won't work for you, please file an issue with details.

ryncsn:
@lkundrak Yes you are right, but I have a question, nm-initrd-generator seems to expect one to pass ip=ibft to work, but the parse-ibft.sh will automatically append the network configure required by ibtf to cmdline when ever it found there is ibft info. So some previous working enviroment is
now broken.
So when using network-manager module, should we let dracut append ip=ibft to cmdline when ibft is detected?

ryncsn:
I think anyway the second commit is useless then, the first commit should still be a valid fix.

lkundrak:
> @lkundrak Yes you are right, but I have a question, nm-initrd-generator seems to expect one to pass ip=ibft to work, but the parse-ibft.sh will automatically append the network configure required by ibtf to cmdline when ever it found there is ibft info. So some previous working enviroment is
> now broken.
> So when using network-manager module, should we let dracut append ip=ibft to cmdline when ibft is detected?

Thanks for the response. You need *some* way to tell the nm-initrd-generator to parse the iBFT. What does your kernel command line look like?

If you were already using rd.iscsi.ibft, then the last commit here https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/393 should fix the issue,

haraldh:
@lkundrak what's the status?

ryncsn:
> > @lkundrak Yes you are right, but I have a question, nm-initrd-generator seems to expect one to pass ip=ibft to work, but the parse-ibft.sh will automatically append the network configure required by ibtf to cmdline when ever it found there is ibft info. So some previous working enviroment is
> > now broken.
> > So when using network-manager module, should we let dracut append ip=ibft to cmdline when ibft is detected?
> 
> Thanks for the response. You need _some_ way to tell the nm-initrd-generator to parse the iBFT. What does your kernel command line look like?
> 
> If you were already using rd.iscsi.ibft, then the last commit here https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/393 should fix the issue,

I remember before the network-manager dracut module is introduced, dracut can just detect and read ibft from firmware interface and don't need extra cmdline.
Now it always need "ip=ibft"? On the machine I can access which have ibft support,  network-legacy doesn't need this cmdline param.

bengal:
> I remember before the network-manager dracut module is introduced, dracut can just detect and read ibft from firmware interface and don't need extra cmdline.
> Now it always need "ip=ibft"? On the machine I can access which have ibft support, network-legacy doesn't need this cmdline param.

What is your kernel command line? With commit [1] (available in NM 1.20.6) `ip=ibft` is not required if you already have `rd.iscsi.ibft` set.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/9fddb395d5ed17f989773ec4ecf78120e6740848

ryncsn:
I think adding rd.iscsi.ibft will just fix the issue now, thanks.

PR #698, opened 2019-12-17
Fix 0-rescue update for a new kernel

alexbe:
It has never updated because vmlinuz-0-rescue-${MACHINE_ID} and initramfs-0-rescue-${MACHINE_ID}.img have always the same names

### DIFF
--- a/51-dracut-rescue.install
+++ b/51-dracut-rescue.install
@@ -75,8 +75,16 @@ ret=0
 
 case "$COMMAND" in
     add)
-        [[ -f "$LOADER_ENTRY" ]] && [[ -f "$BOOT_DIR_ABS/$KERNEL" ]] \
-            && [[ -f "$BOOT_DIR_ABS/$INITRD" ]] && exit 0
+        if [[ -f "$LOADER_ENTRY" ]] ; then
+            #for the same version
+			[[ $(uname -r) == "$KERNEL_VERSION" ]] \
+				&& [[ -f "$BOOT_DIR_ABS/$KERNEL" ]] && [[ -f "$BOOT_DIR_ABS/$INITRD" ]] \
+				&& exit 0
+            #if you want to update "$BOOT_DIR_ABS/$INITRD" just remove it and run
+            #kernel-install add $(uname -r) /lib/modules/$(uname -r)/vmlinuz
+			[[ -f "$BOOT_DIR_ABS/$KERNEL" ]] && mv "$BOOT_DIR_ABS/$KERNEL" "$BOOT_DIR_ABS/$KERNEL.bkp"
+            [[ -f "$BOOT_DIR_ABS/$INITRD" ]] && mv "$BOOT_DIR_ABS/$INITRD" "$BOOT_DIR_ABS/$INITRD.bkp"
+        fi
 
         # source our config dir
         for f in $(dropindirs_sort ".conf" "/etc/dracut.conf.d" "/usr/lib/dracut/dracut.conf.d"); do


51-dracut-rescue.install
```diff
@@ -75,8 +75,16 @@ ret=0
 
 case "$COMMAND" in
     add)
-        [[ -f "$LOADER_ENTRY" ]] && [[ -f "$BOOT_DIR_ABS/$KERNEL" ]] \
-            && [[ -f "$BOOT_DIR_ABS/$INITRD" ]] && exit 0
+        if [[ -f "$LOADER_ENTRY" ]] ; then
+            #for the same version
+			[[ $(uname -r) == "$KERNEL_VERSION" ]] \
+				&& [[ -f "$BOOT_DIR_ABS/$KERNEL" ]] && [[ -f "$BOOT_DIR_ABS/$INITRD" ]] \
+				&& exit 0
```
martinezjavier: @alexbe the idea of the rescue kernel and initrd is to have images that are known to work correctly. In most cases (unless they are removed) these will contain the kernel and modules that were used during installation, since the installation succeeded it is assumed that these work correctly.

But if you re-generate the rescue initrd for every kernel install, then if a kernel has bugs you don't have a rescue image anymore that can be trusted.

alexbe: I support the idea to not remove old kernels, but standard behavior is when you upgrade kernel rpm modules of the old package are removed.

martinezjavier: I don't see how those two things are related. Old kernels would be removed but the `/boot/*-0-rescue-$(cat /etc/machine-id)*` files are not owned by any package so are not affected by rpm packages being installed / removed.

So even when the kernel and initrd that were originally used to copy the rescue images are removed, the rescue images will remain in your system. As explained because the rationale is that if it worked correctly to install the system, it can be used to rescue the system in case of a failure.

With your change, that assumption is changed and the rescue images will always be the last installed kernel version but just with an initrd generated with `--no-hostonly`.

alexbe: I agree with you about these two files, kernel & initramfs, but the kernel uses /lib/modules/<kernel-version> and there are no /lib/modules/0-rescue-* modules. So if we upgrade kernel package  the kernel /boot/vmlinuz-0-rescue-* will lose it's own modules

martinezjavier: > kernel /boot/vmlinuz-0-rescue-* will lose it's own modules

The modules will be present in the rescue initramfs. And since is an initrd image generated with `dracut --no-hostonly', it shouldn't be an issue since the list of modules in the initramfs will be fairly complete. This is how rescue images have always worked.

I agree that we could discuss if getting rescue images to become stale is the best approach or if instead the rescue image should be re-generated after a system wide upgrade or if the current kernel that was originally used as the rescue is deleted, etc.

But both the rationale and then change itself in this PR are not correct in my opinion. Since it will just override the rescue image on every kernel install and you could end with a rescue image that doesn't work.

alexbe: OK, i've understood your point, but it is possible to make the script more flexible, because sometimes we need fix or update system. My proposition is to add third option e.g. "update" along with "add" and "remove"

martinezjavier: Agreed about adding another option for that.

haraldh:
Maybe `dracut --update-rescue`

fadamo:
In my case I know the original rescue image is not working anymore because we update CNA firmware and it needs a new kernel module. I need a proper way to update the rescue image.

haraldh:
> In my case I know the original rescue image is not working anymore because we update CNA firmware and it needs a new kernel module. I need a proper way to update the rescue image.

Just remove them and a new kernel update will install new ones.

PR #702, opened 2020-01-07
modules/network-manager: Install `ip`

cgwalters:
We don't need `ip` but having it is *really* useful for people debugging
in an emergency shell.


thom311:
Also fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1784358

PR #709, opened 2020-01-21
dracut-install: fetch kernel modules from kmoddir

0xdc:
dracut, when passing --kmoddir, would only install modules that exist at
the same location as it will appear in the initramfs.

For most (possibly all) outputs, the output path would be /lib/modules,
so previously all kernel modules must be copied there or the module
installation silently fails.

Now use the original path variable, and install to the shortened path
variable (as calculated by kerneldirlen).

Note: This means that the argument to --kmoddir (and the modules filetree)
should be a directory that ends in 'lib/modules/$kernelversion'.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -915,6 +915,21 @@ esac
 
 [[ $reproducible == yes ]] && DRACUT_REPRODUCIBLE=1
 
+case "${drivers_dir}" in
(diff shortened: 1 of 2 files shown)


haraldh:
> Note: This means that the argument to --kmoddir (and the modules filetree)
> should be a directory that ends in 'lib/modules/$kernelversion'.

Could you add a sanity check in dracut.sh to ensure this?

Issue #398, opened 2018-04-28
fsfreeze can cause problems on Btrfs


cmurf:
```
$ grep -n fsfreeze /usr/bin/dracut
1840:# use fsfreeze only if we're not writing to /
1842:    if ! $(fsfreeze -f $(dirname "$outfile") 2>/dev/null && fsfreeze -u $(dirname "$outfile") 2>/dev/null); then
1843:        dinfo "dracut: warning: could not fsfreeze $(dirname "$outfile")"
[chris@f28s ~]$ 
```

When there is a boot subvolume, and fstab uses mount option subvol=boot to mount at /boot, the above check still allows fsfreeze to happen on the entire Btrfs volume which includes /

If I also decide to do something a bit nutty:

`sudo strace -s 60 -ff -o stracednfinstallkernel.out dnf install -y ~/kernelrpms` where the destination for the 7000+ files this writes out is root fs. This never completes. The fsfreeze -f is issued and it never finishes (why it doesn't finish is uncertain and might be a kernel bug, it'll take more testing). Also the resulting file system is not bootable by default, the freeze starts before grubby has added the initrd line, so the default GRUB entry isn't bootable. (At least on Fedora which uses grubby.)

Anyway, so far in my testing sync() is sufficient on all file systems to successfully flush boot related files as far as the bootloader is concerned, *except* XFS.

cmurf:
sysrq+t issued during the hung freeze
[journal.log](https://github.com/dracutdevs/dracut/files/1957496/journal.log)

cmurf:
I can reproduce this on XFS by making /boot a directory on rootfs; and then modifying dracut do always `fsfreeze -f /boot && fsfreeze -u /boot`

I think this method of working around XFS behaving differently than FAT, ext4, and Btrfs when sync() is called means it should be reverted until the fs devs sort this out. Either XFS devs need sync() to work the same as other file systems; or fsdevs need a new atomic FIFREEZE+FITHAW ioctl. Obviously all /boot related changes care to have fully committed for bootloader purposes is kernel, initramfs, and the bootloader config files, not literally everything on the file system.

The previous dracut behavior used sync on the initramfs file, which sync.c in turn translates into fsync(). So I'm suggesting dracut use only sync, no options and no file specified. That will call sync() which other than XFS, so far in my testing commits files such that the bootloader sees the changes.

fcami:
I'll open a PR so that we don't do anything for btrfs and be done with it.
[...]

PR #713, opened 2020-01-30
dracut.sh: don't call fsfreeze on subvol of root file system

mwilck:
dracut.sh already doesn't call fsfreeze if the output file is on
the root file system. For btrfs, however, this is not sufficient.
Because fsfreeze is a superblock operation, and all btrfs subvolumes
share the same superblock, fsfreeze may freeze the entire system
if the subvolume on which the output file is written and / are
subvolumes of the same file system. Avoid this by comparing file
system UUIDs for btrfs.

This should fix https://github.com/dracutdevs/dracut/issues/398.

Fixes: de576db3c225 ("call fsfreeze(8) on /boot to flush initramfs data & metadata to media")

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2057,6 +2057,40 @@ fi
 
 command -v restorecon &>/dev/null && restorecon -- "$outfile"
 
+btrfs_uuid() {
+    btrfs filesystem show "$1" | sed -n '1s/^.*uuid: //p'
+}
+
+freeze_ok_for_btrfs() {
+    local mnt uuid1 uuid2
+    # If the output file is on btrfs, we need to make sure that it's
+    # not on a subvolume of the same file system as the root FS.
+    # Otherwise, fsfreeze() might freeze the entire system.
+    # This is most conveniently checked by comparing the FS uuid.
+
+    [[ "$(stat -f -c %T -- "/")" == "btrfs" ]] || return 0
+    mnt=$(stat -c %m -- "$1")
+    uuid1=$(btrfs_uuid "$mnt")
+    uuid2=$(btrfs_uuid "/")
+    [[ "$uuid1" && "$uuid2" && "$uuid1" != "$uuid2" ]]
+}
+
+freeze_ok_for_fstype() {
+    local outfile=$1
+    local fstype
+
+    [[ "$(stat -c %m -- "$outfile")" == "/" ]] && return 1
+    fstype=$(stat -f -c %T -- "$outfile")
+    case $fstype in
+        msdos)
+            return 1;;
+        btrfs)
+            freeze_ok_for_btrfs "$outfile";;
+        *)
+            return 0;;
+    esac
+}
+
 # We sync/fsfreeze only if we're operating on a live booted system.
 # It's possible for e.g. `kernel` to be installed as an RPM BuildRequires or equivalent,
 # and there's no reason to sync, and *definitely* no reason to fsfreeze.
@@ -2069,7 +2103,7 @@ if test -d $dracutsysrootdir/run/systemd/system; then
     fi
 
     # use fsfreeze only if we're not writing to /
-    if [[ "$(stat -c %m -- "$outfile")" != "/" && "$(stat -f -c %T -- "$outfile")" != "msdos" ]]; then
+    if [[ "$(stat -c %m -- "$outfile")" != "/" ]] && freeze_ok_for_fstype "$outfile"; then
         if ! $(fsfreeze -f $(dirname "$outfile") 2>/dev/null && fsfreeze -u $(dirname "$outfile") 2>/dev/null); then
             dinfo "dracut: warning: could not fsfreeze $(dirname "$outfile")"
         fi


mwilck:
The failing CI tests are root on NFS - likely unrelated to my patch. Please confirm.

mwilck:
We've had intense internal discussion about this issue at SUSE. Here's a quote of one of our file system experts:

> Using the freeze/thaw interface for this is a hack [...].  Dracut doesn't actually want a frozen file 
> system.  It wants an interface that flushes everything to disk in a way that produces a clean file  
> system.  I don't think that's an unreasonable interface, but I'd suggest going back through the 
> fsdevel, ext4, and xfs lists for discussions on this topic.

This is what I think the long term goal should be. In the short term, we need a user space fix that avoids freezing the system while guaranteeing FS consistency in a majority of cases.

Our debugging has shown that problems occur if some output is written to the frozen filesystem.
The simple command sequence

     fsfreeze -f /some/fs; fsfreeze -u /some/fs

works basically on every file system. But if it's run e.g. though `sh -x 2>/some/fs/logfile` (like when dracut is started with `--debug`), the shell will hang after the `fsfreeze -f`, and not be able to unfreeze again, which is what we need to avoid. The only difference of **btrfs** wrt e.g. xfs is that the system may freeze also if the output file is on a different subvolume.

The most promising idea we've come up with so far is to write a mini-utility that does FIFREEZE / FITHAW (i.e. `fsfreeze -f` / `fsfreeze -u`) in a single program invocation. That should avoid >95% of possible problem cases (exception e.g.: a user could run dracut under `strace -f`).

Comments welcome.

haraldh:
fine with me.. if `fsfreeze` could be patched to do `-f` `-u` in one go or we have an mini-utility, we can alter this code path again

cmurf:
I was part of those upstream conversations. The reality is ext3/4 and especially XFS, do not guarantee full fs metadata update on sync(), initially just the journal is updated. And the journal isn't read by bootloaders.

Btrfs, vfat, ext2 all will do a full fs update on sync(). And in a script you can do 'sync -f initramfs' to get syncfs() on the underlying file system.

For ext3/4 you could just delay and it'll update pretty quickly unless really busy. XFS, busy or not, it might update fs metadata quickly or it might be minutes. I was never able to discover a pattern. Hence XFS developers said use one of: FIFREEZE/FITHAW or remount read-only or umount was necessary. It's not like some dracut people just made this up.

@cgwalters once asked upstream fs devs for an atomic FIFREEZE/FITHAW that was guaranteed to complete, but I can't find it, and I don't think it went anywhere. All I found is this:
https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-sysroot-deploy.c#L1313

mwilck:
> It's not like some dracut people just made this up.

My quote just called it a  "hack" because the current kernel lacks a generic  API (ioctl or what not) to tell journaled file systems to flush the log.

As you opened #398 yourself, you seem to have seen the back side of the coin. Grub not being able to read the FS after a crash or power outage is bad, but a system freezing with no remedy because of an innocently-looking command is bad, too.

> @cgwalters once asked upstream fs devs for an atomic FIFREEZE/FITHAW that was guaranteed to complete, but I can't find it, and I don't think it went anywhere. All I found is this:
> https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-sysroot-deploy.c#L1313

Good point. So probably the simplistic tool I envisioned would need to do something of that kind (fork a child that could unfreeze in case of timeout), too. The same would apply if we are looking for a solution on the shell level, we'd need to do something like the following, which would be the equivalent of the ostree code in sh:

    FREEZE_TIMEOUT=30
    { set +x; exec >/dev/null 2>&1; sleep $FREEZE_TIMEOUT; fsfreeze -u $MNTPNT; } &
    UNFREEZE=$!
    if fsfreeze -f $MNTPNT; then
       fsfreeze -u $MNTPNT && kill $UNFREEZE
    else
       kill $UNFREEZE
    fi

cmurf:
> My quote just called it a "hack" because the current kernel lacks a generic API (ioctl or what not) to tell journaled file systems to flush the log.

And I agree.
 
> As you opened #398 yourself, you seem to have seen the back side of the coin.

My regret is only hindsight being 20/20 and having parroted upstream fs kernel devs advice at face value. However, in their defense, they think bootloaders supporting complex file systems without completely supporting them, i.e. journal replay, is half-assed. And in defense of bootloader devs, that's difficult not least of which is the fs devs don't want to support journal replay in GRUB. Bit of a circular problem, which led one fs dev to opine that maybe downstreams should only be using FAT and ext2 for bootloaders. Kinda funny. However unlikely that is to become reality, a pretty strong argument in favor of keeping things simple makes them more reliable, /boot probably should just be FAT or ext2 or ext4 with journal disabled. Similarly, /boot probably should only be mounted on demand by programs that "own" it like kernel+initramfs installing, and bootloader updates; and umounted when all that's finished.

> Grub not being able to read the FS after a crash or power outage is bad, but a system freezing with no remedy because of an innocently-looking command is bad, too.

I'm completely on-board both being bad.

> Good point. So probably the simplistic tool I envisioned would need to do something of that kind (fork a child that could unfreeze in case of timeout), too. The same would apply if we are looking for a solution on the shell level, we'd need to do something like the following, which would be the equivalent of the ostree code in sh:

I ran into this hang too, when doing `strace -f -ff` I think on dracut. I could always unfreeze from another shell if it was already available, so the fork idea seems like a sane work around for what's kindof a funny problem in the first place. And I also remember if I used `-D` with that strace, the hang never happened either.

mwilck:
> However unlikely that is to become reality, a pretty strong argument in favor of keeping things simple makes them more reliable, /boot probably should just be FAT or ext2 or ext4 with journal disabled.

That doesn't work for SUSE, at least not short-term. It conflicts with SUSE's snapshot/rollback approach which requires `/boot` to be a subvolume of the BTRFS root.

cmurf:
> That doesn't work for SUSE, at least not short-term. It conflicts with SUSE's snapshot/rollback approach which requires `/boot` to be a subvolume of the BTRFS root.

I see Btrfs as an exception. While it's a complex file system, it conforms to a simple and old expectation of completely flushing on sync(), and GRUB has a consistent view of the file system.

cmurf:
fsfreeze deadlock on XFS (the FIFREEZE ioctl and subsequent FITHAW hang indefinitely) 
https://bugzilla.kernel.org/show_bug.cgi?id=207053

(Putting this here because at first glance it seems possibly related, but the bug says sysrq+j doesn't thaw which is inconsistent with `fsfreeze -u` working . Goes back to kernel 4.19.x apparently.)

PR #717, opened 2020-02-06
dracut: Verify multipath config_dir option

mpg-rh:
The 90multipath/module-setup.sh file currently does not check the
dm-multipath config_dir option. This option in multipath.conf file is
used to specify a custom directory/path that contains the multipath
configuration files. It's default value is /etc/multipath/conf.d

Currently install function of module-setup.sh has hardcoded the above
path, but users could change it with config_dir option. So, adding
steps to read the path specified with config_dir option and add
these configuration files to the initial ram disk image.

Signed-off-by: Milan P. Gandhi <mgandhi@redhat.com>
Reviewed-by: Martin Wilck <mwilck@suse.com>

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -79,6 +79,20 @@ install() {
         }
     }
 
+    # Include multipath configuration files from path specified with config_dir
+    while read _k _v; do
+        if [[ "$_k" = config_dir ]]; then
+            _v=${_v#\"}
+            config_dir=${_v%\"}
+        fi
+    done < <(/sbin/multipath -t)
+
+    if [ -d "$config_dir" ]; then
+        config_dir+="/*"
+    else
+        derror "multipath.conf: config_dir - No such directory $config_dir"
+    fi
+
     inst_multiple -o  \
         dmsetup \
         kpartx \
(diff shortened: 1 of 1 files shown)


haraldh:
In a default installation on fedora 31 `/etc/multipath/conf.d` does not exist.
So, this would error all the time?
Note: `inst_multiple -o ` skips non-existing files because of `-o`

haraldh:
Also `/sbin/multipath -t` gives me error messages on a default installation:
```
Feb 07 10:40:29 | /etc/multipath.conf does not exist, blacklisting all devices.
Feb 07 10:40:29 | You can run "/sbin/mpathconf --enable" to create
Feb 07 10:40:29 | /etc/multipath.conf. See man mpathconf(8) for more details
```

mwilck:
> Also `/sbin/multipath -t` gives me error messages on a default installation:
> 
> ```
> Feb 07 10:40:29 | /etc/multipath.conf does not exist, blacklisting all devices.
> Feb 07 10:40:29 | You can run "/sbin/mpathconf --enable" to create
> Feb 07 10:40:29 | /etc/multipath.conf. See man mpathconf(8) for more details
> ```

This is caused by a Red Hat specific patch, I believe. Other distros don't use mpathconf.

danimo:
What is the plan here?

PR #722, opened 2020-02-13
dracut.sh: added help for --regenerate-all

rmetrich:
`--regenerate-all` is not advertised in the help although it should since it's is common that initramfs must be rebuilt for all kernels on the system.

This commit also fixes some help layout.


danimo:
Thank you.

PR #725, opened 2020-02-15
Support the EFI Stub loader's splash image feature.

Neurognostic:
Support the EFI Stub loader's splash image feature.

Checks if `uefi_splash_image` exists in `dracutsysroot` if not unset
`uefi_splash_image`. Alternate Value parameter expansion adds section-vma
for splash image to EFI stub loader when the path to image is valid and
not an empty file.

I did not test on other distributions, but on Arch Linux the `systemd`
package includes a splash image at the path
 `/usr/share/systemd/bootctl/splash-arch.bmp`. Perhaps, if this is a
common practice, a default image could be gathered from that directory.

It is required that the image be in bitmap (`.bmp`) format according to 
`splash.c`.

The code for `stub.c` and `splash.c` can be found at:
https://github.com/systemd/systemd/blob/master/src/boot/efi/stub.c
https://github.com/systemd/systemd/blob/master/src/boot/efi/splash.c

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -236,6 +236,9 @@ Creates initial ramdisk images for preloading modules
   --uefi                Create an UEFI executable with the kernel cmdline and
                         kernel combined
   --uefi-stub [FILE]    Use the UEFI stub [FILE] to create an UEFI executable
+  --uefi-splash-image [FILE]
+                        Use [FILE] as a splash image when creating an UEFI
+                        executable
   --kernel-image [FILE] location of the kernel image
 
 If [LIST] has multiple arguments, then you have to put these in quotes.
@@ -398,6 +401,7 @@ rearrange_params()
         --long loginstall: \
         --long uefi \
         --long uefi-stub: \
+        --long uefi-splash-image: \
         --long kernel-image: \
         --long no-hostonly-i18n \
         --long hostonly-i18n \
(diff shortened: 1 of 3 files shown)


dracut.sh
```diff
@@ -2018,11 +2025,13 @@ if [[ $uefi = yes ]]; then
 
     [[ -s $dracutsysrootdir/usr/lib/os-release ]] && uefi_osrelease="$dracutsysrootdir/usr/lib/os-release"
     [[ -s $dracutsysrootdir/etc/os-release ]] && uefi_osrelease="$dracutsysrootdir/etc/os-release"
+    [[ -s "${dracutsysrootdir}${uefi_splash_image}" ]] || uefi_splash_image=/dev/null
 
     if objcopy \
            ${uefi_osrelease:+--add-section .osrel=$uefi_osrelease --change-section-vma .osrel=0x20000} \
            --add-section .cmdline="${uefi_outdir}/cmdline.txt" --change-section-vma .cmdline=0x30000 \
-           --add-section .linux="$kernel_image" --change-section-vma .linux=0x40000 \
+           --add-section .splash="$uefi_splash_image" --change-section-vma .splash=0x40000 \
```
haraldh:
```
  ${uefi_splash_image: --add-section .splash="$uefi_splash_image" --change-section-vma .splash=0x40000}
```
no need to set it to /dev/null ... just leave it empty

Neurognostic: I updated the code as you suggested.

dracut.sh
```diff
@@ -2018,11 +2025,14 @@ if [[ $uefi = yes ]]; then
 
     [[ -s $dracutsysrootdir/usr/lib/os-release ]] && uefi_osrelease="$dracutsysrootdir/usr/lib/os-release"
     [[ -s $dracutsysrootdir/etc/os-release ]] && uefi_osrelease="$dracutsysrootdir/etc/os-release"
+    [[ -s "${dracutsysrootdir}${uefi_splash_image}" ]] && \
```
Neurognostic: @haraldh, do you think there needs to be a `/` between `${dracutsysroot`} and `${uefi_splash_image}` in order to accept relative paths? _Or_ should only absolute paths pass and leave the code as is?

Issue #734, opened 2020-02-24
`TEST-60-IFCFG` and `TEST-70-BONDBRIDGETEAMVLAN` are broken due to changed persistent NIC names


mrc0mmand:
After getting some debug logs from the test's output, the issue seems to be pretty obvious, however, it's mildly infuriating, since the test uses persistent NIC names, which should be stable, thus preventing such scenarios in the first place.

The client's dracut cmdline for `TEST-60-IFCFG` [contains following line](https://github.com/dracutdevs/dracut/blob/master/test/TEST-60-IFCFG/test.sh#L151):
```
root=nfs:192.168.50.1:/nfs/client bootdev=ens3
```

which suggests the test was written when the NIC naming was in range `ens3-ens7` (this can be also confirmed by looking at cmdlines of following subtests).

However, after adding some debug commands into the dracut image, the NIC names seem to be in range `ens5-ens8`, which, of course, breaks the whole test:

```
[    4.366477] PCI Interrupt Link [LNKD] enabled at IRQ 10
[    4.837600] e1000 0000:00:04.0 eth0: (PCI:33MHz:32-bit) 52:54:00:12:34:01
[    4.839068] e1000 0000:00:04.0 eth0: Intel(R) PRO/1000 Network Connection
[    4.865247] PCI Interrupt Link [LNKA] enabled at IRQ 10
[    5.300681] e1000 0000:00:05.0 eth1: (PCI:33MHz:32-bit) 52:54:00:12:34:02
[    5.302149] e1000 0000:00:05.0 eth1: Intel(R) PRO/1000 Network Connection
[    5.326664] PCI Interrupt Link [LNKB] enabled at IRQ 11
[    5.777708] e1000 0000:00:06.0 eth2: (PCI:33MHz:32-bit) 52:54:00:12:34:03
[    5.778841] e1000 0000:00:06.0 eth2: Intel(R) PRO/1000 Network Connection
[    6.240581] e1000 0000:00:07.0 eth3: (PCI:33MHz:32-bit) 52:54:00:12:34:04
[    6.242012] e1000 0000:00:07.0 eth3: Intel(R) PRO/1000 Network Connection
[    6.695487] e1000 0000:00:08.0 eth4: (PCI:33MHz:32-bit) 52:54:00:12:34:05
[    6.696823] e1000 0000:00:08.0 eth4: Intel(R) PRO/1000 Network Connection
[    6.707139] e1000 0000:00:04.0 ens4: renamed from eth0
[...]

Issue #732, opened 2020-02-22
`network-legacy` module seems to be broken in F31+


mrc0mmand:
In attempt to make the dracut's test suite stable & working again, so it could be possibly extended to cover the IPv6 stack as well, I tried to debug the outstanding issues in #731 with following results:

@lnykryn @lkundrak it looks like there's something wrong with the `network-legacy` module, as the majority of network-related test fails was resolved by switching to NetworkManager (`network-manager` module). From what I could read from [the failing logs](https://travis-ci.org/dracutdevs/dracut/builds/653475914), the initrd with legacy network scripts simply fails to obtain a DHCP lease, whereas with NetworkManager everything works as expected.

Locally (Fedora 31) in `TEST-20-NFS` I managed to reproduce this as follows: 
1) Legacy network scripts (`network-legacy` module)
```diff
diff --git a/test/TEST-20-NFS/test.sh b/test/TEST-20-NFS/test.sh
index 30237dd5..00fb8b6e 100755
--- a/test/TEST-20-NFS/test.sh
+++ b/test/TEST-20-NFS/test.sh
@@ -354,8 +354,8 @@ test_setup() {
 
     # Make client's dracut image
     $basedir/dracut.sh -l -i $TESTDIR/overlay / \
-        -o "plymouth dash" \
-        -a "debug watchdog" \
+        -o "plymouth dash network-manager" \
+        -a "debug watchdog network-legacy" \
         -d "af_packet piix ide-gd_mod ata_piix sd_mod e1000 nfs sunrpc i6300esb" \
         --no-hostonly-cmdline -N \
         -f $TESTDIR/initramfs.testing $KVERSION || return 1
```

```
# make -C test/TEST-20-NFS/ clean setup run
...
[   14.488962] dracut-initqueue[569]: /bin/dracut-initqueue@71(): emergency_shell 'Could not boot.'
[   14.489724] dracut-initqueue[569]: /lib/dracut-lib.sh@1125(emergency_shell): local _ctty
[   14.490436] dracut-initqueue[569]: /lib/dracut-lib.sh@1126(emergency_shell): set +e
[   14.491111] dracut-initqueue[569]: /lib/dracut-lib.sh@1127(emergency_shell): local _rdshell_name=dracut action=Boot hook=emergency
[   14.492033] dracut-initqueue[569]: /lib/dracut-lib.sh@1128(emergency_shell): local _emergency_action
[...]

PR #736, opened 2020-02-24
Fix TEST-99-RPM

mrc0mmand:
I guess I managed to fix the *biggest* offenders, but there are still some outstanding issues which boggle my mind.

What was fixed:

1) network tests now explicitly define NIC names in order to avoid unexpected bumps in persistent naming
    This was the case of #734 where persistent names where bumped *by one*, so, for example, expected interface `ens3` was suddenly named `ens4` and so on.

2) `make rpm` was broken (at least on Fedora) since it used `/tmp` as its RPM buildroot. However, not sure since when, `/tmp` (not only on Fedora) is mounted with `-o noexec`, which breaks the `./configure` step. Moving the buildroot to `/var/tmp` makes the build work once again

3) `TEST-99-RPM` was broken by previous point as well, but also by dnf's new feature - weak dependencies - which polluted the test chroot and caused unexpected failures. Disabling this feature makes the issue go away

4) Travis now installs NetworkManager which allows tests to make use of the `network-manager` module instead of `network-scripts`.
    This technically switches all network-related tests to `network-manager` only mode (except for `TEST-60-IFCFG`, which explicitly uses `network-legacy` module). For now, it's just to make the testsuite pass, due to #732, but in the future it would be good to run the testsuite for both modules.

As for the outstanding issues:

1) sometimes the QEMU VM just gives straight up during early bootup:

```
CLIENT TEST START: rd.luks.uuid=930d0823-20a0-49c4-ac36-f8d8e615eb54 rd.md.uuid=fd76628b:07f18388:f48d3df2:230f7af5 rd.lvm.lv=dracut/root
c[?7l[2J[0mSeaBIOS (version ?-20190727_073836-buildvm-ppc64le-16.ppc.fedoraproject.org-3.fc

31)

Booting from ROM...

Probing EDD (edd=off to disable)... ok

c[?7l[2JCLIENT TEST END: rd.luks.uuid=930d0823-20a0-49c4-ac36-f8d8e615eb54 rd.md.uuid=fd76628b:07f18388:f48d3df2:230f7af5 rd.lvm.lv=dracut/root [FAIL]
TEST: root filesystem on an encrypted LVM PV on a degraded RAID-5  [0;31m [FAILED] [0;39m
```

2) (also) sometimes, the whole job gets stuck and times out after 10 minutes of inactivity (Travis default). I tried to *trick* this by introducing a temporary *watchdog*, which prints a message to stdout every 15 seconds, but even the watchdog stops responding, so it looks like the whole job just gives up (this happens even with a watchdog defined in the top level config file, i.e. `.travis.yml`).
  See, for example, https://travis-ci.org/dracutdevs/dracut/jobs/654907090?utm_medium=notification&utm_source=github_status

The two issues happen randomly to random jobs, so it's a undebuggable race condition, which makes the CI quite unreliable. I'm not even sure on which side the issue is, *sigh*. Any hints would be definitely welcome.

/cc @haraldh @lnykryn


mrc0mmand:
Also, all QEMU tests fail with F32 and F33/rawhide Docker images with:

```
qemu-system-x86_64: error: failed to set MSR 0x48b to 0x159ff00000000
qemu-system-x86_64: /builddir/build/BUILD/qemu-4.2.0/target/i386/kvm.c:2947: kvm_put_msrs: Assertion `ret == cpu->kvm_msr_buf->nmsrs' failed.
```

This should get looked into as well...

mrc0mmand:
@haraldh rebased. I'll wait for the outage to pass and then do another force push to trigger the newly introduced GitHub actions.

I dropped the first patch which fixed the NIC name bump in `TEST-{60,70}` by explicitly configuring NIC names on dracut's cmdline, like:

```
ifname=ens3:52:54:00:12:34:01
ifname=ens4:52:54:00:12:34:02
etc.
```

as it's a different approach for the same issue solved by https://github.com/dracutdevs/dracut/commit/2996d9e56fdb1ef28326d68c55f4b7628787f7d0, not sure which is better in this case.

haraldh:
>     1. network tests now explicitly define NIC names in order to avoid unexpected bumps in persistent naming
>        This was the case of #734 where persistent names where bumped _by one_, so, for example, expected interface `ens3` was suddenly named `ens4` and so on.

The bump was not "unexpected" .. it basically was a HW nic change.

If you define your own interface names, you must not name them to any persistent naming scheme, as they might clash, if you have an unnamed interface and udev wants to give it a persistent name.

haraldh:
> sometimes the QEMU VM just gives straight up during early bootup:

wow! This is bad... haven't seen that one yet

haraldh:
>     1. (also) sometimes, the whole job gets stuck and times out after 10 minutes of inactivity (Travis default). I tried to _trick_ this by introducing a temporary _watchdog_, which prints a message to stdout every 15 seconds, but even the watchdog stops responding, so it looks like the whole job just gives up (this happens even with a watchdog defined in the top level config file, i.e. `.travis.yml`).
>        See, for example, https://travis-ci.org/dracutdevs/dracut/jobs/654907090?utm_medium=notification&utm_source=github_status

This looks like a travis problem.. right?

haraldh:
> Edit: there are also several cases of _RAID-related general flakiness_:
> 
> ```
> Device     Boot Start   End Sectors Size Id Type
> /dev/sda1        2048 12287   10240   5M 83 Linux
> /dev/sda2       12288 32767   20480  10M 83 Linux
> /dev/sda3       32768 53247   20480  10M 83 Linux
> /dev/sda4       53248 71679   18432   9M 83 Linux
> The partition table has been altered.
> Calling ioctl() to re-read partition table.
> Syncing disks.
> mdadm: cannot open /dev/sdb: Device or resource busy
> Powering off.
> [   24.296167] reboot: Power down
> TEST: root filesystem on an encrypted LVM PV on a degraded RAID-5   [FAILED] 
> make[1]: *** [../Makefile.testdir:5: all] Error 1
> make[1]: Leaving directory '/dracut/test/TEST-12-RAID-DEG'
> make: *** [Makefile:5: TEST-12-RAID-DEG] Error 2
> ```

Might need a `udevadm settle` in there

haraldh:
> ```
> mdadm: Defaulting to version 1.2 metadata
> mdadm: RUN_ARRAY failed: Unknown error 524
>   Failed to clear hint file.
>   Device /dev/md0 not found.
> Powering off.
> [    9.733414] reboot: Power down
> TEST: root filesystem over multiple iSCSI   [FAILED] 
> make[1]: *** [../Makefile.testdir:5: all] Error 1
> make[1]: Leaving directory '/dracut/test/TEST-31-ISCSI-MULTI'
> make: *** [Makefile:5: TEST-31-ISCSI-MULTI] Error 2
> ```

`Unknown error` ... might point to some kernel issues? We probably have to remove "quiet" here for the kernel logs.

haraldh:
All in all, this testsuite is a big integration test, too, where we get all the bugs of all components.

haraldh:
> > sometimes the QEMU VM just gives straight up during early bootup:
> 
> wow! This is bad... haven't seen that one yet

yesterday I saw qemu segfault, though...

mrc0mmand:
As all the issues this PR originally addressed were already resolved by various commits in master, the only thing left are fixes for `TEST-99-RPM`, thus renaming this PR accordingly.

haraldh:
please add 99 to the .github workflow

PR #738, opened 2020-02-26
Use $TMPDIR and error out if tmpdir is mounted with 'noexec'

topimiettinen:
If the temporary directory for images is mounted with 'noexec', dracut
may construct unbootable images because most dynamic libraries aren't
installed. Avoid this by using user's $TMPDIR as default temporary
directory and by clearly erroring out if if the temporary directory is
mounted with 'noexec'.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -945,6 +945,12 @@ readonly TMPDIR="$(realpath -e "$tmpdir")"
     printf "%s\n" "dracut: Invalid tmpdir '$tmpdir'." >&2
     exit 1
 }
+
+if findmnt --raw -n --target "$tmpdir" --output=options | grep -q noexec; then
+    [[ $debug == yes ]] && printf "%s\n" "dracut: Tmpdir '$tmpdir' is mounted with 'noexec'."
+    noexec=1
+fi
+
 readonly DRACUT_TMPDIR="$(mktemp -p "$TMPDIR/" -d -t dracut.XXXXXX)"
 [ -d "$DRACUT_TMPDIR" ] || {
     printf "%s\n" "dracut: mktemp -p '$TMPDIR/' -d -t dracut.XXXXXX failed." >&2
@@ -969,7 +975,7 @@ if [[ $early_microcode = yes ]] || ( [[ $acpi_override = yes ]] && [[ -d $acpi_t
     mkdir "$early_cpio_dir"
 fi
 
-[[ -n "$dracutsysrootdir" ]] || export DRACUT_RESOLVE_LAZY="1"
+[[ -n "$dracutsysrootdir" || "$noexec" ]] || export DRACUT_RESOLVE_LAZY="1"
 
 if [[ $print_cmdline ]]; then
     stdloglvl=0


topimiettinen:
The problem is that `ldd` does not work at all if the executable is in `noexec` location.

```
$ cp /bin/ls /tmp
$ ldd /tmp/ls
        not a dynamic executable
$ ldd /bin/ls
        linux-vdso.so.1 (0x00007fff56302000)
        libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 (0x0000754cfb1e7000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000754cfb027000)
        libpcre2-8.so.0 => /lib/x86_64-linux-gnu/libpcre2-8.so.0 (0x0000754cfaf97000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x0000754cfaf92000)
        /lib64/ld-linux-x86-64.so.2 (0x0000754cfb26a000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x0000754cfaf71000)
$ findmnt /tmp -no OPTIONS
rw,nosuid,nodev,noexec,seclabel
```

Alternatively dracut could use the source path of the executable when ldd'ing instead of the copied executable, it could use a different tool for finding the dynamic libraries which doesn't have this limitation (if such a thing exists), or even introduce an own new tool in place of ldd.

Also ldd (or ld-linux.so.2 --list which ldd uses) could be improved to list the libraries without executing (like ld-linux.so.2 --list < exe).

haraldh:
Does it help to unset DRACUT_RESOLVE_LAZY (set on line 972), if you detect noexec?

topimiettinen:
Yes, that fixes the libraries. I updated the PR to do that.

dracut.sh
```diff
@@ -763,6 +763,7 @@ stdloglvl=$((stdloglvl + verbosity_mod_l))
 [[ $dracutbasedir ]] || dracutbasedir=$dracutsysrootdir/usr/lib/dracut
 [[ $fw_dir ]] || fw_dir="$dracutsysrootdir/lib/firmware/updates:$dracutsysrootdir/lib/firmware:$dracutsysrootdir/lib/firmware/$kernel"
 [[ $tmpdir_l ]] && tmpdir="$tmpdir_l"
+[[ $tmpdir ]] || tmpdir="$TMPDIR"
```
haraldh: I am still very uncomfortable about this. Users should really explicitly use `--tmpdir "$TMPDIR"` here. Is that so important for you?

topimiettinen: Not really anymore, since patched dracut works without changing tmpdir. I can drop this line.

In general though it's good idea to avoid public directories /tmp and /var/tmp if more private choices (like /run/user/UID) exist, so I'd still encourage you to make this change, perhaps with some advertising of the fact in release notes.

topimiettinen: In any case, it makes sense to make a new PR for the TMPDIR part. This is a bugfix but that PR is requesting a new feature.

topimiettinen:
One of Travis CI jobs fails (41, root filesystem with NFS) but I suppose that has nothing to do with the PR.

PR #739, opened 2020-02-26
35network-legacy: use $name instead of $env{INTERFACE}

mrc0mmand:
The original behavior of $env{INTERFACE} was undocumented and changed in
the recent udev versions, breaking the ability to bring up networking
reliably. Switching to $name directive should fix this issue.

Related links:
 - https://github.com/systemd/systemd/pull/12700 (udev PR)
 - https://github.com/systemd/systemd/issues/12291 (related udev issue)
 - https://github.com/systemd/systemd/issues/14935 (this issue, udev side)
 - https://github.com/dracutdevs/dracut/issues/732 (this issue, dracut side)

Fixes: #732

---

WIP for now, as I'd like to check the compatibility with older udev, and if it doesn't break something else with the current one.

### DIFF
--- a/modules.d/35network-legacy/net-genrules.sh
+++ b/modules.d/35network-legacy/net-genrules.sh
@@ -65,9 +65,9 @@ command -v fix_bootif >/dev/null || . /lib/net-lib.sh
         bootdev=$(cat /tmp/net.bootdev)
     fi
 
-    ifup='/sbin/ifup $env{INTERFACE}'
+    ifup='/sbin/ifup $name'
 
-    runcmd="RUN+=\"/sbin/initqueue --name ifup-\$env{INTERFACE} --unique --onetime $ifup\""
+    runcmd="RUN+=\"/sbin/initqueue --name ifup-\$name --unique --onetime $ifup\""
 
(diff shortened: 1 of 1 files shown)


yuwata:
Thanks.

mrc0mmand:
- F25 :heavy_check_mark: - CentOS CI (the fail in `TEST-30-ISCSI` (iBTF subtest) happens even without this patch, so it seems to be unrelated and should be dealt with separately)
- F29 :heavy_check_mark: - https://travis-ci.org/dracutdevs/dracut/builds/655397821?utm_source=github_status&utm_medium=notification
- F31: :heavy_check_mark: - https://travis-ci.org/dracutdevs/dracut/builds/655437162?utm_source=github_status&utm_medium=notification (fail in `TEST-99-RPM` is expected, see #736)

Overall it looks like it's working, but it definitely needs eyes of somebody more educated in this matter - @lnykryn, @haraldh PTAL.

mrc0mmand:
Fixed by https://github.com/dracutdevs/dracut/commit/a8ba1c4e25051c3c482bb7a6a754c9d785463917, closing

PR #741, opened 2020-02-28
fcoe/fcoe-genrules.sh: use $name instead of $env{INTERFACE}

mrc0mmand:
Followup to a8ba1c4e25051c3c482bb7a6a754c9d785463917 and 11a5501d0f844b9e52f2f1b62d980b733c73f10a

---

Hopefully the last `$env{INTERFACE}`-related change, managed to completely overlook this one...


mrc0mmand:
I also added an additional commit which fixes `TEST-99-RPM` when run on revisions which SHA begins with a letter (discovered by accident is this PR).

PR #743, opened 2020-03-01
Use TMPDIR if available

topimiettinen:
Use environment variable TMPDIR (typically /run/user/$UID) as default
temporary directory, if available. This should be more private
location than /var/tmp. Path specified with --tmpdir is takes
precedence over TMPDIR and /var/tmp is still used as last resort if
neither TMPDIR is set nor --tmpdir is used.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -140,7 +140,7 @@ Creates initial ramdisk images for preloading modules
   --confdir [DIR]       Specify configuration directory to use *.conf files
                          from. Default: /etc/dracut.conf.d
   --tmpdir [DIR]        Temporary directory to be used instead of default
-                         /var/tmp.
+                         ${TMPDIR:-/var/tmp}.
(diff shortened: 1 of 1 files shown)


topimiettinen:
Previous discussions in PR #738.

CI failures:
- Fedora-30 test 03 is successful for first tests, but fails for "writable root"
- Fedora-31 test 35 fails completely because device `/dev/md0` is not found.

The latter case clearly doesn't seem related at all and in the former case I don't see why this patch would make only one of the tests fail but others would be successful.

haraldh:
> Previous discussions in PR #738.
> 
> CI failures:
> 
>     * Fedora-30 test 03 is successful for first tests, but fails for "writable root"
> 
>     * Fedora-31 test 35 fails completely because device `/dev/md0` is not found.
> 
> 
> The latter case clearly doesn't seem related at all and in the former case I don't see why this patch would make only one of the tests fail but others would be successful.

Yeah, the CI is flaky here and there :cry:

PR #744, opened 2020-03-01
Make lsinitrd usable for images made with Debian mkinitramfs

topimiettinen:
Debian mkinitramfs does not create the file 'early_cpio', so detection
of additional cpio images fails and only the first cpio is listed.

I checked some Arch and Gentoo initramfs files and these didn't have
'early_cpio' either, but they also have only one cpio part.

Fix this so that if check for 'early_cpio' fails, check if directory
'kernel' (as part of '/kernel/x86/microcode') exists.


haraldh:
shouldn't we check for microcode instead then?

topimiettinen:
Updated to check for `/kernel/*/microcode/*.bin`.

PR #750, opened 2020-03-06
Make externally defined CFLAGS work

danimo:
This is required by some distributions such as openSUSE
That define alternate cflags, rather than just appending
new ones.


haraldh:
Huh? What about KMOD_CFLAGS?

haraldh:
We could also merge the two `CFLAGS`

danimo:
Like this?
```
override CFLAGS = -O2 -g -Wall -std=gnu99 -D_FILE_OFFSET_BITS=64 -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2
CFLAGS += $(KMOD_CFLAGS)
```

haraldh:
what's wrong with `?=`

danimo:
Good point. But agree with the rest?

haraldh:
```
CFLAGS ?= -O2 -g -Wall -std=gnu99 -D_FILE_OFFSET_BITS=64 -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2
CFLAGS += $(KMOD_CFLAGS)
```

danimo:
This works for me now.

PR #753, opened 2020-03-10
95znet: Add a rd.znet_ifname= option

ryncsn:
qeth device may have a different IP for each boot, so the rd.ifname=
option will no longer work. So for znet device, introduce a
rd.znet_ifname= options, to subchannel id instead of MAC address as the
identifier and rename the interface.

### DIFF
--- a/modules.d/95znet/parse-ccw.sh
+++ b/modules.d/95znet/parse-ccw.sh
@@ -4,4 +4,22 @@ for ccw_arg in $(getargs rd.ccw -d 'rd_CCW=') $(getargs rd.znet -d 'rd_ZNET='); 
     echo $ccw_arg >> /etc/ccw.conf
 done
 
+for ifname in $(getargs rd.znet_ifname); do
+    IFS=: read ifname_if ifname_subchannels _rest <<< "$ifname"
+    if [ -z "$ifname_if" ] || [ -z "$ifname_subchannels" ] || [ -n "$_rest" ]; then
+        warn "Invalid arguments for rd.znet_ifname="
+    else
+        {
+            ifname_subchannels=${ifname_subchannels//,/|}
+
+            echo 'ACTION!="add|change", GOTO="ccw_ifname_end"'
+            echo 'ATTR{type}!="1", GOTO="ccw_ifname_end"'
+            echo 'SUBSYSTEM!="net", GOTO="ccw_ifname_end"'
+            echo "SUBSYSTEMS==\"ccwgroup\", KERNELS==\"$ifname_subchannels\", DRIVERS==\"?*\" NAME=\"$ifname_if\""
+            echo 'LABEL="ccw_ifname_end"'
+
+        } > /etc/udev/rules.d/81-ccw-ifname.rules
+    fi
+done
+
 znet_cio_free
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -863,6 +863,11 @@ ZNET
     RHEL/Fedora with ccw_init, which is called from udev for certain
     devices on z-series.
     rd.znet can be specified multiple times on the kernel command line.
+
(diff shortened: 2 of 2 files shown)


ryncsn:
Any comment about this PR?

LianboJ:
In most cases, it always provides a persistent MAC address. But for the s390 Arch, sometimes, kernel could run in the LPAR mode and it doesn't provide a persistent MAC address, which caused the kdump failure.  
Adding the 95znet rules should be good to fix it, and the rd.znet_ifname= option doesn't have negative effects. So, this looks good to me. Thanks.

modules.d/95znet/parse-ccw.sh
```diff
@@ -4,4 +4,24 @@ for ccw_arg in $(getargs rd.ccw -d 'rd_CCW=') $(getargs rd.znet -d 'rd_ZNET=');
     echo $ccw_arg >> /etc/ccw.conf
 done
 
+for ifname in $(getargs rd.znet_ifname); do
+    echo "$ifname" | {
```
lnykryn, line 8: I don't think that the pipe to the whole block is necessary here. I would suggest doing
IFS=: read ifname_if ifname_subchannels _ <<< "$ifname"

modules.d/95znet/parse-ccw.sh
```diff
@@ -4,4 +4,24 @@ for ccw_arg in $(getargs rd.ccw -d 'rd_CCW=') $(getargs rd.znet -d 'rd_ZNET=');
...
+        IFS=: read ifname_if ifname_subchannels _
+        if ! [ -n "$ifname_if" ] && [ -n "$ifname_subchannels" ] && [ -z "$_" ]; then
+            warn "Invalid arguments for rd.znet_ifname="
+        else
+            {
+                ifname_subchannels=$(echo $ifname_subchannels | tr "," "|")
```
lnykryn, line 14: I would rather use ${a//,/|}

modules.d/95znet/parse-ccw.sh
```diff
@@ -4,4 +4,24 @@ for ccw_arg in $(getargs rd.ccw -d 'rd_CCW=') $(getargs rd.znet -d 'rd_ZNET=');
     echo $ccw_arg >> /etc/ccw.conf
 done
 
+for ifname in $(getargs rd.znet_ifname); do
+    echo "$ifname" | {
+        IFS=: read ifname_if ifname_subchannels _
+        if ! [ -n "$ifname_if" ] && [ -n "$ifname_subchannels" ] && [ -z "$_" ]; then
```
lnykryn, line 10: I don't think the check is correct. 
[lnykryn@X1 dracut(znet)]$ ifname_if=aaa
[lnykryn@X1 dracut(znet)]$ ifname_subchannels=""
[lnykryn@X1 dracut(znet)]$ _=""
[lnykryn@X1 dracut(znet)]$ if ! [ -n "$ifname_if" ] && [ -n "$ifname_subchannels" ] && [ -z "$_" ]; then echo fail; fi
[lnykryn@X1 dracut(znet)]$ 


I would use 
if [ -z "$ifname_if" ] || [ -z "$ifname_subchannels" ]  || [ -n "$_" ]

ryncsn: Right, I messed up with the operator priority, thanks for pointint it out.

modules.d/95znet/parse-ccw.sh
ryncsn: Good idea, saved a tr call.

modules.d/95znet/parse-ccw.sh
ryncsn: I've no problem to change using read <<< $var style, will update. btw is that a preferred code style or have some advantage?

lnykryn: I think it is used in other places, but personally I find it a bit easier to read.

PR #754, opened 2020-03-10
Fix pre-trigger stage by replacing exit with return in lldpad.sh

rvykydal:
Using exit makes the pre-trigger stage finish after running 03-lldpad.sh
pre-trigger hook.


rvykydal:
The fix assumes that lldpad.sh is just sourced as a hook, not run on its own.
Which seems to be the case.

danimo:
Yes, I missed that when fixing a similar case in 4f194276417237168f5bc929ba847a23b6a4c77f. Thanks.

PR #755, opened 2020-03-10
network: fix glob matching ipv6 addresses

lnykryn:
In this case we want to cover three cases
1) ip_address:ip_address
2) number:macaddress
3) :macaddress

We consider something an IPv6 address if it starts with number and
contains ":", but IPv6 addresses are in hexa.


haraldh:
So, you want a dns1 for IPv6?
Shouldn't it better be enclosed in `[]`?

lnykryn:
> So, you want a dns1 for IPv6?
Yep
> Shouldn't it better be enclosed in `[]`?
It is, but we remove "[" and "]" in the beginning of the function.

haraldh:
> > Shouldn't it better be enclosed in `[]`?
> 
> It is, but we remove "[" and "]" in the beginning of the function.

ah, sorry, missed that.. it's been a long time since I parsed that function completely :)

PR #757, opened 2020-03-11
95fcoe: default rd.nofcoe to false

tblume:
rd.nofcoe should default to false, e.g. fcoe should be enabled unless
overwritten from the command line.
The same applies for lldapd.sh.

### DIFF
--- a/modules.d/95fcoe/lldpad.sh
+++ b/modules.d/95fcoe/lldpad.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if ! getargbool 0 rd.nofcoe ; then
+if getargbool 0 rd.nofcoe ; then
 	info "rd.nofcoe=0: skipping lldpad activation"
 	return 0
 fi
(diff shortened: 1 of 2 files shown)


modules.d/95fcoe/lldpad.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if ! getargbool 0 rd.nofcoe ; then
+if getargbool 0 rd.nofcoe ; then
```
rmetrich, line 3: Sorry for being late, but I think it's very confusing:
- **nofcoe** means _don't enable fcoe_
- 0 for getargbool means _false_
hence
**nofcoe=0** means _don't enable fcoe == false_ which is equivalent to _enable fcoe == true_
which is exactly the opposite of what we want here.

rmetrich: @danimo please ignore, answered in https://github.com/dracutdevs/dracut/commit/6f7823bce65dd4b52497dbb94892b637fd06471a

Issue #694, opened 2019-12-02
35network-manager won't setup network if other modules don't add any "initqueue/finished" hook


ryncsn:
Previously when using 35network-legacy, it will add initqueue/finished hooks to wait for the interface to be configured. However 35network-manager won't do so, and if other modules didn't add any initqueue/finished hook, initqueue won't run, and the nm-run.sh won't have a chance to run either.

Is 35network-manager expecting other module to install initqueue/finished to wait for network to get ready, or perhaps 35network-manager should at lease ensure nm-run.sh got executed for at least once?

ryncsn:
There is a workaround for this problem in kdump, by install another initqueue hook to keep the initqueue running until network is ready:

See: https://lists.fedorahosted.org/archives/list/kexec@lists.fedoraproject.org/thread/ZAYFCE3UOH7UTCBFWZ6KGR5VI625IGZT/

cgwalters:
xref https://github.com/coreos/fedora-coreos-config/pull/259
and in particular https://mail.gnome.org/archives/networkmanager-list/2020-January/msg00000.html
where I argue that we should be running NM as a regular systemd unit.

haraldh:
@lkundrak !

LorbusChris:
Any updates on this?

jomeier:
@cgwalters 
Hi. Any news on this one? Greets, Josef

jomeier:
@haraldh 
Hi Harald. @LorbusChris mentioned that you could come to the rescue here.

Could you give priority to this issue, please? The OKD community needs you 🕺🏻

LorbusChris:
I think I meant @lkundrak might be able to do that.
Anyway, now we've pinged 'em all! :tongue: 
And as @jomeier said, we over at Fedora CoreOS and OKD would appreciate a fix for this issue =)

LorbusChris:
https://bugzilla.redhat.com/show_bug.cgi?id=1809641 is a related BZ for this

tyll:
Hi, The NetworkManager team is happy to help. What is the exact request?

@thom311 or @bengal please take a look.

> https://bugzilla.redhat.com/show_bug.cgi?id=1809641 is a related BZ for this

This is not reported against NetworkManager and also does not include the NM team in CC. Should it?

LorbusChris:
Thanks @tyll! The root cause for that BZ is this issue here :)
[...]

PR #761, opened 2020-03-11
network-manager: ensure that nm-run.sh is executed when needed

bengal:
The network-manager command line hook must install a
initqueue/finished hook to ensure that nm-run.sh is executed when
there are network connections to activate.

Fixes: #694


LorbusChris:
@bengal thank you!!
@haraldh could we get an rpm rpm build with this change for f31? :innocent:

haraldh:
> @bengal thank you!!
> @haraldh could we get an rpm rpm build with this change for f31? innocent

https://bodhi.fedoraproject.org/updates/FEDORA-2020-b36b25de24

PR #776, opened 2020-03-26
url-lib: drop NSS, curl no longer depends on it

t184256:
curl no longer depends on NSS now in Fedora or RHEL, let's stop pulling it in


lnykryn:
I don't think it is a good idea to unconditionally drop it. Maybe we should call curl --version and check if it was built against NSS?

t184256:
Maybe. I've force-pushed my implementation of your suggestion.

PR #778, opened 2020-03-27
Retrieve service, rpc and protcol entries via getent

danimo:
Sometimes, systems have more elaborate backends in nsswitch.conf,
the complexity of which we do not want in the initrd. SUSE systems
with atomic for instance, where those files may also reside in /usr/etc.

With this change, the unavaiable nsswitch.conf backends will be silently
ignored, and we still end up with all the information we need.

Reference: boo#1165366

### DIFF
--- a/modules.d/95cifs/module-setup.sh
+++ b/modules.d/95cifs/module-setup.sh
@@ -37,7 +37,13 @@ install() {
     local _i
     local _nsslibs
     inst_multiple -o mount.cifs
-    inst_multiple /etc/services /etc/nsswitch.conf /etc/protocols
+    inst_multiple /etc/nsswitch.conf
+    if [ $hostonly ]; then
+        getent services > ${initdir}/etc/services
+        getent protocols > ${initdir}/etc/protocols
+    else
+        inst_multiple /etc/services /etc/protocols
+    fi
 
     inst_libdir_file 'libcap-ng.so*'
 
--- a/modules.d/95nfs/module-setup.sh
+++ b/modules.d/95nfs/module-setup.sh
@@ -77,8 +77,16 @@ cmdline() {
 install() {
     local _i
     local _nsslibs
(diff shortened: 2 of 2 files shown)


danimo:
This also has the neat side effect that we no longer ignore NIS or LDAP-based entries for those files.

danimo:
Downstream Bug-Report: https://bugzilla.suse.com/show_bug.cgi?id=1165366

haraldh:
> Downstream Bug-Report: https://bugzilla.opensuse.org/show_bug.cgi?id=1165366

https://bugzilla.opensuse.org/show_bug.cgi?id=1165366#c18

haraldh:
> This also has the neat side effect that we no longer ignore NIS or LDAP-based entries for those files.

Aren't these dynamic? I think this be dependent on `$hostonly`

danimo:
Done. Also removed the changes for the tests, as they are not hostonly.

modules.d/95cifs/module-setup.sh
```diff
@@ -37,7 +37,13 @@ install() {
     local _i
     local _nsslibs
     inst_multiple -o mount.cifs
-    inst_multiple /etc/services /etc/nsswitch.conf /etc/protocols
+    inst_multiple /etc/nsswitch.conf
+    if [ $hostonly ]; then
+        getent services > ${initdir}/etc/services
+        getent protocols > ${initdir}/etc/services
```
haraldh, line 43:
```suggestion
        getent protocols > ${initdir}/etc/protocols
```

modules.d/95cifs/module-setup.sh
```diff
@@ -37,7 +37,13 @@ install() {
...
+    else
+        inst_multiple /etc/services /etc/services
```
haraldh, line 45: what happened to `/etc/protocols` ??

haraldh (changes requested):

PR #780, opened 2020-04-03
multipath: add automatic configuration for multipath

darkmuggle:
Add support for 'rd.multipath=default' for using the default
configuration on boot. The intended purpose for this is to help support
ostree-based image boots from multipathed devices (such as Fedora and
Red Hat CoreOS).

### DIFF
--- a/modules.d/90multipath/multipathd-configure.service
+++ b/modules.d/90multipath/multipathd-configure.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=Device-Mapper Multipath Default Configuration
+Before=iscsi.service iscsid.service lvm2-activation-early.service
+Wants=systemd-udev-trigger.service systemd-udev-settle.service local-fs-pre.target
+After=systemd-udev-trigger.service systemd-udev-settle.service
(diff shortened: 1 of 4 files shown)


darkmuggle:
For background, on RHCOS is part of OpenShift. The user generally does not configure the system in the classic way. Further the initramfs is distributed as part of the RHCOS OStree. By having a kernel flag for `rd.multipath=default` the user would have the ability to use the default pathing configuration.

modules.d/90multipath/multipathd-configure.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Device-Mapper Multipath Default Configuration
+Before=iscsi.service iscsid.service lvm2-activation-early.service
+Wants=systemd-udev-trigger.service systemd-udev-settle.service local-fs-pre.target
+After=systemd-udev-trigger.service systemd-udev-settle.service
```
dtardon, line 5: Are the deps on systemd-udev-*.service really needed? AFAICS `mpathconf --enable` doesn't access any HW; it just writes a config. file...

PR #781, opened 2020-04-05
configure: Find FTS library with --as-needed

awilfox:
When LDFLAGS contains ``-Wl,--as-needed``, the FTS library will be
ignored if it is specified before the .c source.

Fixes: 62f27ee6f145 ("configure: find cflags and libs for fts on musl")


danimo:
Rebased in #970, which will be merged. Closing.

PR #787, opened 2020-04-09
Add a --no-switch-root option

ryncsn:
Previously I added a dracut_no_switch_root helper to indicate some
    modules that the initramfs will no do swith root. To further extend the
    support for switch-root less initramfs, introduce an argument to avoid
    install some files.
    
    Some special user like kdump may run into strange race condition when
    systemd try to reload the config in the real root, this will mask the
    service for doing that, also drop other unneeded services.
    
### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -416,6 +416,7 @@ rearrange_params()
         --long no-hostonly-i18n \
         --long hostonly-i18n \
         --long no-machineid \
+        --long no-switch-root \
         -- "$@")
 
     if (( $? != 0 )); then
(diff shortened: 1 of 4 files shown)


danimo:
Could you add some documentation for the parameter to the man page, please?

danimo:
Are you still considering to update this or can it be closed? If not, please also add the parameter to dracut-bash-completion.sh.

ryncsn:
Sorry for the late response, I nearly forgot this PR and missed the notification... I'll update it.

PR #788, opened 2020-04-09
dracut-initqueue: Print more useful info in case of timeout

ryncsn:
Currently when initqueue timeout, it span the console with
"dracut-initqueue timeout - starting timeout scripts", which isn't very
helpful as we still don't know what actually happened. Try to improve
this by print what is actually being waited.

Besides, only print "starting timeout scripts" when there are
actual timeout scripts to use.


haraldh:
test-99:
```
file /etc/dnf/modules.d/eclipse.module.rpmmoved is not owned by any package
```
oh my...

haraldh:
> `file /etc/dnf/modules.d/eclipse.module.rpmmoved is not owned by any package`

https://github.com/dracutdevs/dracut/pull/790

PR #791, opened 2020-04-10
Don't start rngd if there is no hardware entropy generator

mikhailnov:
rngd requires a hardware random generator device:
```
-r, --rng-device=file      Kernel device used for random number input
                             (default: /dev/hw_random)
```
But it is not present in virtual machines and multiple hardware paltforms.
It leads to rngd.service being failed in initrd.
Let's stop trying to run rngd if default device /dev/hw_random does not exist.

### DIFF
--- a/modules.d/06rngd/rngd.service
+++ b/modules.d/06rngd/rngd.service
@@ -2,6 +2,7 @@
 Description=Hardware RNG Entropy Gatherer Daemon
 DefaultDependencies=no
 Before=systemd-udevd.service
+ConditionPathExists=/dev/hw_random
 
 [Service]
 ExecStart=/usr/sbin/rngd -f


haraldh:
Hmm, seems like we are talking about different device names.

```console
❯ cat /sys/class/misc/hw_random/uevent 
MAJOR=10
MINOR=183
DEVNAME=hwrng

❯ ls -l /dev/hwrng 
crw-------. 1 root root 10, 183 20. Apr 14:17 /dev/hwrng

❯ ls -l '/dev/char/10:183'
lrwxrwxrwx. 1 root root 8 20. Apr 14:17 /dev/char/10:183 -> ../hwrng
```

Since this is a static assigned MAJOR:MINOR, I would change the systemd file to point to:
`/dev/char/10:183` if your device name differs.

mikhailnov:
Hm, who are you talking with..? :)
`/dev/hw_random` != `/dev/hwrng`, sorry, I did not understand what you mean

haraldh:
> Hm, who are you talking with..? :)
> `/dev/hw_random` != `/dev/hwrng`, sorry, I did not understand what you mean

https://github.com/torvalds/linux/commit/e9e23158f2f5a0992f27d367aa1f11ba12b0b54c

haraldh:
So my understanding is, they renamed `/dev/hw_random` to `/dev/hwrng`

nhorman:
To be clear, it is absolutely wrong to disable rngd just because a hwrng device doesn't exist

mikhailnov:
> With the addition of jitterentropy several releases ago, rngd is guaranteed to have at least that entropy source available even in virtual guests, and so the failure to get /dev/hwrng to open is a failure, but not a fatal one.

@nhorman Thanks for pointing to it, it was disabled in rng-tools, I have enabled it there.

PR #792, opened 2020-04-15
01bootetc: introduce module for updating dracut etc

darkmuggle:
This provides support for updating '/etc/' in the initramfs using
'bootetc=<DEVICE>'. If the device has the path '/initrd-etc', then the
tree is copied into the initramfs.

The purpose of bootetc is to aide immutable operating systems like
Fedora and Red Hat CoreOS that ship static initrds with a path for
user-provided configuration. Example use cases include udev rules, iSCSI
and multipath configuration, fips and Network Manager configurations.

### DIFF
--- a/modules.d/01bootetc/boot-etc.sh
+++ b/modules.d/01bootetc/boot-etc.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+. /lib/dracut-lib.sh
+
+bootetc=$(getarg bootetc=)
+if [ "${bootetc}x" == "x" ]; then
+    exit 0
+fi
+
+mount_boot bootetc
+if [ -d /boot/initrd-etc ] && [ ! -f /run/bootetc.done ]; then
+    info "bootetc: Updating initramfs etc from ${bootetc}/initrd-etc"
+    copytree /boot/initrd-etc /etc
+    touch /run/bootetc.done
+
+    if [ -z "$DRACUT_SYSTEMD" ]; then
+        systemctl try-restart dracut-cmdline.service
+        systemctl try-restart systemd-udev-trigger.service
+    fi
+fi
+
+umount /boot >/dev/null 2>&1
--- a/dracut.spec
+++ b/dracut.spec
@@ -343,6 +343,7 @@ install -m 0755 51-dracut-rescue-postinst.sh $RPM_BUILD_ROOT%{_sysconfdir}/kerne
 %if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version}
 %{dracutlibdir}/modules.d/01fips
 %endif
+%{dracutlibdir}/modules.d/01bootetc
 %{dracutlibdir}/modules.d/01systemd-initrd
 %{dracutlibdir}/modules.d/03modsign
 %{dracutlibdir}/modules.d/03rescue
--- a/modules.d/01bootetc/boot-etc.service
+++ b/modules.d/01bootetc/boot-etc.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=Update initramfs etc
+Before=iscsi.service iscsid.service lvm2-activation-early.service
+Wants=systemd-udev-trigger.service systemd-udev-settle.service local-fs-pre.target
+After=systemd-udev-trigger.service systemd-udev-settle.service
+Before=local-fs-pre.target dracut-initqueue.service
+
+DefaultDependencies=no
+Conflicts=shutdown.target
+
+ConditionKernelCommandLine=bootetc
+ConditionPathExists=!/run/bootetc.done
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/boot-etc
+ExecStart=/usr/bin/touch /run/bootetc.done
+
+[Install]
+WantedBy=sysinit.target
--- a/modules.d/01bootetc/module-setup.sh
+++ b/modules.d/01bootetc/module-setup.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    return 0
+}
+
+# called by dracut
+depends() {
+    return 0
+}
+
+# called by dracut
+install() {
+    inst_simple "$moddir/boot-etc.sh" "/sbin/boot-etc"
+
+    if dracut_module_included "systemd"; then
+        inst_simple "${moddir}/boot-etc.service" "${systemdsystemunitdir}/boot-etc.service"
+        systemctl -q --root "$initdir" enable boot-etc.service
+    else
+        inst_hook pre-trigger 01 "$moddir/boot-etc.sh"
+    fi
+
+}
--- a/modules.d/01fips/fips-boot.sh
+++ b/modules.d/01fips/fips-boot.sh
@@ -6,7 +6,7 @@ elif [ -z "$fipsmode" ]; then
     die "FIPS mode have to be enabled by 'fips=1' not just 'fips'"
 elif getarg boot= >/dev/null; then
     . /sbin/fips.sh
-    if mount_boot; then
+    if mount_boot boot fips=1; then
         do_fips || die "FIPS integrity test failed"
     fi
 fi
--- a/modules.d/01fips/fips-noboot.sh
+++ b/modules.d/01fips/fips-noboot.sh
@@ -6,6 +6,6 @@ elif [ -z "$fipsmode" ]; then
     die "FIPS mode have to be enabled by 'fips=1' not just 'fips'"
 elif ! [ -f /tmp/fipsdone ]; then
     . /sbin/fips.sh
-    mount_boot
+    mount_boot boot fips=1
     do_fips || die "FIPS integrity test failed"
 fi
--- a/modules.d/01fips/fips.sh
+++ b/modules.d/01fips/fips.sh
@@ -1,58 +1,5 @@
 #!/bin/sh
 
-mount_boot()
-{
-    boot=$(getarg boot=)
-
-    if [ -n "$boot" ]; then
-        case "$boot" in
-        LABEL=*)
-            boot="$(echo $boot | sed 's,/,\\x2f,g')"
-            boot="/dev/disk/by-label/${boot#LABEL=}"
-            ;;
-        UUID=*)
-            boot="/dev/disk/by-uuid/${boot#UUID=}"
-            ;;
-        PARTUUID=*)
-            boot="/dev/disk/by-partuuid/${boot#PARTUUID=}"
-            ;;
-        PARTLABEL=*)
-            boot="/dev/disk/by-partlabel/${boot#PARTLABEL=}"
-            ;;
-        /dev/*)
-            ;;
-        *)
-            die "You have to specify boot=<boot device> as a boot option for fips=1" ;;
-        esac
-
-        if ! [ -e "$boot" ]; then
-            udevadm trigger --action=add >/dev/null 2>&1
-            [ -z "$UDEVVERSION" ] && UDEVVERSION=$(udevadm --version)
-            i=0
-            while ! [ -e $boot ]; do
-                if [ $UDEVVERSION -ge 143 ]; then
-                    udevadm settle --exit-if-exists=$boot
-                else
-                    udevadm settle --timeout=30
-                fi
-                [ -e $boot ] && break
-                sleep 0.5
-                i=$(($i+1))
-                [ $i -gt 40 ] && break
-            done
-        fi
-
-        [ -e "$boot" ] || return 1
-
-        mkdir /boot
-        info "Mounting $boot as /boot"
-        mount -oro "$boot" /boot || return 1
-    elif [ -d "$NEWROOT/boot" ]; then
-        rm -fr -- /boot
-        ln -sf "$NEWROOT/boot" /boot
-    fi
-}
-
 do_rhevh_check()
 {
     KERNEL=$(uname -r)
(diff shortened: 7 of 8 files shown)


darkmuggle:
@cgwalters this is the end result of the idea to support `/etc` coming from boot.

modules.d/01bootetc/boot-etc.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+. /lib/dracut-lib.sh
+
+bootetc=$(getarg bootetc=)
```
cgwalters, line 4: OK, so we require an explicit opt-in karg like `bootetc=1`; and you see us changing FCOS to pass that by default?  Or it only gets injected when e.g. dm-multipath is in use?

modules.d/01bootetc/boot-etc.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+mount_boot bootetc
+if [ -d /boot/initrd-etc ] && [ ! -f /run/bootetc.done ]; then
+    info "bootetc: Updating initramfs etc from ${bootetc}/initrd-etc"
+    copytree /boot/initrd-etc /etc
+    touch /run/bootetc.done
+
+    if [ -z "$DRACUT_SYSTEMD" ]; then
+        systemctl try-restart dracut-cmdline.service
```
cgwalters, line 16: This part is probably the most unfortunate.  I don't know all of what's going on in `dracut-cmdline.service` but it'll need to be prepared to be killed at arbitrary points.

Can't we have this bit run `Before=dracut-cmdline.service`?

modules.d/01bootetc/boot-etc.sh
```diff
@@ -0,0 +1,21 @@
...
+        systemctl try-restart systemd-udev-trigger.service
```
cgwalters, line 17: Hm, why does udev need to be retriggered?

dustymabe: probably so the device that contains the boot partition can be found?

dustymabe: oh `re-triggered`.. not sure

cgwalters:
At a high level, I think the original idea was just that we support e.g. `/boot/etc/multipath.conf`, which would be checked if e.g. `rd.multipath=1` is set or so and used for the rootfs mounting path.  This would be less flexible, but would avoid any need to restart any parts of dracut.

modules.d/01bootetc/boot-etc.sh
darkmuggle: Udev rules :)

modules.d/01bootetc/boot-etc.sh
darkmuggle: Nope, I looked at the boot-path. `dracut-cmdline.service` is run before Udev. we need udev to find the boot device.

jlebon:
I think the approach here has potential. A few questions/comments:
- Do we have any concerns here about trying to wait for `/boot` early on and holding up boot?
- We're modifying boot configuration while booting, which means we'll race with any other code which is reading those files. I agree it probably covers a lot of use case, though it doesn't seem as complete as the approaches discussed in https://github.com/coreos/rpm-ostree/issues/1930. Would you agree?

modules.d/01bootetc/boot-etc.sh
darkmuggle: It's more like `bootetc=LABEL=....`. I envision this being an opt-in. We _could_ use it for FCOS, but I'm particularly keen on that making that decision today.

cgwalters:
> though it doesn't seem as complete as the approaches discussed in coreos/rpm-ostree#1930. Would you agree?

Well I was pushing back against using initramfs mutation for this because it conflicts with any approach to signing, it's mixing *code* and *configuration*.  The initramfs is *code* that should be parsing configuration from e.g. the kernel cmdline.  The problem with multipath is that its configuration is too big for the cmdline.

darkmuggle:
> * Do we have any concerns here about trying to wait for `/boot` early on and holding up boot?

Its a pretty-fast failure and its opt-in only. Since its run after `udev`, it fairly quick.,

> * We're modifying boot configuration while booting, which means we'll race with any other code which is reading those files. I agree it probably covers a lot of use case, though it doesn't seem as complete as the approaches discussed in [coreos/rpm-ostree#1930](https://github.com/coreos/rpm-ostree/issues/1930). Would you agree?

Sort of. I think they are orthogonal. Support in `rpm-ostree` is one question. This could be used for traditional systems and CoreOS now, while support in `coreos/rpm-ostree#1930 ` is done. But the bigger question is first boot, for things like `multipath` or things that are needed to install the system in the first place.

darkmuggle:
> Well I was pushing back against using initramfs mutation for this because it conflicts with any approach to signing, it's mixing _code_ and _configuration_. The initramfs is _code_ that should be parsing configuration from e.g. the kernel cmdline. The problem with multipath is that its configuration is too big for the cmdline.

Multipath configuration can't be used on the commandline. The configuration must originate from `/etc/multipath.conf`. The only way to narrowly scope it multipath would be to bind it to the multipath module; that felt a bit too hacky. I had to step back and realize that Dracut is scoped to Linux et al, not just CoreOS. CoreOS may need it for one thing, but other Linux's could use it for other things.  

This PR is a general solution for general Linux.  And having been a support engineer, I would have killed for the features that this enables such as the ability to change the configuration in the initramfs without having to play the regeneration of the initramfs (because everyone loves the bind mount, chroot, regenerate, reboot game). 

Something else to consider: it was requested that this idea be put in Dracut. If this is too much for CoreOS, then we should carry the narrowly scoped module _in_ our own trees for building CoreOS.

cgwalters:
> The only way to narrowly scope it to multipath would be to bind it to the multipath module; that felt a bit too hacky. 

Right, we also have the iSCSI case which is similar.  But any dracut module that wants to use this shares all the bits of mounting /boot - the difference just comes in that the individual modules would cherry-pick their own config files; IOW:

dm-multipath-pre.service
```
ConditionKernelCmdline=boot
RequiresMountsFor=/boot
ExecStartPre=/bin/cp /boot/etc/multipathd.conf /etc
...
```

iscsid-pre.service:
```
ConditionKernelCmdline=boot
RequiresMountsFor=/boot
ExecStartPre=/bin/cp /boot/etc/iscsid.conf /etc
```

etc.  (Written as a systemd unit instead of the shell, but the same idea applies in both).

I agree it's a good idea to find solutions that will apply to traditional systems too!  And nothing about the "narrow" approach excludes that - IOW once that change landed at some point e.g. Anaconda could be changed for "traditional" Fedora systems to set up dm-multipath that way instead of regenerating the initrd.  But..the *motivation* for doing that is a lot lower if one is always regenerating the initrd locally anyways.

Or to rephrase this, either way something different is happening; CoreOS uses `initrd-etc` and traditional doesn't, or CoreOS uses `/boot/etc/multipath.conf` and traditional doesn't.

cgwalters:
To ensure this is all xref'd; Lennart suggested [using EFI variables](https://lists.freedesktop.org/archives/systemd-devel/2019-September/043486.html) ([my followup](https://lists.freedesktop.org/archives/systemd-devel/2019-November/043754.html)) but it just doesn't work for us because e.g. AWS is BIOS only and we need a uniform mechanism (not to mention e.g. needs to work on s390x too etc.).

cgwalters:
Another angle on this code/configuration thing; as this stands today someone could write an `/boot/initrd-etc/systemd/system/mycustom.service` (and I don't think it would work because systemd would already have loaded the configuration, i.e. something would need to `systemctl daemon-reload`) - but systemd units are arbitrary code that are being injected into the initramfs in this approach.  Whereas if individual services know to look for just their configuration, we avoid the injected code.

darkmuggle:
> Another angle on this code/configuration thing; as this stands today someone could write an `/boot/initrd-etc/systemd/system/mycustom.service` (and I don't think it would work because systemd would already have loaded the configuration, i.e. something would need to `systemctl daemon-reload`) - but systemd units are arbitrary code that are being injected into the initramfs in this approach. Whereas if individual services know to look for just their configuration, we avoid the injected code.

That's why I didn't do a `daemon-reload`. I considered that part of the idea. And no, it doesn't work. I tested that. 

Given the strong distaste for this PR, I'll rework this.

cgwalters:
> Given the strong distaste for this PR, I'll rework this.

Well, we can see what the other drauct devs think too!

darkmuggle:
xref to https://github.com/coreos/ignition-dracut/issues/154 and the commend that inspired this PR https://github.com/coreos/ignition-dracut/issues/154#issuecomment-606881880 for background

jlebon:
I guess one other concern I have with having configs in `/boot` in general is that now we'd have two sources of truth: `/etc` (where dracut sources configs from today) and `/boot/initrd-etc`.

If we constrain this to *just* solving the firstboot case, I think doing something similar to https://github.com/coreos/fedora-coreos-config/pull/346 could make sense. (I.e. apply in initrd, propagate to real root, delete directory). Would that solve the multipath scenario?

cgwalters:
> If we constrain this to just solving the firstboot case, I think doing something similar to coreos/fedora-coreos-config#346 could make sense. (I.e. apply in initrd, propagate to real root, delete directory). Would that solve the multipath scenario?

We need to find the multipath configuration on every boot.

darkmuggle:
> I guess one other concern I have with having configs in `/boot` in general is that now we'd have two sources of truth: `/etc` (where dracut sources configs from today) and `/boot/initrd-etc`.

For an RHCOS in the scope of OCP, there is no easy mechanism to regenerate the initrd.  

The two-sources of truth concern, IMO, is mitigated by the fact that this is an opt-in (off-by-default) option. Turning it is explicitly stating you want to use the mechanism. 

> If we constrain this to _just_ solving the firstboot case, I think doing something similar to [coreos/fedora-coreos-config#346](https://github.com/coreos/fedora-coreos-config/pull/346) could make sense. (I.e. apply in initrd, propagate to real root, delete directory). Would that solve the multipath scenario?

No. Not at all. If root is on a multipath, then it has to be enabled in the initramfs before root is mounted. If the root is mounted first, then the multipath is not activated.  So you have to regenerate the initramfs. I didn't pursue using the MCO to trigger an initramfs regeneration because the MCO can't deliver payloads on a node level; multipath configuration can be/often-is a node-level configuration.  

If:
- appending or regenerating the initrd is out (due to desires to sign, don't want to mutate it)
- reading the files from /boot alternative path is out
- opening the root device and fetching a while-list is worse, especially in the case of LUKS (open, read files, close LUKS, bounce a few things, re-open....)

Then multipath, iSCSI and other complicated disk setups can't be supported by CoreOS.

darkmuggle:
I fear that we might wear our welcome here by debating CoreOS specific concerns here. 

@jlebon @cgwalters @dustymabe let's take the debate to https://github.com/coreos/fedora-coreos-tracker/issues/464 And the Dracut Developers are more than welcome to come to our space. 

I'm inclined to close until a decision is made at the CoreOS level. If the idea wouldn't be merged into CoreOS's own codebase, why ask Dracut to consider it?

cgwalters:
I don't think this problem domain is solely CoreOS specific - it applies to any "image based" system which wants to ship a server-side generated initramfs, right?  And it's good if solutions we come up with can be adopted by other distributions using non-OSTree mechanisms.

darkmuggle:
> I don't think this problem domain is solely CoreOS specific - it applies to any "image based" system which wants to ship a server-side generated initramfs, right? 

Agreed. And that's what I tried to do with this, and why I liked your suggestion to cargo-cult in `/boot/etc`.  And it was your suggestion that I submitted here first. 

> And it's good if solutions we come up with can be adopted by other distributions using non-OSTree mechanisms.

I  suggested that we move the discussion because the scope of conversation had become one about OSTree needs. I had proposed the general solution and it quickly rabbit holed. I would be happy to consider discussing the best implementation for the feature within the scope of the generic Linux community for image-based workflows. CoreOS' needs were indeed the opening justification, but it has dominated the discussion. I'd rather we have our myopic discussions in our own space.

To be clear -- your counter idea of targeted and per-module overrides makes sense, and there's merit to it. As does the merits of Jonathan's concerns about not having N sources of truth (although, having 18 ways to the same thing is a thing in Linux). The solution as proposed puts the power in the hands of the user by allowing for easy side-loading. The tension between the target path and this solution is really whether we want a side-loading functionality for users and to which degree. Not debating OSTree specific concerns.  

My position is that once we enable if Multipath and iSCSI, someone else will want it for something else. In looking at it for Multipath, I concluded that a general side-loading mechanism is pro-user and gives a good balance between image base workloads against image-based workloads. Clearly there are downsides.  My argument is that reducing friction for users will increase adoptions while image masters do their arcane and byzantine dark magic. 

I do think that we should continue the debate for CoreOS in the Fedora CoreOS. Trust me, had I submitted it to Fedora CoreOS or Ignition Dracut this would have looked very different. The general side loading of `/etc` is a starting point and something that the Dracut community can decide. CoreOS is a member of the Dracut community, too. The danger I am trying to avoid is having a change that only suits our needs or having an exclusively CoreOS discussion. 

With that, let's continue the debate on the generic solution here. And then figure it how/if/what to use in our ivory towers.

darkmuggle:
@haraldh Can I bother to weigh in here?  There's a lot going on in the conversations, and without understanding how the Dracut Community feels, this idea is pretty much dead in my opinion.

haraldh:
For the reference.. @darkmuggle and I chatted in private.

I opted for either a `unsecure-config` source, which has to be treated carefully and/or a signed config source (signed by the Machine Owner Key or similar).

IMHO, ideally it would be a signed config cpio, which could be checked by grub, so we would avoid any reload of systemd services.

Issue #727, opened 2020-02-17
How to use cmdline.d on host without creating cmdline.d/90crypt.conf on initramfs


Neurognostic:
## Issue

How can I use the `/etc/cmdline.d/` directory on the host for specifying kernel parameters without `/etc/cmdline.d/90crypt.conf` being created on the initramfs?

### Details

When my root device is a LUKS encrypted volume and `hostonly=yes` is set the `crypt` module automatically creates `/etc/cmdline.d/90crypt.conf` on the initramfs with the line `rd.luks.uuid=<MY-UUID>`. This causes an error to printed during boot right on top of where I type my LUKS passphrase when `dracut-initqueue` is trying to find root, because root was already found by `systemd-cryptsetup-generator(8)` from kernel parameter specified elsewhere (e.g. boot loader configuration or with the `kernel_cmdline` option in `dracut.conf(5)`).

```shell
dracut-initqueue[518]: Failed to start systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service: Unit systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service: Unit systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service not found.
```

The error does not prevent my root LUKS volume from being opened and mounted.

---

The only way I have found to prevent this error is to either use `hostonly=no` *or* `hostonly_cmdline=no`, but when I use either of these options `dracut(8)` no longer parses the `cmdline.d` directory. Which forces me to use the `dracut.conf(5)` option `kernel_cmdline` when embedding kernel parameters.

Because of this, the `cmdline.d` directory is useless. I have not found anyway to prevent `cmdline.d/90crypt.conf` from being created on the initramfs while still allowing `cmdline.d/` to be parsed with `hostonly_cmdline=yes`.

This issue is occurs with the latest master branch.

haraldh:
might be fixed with 16abea2592d7fee6166660098b647356d5a4cc14

Btw, systemd does not read `/etc/cmdline.d`.

diabonas:
See your previous issue #727, I think the
```
dracut-initqueue[518]: Failed to start systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service: Unit systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service: Unit systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service not found.
```
error should hopefully be fixed by #794.

Issue #723, opened 2020-02-13
Failed to start systemd-cryptsetup@<uuid>.service


Neurognostic:
This is an issue with the master branch version on Arch Linux.

I was looking to try out a new feature (secure boot signing), but unfortunately I am seeing this weird error printed on top of where I type my LUKS password. The same exact `dracut.conf(5)` and `dracut.cmdline(7)` used on v49 works without issue (except missing features).

The strange thing is I can still type in my password and successfully decrypt and mount root without issue.

```log
dracut-initqueue[518]: Failed to start systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service: Unit systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service: Unit systemd-cryptsetup@luksx2d<<CRYPT-UUID>>.service not found.
```

I ran with `rd.debug` parameter and did not see anything additional that would help, just this same error. It did appear to be looping the same debug output while I typed my password, which I have not seen when running in debug mode on v49.

Here is the full debug log for the initramfs minus kernel messages: 
[dracut-master.debug.log.zip](https://github.com/dracutdevs/dracut/files/4197774/dracut-master.debug.log.zip)

How to reproduce:
```conf
# /etc/dracut.conf.d/myconf.conf
hostonly="yes"
hostonly_cmdline="yes"
compress="lz4"
show_modules="yes"
add_drivers+="lz4 lz4_compress"
uefi_stub="/usr/lib/systemd/boot/efi/linuxx64.efi.stub"
uefi_secureboot_cert="/etc/efi-keys/db.crt"
uefi_secureboot_key="/etc/efi-keys/db.key"
LINUX_CMDLINE=(
#       rw.info
#       rd.shell
#       rd.debug
       rw
#       rd.cmdline=ask
       rd.luks.timeout=60
       rd.luks.crypttab=0
       rd.luks.uuid=$(cryptsetup luksUUID /dev/disk/by-partlabel/cryptsystem)
       root=UUID=$(blkid -o value -s UUID /dev/disk/by-label/system)
)
kernel_cmdline="${LINUX_CMDLINE[*]}"
unset LINUX_CMDLINE
```

```shell
dracut --force --verbose --uefi --kver 5.5.3-arch1-1 /efi/initramfs-linux-unified.efi.signed
```

Any idea what is going on here?

Neurognostic:
[...]

PR #794, opened 2020-04-18
dracut-lib.sh: quote variables in parameter expansion patterns

diabonas:
According to [POSIX.1-2017, 2.6.2 Parameter Expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02):
> *${parameter%<b>[</b>word<b>]</b>}*
> [...] The *word* shall be expanded to produce a pattern.

This means if word contains variables that itself contain special characters like asterisks or backslashes, these are treated as pattern characters unless the variable is quoted. Try e.g. the following example in bash, dash or (busybox) ash:
```
i='a\c'; j='\'; echo "${i%$j*}"
```
This prints `a\c` because `$j*` is expanded to `\*`, escaping the asterisk. In contrast,
```
i='a\c'; j='\'; echo "${i%"$j"*}"
```
produces the expected result `a` because the backslash is not specially treated any more after quoting.

The quotes that this commit adds have been previously removed in commit f9c96cf56fed390841eac05c43826e62014c9188, citing issues with busybox hush without further specifying the actual error. I tested a recent busybox build (upstream commit [9aa751b08ab03d6396f86c3df77937a19687981b](https://git.busybox.net/busybox/commit/?id=9aa751b08ab03d6396f86c3df77937a19687981b)) and couldn't find any problems. Note that the above example always produces `a\c` in hush regardless of quoting `$j`, making hush unsuitable
for use with dracut, but using quotes in parameter expansions generally works.

The unquoted variables break the "rd.luks.uuid/name" kernel command line options in dracut 050 because
https://github.com/dracutdevs/dracut/blob/25c7a13985473ff9714245eaa602225b1a0f9ecb/modules.d/90crypt/parse-crypt.sh#L68
is not able to escape the backslashes any more, see #723, #727: backslashes in the `systemd-cryptsetup@.service` unit name stay unescaped for use in udev (cf. commit 0f6d93eb9d63695a64002ec8b0421fbc9fc8a7a3), leading to failures in starting the unit.

This partially reverts commit f9c96cf56fed390841eac05c43826e62014c9188.

### DIFF
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -24,7 +24,7 @@ debug_on() {
 
 # returns OK if $1 contains literal string $2 (and isn't empty)
 strstr() {
-    [ "${1##*$2*}" != "$1" ]
+    [ "${1##*"$2"*}" != "$1" ]
 }
 
 # returns OK if $1 matches (completely) glob pattern $2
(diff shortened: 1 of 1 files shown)


haraldh:
Thank you!

PR #801, opened 2020-04-21
Always pull in machinery to read ifcfg files

pjgeorg:
So far machinery is only pulled in if the user has not yet included any ifcfg files.

Noticed when boot failed after changing my custom module's prefix to be loaded before other modules.
The current code contradicts the code and depends on when user decides  to add ifcfg files.
I therefore believe that it would be better to simply always pull in the required library.


haraldh:
thanks!

PR #805, opened 2020-04-30
95nfs: only install rpc services for NFS < 4 when hostonly is strict

ryncsn:
Most rpc services are not needed for NFSv4, so skip these components
to save space. rpc.idmapd is still needed in case of
nfs4_disable_idmapping is set to 0.

### DIFF
--- a/modules.d/95nfs/nfs-start-rpc.sh
+++ b/modules.d/95nfs/nfs-start-rpc.sh
@@ -1,8 +1,6 @@
 #!/bin/sh
 
-
 if modprobe sunrpc || strstr "$(cat /proc/filesystems)" rpc_pipefs; then
-
     [ ! -d /var/lib/nfs/rpc_pipefs/nfs ] && \
         mount -t rpc_pipefs rpc_pipefs /var/lib/nfs/rpc_pipefs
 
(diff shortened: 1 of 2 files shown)


modules.d/95nfs/nfs-start-rpc.sh
```diff
@@ -17,8 +15,8 @@ if modprobe sunrpc || strstr "$(cat /proc/filesystems)" rpc_pipefs; then
     # Start rpc.statd as mount won't let us use locks on a NFSv4
     # filesystem without talking to it. NFSv4 does locks internally,
     # rpc.lockd isn't needed
-    [ -z "$(pidof rpc.statd)" ] && rpc.statd
-    [ -z "$(pidof rpc.idmapd)" ] && rpc.idmapd
+    command -v rpc.statd >/dev/null && [ -z "$(pidof rpc.statd)" ] && rpc.statd
+    command -v rpc.idmapd >/dev/null && [ -z "$(pidof rpc.idmapd)" ] && rpc.idmapd
```
haraldh, line 19: you might want `type -p` here instead of `command -v`

ryncsn: Hmm, there are a few other `command -v` usage in this script file, so I used `command -v` to keep a consistent code style, any advantage of using `type -p` instead? If `type -p` is superior, perhaps I should change all `command` usage in this script to `type`.

puleglot: Also `type -p` doesn't work in `dash`

PR #809, opened 2020-05-04
Fix test in lsinitrd

Vogtinator:
If `$uefi` is empty, this evaluated to true previously,
resulting in `initrd in UEFI: : 13M`.


danimo:
Thanks!

haraldh:
`[[ -n $uefi ]]` would have not required `""` ... just saying

danimo:
We're in bash land here. So yeah, would probably have been more elegant.

Vogtinator:
It's a mixture of `[[]]` and `[]` already, I didn't want to put any unrelated change into the diff. In fact, even then I would've used `[[ -n "$uefi" ]]` for consistency, as that's how the line above also uses `-z`.

PR #813, opened 2020-05-07
network-manager: set kernel hostname from the command line

bengal:
Since commit ff70adf873ef ("initrd: save hostname to a file in /run"),
the initrd generator of NetworkManager parses the hostname from 'ip='
options of the kernel command line and writes it to
/run/NetworkManager/initrd/hostname.

When that file exists, set the kernel hostname.

In presence of multiple hostnames in the command line, the last one
wins. Hostnames from command line always have precedence over ones
received through DHCP. This is a bit different from the legacy network
module that gives higher precedence to the hostname (from DHCP or
command line) of the last interface that is brought up, which depends
on the udev order.


haraldh:
please rebase

haraldh:
will merge via
https://github.com/dracutdevs/dracut/pull/831

PR #814, opened 2020-05-11
rngd: do not start inside container

tpgxyz:
Host should take care of good entropy pool
https://github.com/dracutdevs/dracut/pull/791

### DIFF
--- a/modules.d/06rngd/module-setup.sh
+++ b/modules.d/06rngd/module-setup.sh
@@ -33,6 +33,8 @@ check() {
 install() {
     inst rngd
     inst_simple "${moddir}/rngd.service" "${systemdsystemunitdir}/rngd.service"
+    # make sure dependant libs are installed too
+    inst_libdir_file opensc-pkcs11.so
 
     systemctl -q --root "$initdir" add-wants sysinit.target rngd.service
 }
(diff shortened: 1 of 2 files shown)


danimo:
Excuse my ignorance, but why would you run dracut inside of a container in the first place? Or is this about virtualized containers? In that case, is there really a guarantee that the guest always has entropy available?

tpgxyz:
@danimo i think in case of any virtualizations, its about host to take care of entropy pool.
If host entropy is 0 then running any entropy gathering tool does not make it any better.

danimo:
True indeed. Forgot that rngd does not try to generate pseudo randomness like haveged.

modules.d/06rngd/module-setup.sh
```diff
@@ -33,6 +33,8 @@ check() {
 install() {
     inst rngd
     inst_simple "${moddir}/rngd.service" "${systemdsystemunitdir}/rngd.service"
+    # make sure dependant libs are installed too
+    inst_libdir_file opensc-pkcs11.so
```
danimo, line 37: This is not a hard dependency on all platforms, e.g. openSUSE, so now, this line will fail there.

nhorman: start rngd with the -x pkcs11 option, that should prevent that entropy source from initializing, and  squash the library dependency, so it need not be included in the initramfs

tpgxyz:
@danimo  I've prepared new PR https://github.com/dracutdevs/dracut/pull/905 as i lost that branch... Please review it, as this PR i'm closing.

PR #815, opened 2020-05-11
rngd: install dependant libs too

tpgxyz:
By default rng-tools are compiled with pkcs11 support.
Make sure opensc-pkcs11.so library is installed inside initramfs to prevent error on boot

### DIFF
--- a/modules.d/06rngd/module-setup.sh
+++ b/modules.d/06rngd/module-setup.sh
@@ -33,6 +33,8 @@ check() {
 install() {
     inst rngd
     inst_simple "${moddir}/rngd.service" "${systemdsystemunitdir}/rngd.service"
+    # make sure dependant libs are installed too
+    inst_libdir_file opensc-pkcs11.so
 
     systemctl -q --root "$initdir" add-wants sysinit.target rngd.service
 }


modules.d/06rngd/module-setup.sh
```diff
@@ -33,6 +33,8 @@ check() {
 install() {
     inst rngd
     inst_simple "${moddir}/rngd.service" "${systemdsystemunitdir}/rngd.service"
+    # make sure dependant libs are installed too
+    inst_libdir_file opensc-pkcs11.so
```
danimo, line 37: Please make this optional. It's at least not around by default on openSUSE.

tpgxyz: Found no errors when i run dracut -f without opensc-pkcs11.so on my system. Initramfs regenerated correctly

danimo: True. This command implies `-o`. @haraldh was this an explicit design choice? Is there a rule to tell which inst rules imply `-o`?

nhorman:
lazy binding should be the default for rng-tools, if you run rngd with the -x pkcs11 option, it shouldn't attempt to resolve the library and you can leave it out

modules.d/06rngd/module-setup.sh
haraldh: Yeah, it was always defined to be sloppy. So, I didn't change the behavior over time.

PR #825, opened 2020-05-25
Fixes for find_binary() and busybox module

puleglot:
Fixes: a01204202b30

https://bugzilla.redhat.com/show_bug.cgi?id=1810504
https://bugs.gentoo.org/714838


puleglot:
The second patch is optional. Feel free to drop it if you want to keep compatibility with ancient versions of busybox.

PR #826, opened 2020-05-26
98dracut-systemd: Display stdout on console as well

danimo:
This fixes the behavior of info() on systems with systemd,
and matches it to the SysVinit behavior.

Previously, stdout would go exclusively to journal/syslog,
rendering info messages useless for situations where the system
would fail to boot up (e.g. fips).

### DIFF
--- a/modules.d/98dracut-systemd/dracut-cmdline.service
+++ b/modules.d/98dracut-systemd/dracut-cmdline.service
@@ -23,7 +23,7 @@ Environment=NEWROOT=/sysroot
 Type=oneshot
 ExecStart=-/bin/dracut-cmdline
 StandardInput=null
-StandardOutput=syslog
+StandardOutput=syslog+console
 StandardError=syslog+console
 KillMode=process
 RemainAfterExit=yes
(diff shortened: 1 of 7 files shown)


haraldh:
Hmm, just wondering how often this causes the splash screen to disappear for no reason ...

danimo:
The obvious alternative is to let info() go to stderr as well, although that'd be somewhat a lie. Pick your poison.

haraldh:
> rendering info messages useless for situations where the system would fail to boot up (e.g. fips).

I have the feeling, that these messages should be `warn` or `error` then...

danimo:
This is about displaying progresses:

```
    info "Self testing crypto algorithms"
    info "Checking integrity of kernel"
    info "All initrd crypto checks done"
```

Sometimes if the FIPS checks fail for spurious reasons, we would like reporters to tell us how far they got. Seems to me the better fix is to rewrite the part to `error` whenever we currently `return 1` instead of printing progress that nobody is able to see because they are denied an emergency shell under the FIPS rules.

haraldh:
> Seems to me the better fix is to rewrite the part to `error` whenever we currently `return 1` instead of printing progress that nobody is able to see because they are denied an emergency shell under the FIPS rules.

yes

danimo:
Superseded by #851 .

PR #838, opened 2020-06-10
Fips workaround for Fedora/RHEL

ryncsn:
On Fedora/RHEL, libcryto will verify both itself and libssl on start, if
libssl is missing, FIPS self test will fail. However libssl is not a
dependency of libcryto so dracut will not install it, unless some other
binary or library pulls it in. Systemd requires libssl, so in most cases
it just worked, but could fail in some corner cases where systemd is not
used.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1941,6 +1941,17 @@ if [[ $kernel_only != yes ]]; then
             break 2
         done
     done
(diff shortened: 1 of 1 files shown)


haraldh:
Thanks!

PR #844, opened 2020-06-17
95nvmf: add module for NVMe-oF

hreinecke:
Add a module for booting from NVMe-oF devices.


danimo:
Two are missing:

```
In parse-nvmf-boot-connections.sh line 53:
    [ -z "hosttraddr" ] && hosttraddr="none"
          ^--------^ SC2157: Argument to -z is always false due to literal strings.


In parse-nvmf-boot-connections.sh line 54:
    [ -z "trsvcid" ] && trsvcid="none"
          ^-----^ SC2157: Argument to -z is always false due to literal strings.
```
Also:
```
In parse-nvmf-boot-connections.sh line 84:
    parse_nvmf_discover $d
                        ^-- SC2086: Double quote to prevent globbing and word splitting.
```

May be a good idea if the arguments can contain spaces.

hreinecke:
> Two are missing:
> 
> ```
> In parse-nvmf-boot-connections.sh line 53:
>     [ -z "hosttraddr" ] && hosttraddr="none"
>           ^--------^ SC2157: Argument to -z is always false due to literal strings.
> 
> 
> In parse-nvmf-boot-connections.sh line 54:
>     [ -z "trsvcid" ] && trsvcid="none"
>           ^-----^ SC2157: Argument to -z is always false due to literal strings.
> ```

Fixed now.

> 
> Also:
> 
> ```
> In parse-nvmf-boot-connections.sh line 84:
>     parse_nvmf_discover $d
>                         ^-- SC2086: Double quote to prevent globbing and word splitting.
> ```
> 
> May be a good idea if the arguments can contain spaces.

It can't, but I've enclosed it nevertheless.

PR #845, opened 2020-06-25
Support NetworkManager in the cms module

bengal:
Change the cms module to support operation with NetworkManager instead of network-legacy.


haraldh:
Thank you very mucho!

PR #853, opened 2020-07-08
Send DHCP on single device to improve boot time

anjalidk:
devices, on which DHCP is sent. However, in most cases, only the
first one succeeds (the one for the minimum BDF device).
The second one keeps re-trying DHCP, adding almost a minute to
the bootup time, before failing. This patch sends DHCP only for
the minimum BDF device.
This is enabled only when we add ip=single-dhcp in kernel command
parameters. If the DHCP fails on minimum BDF device for some
reason, we keep note of this failed device in
/run/initramfs/dhcpfaildev. This will ensure correctness for the
case when the first device passed to ifup.sh was the minimum BDF,
but it failed; then subsequent calls to ifup.sh for other devices
will skip using the minimum BDF device for sending DHCP, and
choose another device.
Where applicable, this patch gives an improvement of about 1
minute to the system bootup time.

### DIFF
--- a/modules.d/35network-legacy/ifup.sh
+++ b/modules.d/35network-legacy/ifup.sh
@@ -23,6 +23,75 @@ if [ "$netif" = "lo" ] ; then
     exit 0
 fi
 
+min_device=""
+DHCP_FAIL="/run/initramfs/dhcpfaildev"
+UDEVLOG="/run/initramfs/udevlog"
+
+min_bdf_device() {
+    devs=$(ls /sys/class/net)
+
+    # Get the udevadm output in file $UDEVLOG
+    udevadm info -e > "$UDEVLOG"
+
+    for netdev in $devs; do
+        if [ "$netdev" = "lo" ]; then
+            continue
+        fi
+        if [ -e  "$DHCP_FAIL" ]; then
+            faildh=$(awk -v patrn="$netdev" '$0==patrn' "$DHCP_FAIL")
+            if [ "$faildh" = "$netdev" ]; then
+                warn "DHCP failed on $netdev earlier, skipping"
+                continue
+            fi
+        fi
+
+        # As an example, relevant output of udevadm when searching for a
+        # string ending with net/eno2 (eg. for device eno2), gives:
+        # P: /devices/pci0000:3a/0000:3a:00.0/0000:3b:00.0/net/eno2
+        # E: DEVPATH=/devices/pci0000:3a/0000:3a:00.0/0000:3b:00.0/net/eno2
+        # Choosing 5th field (with FS=/) and then the first line gives us the
+        # domain:bus:slot.func (in $pcinfo below)
+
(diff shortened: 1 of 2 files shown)


bengal:
> devices, on which DHCP is sent. However, in most cases, only the
> first one succeeds (the one for the minimum BDF device).

Can you elaborate on what is a BDF, and why is it common that only the lowest BDF succeeds? It seems arbitrary to me, why not the second one or just a random one?

Also, if you know beforehand that the minimum BDF will succeed, can't you just pass a proper command line as `ip=enp1s0:dhcp`?

anjalidk:
Hi,

The BDF is the Bus, Device, Function used to describe PCI devices.

Generally devices with min BDF are the primary NICs which will succeed 
with DHCP (atleast on OCI cloud this is true). I guess this min BDF 
could have been made a configurable function by a user, where instead of 
min BDF, they could choose some other characteristic, which they know 
will succeed for their environment.

Regarding not using the command line with the interface name, the 
solution then becomes dependent on the interface name, and not a generic 
solution for many thousands of servers, where interface name could be 
different depending on server type or something else. We want a generic 
solution (single image) which works for all servers.

Anjali

bengal:
I agree that it is inconvenient to regenerate the image for different environments; for the same reason, the policy used to chose the boot interface should not be hardcoded in the image itself. "The device with lowest PCI number" could be the right choice for your environment, but not for others. I don't know how OCI works, but I imagine that the entity that launches the image knows what the right interface is, and could pass it as a kernel parameter?

BTW, to mitigate the 1-minute delay, I guess you can add `rd.net.timeout.dhcp=` to the command line to lower the DHCP timeout; but of course a better solution would be to pass the right interface name.

Also note that your commit only changes the 'network-legacy' module. There is another 'network-manager' module that can be used in alternative (and is enabled by default in Fedora). Ideally they should be in sync and provide the same functionality.

Alternatively, there could be a new parameter that tells dracut to continue after one interface is successfully configured.

anjalidk:
Hi,

Sorry for the late response - I was trying out a different solution for 
this - I have implemented a solution to send DHCP requests out on all 
interfaces in parallel. Once DHCP is successful on a particular 
interface, no more DHCP requests are sent out and existing dhclient 
requests are killed. Since all requests are sent out in parallel at the 
same time, there isn't a sequential wait for each interface to succeed 
or fail before DHCP is sent on next interface.

This is helping to reduce boot times not only on the original case, 
where there are 2 interfaces, but also on a 16 interface box, where even 
adding a dhcp timeout, |rd.net.timeout.dhcp=| to the command line, 
results in a 5 mins boot time. With this fix, boot time is reduced to 
about 1min, a drastic reduction in boot time.

This makes the solution generic and usable by anyone having more than 1 
interface and looking to enhance boot times.

Let me know if that sounds ok, and I will submit the patch for verification.

Thanks

Anjali

PR #859, opened 2020-07-15
Systemd type unit logfixes

johannbg:
As of systemd v246 the StandardError= and StandardOutput= in unit files no longer support the "syslog" and "syslog+console" switches. They were long removed from the documentation but are now resulting in warnings ( which for example can be seen in F33 ) when used, and are converted to "journal" and "journal+console" automatically. 

This pull request removes any "StandardError=syslog" and StandardOutput=syslog" in type unit files since they are unnecessary due to the default being "StandardError=journal" and "StandardOutput=journal" in systemd as well as replace any occurrences of syslog+console with journal+console where applicable, which silences those deprecation warnings.


johannbg:
Note that test 36 failing with "Filesystem too small for a journal" has nothing to do with my changes or systemd's journal. It's failing due to the filesystems blocksize not being set to 2048 or higher when it's being created and or in mke2fs.conf. If it's omitted it will default to the size of "device" and if the partition size is equal to or less than 512MB it is treated as "small" in mke2fs.conf which has a blocksize of "1024" by default hence either increase the blocksize in mk2fs.conf to 2048 or higher or run the test with -b 2048 or higher to prevent this test from failing.

johannbg:
@haraldh, @danimo is there something preventing this from being review/merged?

haraldh:
Thank you!

johannbg:
@haraldh no problem and encase you are coming back from holiday and thus might be unaware of but machine-id behaviour is being poked upstream which probably will end up breaking something in Dracut either directly if it expects it to exist or contain a a value or indirectly by changed behaviour in bootctl/kernel-install which probably will need to be changed along side it I'm not sure why things got written in such manner they relied on value in machined-id instead of $UUID or $PARTUUID of the efi partition to begin with. In anycase the requirement seems to have come up that the file can be empty or non-existent which will break whole bunch of things so you might want to ping Lennart to make sure he's on top of those changes ( he's not making it ) so things dont accidentally get merged which could end up breaking things in Dracut and probably Anaconda and perhaps other places as well.

haraldh:
@johannbg Thanks for the info... Not on holiday, but busy with coding for https://github.com/enarx/enarx/

Issue #849, opened 2020-07-02
Support globbing firmware files


danimo:
Some firmware files or firmware auxiliary files (such as the firmware for brcm80211) are non-exhaustive and can thus not be expressed in terms of a fully enumerated firmware list in the module. This was discussed on [LKML](https://lore.kernel.org/lkml/338e3cff-dfa0-c588-cf53-a160d75af2ee@redhat.com/).

I hence propose (and plan to implement) supporting globbing for `dracut_install`'s `install_firmware()` instead.

haraldh:
sure, makes sense... go ahead :)

PR #860, opened 2020-07-17
dracut-install: Globbing support for resolving "firmware:"

danimo:
Implements #849

### DIFF
--- a/install/dracut-install.c
+++ b/install/dracut-install.c
@@ -27,6 +27,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
+#include <glob.h>
 #include <libgen.h>
 #include <limits.h>
 #include <stdbool.h>
@@ -123,6 +124,7 @@ static inline void fts_closep(FTS **p) {
 }
 #define _cleanup_fts_close_ _cleanup_(fts_closep)
 
+#define _cleanup_globfree_ _cleanup_(globfree)
 
 
 static size_t dir_len(char const *file)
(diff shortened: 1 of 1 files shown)


install/dracut-install.c
```diff
@@ -1362,27 +1386,22 @@ static int install_firmware(struct kmod_module *mod)
                                 exit(EXIT_FAILURE);
                         }
 
-                        fw = fwpath;
-                        if (stat(fwpath, &sb) != 0) {
-                                r = asprintf(&fwpath_xz, "%s.xz", fwpath);
-                                if (r < 0) {
-                                        log_error("Out of memory!");
-                                        exit(EXIT_FAILURE);
+                        if (strstr(fwpath, "*") != 0 || strstr(fwpath, "?") != 0 || strstr(fwpath, "[") != 0) {
```
haraldh, line 1389: With the assumption that `*` `?` `[` will never occur in the filename? At least, check that is was not escaped :-)
Also, I would check `value` instead of `fwpath`.

haraldh: Ah, maybe check if `fwpath` exists, if not check for the regex chars, then do the `glob`

haraldh (changes requested):

install/dracut-install.c
danimo: Implemented, but in reverse (string comparison should be cheaper than a stat)

danimo:
Merging since this has been stable in the SUSE branch for a while.

PR #864, opened 2020-07-25
dracut.sh: fix ia32 detection for uefi executables

q66:
As far as I can tell, uname -m will never output ia32, only i386 or i686.


danimo:
Good catch! But slow runners... I'll merge once they are done.

danimo:
Can you update, please?

q66:
done

danimo:
here as well...

q66:
done

q66:
by the way, chances are this should be supported for aarch64 as well - i might get a chance to test it later, so we'll see, i'll make a new PR for that if it works

PR #865, opened 2020-07-25
dracut.sh: fix early microcode detection logic

q66:
This fixes two issues:

1) on non-x86 systems in non-hostonly config this would cause
   an annoying warning on every initramfs generation
2) on non-x86 systems in hostonly config this would result in
   early microcode not getting disabled

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1194,19 +1194,26 @@ fi
 
 if [[ $early_microcode = yes ]]; then
     if [[ $hostonly ]]; then
-        [[ $(get_cpu_vendor) == "AMD" ]] \
-            && ! check_kernel_config CONFIG_MICROCODE_AMD \
-            && unset early_microcode
-        [[ $(get_cpu_vendor) == "Intel" ]] \
-            && ! check_kernel_config CONFIG_MICROCODE_INTEL \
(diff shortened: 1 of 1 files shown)


dracut.sh
```diff
@@ -1194,19 +1194,26 @@ fi
...
     else
         ! check_kernel_config CONFIG_MICROCODE_AMD \
             && ! check_kernel_config CONFIG_MICROCODE_INTEL \
             && unset early_microcode
     fi
-    [[ $early_microcode != yes ]] \
-        && dwarn "Disabling early microcode, because kernel does not support it. CONFIG_MICROCODE_[AMD|INTEL]!=y"
+    # Do not complain on non-x86 architectures as it makes no sense
+    case $(uname -m) in
+        x86_64|i*86)
```
danimo, line 1211: Any reason not to use ? here? Effectively, this can only be 3 (older kernels), or 4-6...

q66: you're right, though it shouldn't really matter (there aren't any other unames that would match as false positive)

q66: fixed now, fixed it in the other PRs as well

danimo:
Thanks!

PR #866, opened 2020-07-25
mkinitrd-dracut.sh: use vmlinux regex for ppc*, vmlinuz for i686

q66:
Previously this would not catch ppc64le, now it does; same with i686.

### DIFF
--- a/mkinitrd-dracut.sh
+++ b/mkinitrd-dracut.sh
@@ -57,10 +57,10 @@ default_kernel_images() {
         s390|s390x)
             regex='image'
             ;;
-        ppc|ppc64)
+        ppc*)
             regex='vmlinux'
             ;;
-        i386|x86_64)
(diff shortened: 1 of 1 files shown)


mkinitrd-dracut.sh
```diff
@@ -57,10 +57,10 @@ default_kernel_images() {
         s390|s390x)
             regex='image'
             ;;
-        ppc|ppc64)
+        ppc*)
             regex='vmlinux'
             ;;
-        i386|x86_64)
+        i*86|x86_64)
```
danimo, line 63: Same as in other commit, ? would be preferable.

danimo:
Can you update, please?

q66:
done

danimo:
Hmm, again please. Not sure why It can't rebase...

q66:
try now

PR #867, opened 2020-07-25
90ppcmac: respect DRACUT_ARCH, don't exclude ppcle

q66:
respect `DRACUT_ARCH` like other modules, and 32-bit PowerPC Macs can possibly boot in LE mode (and there is some work undergoing), so don't exclude ppcle


danimo:
Thanks. Looks good, let's wait for the runners...

danimo:
Can you update, please?

q66:
done

PR #870, opened 2020-07-29
99shutdown: fix $ACTION failures if underlying systemctl command does not go through

### LINKED ISSUE #869: 99shutdown: $ACTION fails if systemctl dbus is down
Hey Dracut Dev's:
I'm working on a project to upgrade from the RHEL based kernel in a service that uses CentOS 7.8 to an upstream kernel. The project is targeting TOT Linux 5.4.y, currently 5.4.53.

I've got everything working great except for one small issue. I can't reboot the system :) 

Happy to file an RH BZ; however, I wanted to discuss this here and propose what I found as a fix upstream first.

_**Background**_
This service runs in a virtual machine that could be deployed on ESXi, Hyper-V, KVM, or AWS "metal" instances, depending on the customer requirements. Disk devices for this VM are presented via PCIe passthrough devices, such disks hanging on mpt3sas device(s) or NVMe device(s), with one or more disks depending on the service t-shirt size. This makes the actual storage devices the same across all four deployment modes.

Given these are direct disk devices, we have to have some way to protect the operating system partitions (e.g. root, home, blah blah blah), so we use mdraid to mirror them across two SSDs. 

_**Problem Statement**_
If I do `sudo init 6` or `sudo init 0`, or just about any other incantation of `reboot`, `poweroff`, `shutdown`, etc etc, it does not work.

The system will say dracut Warning reboot failed, and drop into the emergency shell.

In the rd.debug shell, attempting to type `reboot` or `reboot -f` or `reboot -f -d -n` do not do anything, and return silently/immediately with status code zero. The only way out is to either type exit several times, sysrq reset the system, or reset from the hypervisor level.

_**Debugging information**_
I've turned up all of the debug knobs I could find, namely: 
`rd.debug rd.break=pre-shutdown rd.break=shutdown rd.shell log_buf_len=1M systemd.log_level=debug systemd.log_target=kmsg printk.devkmsg=on debug`

When I drop into the debug shell (just using rd.debug for example), the console will show that the reboot has failed and shows me the path followed, which is just plain old 99shutdown, nothing remarkable.

Given I have mdraid, I'm automagically populating `/run/initramfs/.need_shutdown`, so it is following the 99shutdown code path, which is fine. The mdraid appears to shutdown just fine, and everything appears to be unmounted just fine, it simply doesn't reboot.

Upon investigation, and much fussing around with the 99shutdown script and doing `dracut -fv`, the failure point is simply that the reboot command itself is not working right here: https://github.com/dracutdevs/dracut/blob/RHEL-7/modules.d/99shutdown/shutdown.sh#L112
```
case "$ACTION" in
    reboot|poweroff|halt)
        $ACTION -f -d -n      <<< this doesn't do anything in my case, it just returns and immediately drops to warn below
        warn "$ACTION failed!"
        ;;
    kexec)
        kexec -e
        warn "$ACTION failed!"
        reboot -f -d -n
        ;;
    *)
        warn "Shutdown called with argument '$ACTION'. Rebooting!"
        reboot -f -d -n
        ;;
esac
```

In the rd.debug shell, if I type reboot, I get the same behavior as the script, a no-op. I found that the reboot command simply appears to be an alias that does this: 
```
#!/bin/bash

((sleep 1; systemctl reboot) &>/dev/null </dev/null & exit 0)
```

Naturally, I tried `systemctl reboot` (as well as with `-f -d -n`) and all of them fail as they want to go through D-Bus, but d-bus it one of the first services shutdown, so they are not available.

**HOWEVER, if I do `systemctl reboot -ff`, it reboots like a charm.** 

Doing some research around that, I found this post from one of the systemd devs @poettering https://github.com/systemd/systemd/issues/3282#issuecomment-220006014, which describes the differences between `systemctl reboot -f` and `-ff`:

> "systemctl poweroff" shuts down going through the usual clean service termination logic.
> "systemctl poweroff -f" skips the clean service termination logic, and goes directly into the final killing and umount spree. If this is used, after a reboot file systems will be clean.
> "systemctl poweroff -ff" will sync and immediately reboot. In this case, the file system will be considered dirty after the reboot, and fsck will run on those file systems that need that concept.
> The first two listed above require D-Bus, and if that's hosed, they will fail. However, you can achieve a similar effect simply by sending a signal to PID 1. Specifically: SIGRTMIN+4 is the fully clean shutdown, and SIGRTMIN+14 is the expedited shutdown, that still results in clean file systems.
> Things are similar for reboot. "systemctl reboot" is the fully clean reboot, "systemctl reboot -f" is the one where clean service shutdown is skipped and the killing and umounting spree is entered directly. "systemctl reboot -ff" will sync() and reboot() immediately without any killing spree or unmounting.

_**Proposed Solution**_
in the 99shutdown/shutdown.sh script, simply replace `$ACTION -f -d -n` with `systemctl $ACTION -ff`

_**Train of thought on this solution "feels right"**_
I took a look at the srpm for systemd on CentOS7: http://vault.centos.org/7.8.2003/updates/Source/SPackages/systemd-219-73.el7_8.8.src.rpm

I see that in `src/systemctl/systemctl.c`, the reboot command has the following assertions that align with how `$ACTION -f -d -n` is being called in both the RHEL-7 and RHEL-8 branches of dracut.

1. `-f` is force, simple stuff
2. `-n` in rhel 7 systemd does nothing at all, and it is simply a nop, only there for forward compatibility. In the RHEL8 systemd srpm, -n means no sync, so we will not do a file system sync at all
3. -d is not write a wtmp record

From the perspective of 99shutdown, by the time we reach `$ACTION -f -d -n` we've successfully unmounted everything, and we've shutdown all DM's and MD's, along with all system services and user processes. 

With that in mind, what I believe dracut is declaring in 99shutdown is:

- I've already done all other shutdown processes via my normal shutdown paths in systemd
- I'm in 99shutdown because `.need_shutdown` is present
- 99shutdown is taking care of all "stragglers" that need some graceful work, such as md-raid disassembly, etc (all prior work in shutdown script, i.e. `_check_shutdown` and everything before it)
- When all of that is done, do `reboot|poweroff|halt` with the following assertions:
- I want this system off **_now_** (`-f`)
- I do not care about file systems syncs (`-n`) ... as everything is `umount`'d and disassembled already
- Do not write any records of this in wtmp record (`-d`) ... because we're about to kill this thing

It seems at that point, the only thing left to do is to shoot the system dead and come back up.

At this point, we exec $ACTION (i.e reboot command); however if for some reason the reboot command fails, such as situations where it is a wrapper around `systemctl reboot` (which requires D-Bus but D-Bus is down at this point) the system will fail to restart, and we'll dump to `warn` path.

**Therefore** ... to deterministically remove the dependency on D-Bus, we should use `systemctl $ACTION -ff` 

Thoughts, comments, questions? Happy to provide any and all logging, additional info, etc etc

JonKohler:
Problem:
In 99shutdown, it is possible for $ACTION -f -d -n to fail for reboot/poweroff,
in a situation where the $ACTION depends on systemctl D-Bus to be available.

At this point in the shutdown process, all processes/services are down, all FSs
are umount'd and all services that dictate .need_shutdown are downed by the
_check_shutdown loop. All that remains is to reboot|poweroff|halt the system
and there is no remaining sync, umount, or other shutdown work to be done.

Fix:
Change $ACTION -f -d -n to use systemctl $ACTION -ff to --force --force the
$ACTION. This will bypass any requirement on D-Bus and kick the system in gear.

Closes #869. See #869 for details.

### DIFF
--- a/modules.d/99shutdown/shutdown.sh
+++ b/modules.d/99shutdown/shutdown.sh
@@ -118,7 +118,7 @@ getarg 'rd.break=shutdown' && emergency_shell --shutdown shutdown "Break before 
 
 case "$ACTION" in
     reboot|poweroff|halt)
-        $ACTION -f -n
+        systemctl $ACTION -ff
         warn "$ACTION failed!"
         ;;
     kexec)


haraldh:
In a systemd environment, the $ACTION is symlinked anyway to `systemctl`. I don't want to use the force flag here, because it works for everybody for years now in a normal environment.

PR #871, opened 2020-08-03
dracut.sh: correct x86 machine hardware name

sgn:
`uname -m` reports i[34567]86 on x86 machine instead of ia32.


sgn:
Duplicate of #864

PR #876, opened 2020-08-04
Change the order of NFS servers during the boot

merlyel:
NFS server provided by DHCP in next-server option has higher priority than DHCP-server itself.

RH Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1859513


danimo:
Thanks!

PR #877, opened 2020-08-05
35network-legacy: simplify fallback dhcp setup

tblume:
combine code for "no ip option directed at our interface" and
"No ip lines default to dhcp"
correct evaluation of return code
fix application of "load_ipv6" call to ipv6 setup only


danimo:
Thanks

PR #883, opened 2020-08-06
Labels

johannbg:
This adds automate pull request labels via github actions for master and all RHEL branches.

### DIFF
--- a/.github/master-label-triggers.yml
+++ b/.github/master-label-triggers.yml
@@ -1,0 +1,313 @@
+# Add 'needs review' label to any pull request
+needs review:
+- "*"
+
+# Add 'repository' label to any root file changes
+repository:
+- ./*
+
+# Add 'github' label to any .github/ file changes
+github:
+- .github/*
+- .github/**/*
+
+# Add 'mkosi' label to any .mkosi/ file changes
+mkosi:
+- .mkosi/*
+- .mkosi/**/*
+- .mkosi/**/**/*
+
+# Add 'dracut install' label to any install/  file changes
+dracut install:
+- install/*
+- install/**/*
+
+# Add 'example' label to any examples/ file changes
+example:
+- examples/*
+- examples/**/*
+- examples/**/**/*
+
+# Add a 'modules' label to any modules.d/ file changes
+modules:
+- modules.d/*
+- modules.d/**/*
+
+# Add a 'bash' label to any bash module changes
+- modules.d/00bash/*
+
+# Add a 'bootchart' label to any bootchart module changes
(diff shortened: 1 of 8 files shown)


.github/master-label-triggers.yml
```diff
@@ -0,0 +1,313 @@
...
+# Add 'repository' label to any root file changes
+repository:
+- ./*
+
+# Add 'github' label to any .github/ file changes
+github:
+- .github/*
+- .github/**/*
+
+# Add 'mkosi' label to any .mkosi/ file changes
```
danimo, line 14: This looks like it would not apply to dracut. Is that from systemd?

johannbg: I'm planning on submitting mkosi setup pr request for dracut project to make it somewhat easer develop and test dracut on multiple distributions hence I ( prematurely ) including that line from having to do it later in the process.

johannbg:
> The rhel triggers look very repetitive. If we can't reuse one yml file for all distros (e.g. because some distros ship their own modules), I think we should at least generate them to keep the error rate low.

We can have a single distro file for example rhel-labeler.yml below at the cost of not being able to lookup issues on individual branches ( label:branches label:rhel-6 )(label:branches label:rhel-7 etc ) which may or may not be a wanted behaviour. Given that I'm not part of Red hat's dracut team or Suse I have no insight how downstream vendors might want to have this or what value it is general from upstream as in we might even want to just skip the vendor labelers altogether but having them might give downstream some additional freedom for their teams to adapt to upstream. ( upstream has one set of labels and workflows while suse has one for their team and red hat another for theirs etc ).

name: rhel-labeler
on: 
  pull_request:
    branches:
      - RHEL-6
      - RHEL-7
      - RHEL-8

jobs:
  label:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/labeler@v2
      with:
        repo-token: "${{ secrets.GITHUB_TOKEN }}"
        configuration-path: .github/rhel-label-triggers.yml

johannbg:
Note that you can merge this since I will be doing a followed up pull request to test if labelling works ( github actions can be restrictive on forked repo's which may or may not affect this ) in which I can clean up whatever is required in that follow up PR. 

In the second PR I was also planning to use a different plugin which re-uses the same files but uses cron job instead of pull requests and might actually label all existing pull requests when it runs thus saving us ( or me ) from having to go through ca 590 pull request that have not been labelled yet. ( work that would otherwise probably take about a week on and off. I spent couple of hours yesterday and I only went through ca 70 pull requests )

danimo:
That's ok for me.

PR #894, opened 2020-08-12
40network: Keep nameservers from all interfaces

travier:
Make sure that all nameservers configured by each interface are kept in
/etc/resolv.conf. The current code will only keep the nameservers
configured for the latest interface.

This matters especially in the case where the latest interface is
configured via DHCP and if the DHCP server does not send a nameserver.
The current code will remove all previously configured nameservers from
/etc/resolv.conf leaving the system unable to resolve hostnames.

### DIFF
--- a/modules.d/40network/net-lib.sh
+++ b/modules.d/40network/net-lib.sh
@@ -122,7 +122,12 @@ setup_net() {
     [ -e /tmp/dhclient.$netif.dhcpopts ] && . /tmp/dhclient.$netif.dhcpopts
     # set up resolv.conf
     [ -e /tmp/net.$netif.resolv.conf ] && \
-        awk '!array[$0]++' /tmp/net.$netif.resolv.conf > /etc/resolv.conf
+        # save existing entries
+        touch /etc/resolv.conf
+        cp /etc/resolv.conf /tmp/net.$netif.resolv.conf.orig
+        # prepend new DNS servers while making sure that each appear only once
+        cat /tmp/net.$netif.resolv.conf /tmp/net.$netif.resolv.conf.orig \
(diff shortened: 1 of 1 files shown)


travier:
Test failures do no appear to be related to this PR as far as I understand.

danimo:
While I see the general benefit of this, this will break with split-brain DNS.

Assume the following:

1. eth0: static, eth1: dhcp
2. eth0 comes up, ships nameserver entry with "external" DHCP view -> gets inserted first
3. eth1 comes up, ships nameserver entry with "internal" DHCP view" -> gets inserted last

now when nfs/iscsi/whatever try to resolve an internal domain name, it will by default perform the lookup up via first entry, and will fail.

I think neither strategy is sufficient here.

//cc @tblume

travier:
This is a valid point. It could potentially be resolved by having a resolver that query all DNS servers such as systemd-resolved.

travier:
I'm afraid that completely fixing this case would require changes that are not clearly defined and partially already done when using NetworkManager in the initramfs:
  - https://fedoraproject.org/wiki/Changes/NetworkManager_1.16
  - https://developer.gnome.org/NetworkManager/stable/nm-initrd-generator.html
  - https://fedoraproject.org/wiki/Changes/systemd-resolved (upcoming in Fedora 33 and it looks like inclusion in the initramfs has not been discussed as far as I know)

I can make a change that will prepend the new servers instead of append like this change does.
This will:
  - keep all nameservers received for each interface thus making the case where we receive only one nameserver work
  - keep the current ordering behavior (last goes first) to avoid breaking the setups that work now.

WDYT?

travier:
Updated to make the change described above.

travier:
This is waiting for upstream review / decision.

PR #895, opened 2020-08-14
WIP: Adding networkd resolved timesyncd hostnamed systemd services as modules.

johannbg:
Decided to split that networkd early? work into separated modules.

### DIFF
--- a/modules.d/03systemd-hostnamed/module-setup.sh
+++ b/modules.d/03systemd-hostnamed/module-setup.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    [[ $mount_needs ]] && return 1
+
+    if ! dracut_module_included "systemd"; then
+        derror "systemd-hostnamed needs systemd in the initramfs"
+        return 1
+    fi
+
+    return 0
+}
+
+# called by dracut
+depends() {
+    echo "systemd systemd-networkd"
+}
+
+installkernel() {
+    return 0
+}
+
+# called by dracut
+install() {
+    local _mods
+
+    inst_multiple -o \
+       $systemdutildir/systemd-hostnamed \
+       $systemdsystemunitdir/systemd-hostnamed.service \
+       hostname hostnamectl
+
+    if [[ $hostonly ]]; then
+       inst_multiple -H -o \
+           /etc/hosts \
+           /etc/hostname \
+    	   /etc/machine-info \
+           ${NULL}
+    fi
+
+    _arch=${DRACUT_ARCH:-$(uname -m)}
+    inst_libdir_file {"tls/$_arch/",tls/,"$_arch/",}"libnss_myhostname.so.*"
+
+    systemctl -q --root "$initdir" enable systemd-hostnamed
+
+}
--- a/modules.d/02systemd-networkd/module-setup.sh
+++ b/modules.d/02systemd-networkd/module-setup.sh
@@ -9,7 +9,7 @@ check() {
         return 1
     fi
 
-    return 255
+    return 0
 }
 
 # called by dracut
@@ -23,6 +23,8 @@ installkernel() {
 
 # called by dracut
 install() {
+    local _mods
+
     inst_multiple -o \
         $systemdutildir/systemd-networkd \
         $systemdutildir/systemd-networkd-wait-online \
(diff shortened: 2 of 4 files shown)


johannbg:
@haraldh great if you could have a look at this and maybe give me some direction since these are my first modules.

LaserEyess:
Might be out of scope for this PR but would you be interested in making networkd a proper network module like `35network-legacy` and `35network-manager`? I am attempting something like that in a private fork https://github.com/LaserEyess/dracut/pull/1 but I don't have a solid grasp on everything needed to be a full network module in dracut, nor a good environment to dump debugging/build artifacts from testing.

johannbg:
> Might be out of scope for this PR but would you be interested in making networkd a proper network module like `35network-legacy` and `35network-manager`? I am attempting something like that in a private fork [LaserEyess#1](https://github.com/LaserEyess/dracut/pull/1) but I don't have a solid grasp on everything needed to be a full network module in dracut, nor a good environment to dump debugging/build artifacts from testing.

I'll probably end up going through most of the systemd integration compare it with upstream systemd and make modules an better integrated in dracut however systemd components can and do have dependency on other systemd components so things can and will break if you would omit some of those modules ( as Lukáš has pointed out ). 
Once I'm done with that I'll probably proceed on making IWD module for wifi support and or general networking as well and maybe look at bluetooth.

Also note that dont use anything in this pull request in your fork since it's outdated ( I've gone through several iterations on those modules ) and I'll end up submitting those modules individually for easier reviewing.

LaserEyess:
FWIW I don't actually think these dependencies are as hard as is assumed. For example, you *can* have networkd working *without* resolved, timesyncd, and hostnamed. However using features like [this](https://www.freedesktop.org/software/systemd/man/systemd.network.html#DNS=) and [this](https://www.freedesktop.org/software/systemd/man/systemd.network.html#NTP=) (among others if you look at the manpage for `systemd.network`) won't work properly. I currently use networkd by itself for my initrd internet and I don't have resolved/timesyncd.

modules.d/03systemd-hostnamed/module-setup.sh
```diff
@@ -0,0 +1,46 @@
...
+        derror "systemd-hostnamed needs systemd in the initramfs"
+        return 1
+    fi
+
+    return 0
+}
+
+# called by dracut
+depends() {
+    echo "systemd systemd-networkd"
```
bengal, line 17: I don't think this module should depend on systemd-networkd. I did a quick try and it seems hostnamed works fine without networkd.

NetworkManager, when running in the initramfs, currently sets the hostname directly in the kernel. However we are planning to run NM as a systemd service with D-Bus support; in that case it would preferably set the hostname via D-Bus through hostnamed.

In my opinion it should be possible to make other services depend on hostnamed without pulling networkd into the initrd.

modules.d/03systemd-hostnamed/module-setup.sh
```diff
@@ -0,0 +1,46 @@
...
+       hostname hostnamectl
+
+    if [[ $hostonly ]]; then
+       inst_multiple -H -o \
+           /etc/hosts \
+           /etc/hostname \
+    	   /etc/machine-info \
+           ${NULL}
+    fi
+
```
bengal, line 40: Shouldn't we also install:

```
/usr/share/dbus-1/system.d/org.freedesktop.hostname1.conf
/usr/share/dbus-1/system-services/org.freedesktop.hostname1.service
/usr/lib/systemd/system/dbus-org.freedesktop.hostname1.service
```

?

modules.d/03systemd-hostnamed/module-setup.sh
```diff
@@ -0,0 +1,46 @@
...
+    _arch=${DRACUT_ARCH:-$(uname -m)}
+    inst_libdir_file {"tls/$_arch/",tls/,"$_arch/",}"libnss_myhostname.so.*"
+
+    systemctl -q --root "$initdir" enable systemd-hostnamed
+
+}
```
bengal, line 46: Please also add the module in the `network` section of `dracut.spec`.

modules.d/03systemd-hostnamed/module-setup.sh
johannbg: @bengal Not needed afaik

modules.d/03systemd-hostnamed/module-setup.sh
johannbg: @bengal it's up to the package maintainers to add or remove any relevant modules for their distro and this work is far from completed. On top of that the dracut spec file should arguably be split into three separated files ( Fedora/RHEL/SUSE )

modules.d/03systemd-hostnamed/module-setup.sh
bengal: Oh, right; they are only needed if there is D-Bus support in the initrd (which is not there yet, but is being added in #900).

modules.d/03systemd-hostnamed/module-setup.sh
bengal: > @bengal it's up to the package maintainers to add or remove any relevant modules for their distro and this work is far from completed. 

Fair enough.

> On top of that the dracut spec file should arguably be split into three separated files ( Fedora/RHEL/SUSE )

As far as I understand, the spec file already supports those 3 distros. A single file with conditionals (as it is now) seems better than having to maintain 3 files with a lot of duplication.

johannbg: > As far as I understand, the spec file already supports those 3 distros. A single file with conditionals (as it is now) seems better than having to maintain 3 files with a lot of duplication.

The only pros you have by maintaining multiple distribution in a single rpm spec file is if it's being built by the same buildsystem and used by the same test system and test cases which may or may not be the case here since I have heard RH sales person claiming that RH build and maintained majority of Suse packages ( which was more likely to have been a sales pitch to get company to chose RHEL over Suse enterprise ) but since I dont work for either of those company's I cant say if that's really the case or not.

It's unmaintainable for upstream in general trying to support all the what ca 50 packaging solution in the fragmented downstream linux distribution ecosystem which is why I personally feel that dracut should just support/ship packaging example ( and tests ) for the primary's as in Arch,Debian,Fedora,Gentoo and Opensuse unless downstream provides resources to support/maintain those files upstream

On top of that if I was a downstream package maintainer I would ensure that each of the dracut modules would be sub package of the component they represent since it's the only way to keep the package dependency chain intact and installs small which is not the case in the dracut spec file.

LaserEyess:
What's the status of this now that there's dbus support in dracut?

johannbg:
.

johannbg:
.

PR #896, opened 2020-08-19
Add 90overlay-root module on request of Debian Maintainer

danimo:
Putting this here for discussion.

### DIFF
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -155,6 +155,10 @@ Misc
     started. If you have hardware, which takes a very long time to announce its
     drives, you might want to extend this value.
 
+**rd.root.overlay**::
+    make a read-only rootfs writeable by adding a tmpfs on top via overlayfs.
+    This is useful for read-only NFS root file systems.
+
 **rd.timeout=**__<seconds>__::
     specify how long dracut should wait for devices to appear. The
     default is '0', which means 'forever'. Note that this timeout
--- a/modules.d/90overlay-root/module-setup.sh
+++ b/modules.d/90overlay-root/module-setup.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    return 255
+}
+
+# called by dracut
+depends() {
+    # We do not depend on any modules - just some root
+    return 0
+}
+
+# called by dracut
+installkernel() {
+    hostonly='' instmods overlay
+}
+
+# called by dracut
+install() {
+    inst_hook pre-pivot 10 "$moddir/overlay-mount.sh"
+}
--- a/modules.d/90overlay-root/overlay-mount.sh
+++ b/modules.d/90overlay-root/overlay-mount.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+# make a read-only nfsroot writeable by using overlayfs
+# the nfsroot is already mounted to $NEWROOT
+# add the parameter rd.root.overlay to the kernel to activate this feature
+
+. /lib/dracut-lib.sh
+
+if ! getargbool 0 rd.root.overlay -d rootovl; then
+    return
+fi
+
+upperdir=/run/overlay-root/upper
+lowerdir=/run/overlay-root/lower
+workdir=/run/overlay-root/work
+
+mkdir -p "$upperdir" "$lowerdir" "$workdir"
+
+# Move root
+mount --move "$NEWROOT" "$lowerdir"
+
+# Create tmpfs
+mount -t tmpfs -o mode=0755 tmpfs "$upperdir"
+mount -t tmpfs -o mode=0755 tmpfs "$workdir"
+
+# Merge both to new Filesystem
+mount -t overlay -o "lowerdir=$lowerdir,upperdir=$upperdir,workdir=$workdir,redirect_dir=on,metacopy=on,default_permissions" overlay "$NEWROOT"
+
+# Let filesystems survive pivot
+mkdir -p "$NEWROOT$upperdir" "$NEWROOT$lowerdir" "$NEWROOT$workdir"
+mount --move "$upperdir" "$NEWROOT$upperdir"
+mount --move "$lowerdir" "$NEWROOT$lowerdir"
+mount --move "$workdir" "$NEWROOT$workdir"
--- a/modules.d/90overlay-root/README
+++ b/modules.d/90overlay-root/README
@@ -1,0 +1,14 @@
+dracut rootfs overlayfs module
+
+Make any rootfs ro, but writable via overlayfs.
+This is convenient, if for example using an ro-nfs-mount.
+
+Add the parameter "rd.root.overlay" to the kernel to activate this feature.
+
+This happens pre-pivot. Therefore the final root file system is already
+mounted. It will be set ro, and turned into an overlayfs mount with an
+underlying tmpfs.
+
+The original root and the tmpfs will be mounted at /live/image and
+/live/sysroot-overlay in the final rootfs.
+
--- a/pkgbuild/dracut.spec
+++ b/pkgbuild/dracut.spec
@@ -368,6 +368,7 @@ echo 'dracut_rescue_image="yes"' > $RPM_BUILD_ROOT%{dracutlibdir}/dracut.conf.d/
 %{dracutlibdir}/modules.d/90mdraid
 %{dracutlibdir}/modules.d/90multipath
 %{dracutlibdir}/modules.d/90nvdimm
+%{dracutlibdir}/modules.d/90overlay-root
 %{dracutlibdir}/modules.d/90ppcmac
 %{dracutlibdir}/modules.d/90qemu
 %{dracutlibdir}/modules.d/91crypt-gpg


modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+# make a read-only nfsroot writeable by using overlayfs
+# the nfsroot is already mounted to $NEWROOT
+# add the parameter rootovl to the kernel, to activate this feature
+
+. /lib/dracut-lib.sh
+
+if ! getargbool 0 rootovl ; then
```
danimo, line 9: This should probably become `rd.rootovl` (and keep `rootovl` as a compat parameter). @haraldh?

danimo:
//cc @Vogtinator This may also be of interest to you

modules.d/90kernel-modules-pinctrl/module-setup.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
Vogtinator, line 1: Unrelated?

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,39 @@
...
+    return
+fi
+
+modprobe overlay
+
+# a little bit tuning
+mount -o remount,nolock,noatime $NEWROOT
+
+# Move root
```
Vogtinator, line 18: `mount --make-private /` would fix this. Though I don't know whether that breaks something else.

Vogtinator: If making `/` private is not an option, please move the pseudo-move into a function

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+# make a read-only nfsroot writeable by using overlayfs
+# the nfsroot is already mounted to $NEWROOT
+# add the parameter rootovl to the kernel, to activate this feature
+
+. /lib/dracut-lib.sh
+
```
Vogtinator, line 8: I recommend `set -eu`

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,39 @@
...
+# Move root
+# --move does not always work. Google >mount move "wrong fs"< for
+#     details
+mkdir -p /live/image
+mount --bind $NEWROOT /live/image
+umount $NEWROOT
+
+# Create tmpfs
+mkdir /cow
+mount -n -t tmpfs -o mode=0755 tmpfs /cow
```
Vogtinator, line 27: Why `-n` here?

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,39 @@
...
+mkdir /cow/work /cow/rw
+
+# Merge both to new Filesystem
+mount -t overlay -o noatime,lowerdir=/live/image,upperdir=/cow/rw,workdir=/cow/work,default_permissions overlay $NEWROOT
```
Vogtinator, line 31: Why `noatime`?

Vogtinator (changes requested):

Vogtinator:
I've used something like this for years several years ago, but could not use overlayfs on NFSv4 due to ACLs in xattrs. Is that fixed meanwhile or does this only work with either NFSv3 or NFSv4 without ACLs?

modules.d/90kernel-modules-pinctrl/module-setup.sh
danimo: Indeed. my fault.

modules.d/90overlay-root/module-setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+check() {
+    # do not add modules if the kernel does not have overlayfs support
+    [ -d $dracutsysrootdir/lib/modules/$kernel/kernel/fs/overlayfs ] || return 1
```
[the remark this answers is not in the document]
danimo: I was also wondering about this. What's a good way to check here? Purely checking for functionality might not work, because it might not be loaded.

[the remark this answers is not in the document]
Vogtinator: Just the `instmods` in `installkernel` should be enough. It's impossible to guarantee 100% that if `check` succeeds that it actually works anyway.

modules.d/90overlay-root/module-setup.sh
```diff
@@ -0,0 +1,20 @@
...
+}
+
+depends() {
+    # We do not depend on any modules - just some root
+    return 0
+}
+
+# called by dracut
+installkernel() {
+    instmods overlay
```
Vogtinator, line 15: Might need `hostonly=''`

modules.d/90overlay-root/module-setup.sh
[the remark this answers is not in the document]
danimo: Good point.

modules.d/90overlay-root/module-setup.sh
Mrfai: You  are right, overlayfs and NFSv4 is a little bit tricky.

As of 2019 overlayfs  can only work with NFS v4 if acl(5) support is disabled on the filessystem on the NFS server side. This can be done by remounting the nfsroot on the NFS server with noacl. I'm not sure if newer kernel version still need this.

danimo: Included. Keeping open because of the unrelated comment.

modules.d/90overlay-root/module-setup.sh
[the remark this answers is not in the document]
danimo: Removed.

modules.d/90overlay-root/overlay-mount.sh
Mrfai: IIRC, at  this boot time /etc/mtab is not writeable.

Mrfai: I my usecase, I always mount a read-only nfsroot.

modules.d/90overlay-root/overlay-mount.sh
Mrfai: noatime is just a small optimisation. You are remove this.

modules.d/90overlay-root/overlay-mount.sh
Vogtinator: AFAIK / is the initrd here, so tmpfs. It's post likely linked to `/proc/self/mounts` anyway and I'd expect that the `mount --bind` calls also write to mtab.

Mrfai: I'll try to check if it works without -n.

Mrfai: -n can be removed, also noatime is not needed

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,39 @@
...
+# the nfsroot is already mounted to $NEWROOT
+# add the parameter rootovl to the kernel, to activate this feature
+
+. /lib/dracut-lib.sh
+
+if ! getargbool 0 rootovl ; then
+    return
+fi
+
+modprobe overlay
```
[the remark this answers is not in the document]
Mrfai: modprobe is not needed. I just did a test.

modules.d/90overlay-root/overlay-mount.sh
Vogtinator: Unfortunately, yes... IMO it's still better to bail out early if something fails than to continue like nothing bad happened. In the case where the mount fails, it would actually exit 1 (by chance?) because the last `mount --move` would fail and such set the script's return value. The errors until that point would be completely ignored though.

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,29 @@
...
+mkdir -p /live/image
+mount --move $NEWROOT /live/image
+
+# Create tmpfs
+mkdir /cow
+mount -t tmpfs -o mode=0755 tmpfs /cow
+mkdir /cow/work /cow/rw
+
+# Merge both to new Filesystem
+mount -t overlay -o lowerdir=/live/image,upperdir=/cow/rw,workdir=/cow/work,default_permissions overlay $NEWROOT
```
Vogtinator, line 23: It might be a performance (and memory) improvement to use `redirect_dir=on,metacopy=on` here to avoid copying of data to the tmpfs. Those options are used from the kernel configuration if unspecified, so depending on the setup that might be the default.

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Move root
+# --move does not always work. Google >mount move "wrong fs"< for
+#     details
+mkdir -p /live/image
+mount --bind $NEWROOT /live/image
+umount $NEWROOT
+
+# Create tmpfs
+mkdir /cow
```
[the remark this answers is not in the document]
Vogtinator: `/sysroot-overlay` maybe?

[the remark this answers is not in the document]
Mrfai: Good idea. But I leave this to the RH people what to choose.

modules.d/90overlay-root/overlay-mount.sh
Mrfai: First test using set -eu:
lib/dracut-lib.sh: line 61: DRACUT_SYSTEMD unbound variable
_newoption: unbound variable
DRACUT_SYSTEMD: unbound variable
then kernel panic
 :-(

Mrfai:
> Great, I will also look at the README. -- regards Thomas

I forgot what we wanted to change in README. Any hints?

modules.d/90overlay-root/overlay-mount.sh
Vogtinator: It may have to be done after sourcing `lib/dracut-lib.sh`...

Vogtinator:
> > Great, I will also look at the README. -- regards Thomas
> 
> I forgot what we wanted to change in README. Any hints?

No idea, but I noticed that it doesn't mention that changes end up in tmpfs and are volatile.

haraldh (changes requested):
Needs some documentation in `dracut.cmdline.7.asc` about the command line parameter mentioning the specific dracut module

haraldh:
nobody?

Mrfai:
This is my proposal for the dracut.cmdline.7.asc documentation

** rootovl**::
    Make a read-only rootfs writeable by adding a tmpfs on top via overlayfs.
    This is useful for read-only NFS root file systems.

Mrfai:
I'm not sure about the workflow to update this PR.

haraldh:
ah, it would require @danimo to make those changes, if @Mrfai file a new PR

danimo:
Updated pull-request with all the feedback, shy of setting `+eu`. Graduate to full PR.

danimo:
> Is the newline after mount -o in [f97baf7](https://github.com/dracutdevs/dracut/commit/f97baf7aea4074c10e00e8e62c199f7a70237d25) intentional?

Nope, that was a mistake. Fix.

danimo:
> Needs some documentation in `dracut.cmdline.7.asc` about the command line parameter mentioning the specific dracut module

Can you switch to approval for this PR?

Mrfai:
A minor change for README
/live/cow -> /live/sysroot-overlay

danimo:
@haraldh @johannbg ready for merge

modules.d/90overlay-root/module-setup.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+depends() {
+    # We do not depend on any modules - just some root
+    return 0
```
haraldh, line 5: Please make this `return 255`, so it is only included, if requested.

haraldh (changes requested):

modules.d/90overlay-root/module-setup.sh
Mrfai: I cannot find any other dracut module that returns 255 here, so I don't understand why this module should do it.

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,29 @@
...
+mkdir /sysroot-overlay/work /sysroot-overlay/rw
+
+# Merge both to new Filesystem
+mount -t overlay -o lowerdir=/live/image,upperdir=/sysroot-overlay/rw,workdir=/sysroot-overlay/work,redirect_dir=on,metacopy=on,default_permissions overlay "$NEWROOT"
+
+# Let filesystems survive pivot
+mkdir -p "$NEWROOT/live/sysroot-overlay"
+mkdir -p "$NEWROOT/live/image"
+mount --move /sysroot-overlay/rw "$NEWROOT/live/sysroot-overlay"
+mount --move /live/image "$NEWROOT/live/image"
```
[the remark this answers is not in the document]
Mrfai: Good idea. @danimo can you fix this?

modules.d/90overlay-root/module-setup.sh
haraldh: Oops.. my error... I meant in `check()`

modules.d/90overlay-root/module-setup.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+depends() {
```
haraldh, line 3: This should have:

```
check() {
    return 255
}
```

haraldh (changes requested):

modules.d/90overlay-root/module-setup.sh
haraldh: @danimo @Mrfai ... sorry, not in `depends()` but in `check()`

modules.d/90overlay-root/overlay-mount.sh
```diff
@@ -0,0 +1,33 @@
...
+mount -t tmpfs -o mode=0755 tmpfs "$workdir"
+
+# Merge both to new Filesystem
+mount -t overlay -o "lowerdir=$lowerdir,upperdir=$upperdir,workdir=$workdir,redirect_dir=on,metacopy=on,default_permissions" overlay "$NEWROOT"
+
+# Let filesystems survive pivot
+mkdir -p "$NEWROOT$upperdir" "$NEWROOT$lowerdir" "$NEWROOT$workdir"
+mount --move "$upperdir" "$NEWROOT$upperdir"
+mount --move "$lowerdir" "$NEWROOT$lowerdir"
+mount --move "$workdir" "$NEWROOT$workdir"
```
haraldh, line 33: This does not work and should be removed.

haraldh (changes requested):

modules.d/90overlay-root/overlay-mount.sh
haraldh, line 33: This does not work and should be removed, because it is unnecessary with `/run`

haraldh (changes requested):

modules.d/90overlay-root/overlay-mount.sh
haraldh: `/run` will be moved by switch_root to `$NEWROOT/run` anyway

johannbg (approved):

man/dracut.cmdline.7.asc
```diff
@@ -155,6 +155,10 @@ Misc
     started. If you have hardware, which takes a very long time to announce its
     drives, you might want to extend this value.
 
+**rd.root.overlay**::
```
LaszloGombos, line 158: It would be useful to provide guidance how this option is different from systemd.volatile=overlay and what are the tradeoffs ? One difference I noticed is that systemd.volatile=overlay would not leave the mounted filesystem visible. Is that the only difference ?

Mrfai: This module runs without systemd. It is used by the FAI project for several years. It seems that systemd.volatile only makes /etc and /var writeable, rd.root.overlay makes the whole / writeable.

LaszloGombos: Thanks for the clarification. I have not realized that systemd is an optional dependency for dracut.

PR #897, opened 2020-08-19
Remove stratis module

mulkieran:


mulkieran:
It does not look to me like failures are related to change so far.

PR #900, opened 2020-08-21
Support for wicked network manager

danimo:
This introduces support for [wicked](https://github.com/openSUSE/wicked) as network management module.

Todo:

* [x] compare dbus module with the one in the [bluetooth HiD PR](https://github.com/dracutdevs/dracut/pull/544).
* [ ] add tests if possible (wicked is probably not available on fedora)
* [ ] check if all parameters are properly accepted
* [ ] add documentation

### DIFF
--- a/modules.d/06dbus/module-setup.sh
+++ b/modules.d/06dbus/module-setup.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+  require_binaries dbus-daemon || return 1
+
+  return 255
+}
+
+depends() {
+  echo systemd
+  return 0
+}
+
+install() {
+  local DBUS_SERVICE=/usr/lib/systemd/system/dbus.service
+  if [[ -e $DBUS_SERVICE ]]; then
+    if [[ -L $DBUS_SERVICE ]]; then
+      DBUS_SERVICE=$(readlink $DBUS_SERVICE)
+    fi
+  else
+    DBUS_SERVICE=/etc/systemd/system/dbus.service
+    if [[ -e $DBUS_SERVICE ]]; then
+      if [[ -L $DBUS_SERVICE ]]; then
+        DBUS_SERVICE=$(readlink $DBUS_SERVICE)
+      fi
+    else
+      echo "Could not find dbus.service";
+      exit 1
+    fi
+  fi
+
+  inst_multiple \
+    $DBUS_SERVICE \
+    /usr/lib/systemd/system/dbus.socket \
+    /usr/bin/dbus-daemon \
+    /usr/bin/dbus-send
+
+  dbus_system_services="
+        org.freedesktop.systemd1
+        org.freedesktop.timedate1
+        org.freedesktop.hostname1
+  "
+  inst_dir      /etc/dbus-1/system.d
+  inst_dir      /usr/share/dbus-1/services
+  inst_dir      /usr/share/dbus-1/system-services
+  inst_multiple /etc/dbus-1/system.conf
+  inst_multiple /usr/share/dbus-1/system.conf \
+                /usr/share/dbus-1/services/org.freedesktop.systemd1.service
+  for service in $dbus_system_services ; do
+      inst_multiple        /etc/dbus-1/system.d/${service}.conf \
+              /usr/share/dbus-1/system-services/${service}.service
+  done
+  inst_multiple $(find /var/lib/dbus)
+
+  inst_hook cleanup 99 "$moddir/dbus-cleanup.sh"
+
+  grep '^messagebus:' /etc/passwd >> "$initdir/etc/passwd"
+  grep '^messagebus:' /etc/group >> "$initdir/etc/group"
+
+  # do not enable -- this is a static service
+  #systemctl --root "$initdir" enable $DBUS_SERVICE > /dev/null 2>&1
+
+  sed -i -e \
+'/^\[Unit\]/aDefaultDependencies=no\
+Conflicts=shutdown.target\
+Before=shutdown.target' \
+    "$initdir"$DBUS_SERVICE
+
+  sed -i -e \
+'/^\[Unit\]/aDefaultDependencies=no\
+Conflicts=shutdown.target\
+Before=shutdown.target
+/^\[Socket\]/aRemoveOnStop=yes' \
+    "$initdir"/usr/lib/systemd/system/dbus.socket
+
+  #We need to make sure that systemd-tmpfiles-setup.service->dbus.socket will not wait local-fs.target to start,
+  #If swap is encrypted, this would make dbus wait the timeout for the swap before loading. This could delay sysinit
+  #services that are dependent on dbus.service.
+  sed -i -Ee \
+    '/^After/s/(After[[:space:]]*=.*)(local-fs.target[[:space:]]*)(.*)/\1-\.mount \3/' \
+    "$initdir"/usr/lib/systemd/system/systemd-tmpfiles-setup.service
+}
--- a/modules.d/06dbus/module-setup.sh
+++ b/modules.d/06dbus/module-setup.sh
@@ -36,21 +36,12 @@ install() {
     /usr/bin/dbus-daemon \
     /usr/bin/dbus-send
 
-  dbus_system_services="
-        org.freedesktop.systemd1
-        org.freedesktop.timedate1
-        org.freedesktop.hostname1
-  "
   inst_dir      /etc/dbus-1/system.d
   inst_dir      /usr/share/dbus-1/services
   inst_dir      /usr/share/dbus-1/system-services
   inst_multiple /etc/dbus-1/system.conf
   inst_multiple /usr/share/dbus-1/system.conf \
                 /usr/share/dbus-1/services/org.freedesktop.systemd1.service
-  for service in $dbus_system_services ; do
-      inst_multiple        /etc/dbus-1/system.d/${service}.conf \
-              /usr/share/dbus-1/system-services/${service}.service
-  done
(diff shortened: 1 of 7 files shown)


rtorrero:
#### About your first point:
```
$ find /var/lib/dbus
/var/lib/dbus
/var/lib/dbus/machine-id
``` 
Maybe we could just add them explicitely, since it's just two files?

#### About the second point:

Do you mean the systemd service files for the wickedd parts? it's less messy to start it like this vs explicit calls since there are dependencies between the multiple wicked components and also due to the dependency with dbus. We tried multiple combinations and we had trouble to get get wicked started only when dbus was ready to be used. This way it seemed it worked fine.

danimo:
> ```
> $ find /var/lib/dbus
> /var/lib/dbus
> /var/lib/dbus/machine-id
> ```
> 
> Maybe we could just add them explicitely, since it's just two files?

Yes. If there is no deeper reason behind that, we should go for it. Also, we need to ensure the module either becomes host only safe.

> #### About the second point:
> 
> Do you mean the systemd service files for the wickedd parts? it's less messy to start it like this vs explicit calls since there are dependencies between the multiple wicked components and also due to the dependency with dbus. We tried multiple combinations and we had trouble to get get wicked started only when dbus was ready to be used. This way it seemed it worked fine.

I was referring to this:

```
-  dbus_system_services="
-        org.freedesktop.systemd1
-        org.freedesktop.timedate1
-        org.freedesktop.hostname1
-  "
```

danimo:
@rtorrero ping?

rtorrero:
> @rtorrero ping?

I'm currently recovering the setup I used to test this. I think I remember it had something to do with the resulting lease file in case of DHCP (that will be used after the initrd stage too).

modules.d/06dbus/module-setup.sh
```diff
@@ -0,0 +1,83 @@
...
+    else
+      echo "Could not find dbus.service";
+      exit 1
+    fi
+  fi
+
+  inst_multiple \
+    $DBUS_SERVICE \
+    /usr/lib/systemd/system/dbus.socket \
+    /usr/bin/dbus-daemon \
```
bengal, line 36: This doesn't work on Fedora, where `/etc/systemd/system/dbus.service` is a link to `/usr/lib/systemd/system/dbus-broker.service` and the service requires binaries `/usr/bin/dbus-broker` and
`/usr/bin/dbus-broker-launch`.

Either install the `dbus-daemon.service` and the `dbus-daemon` binary, or install `dbus.service` and the matching binaries. I would prefer the latter.

modules.d/06dbus/module-setup.sh
```diff
@@ -0,0 +1,83 @@
...
+  for service in $dbus_system_services ; do
+      inst_multiple        /etc/dbus-1/system.d/${service}.conf \
+              /usr/share/dbus-1/system-services/${service}.service
+  done
+  inst_multiple $(find /var/lib/dbus)
+
+  inst_hook cleanup 99 "$moddir/dbus-cleanup.sh"
+
+  grep '^messagebus:' /etc/passwd >> "$initdir/etc/passwd"
+  grep '^messagebus:' /etc/group >> "$initdir/etc/group"
```
bengal, line 59: On Fedora the service is run as `dbus` user instead of `messagebus`.

modules.d/06dbus/module-setup.sh
```diff
@@ -0,0 +1,83 @@
...
+        org.freedesktop.hostname1
+  "
+  inst_dir      /etc/dbus-1/system.d
+  inst_dir      /usr/share/dbus-1/services
+  inst_dir      /usr/share/dbus-1/system-services
+  inst_multiple /etc/dbus-1/system.conf
+  inst_multiple /usr/share/dbus-1/system.conf \
+                /usr/share/dbus-1/services/org.freedesktop.systemd1.service
+  for service in $dbus_system_services ; do
+      inst_multiple        /etc/dbus-1/system.d/${service}.conf \
```
bengal, line 51: These are in `/usr/share/dbus-1/system.d` on Fedora.

modules.d/06dbus/module-setup.sh
```diff
@@ -0,0 +1,83 @@
...
+      echo "Could not find dbus.service";
+      exit 1
+    fi
+  fi
+
+  inst_multiple \
+    $DBUS_SERVICE \
+    /usr/lib/systemd/system/dbus.socket \
+    /usr/bin/dbus-daemon \
+    /usr/bin/dbus-send
```
bengal, line 37: Maybe install `busctl` instead? Nowadays I find it more useful.

modules.d/06dbus/module-setup.sh
```diff
@@ -0,0 +1,83 @@
...
+  grep '^messagebus:' /etc/group >> "$initdir/etc/group"
+
+  # do not enable -- this is a static service
+  #systemctl --root "$initdir" enable $DBUS_SERVICE > /dev/null 2>&1
+
+  sed -i -e \
+'/^\[Unit\]/aDefaultDependencies=no\
+Conflicts=shutdown.target\
+Before=shutdown.target' \
+    "$initdir"$DBUS_SERVICE
```
bengal, line 68: I wonder if such replacements could be made more readable by using drop-in snippets:
```
    mkdir -p "${initdir}/$systemdsystemunitdir/dbus.socket.d"
    (
        echo "[Unit]"
	echo "DefaultDependencies=no"
        ...
    ) > "${initdir}/$systemdsystemunitdir/dbus.socket.d/dracut.conf"

File: modules.d/06dbus/module-setup.sh (hunk 2 above)
Comment (rtorrero) [reply modules.d/06dbus/module-setup.sh:59]: @danimo I think that the las PR that I send you was missing some of the changes. This was addressed in: https://github.com/rtorrero/dracut/commit/cc78beb4045ed902c47ad910908775264fb5f414

Could you add those changes here too?

File: modules.d/06dbus/module-setup.sh (hunk 3 above)
Comment (rtorrero) [reply modules.d/06dbus/module-setup.sh:51]: I noticed that in our latest Tumbleweed snapshot the location has also changed to be /usr/share/dbus-1/system.d, we probably need to use `inst_any` here?

**Comment by rtorrero** [comment]
> > @rtorrero ping?
> 
> I'm currently recovering the setup I used to test this. I think I remember it had something to do with the resulting lease file in case of DHCP (that will be used after the initrd stage too).

After doing some tests, I can confirm that
``` dbus_system_services="
       org.freedesktop.systemd1
       org.freedesktop.timedate1
       org.freedesktop.hostname1
 "```
is not really needed. It made no difference having them there or not.

File: modules.d/35network-wicked/module-setup.sh [hunk 6]
```diff
@@ -0,0 +1,61 @@
...
+    inst_dir /etc/wicked/extensions
+    inst_dir /usr/share/wicked/schema
+    inst_dir /usr/lib/wicked/bin
+    inst_dir /var/lib/wicked
+
+    inst_multiple /etc/wicked/*.xml
+    inst_multiple /etc/wicked/extensions/*
+    inst_multiple /etc/dbus-1/system.d/org.opensuse.Network*
+    inst_multiple /usr/share/wicked/schema/*
+    inst_multiple /usr/lib/wicked/bin/*
```
rtorrero, line 40: I've noticed that due to a macro change in our build service, now the wickedd-* files could land either on `usr/lib/wicked/bin/` or on `/usr/libexec/wicked/bin/` depending on when the RPM was build. 
Should we change it for:
   ```
    inst_any /usr/lib/wicked/bin/wickedd-auto4 /usr/libexec/wicked/bin/wickedd-auto4
    inst_any /usr/lib/wicked/bin/wickedd-dhcp4 /usr/libexec/wicked/bin/wickedd-dhcp4
    inst_any /usr/lib/wicked/bin/wickedd-dhcp6 /usr/libexec/wicked/bin/wickedd-dhcp6
```
?

modules.d/06dbus/module-setup.sh
bengal: Yes, I guess so.

bengal:
Hi, any progress on this?

I'm reworking the NetworkManager dracut module to run as a systemd
service with D-Bus support and I'm looking forward to see the dbus
module in this pull request merged.

Maybe the dbus module could be a separate PR, so that it can be
reviewed and merged independently from wicked support?

modules.d/06dbus/module-setup.sh
danimo: Fixed

modules.d/06dbus/module-setup.sh
danimo: Done

modules.d/06dbus/module-setup.sh
danimo: I removed them. They do not seem to be required in the first place. If we find out we do, we can still re-add them with the new paths.

modules.d/06dbus/module-setup.sh
danimo: I agree it's not pretty, but I would like to perform the clean up in a second step. Removing all default dependencies feels a bit of a crass measure.

modules.d/35network-wicked/module-setup.sh
danimo: I am now accounting for both contingencies.

modules.d/06dbus/module-setup.sh
danimo: I have made the required changes, but tests on fedora is missing atm.

danimo:
@bengal Sorry for the slow progress. I've done some testing, addressed (most of) your concerns and will be doing some more testing tomorrow, before merging this.

PR #905, opened 2020-08-26
Do not start inside container

tpgxyz:
Host should take care of good entropy pool
#791

This PR replaces https://github.com/dracutdevs/dracut/pull/814 as i lost that branch...


danimo (approved):

mikhailnov:
Is there any usecase when initrd can be started in container?

haraldh:
> Is there any usecase when initrd can be started in container?

Not that I know of.. That would basically be an initrd as a container runtime...

PR #906, opened 2020-08-26
Adding dbusdir as global var

johannbg:
Should simplify writing dbus related modules as in
inst_multiple -o \
$dbusdir/org.freedesktop.foo.conf
$systemdutildir/system/org.freedesktop.foo.service
...


johannbg:
Brainfarted this idea into pull request and burned all the test cases in the process. need to rework-it and resubmit or maybe not anyway closing.

PR #909, opened 2020-08-28
45url-lib: improve ca-bundle detection

danimo:
The current detection routine for openssl-based libcurl assumes that
libcurl has its own hardcoded path to the ca-bundle. Fix the
cases where curl is compiled with
```
  --with-ca-fallback --without-ca-path --without-ca-bundle
```
In this case, we must also grep in OpenSSLs libcrypto.

Other changes:
  - filter reported but non-existant pathes
  - strip nul bytes returned by grep
  - consider that ca-bundles might use '.pem' instead of '.crt'

Downstream report: https://bugzilla.suse.com/show_bug.cgi?id=1175892

Reference: boo#1175892


qby-wenzel:
Hi, is there any progress on this topic? Maybe somewhere else than here? In the suse bugtracker it's also kind of stale.

We're using dracut/ignition to bootstrap our devices over PXE. Currently the call to the ignition service (https) is proxied through a local nginx to remove the need of a direct ssl connection, but a system without this workaround would be even nicer :)

johannbg (changes requested):
You forgot to add sed as a required binary to the check() section for the module. 
And while you are add it, it would be nice if you could add ctorrent,grep and echo since they are missing from there as well.

haraldh:
`ctorrent` is already there:
```
     inst_multiple -o ctorrent
```

johannbg:
> and it is optional.

Hmm I'm not seeing how anything that is used in modules or related helper script to a module can be considered optional unless ofcourse you want to include those optionals in some kind of dracut core/optionals module but at that point you are just introducing dependency on that module instead.

haraldh:
This is a shell function library. Modules, which really need `ctorrent_fetch_url()` should require it.
Otherwise it is 
```
command -v ctorrent >/dev/null \
    && add_url_handler ctorrent_fetch_url torrent
```

So, you could gracefully failover to another mechanism to get your data.

johannbg:
Regardless whether it is a library or not, there is something that has to check for ( in this case ) ctorrents existence and make sure it is included/installed into the initramfs for it to work and error out otherwise. 

And at the same time that "magic" has to be implemented in such manner that it is obvious to downstream where the thing that reside on their initramfs originally came from right ( and hard checks/dependency/failures are pretty obvious to everyone in most cases), so buried shebangs like this "command -v "FOO" >/dev/null", that forces downstream/administrator grep recursively through the source, just to figure out where that magic is, does not exactly spell out user friendly or captain obvious to people.

If we want to keep doing hack like these, we will just in the end being forced to implement something like "modinfo" for dracut modules and files that reside on the initramfs right.

haraldh:
so, to reiterate the historic contents:
* nothing needs ctorrent for now
* the url-lib library was "sponsored" by anaconda stage-1 and is shared with other parts
* if ctorrent is present, it can be used with the url-lib with anaconda

qby-wenzel:
This would also mean, that the [nfs part](https://github.com/dracutdevs/dracut/blob/master/modules.d/45url-lib/url-lib.sh#L157) is questionable? It is using the same method to check for the existance of the `nfs_to_var` function.
```shell
command -v nfs_to_var >/dev/null && add_url_handler nfs_fetch_url nfs nfs4
```
But to keep the action flowing, should this topic of handling those requirements better be based in another issue? The initial reason for this PR was the better _ca-bundle detection_. I guess with the current discussion, this could point to a segmentation (?) of this module to help going along with its requirements. Like it's done in the [network module](https://github.com/dracutdevs/dracut/blob/master/modules.d/40network/module-setup.sh#L12) and it's "submodules" (nm, wicked, legacy).

johannbg:
Yeah let's keep that separated. Sorry for the confusion

johannbg (approved):

qby-wenzel:
@danimo Hi :) Is there any reason -- you can share with us -- why this PR was switched back to draft again? I would randomly guess, that it did not fit into the last release / feature freeze cycle? 'cause v52 is just on the horizon. From my PoV this PR was quite ready ;) ...separate from maybe some checks like the PR name. Thanks in advance.

haraldh:
ping?

rorosen:
Hi, is there any progress on this issue or any information you can share why it stalls?

aafeijoo-suse:
@johannbg @haraldh how can we move this fix forward? Thank you in advance.

johannbg:
@aafeijoo-suse I cant recall why @danimo never finished this PR and I think it's safe to assume at this point that he no longer has this branch ( it has conflicts which need to be resolved ) so it's probably best that you go over his changes see if something is missing and file a new PR which we can go through.

PR #915, opened 2020-08-29
90crypt: make `rd.luks.key` usable with encrypted keydev.

the-antz:
Add optional prefix `keysource:` for the values of `rd.luks.partuuid`,
`rd.luks.serial` and `rd.luks.uuid`.
If specified, ask for passphrase instead of waiting for keydevs to come
online.

This makes it possible to use an encrypted *keydev* with `rd.luks.key`.

The `keysource:` keyword essentially marks an `rd.luks.partuuid` / `rd.luks.serial` / `rd.luks.uuid` argument as always-ask-for-passphrase. Without it, and with `rd.luks.key` specified, dracut would wait for an appropriate key entry to appear in `/tmp/luks.keys` (which won't happen) until timeout, resulting in the asking-for-passphrase fallback for all devices.

What I'm not yet sure about is if the priority of 31 for a generated `cleanup`-hook is appropriate (another cleanup hook of the module uses 30-*). It seems to work fine for me but, well, feedback welcome :).

## Changes

`90crypt`-module and related documentation.

## Checklist

### DIFF
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -302,6 +302,8 @@ crypto LUKS
     The comparisons also matches, if _<luks uuid>_ is only the beginning of the
     LUKS UUID, so you don't have to specify the full UUID.
     This parameter can be specified multiple times.
+    _<luks uuid>_ may be prefixed by the keyword `keysource:`, see
+    _rd.luks.key_ below.
 
 **rd.luks.allow-discards=**__<luks uuid>__::
     Allow  using  of discards (TRIM) requests for LUKS partitions with the given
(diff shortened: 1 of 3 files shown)


danimo:
//cc @mwilck for review

mwilck:
The man page still says "*If you do need rd.luks.key to work, you will have to exclude the "systemd" dracut module and any modules that depend on it.*" Does this still hold with this patch applied? If the whole functionality is only available without systemd, it won't be helpful for many users.

the-antz:
> The man page still says "_If you do need rd.luks.key to work, you will have to exclude the "systemd" dracut module and any modules that depend on it._" Does this still hold with this patch applied? If the whole functionality is only available without systemd, it won't be helpful for many users.

Unfortunately that still applies, the patch just modifies the existing *rd.luks.key* code a bit to allow the new use case.

On a different note:
I noticed that neither `rd.luks.partuuid` nor `rd.luks.serial` create a finished-hook which seems wrong (`rd.luks.uuid` does create one). The two options aren't mentioned on the wiki, too. Are these deprecated? If they are I should probably change the wiki-update to not reintroduce them. The PR should not affect any possible brokenness in regards to those options. If it's indeed a bug I could look into creating a separate PR for this.

mwilck (approved):
The patch looks good to me. I think that we do need a more general solution for the various possible ways to unlock encrypted disks in the future, but with the limitations stated (in particular, no systemd), this is fine.

the-antz:
Should I remove `rd.luks.partuuid` and `rd.luks.serial` from the new wiki part? In hindsight, it might not have been the best idea to add them as they aren't mentioned anywhere currently.

danimo:
@the-antz a damn, missed your comment. what do you mean wiki part?

the-antz:
I added to the docu that the `keysource` keyword can be used with `rd.luks.partuuid`, `rd.luks.serial` and `rd.luks.uuid`, which is correct but `rd.luks.partuuid` and `rd.luks.serial` itself are not explained in the documentation, only `rd.luks.uuid` is. In hindsight it seems a bit inconsistent to mention them when they're not itself documented.

I was wondering if I should change it only mention `keysource` as a prefix for `rd.luks.uuid`.

the-antz:
Nevermind, just reread the patch and noticed that I didn't mention the two keywords at all. Sry, seems my memory failed me on that one.

Thanks for the merge.

PR #916, opened 2020-08-31
Adding repo badges

johannbg:
This pull request adds repo badges which will give some oversight of downstream activity. 
Note the repo downstream list should just be kept strictly to the primary's ( Arch,Debian,Fedora,Gentoo,OpenSuse).
Any future patches that try to extend the list beyond that  ( My $DISTRO is missing stuff ) should arguably be rejected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,17 @@ dracut
 
 dracut is an event driven initramfs infrastructure.
 
-[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-v2.0%20adopted-ff69b4.svg)](.github/CODE_OF_CONDUCT.md) 
-[![Build Status](https://travis-ci.org/dracutdevs/dracut.svg?branch=master)](https://travis-ci.org/dracutdevs/dracut)
-[![Fedora-31](https://github.com/dracutdevs/dracut/workflows/Fedora-31/badge.svg?branch=master)](https://github.com/dracutdevs/dracut/actions?query=workflow%3AFedora-31)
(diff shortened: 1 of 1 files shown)


danimo (changes requested):
I would like to separate CI status from package availability. I would also move the Contributor Covenant into its own section where we explain it in a few sentences.

johannbg:
@danimo well I would like to keep the readme file short and simple in the long run since README files are not meant to be projects documentations so they should not be too long and hard to digest so they should contain a brief oversight/status of the project via labels + and something short like 

Documentation is available on dracut's web site ( or the github wiki since we dont have a domain and the kernel one is old and outdated and harder to maintain ).
Consult our NEWS file for information about what's new in the most recent dracut versions.
Consult our Contribution Guidelines for contributing ( pointing to ./github/CONTRIBUTE.md ).
Consult our Community CoC ( pointing to .github/CODE_OF_CONDUCT.md ).
Consult our Security policy ( pointing to .github/SECURITY.md )
etc...

And hopefully we will be adding more labels with more checks and manage to get the project aligned with CII Best Practices. Also these are not changes that have to happen all at once we can implement these changes and then change simply change them later as in merge this then you can simply file a PR as you would like to see it and we just merged that ( + I have no idea how you are envisioning this hence I cant make those changes thus it's faster for you to make them instead of spending time going back and fourth about it ;) ).

johannbg:
@danimo are we at impasse with this one? If so it's better to just close it rather than leaving it hanging.

danimo:
I'd really like a nicer solution to this without being hit by a sack of tags. Let's restructure the README first (I have an idea or two) and then reapproach this.

PR #919, opened 2020-09-04
dracut-emergency: run emergency hooks before rdsosreport

ddiss:
Match non-systemd _emergency_shell() behaviour in running emergency
hooks prior to rdsosreport.
On my system this trims ~30ms from boot to emergency hook invocation,
which can be important when critical steps are performed by the hook.

Signed-off-by: David Disseldorp <ddiss@suse.de>

### DIFF
--- a/modules.d/98dracut-systemd/dracut-emergency.sh
+++ b/modules.d/98dracut-systemd/dracut-emergency.sh
@@ -14,9 +14,9 @@ export _rdshell_name="dracut" action="Boot" hook="emergency"
 _emergency_action=$(getarg rd.emergency)
 
 if getargbool 1 rd.shell -d -y rdshell || getarg rd.break -d rdbreak; then
+    source_hook "$hook"
     FSTXT="/run/dracut/fsck/fsck_help_$fstype.txt"
     RDSOSREPORT="$(rdsosreport)"
-    source_hook "$hook"
     while read _tty rest; do
         (
             echo


danimo:
I agree that the hooks should be performed at the same time, and not depend on whether or not systemd is used. However, this will break any fancy construct that people might use to automatically get the rdsosreport.txt off the system...

ddiss:
> I agree that the hooks should be performed at the same time, and not depend on whether or not systemd is used.However, this will break any fancy construct that people might use to automatically get the rdsosreport.txt off the system...

Yes, although those constructs would already be broken in a non-systemd setting, so I think it's safe enough to require that any such hooks (I'm not aware of any) call rdsosreport themselves.

ddiss:
> Yes, although those constructs would already be broken in a non-systemd setting, so I think it's safe enough to require that any such hooks (I'm not aware of any) call rdsosreport themselves.

Thinking about this a little more... If you feel strongly about running `rdsosreport` beforehand, how about we move invocation and output into `hookdir/emergency/01-die.sh`? This would allow `00-` emergency hooks to still run prior, while removing some duplicate `_emergency_shell()` and `dracut-emergency.sh` code. `RDSOSREPORT` could also be set for consumption by any subsequent hooks. E.g.

```
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -464,6 +464,11 @@ die() {
     {
         echo "warn dracut: FATAL: \"$*\"";
         echo "warn dracut: Refusing to continue";
+        RDSOSREPORT="$(/sbin/rdsosreport)"
+        echo
+        echo $RDSOSREPORT
+        echo 'You might want to save "/run/initramfs/rdsosreport.txt" to a USB stick or /boot'
+        echo 'after mounting them and attach it to a bug report.'
     } >> $hookdir/emergency/01-die.sh
     [ -d /run/initramfs ] || mkdir -p -- /run/initramfs
 
[remainder of patch removes duplicate `_emergency_shell()` and `dracut-emergency.sh` code]
```

ddiss:
Ping, any more feedback here?

johannbg:
@ddiss currently everyone is busy with life so anyone should expect high latency ( read as up to two weeks between ) responses and other related activities.

danimo:
Sounds like a good idea from my POV, but I would like a second opinion.

ddiss:
> Conceptually sounds like a good idea from my POV, but I would like a second opinion. die() is the only way to an emergency console, and in those situations the rdsosreport would no longer get generated...

Thanks for the review. I'll keep this open as-is, given that I think the current proposal is simple, consistent and safe. Happy to hear other opinions though.

ddiss:
@haraldh is anything blocking a merge for this one-line change?

haraldh (approved):

haraldh:
> > Conceptually sounds like a good idea from my POV, but I would like a second opinion. die() is the only way to an emergency console, and in those situations the rdsosreport would no longer get generated...
> 
> Thanks for the review. I'll keep this open as-is, given that I think the current proposal is simple, consistent and safe. Happy to hear other opinions though

oh.. forgot about the report being generated before the shell.

haraldh:
how about making the report part of the hook?

ddiss:
> how about making the report part of the hook?

Sure, I'll rework this to generate the report as part of the hook, as outlined in https://github.com/dracutdevs/dracut/pull/919#issuecomment-689124999 . Thanks for the feedback

ddiss:
I have some rough changes to provide the functionality mentioned in https://github.com/dracutdevs/dracut/pull/919#issuecomment-689124999 . I plan on cleaning them up and submitting them in the coming days.

PR #922, opened 2020-09-08
Add --uefi-output for custom output filename

### LINKED ISSUE #921: Customize output filename
Hello!

I am generating an UEFI application using dracut and for now that one is automatically installed on the system, however that is not what I want, I would just want the UEFI application to be written somewhere so I can do what I have to do with it later.

The manual says:

> The default output filename is <EFI>/EFI/Linux/linux-$kernel$-<MACHINE_ID>-<BUILD_ID>.efi. <EFI> might be /efi, /boot or /boot/efi depending on where the ESP partition is mounted. The <BUILD_ID> is taken from BUILD_ID in /usr/lib/os-release or if it exists /etc/os-release and is left out, if BUILD_ID is non-existant or empty.

I would like not to use the default, it says it's default but that means it should be possible to not use the default?

How could I modify the output filename?

Thank you

leo-lb:
Closes #921

This pull request adds support for an `--uefi-output` command line parameter and an `uefi_output` configuration directive.

## Checklist

### DIFF
--- a/dracut-bash-completion.sh
+++ b/dracut-bash-completion.sh
@@ -38,7 +38,7 @@ _dracut() {
                               --fwdir --libdirs --fscks --add-fstab --mount --device --nofscks
                               --kmoddir --conf --confdir --tmpdir --stdlog --compress --prefix
                               --kernel-cmdline --sshkey --persistent-policy --install-optional
-                              --loginstall --uefi-stub --kernel-image
+                              --loginstall --uefi-stub --kernel-image --uefi-output
                               '
         )
 
--- a/dracut.8.asc
+++ b/dracut.8.asc
@@ -524,6 +524,9 @@ will not be able to boot.
     Specifies the UEFI stub loader's splash image. Requires bitmap (**.bmp**) image
     format.
 
+**--uefi-output _<FILE>_**::
+    Specifies the UEFI application output filename, uses default if not set.
+
 **--kernel-image _<FILE>_**::
     Specifies the kernel image, which to include in the UEFI executable. The default is
     _/lib/modules/<KERNEL-VERSION>/vmlinuz_ or _/boot/vmlinuz-<KERNEL-VERSION>_
@@ -673,6 +676,8 @@ Jeremy Katz
 David Dillow
 
 Will Woods
+
+Léo Le Bouter
 
 SEE ALSO
 --------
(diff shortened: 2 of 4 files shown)


leo-lb:
Do we need unit tests for this? I have tested it locally and it runs fine.

leo-lb:
Does this need unit tests? I have tested it locally in these cases:

- [x] `--uefi-output` command line parameter present, all others absent
- [x] `uefi_output` configuration directive present, all others absent
- [x] all new options absent

And it works fine.

leo-lb:
Failing tests look to be unrelated with this PR

leo-lb:
@ryncsn I see you've accidentally unchecked some checkbox in my first post :-)

What would be missing for a review? I see there's a good amount of outstanding issues and PRs

danimo (changes requested):
sorry for the late reply. Looks good, but can you add the parameter to dracut-bash-completion.sh as well?

leo-lb:
@danimo Thanks for the review, I made the change you requested!

danimo:
@haraldh what's the attribution policy?

haraldh:
> @haraldh what's the attribution policy?

What do you mean?

haraldh:
> @haraldh what's the attribution policy?

@danimo what do you mean?

danimo:
> > @haraldh what's the attribution policy?
> @danimo what do you mean?

The patch adds the author's name to the man pages, and I realized that we have no real attribution policy there.

haraldh:
> > > @haraldh what's the attribution policy?
> > > @danimo what do you mean?
> 
> The patch adds the author's name to the man pages, and I realized that we have no real attribution policy there.

Yeah, it shouldn't do that, unless the author belongs to the top committers.

Rule of thumb: running `make AUTHORS; head -n 10  AUTHORS` would output those.

haraldh:
Or we just point to the shipped `AUTHORS` file.

leo-lb:
I added it because I saw quite a few names there and that I edited the manual. To me that was the manual's authors, not the software authors.

danimo:
Iet's update the authors file as part of the release process.

leo-lb:
Thank you for merging, feel free to remove my name in the manual if that's not OK with you.

barmadrid:
@danimo please review [this](https://github.com/dracutdevs/dracut/issues/921#issuecomment-733838751) and confirm if this merge should be reverted. Thank you!

PR #923, opened 2020-09-10
04watchdog: split the watchdog module install

ryncsn:
In some cases, user only want to include the watchdog module, not the
wdctl or any other userspace helper. For example, systemd have a
RebootWatchdogSec option that use watchdog to prevent reboot hangs. And
it can help prevent machines hangs when reboot directly within the
initramfs stage. So split the module installation to a standlone module.

Also when watchdog-module get included, install driver for all loaded
watchdog instaed of only install driver for active watchdog. Both
watchdog and watchdog-module return 255 in check(), so it's enabled only
when manually included, the watchdog may get configured/activated later.

This pull request changes...

## Checklist

Fixes #


danimo:
LGTM

PR #927, opened 2020-09-20
RFC: network-manager: start NM as systemd service with D-Bus support

bengal:
This commit changes how NM is started inside the initramfs. Instead of
synchronously running NM from the initqueue hook in the special
`--configure-and-quit=initrd` mode, start it from a systemd service.

This has multiple advantages. First, we no longer need to run NM in a
special mode that requires additional testing and
maintenance. NetworkManager works exactly as in the real root.

Also, now NM runs with D-Bus support and therefore its API can be used
by other modules. This open the possibility, for example, to integrate
nm-cloud-setup to automatically configure networking in the initramfs
based on cloud metadata.

This commit also enables the NetworkManager dispatcher service, that
allows to run shell scripts when an interface state changes. In this
way, we can call the various dracut hooks (online, netroot) as soon as
one interface gets activated instead of waiting that all interfaces
were configured (as the current implementation does).

Requirements for this commit are:

 - the D-Bus dracut module submitted at:
   https://github.com/dracutdevs/dracut/pull/544
   https://github.com/dracutdevs/dracut/pull/900

 - the systemd-hostnamed dracut module submitted at
   https://github.com/dracutdevs/dracut/pull/895 , which allows NM to change the hostname via D-Bus
   through hostnamed


danimo:
Can you modify the patch to work with the new `06dbus` module? If it needs adjustments, feel free to make them.

zboszor:
This would be nice to be revived and included. Currently, if NM is included in initramfs, the hostname is acquired early from DHCP but then it's reset to localhost because systemd-hostnamed is not included and when systemd (re)starts after switch_root, the static hostname (localhost.localdomain) is reloaded.

PR #928, opened 2020-09-22
Include devfreq drivers in initrd

nullr0ute:
Some SoCs now have drivers that user devfreq in early init and fail
if the drivers are missing so make sure we have them in the initrd.

This pull request changes...
kernel modules in initrd
## Checklist

Fixes #
Some arm devices that have GPU drivers that use devfreq for scaling


johannbg:
Not sure if we should start including various drivers for the plethora of SoCs out there by default upstream which in turns leads to downstream not supporting those SoCs having to patch that out but otherwise LGTM @danimo @haraldh

haraldh (approved):
LGTM

PR #935, opened 2020-10-01
95nvmf: rework FC connection handling

hreinecke:
This pull request reworks the FC-NVMe connection handling. The original design of creating custom systemd services on the fly and activate them via udev doesn't work in the initrd due to the lack of dbus.
This patchset reworks the connection handling by using initqueue jobs to establish the initial connections.
Also the manpage is updated to document the nvmf commandline options.


danimo (approved):

danimo:
Note: we will have (optional) dbus support in dracut master soon, but I agree it should work without, esp. if you want to backport the module to earlier versions.

PR #937, opened 2020-10-01
dracut-install: fix edge-case regression with weak modules

danimo:
This was introduced with 6dafdda4a6bdb8721133e4267553c5d86564f9e8, but
is actually caused by the fact that modules that have already been
installed cause dracut_install() to return without adding the module
to the hashmap. This can happen if an earlier-run dracut module chose
to install the same module. Now modprobe statements like:

softdep usb_storage post: uas
softdep uas pre: usb_storage

(which look weird, but are perfectly valid), cause
dracut_install() to enter an infinite recursion if and only if
at least one of the files has previously been installed by another
module.

Fix this by also adding already installed modules to the hashmap.

## Checklist

### DIFF
--- a/install/dracut-install.c
+++ b/install/dracut-install.c
@@ -807,123 +807,123 @@ static int dracut_install(const char *orig_src, const char *orig_dst, bool isdir
                         log_debug("'%s' already exists", fulldstpath);
 
                 /* dst does already exist */
-                return ret;
-        }
-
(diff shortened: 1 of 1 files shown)


haraldh:
Thanks!

danimo:
> Thanks!

I take that as an approval :)

PR #940, opened 2020-10-03
rootfs-block: only write root argument for block device

sgn:
Some filesystem (e.g. ZFS, and btrfs subvolumes) don't use block
devices. Should they be mounted as `/`, `find_root_block_device`
yields nothing, hence dracut will append this problematic argument
to kernel cmdline:

	root=/dev/block

On a machine that employ root ZFS on LUKS, which was setup with
an OpenPGP-encrypted key file, this argument renders that machine
unbootable. Remove that `root=/dev/block` manually could boot the
machine.

Let check if that device is a block device before write down `root`
argument. This is consistent with the check for block device in
`find_block_device`.

Signed-off-by: Đoàn Trần Công Danh <congdanhqx@gmail.com>

---

This pull request changes...

## Changes
- rootfs-block: only write root argument for block device

## Checklist

### DIFF
--- a/modules.d/95rootfs-block/module-setup.sh
+++ b/modules.d/95rootfs-block/module-setup.sh
@@ -30,7 +30,8 @@ cmdline_journal() {
 }
 
 cmdline_rootfs() {
-    local _dev=/dev/block/$(find_root_block_device)
+    local _block=$(find_root_block_device)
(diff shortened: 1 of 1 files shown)


sgn:
I intentionally keep the `-e` check for `rootfstype` and `rootflags` because it still works

PR #943, opened 2020-10-07
90kernel-modules: arm: add drivers/hwmon for arm/arm64

kwizart:
In the case of the s805x the drivers/hwmon directory contains the
scpi_hwmon kernel module.
On a running system, lsmod would output the following dependencies:
 arm_scpi               24576  2 clk_scpi,scpi_hwmon

It means that if the clock driver is bundled in the initramfs it will
bring arm_scpi. But if scpi_hwmon is missing the scpi will be incomplete
and it can lead to crashes.

When the hwmon is bundled, no crash occurs

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -69,6 +69,7 @@ installkernel() {
                 "=drivers/dma" \
                 "=drivers/extcon" \
                 "=drivers/gpio" \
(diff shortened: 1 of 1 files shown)


kwizart:
Note: I'm currently requesting comment for this change, specially trying to contact the upstream maintainers.

I will report the exact crash at some point.

Issue #947, opened 2020-10-16
btrfs raid not ready but systemd tries to mount it anyway


danieljrmay:
**Describe the bug**
I have a machine with a EXT4 / partition and a 10 HDD BTRFS RAID1 volume which I have mounting at `/srv` in `/etc/fstab`. Out of the 10 HDDs which make up the BTRFS volume 4 were connected directly to the motherboard, 6 were connected via an HBA card. In this configuration the the BTRFS volume would fail to mount automatically at boot time. 

I initially reported this issue at ask.fedoraproject.org:

https://ask.fedoraproject.org/t/btrfs-volume-mount-fails-at-boot-but-works-once-system-is-up/9593

The same issue was then reported to the systemd-devel mailing list:

https://lists.freedesktop.org/archives/systemd-devel/2020-October/045399.html

Where it has been suggested that the issue may be with dracut:

https://lists.freedesktop.org/archives/systemd-devel/2020-October/045440.html

Hence I am reporting this here.

**Distribution used**
Fedora 32.

**Dracut version**
```
$ sudo rpm -q dracut
dracut-050-61.git20200529.fc32.x86_64
```

**Init system**
```
$ sudo rpm -q systemd
systemd-245.8-2.fc32.x86_64
```

**To Reproduce**
This may be due to a race condition so it might be difficult to reproduce. However, a setup similar to mine may well have the same problem: non-BTRFS / filesystem, a multi-device BTRFS filesystem which is split between direct connection to the motherboard and HBA card and is configured to mount in `/etc/fstab` 

The BTRFS volume *mounts successfully at boot time* when:
- `udev.log_priority=debug systemd systemd.log_level=debug` is added to the grub boot parameters.
- All 10 HDDs are connected to the HBA card (with normal grub boot parameters).
 
**Expected behavior**
The BTRFS volume should mount automatically during the boot process.

**Additional context**
This is a Fedora 32 machine which I think was upgraded via `dnf` from Fedora 31 sometime ago. The BTRFS volume was added recently when migrating storage from an `mdadm` based RAID6 XFS filesystem.

I hope this makes some kind of sense :wink:

cmurf:
I agree with Lennart. Either include the btrfs udev rule unconditionally or if fstab contains any btrfs reference.

danieljrmay:
> It's definitely not in @danieljrmay 's initramfs when inspected with `lsinitrd`.

I should have mentioned this in the original report:

```
$ sudo uname -a
[...]

PR #948, opened 2020-10-16
Adding btrfs udev rules might fix issue 947

johannbg:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/95udev-rules/module-setup.sh
+++ b/modules.d/95udev-rules/module-setup.sh
@@ -38,7 +38,9 @@ install() {
         60-pcmcia.rules \
         60-persistent-storage.rules \
         61-persistent-storage-edd.rules \
-        70-uaccess.rules \
+	64-btrfs-dm.rules \
+	64-btrfs.rules \
+	70-uaccess.rules \
         71-seat.rules \
         73-seat-late.rules \
         75-net-description.rules \


modules.d/95udev-rules/module-setup.sh
```diff
@@ -38,7 +38,9 @@ install() {
         60-pcmcia.rules \
         60-persistent-storage.rules \
         61-persistent-storage-edd.rules \
-        70-uaccess.rules \
+	64-btrfs-dm.rules \
```
danimo, line 41: Please fix indentation.

danimo (changes requested):
See comment.

haraldh:
looks good, although I would add them in `modules.d/90btrfs`

johannbg:
@haraldh yeah I came to the same conclusion along with remove all this legacy cruff that is in that module. From my pov dracut should do a release then after that we need to seriously re-facture-it/rework it and remove all that legacy cruff like that old ( btrfs ) rules file etc. On top of that I took a wee bit closer look at that 64-btrfs-dm.rules I included and is being shipped with btrfs-progs in fedora and seems to be broken so that needs fixing in the *right* place as well so arguably we should not be including it. @haraldh @danimo thanx for reviewing this.

danimo:
@haraldh fine with merging as-is?

haraldh:
> @haraldh fine with merging as-is?

@danimo yes

johannbg:
@danimo @haraldh Note that 64-btrfs-dm.rules ( which come from btrfs-progs ) and I included are outdated and probably break with systemd 247+ so by adding it we might be signing ourselves up for more issues.

johannbg:
.

cmurf:
I don't see how /usr/lib/udev/rules.d/64-btrfs-dm.rules would fix #947. What's missing from the initrd in that bug is /usr/lib/udev/rules.d/64-btrfs.rules to make sure systemd waits for all btrfs devices to be ready before trying to mount. 

/usr/lib/udev/rules.d/64-btrfs.rules is part of systemd

/usr/lib/udev/rules.d/64-btrfs-dm.rules is part of btrfs-progs, if that's somehow stale or out of date, it needs an issue filed with [btrfsprogs](https://github.com/kdave/btrfs-progs/issues)

johannbg:
Let's start by getting the udev rules into the reporters initramfs before claiming that it's not the problem and I'm very well aware of where I could file reports against btrfs-prog and those rules are broken ( as well as bunch of other udev rules ) but thanks for the info. We will look at this properly when we have the time...

cmurf:
The [problem is resolved](https://ask.fedoraproject.org/t/btrfs-volume-mount-fails-at-boot-but-works-once-system-is-up/9593/23?u=chrismurphy) when `dracut -f --add btrfs` is used, which results in usr/lib/udev/rules.d/64-btrfs.rules being included in the initramfs. Whereas /usr/lib/udev/rules.d/64-btrfs-dm.rules is absent. In the original report, as well as my testing, dm isn't being used for anything.

PR #957, opened 2020-10-21
Introduce a --hostonly-nics option

ryncsn:
This pull request changes...

## Changes
Introduce a --hostonly-nics param, with this param, if strict hostonly mode is used, it will only install driver for specified **NIC.000p**

## Checklist

Fixes #


danimo:
Out of curiosity: What use case necessitates that patch?

ryncsn:
Hi, it's mostly for kdump, and so are many other --hostonly-xxx options. Kdump uses dracut, but have a very limited memory, so it expects the initramfs to be as small as possible. In some case, where there are multiple NIC cards using different drivers, and kdump knows which NIC is needed, it can specify the needed NICs and drop all other unnecessary drivers.

PR #958, opened 2020-10-21
Improve squash module installation

ryncsn:
This pull request changes...

## Changes
Make squash module installation covers more cases. Now it works when needed module is built-in, and fail earlier if there are missing modules. 

## Checklist

Fixes #


danimo:
LGTM

PR #930, opened 2020-09-23
00systemd: add missing cryptsetup-related targets

jlebon:
We want these in the initramfs. Things related to clevis and systemd's
`cryptsetup-generator` reference these targets.

PR #959, opened 2020-10-21
90crypt: ship initrd-cryptsetup.target

jlebon:
This was added in https://github.com/systemd/systemd/pull/17149 and is
the designated cryptsetup target for all encrypted volumes that need to
be opened in the initrd.

So it effectively replaces `cryptsetup.target` and
`remote-cryptsetup.target` there. I've removed the latter since it was
added recently (by me) in #930, but kept `cryptsetup.target` since we've
been shipping it for a long time now.


danimo:
LGTM

PR #965, opened 2020-11-07
fix the bug that cannot be stopped before swtich_root in CentOS 8.2

linux-gcc:
1、fix the bug that cannot be stopped before switch_root in CentOS 8.2 
2、commit id: c289e35


linux-gcc:
> 1、fix the bug that cannot be stopped before switch_root in CentOS 8.2
> 2、commit id: [c289e35](https://github.com/dracutdevs/dracut/commit/c289e359c5202af2515a947e15769e9aa7ce9b5e)

linux-gcc:
> 1、fix the bug that cannot be stopped before switch_root in CentOS 8.2
> 2、commit id: [c289e35](https://github.com/dracutdevs/dracut/commit/c289e359c5202af2515a947e15769e9aa7ce9b5e)

PR #966, opened 2020-11-10
fix the breakpoint failure before swtich_root in centos 8.2

linux-gcc:
1、fix the breakpoint failure before swtich_root in centos 8.2;
2、whel the system starts with the rd.break parameter,the system cannot stop before ;
3、fix patch: dadfdc0

### DIFF
--- a/modules.d/50drm/module-setup.sh
+++ b/modules.d/50drm/module-setup.sh
@@ -33,7 +33,7 @@ installkernel() {
     if [[ $hostonly ]]; then
         for i in /sys/bus/{pci/devices,virtio/devices,soc/devices/soc?}/*/modalias; do
             [[ -e $i ]] || continue
-            if hostonly="" dracut_instmods --silent -s "drm_crtc_init" -S "iw_handler_get_spy" $(<$i); then
+            if hostonly="" dracut_instmods --silent -s "drm_crtc_init|drm_dev_register" -S "iw_handler_get_spy" $(<$i); then
                 if strstr "$(modinfo -F filename $(<$i) 2>/dev/null)" radeon.ko; then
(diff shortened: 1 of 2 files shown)


linux-gcc:
can you take the time help review the code  ? dadfdc0
@haraldh @ryncsn @johannbg @danimo @lnykryn

linux-gcc:
can you take the time help review the code ? dadfdc0
@dtardon 

thanks !

linux-gcc:
1、 can you help review this code? dadfdc0
2、I have committed the bugzilla.
     bugzilla url: https://bugzilla.redhat.com/show_bug.cgi?id=1896591 
     
@haraldh @lnykryn

dtardon:
@linux-gcc please stop pinging random people.

linux-gcc:
what should  I do with this PR ?  looking forward to your reply. thanks !
@haraldh

haraldh:
So, you want to revert 556ff7c76a5c365889ea9972c1878eaf7073bde8 ..

Specifying `rd.break` on the kernel command line will halt dracut just fine at this point.
You can even specify `rd.break=cleanup rd.break` and it halt in the cleanup phase and just before switch_root.

Seems to work for me, and other didn't report anything similar in the last 4 years.

linux-gcc:
First of all, thank you very much for your reply！
 1、It means that there is a difference in the breakpoint function usage of initramfs between centos 7.x and centos 8.x?
 2、The centos 7.x series should not incorporate this patch,（556ff7c ）， so this issue should not have been reported in the last 4 years.
 3、In addition, what is the purpose of incorporating this patch before incorporating it?
@haraldh

PR #967, opened 2020-11-12
Allow LiveOS image from a nbd device

luizluca:
This pull request changes...

## Changes

Allows to use an nbd device exporting an squashfs image in a overlay setup.

## Checklist
- [  ] I have reviewed and updated any documentation if relevant
- [  ] I am providing new code and test(s) for it

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -78,7 +78,11 @@ if [ -f $livedev ]; then
     esac
     [ -e /sys/fs/$fstype ] || modprobe $fstype
 else
-    if [ "$(blkid -o value -s TYPE $livedev)" != "ntfs" ]; then
+    livedev_fstype=$(blkid -o value -s TYPE $livedev)
+    if [ "$livedev_fstype" = "squashfs" ]; then
+       # no mount needed - we've already got the LiveOS image in $livedev
+       SQUASHED=$livedev
+    elif [ "$livedev_fstype" != "ntfs" ]; then
         mount -n -t $fstype -o ${liverw:-ro} $livedev /run/initramfs/live
     else
         # Symlinking /usr/bin/ntfs-3g as /sbin/mount.ntfs seems to boot
--- a/modules.d/95nbd/nbdroot.sh
+++ b/modules.d/95nbd/nbdroot.sh
@@ -109,25 +109,9 @@ if strstr "$(nbd-client --help 2>&1)" "systemd-mark"; then
 fi
 
 if [ "$nbdport" -gt 0 ] 2>/dev/null; then
-    if [ -z "$DRACUT_SYSTEMD" ]; then
-        nbd-client "$nbdserver" $nbdport /dev/nbd0 $preopts $opts || exit 1
-    else
-        systemd-run --no-block --service-type=forking --quiet \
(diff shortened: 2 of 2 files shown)


danimo:
@haraldh as this proposes to revert one of your commits, you may want to review this one.

luizluca:
> This issue is being marked as stale because it has not had any recent activity. It will be closed if no further activity occurs. If this is still an issue in the latest release of Dracut and you would like to keep it open please comment on this issue within the next 7 days. Thank you for your contributions.

Ping

luizluca:
Keep open

johannbg:
@luizluca any particular reason you are removing the systemd logic?

luizluca:
> @luizluca any particular reason you are removing the systemd logic?

Yes, I described it in commit message (3dcd76a893f7caddb4ae723e3a5467bd18444bcf).

systemd runs everything at the same time in parallel when not explicitly restricted, including
a systemd-run job. It uses its dependency system to synchronize them. However, here,
systemd is used without using dependencies. With systemd, you either go all-in or it will be
broken the moment anything depends on a product from a systemd service.

I don't know the best solution for fixing systemd usage here. systemd-run could be executed synchronously,
but it would be no different from a direct call. It's complicated to synchronize from a non-systemd process
without adding polling everywhere a resource that might be provided by systemd is in use.

johannbg:
@luizluca upstream recently did changes to it's type unit see https://github.com/NetworkBlockDevice/nbd/issues/91 so could you see if that affects your changes and even test with the following changes
```
systemd-run --no-block --service-type=oneshot --quiet \
                    --description="nbd nbd0" \
                    -p 'DefaultDependencies=no' \
                    -p 'RemainAfterExit=yes' \
                    ...
```

luizluca:
> @luizluca upstream recently did changes to it's type unit see [NetworkBlockDevice/nbd#91](https://github.com/NetworkBlockDevice/nbd/issues/91) so could you see if that affects your changes and even test with the following changes
> 
> ```
> systemd-run --no-block --service-type=oneshot --quiet \
>                     --description="nbd nbd0" \
>                     -p 'DefaultDependencies=no' \
>                     -p 'RemainAfterExit=yes' \
>                     ...
> ```

I didn't test but it should not fix the issue, at least not alone. The issue is that nbd jobs runs in background and there is nothing afterwards that check if it worked or not. If nbdclient can now work better informing if it worked of not to systemd. The question is who would check for that state? Today, no module will check for systemd services.

Every module that might use a device should be built considering that that device it will use is still not available. There is already this type of check but only for /dev/root, doing a loop with a timeout. If every module was built as a systemd unit, it could simply use systemd dependency system, waiting for a xxx.device that it will use, for example. It needed some kind of check at https://github.com/dracutdevs/dracut/blob/4026cd3b012be7f06c7572bdf50dd54a18ecb684/modules.d/90dmsquash-live/dmsquash-live-root.sh#L37 for example. If blkid is called on a still missing /dev/nbd0, it will fail.

I'm not sure if systemd-run is designed to use the dependency system (either providing or requiring them). However, as it creates a service, it could theoretically be used that way. Maybe if you run a dummy service synchronous systemd-run depending on a device might work. Until then, the safest solution is to not run nbdclient in parallel.

johannbg (approved):

haraldh (approved):

PR #972, opened 2020-11-18
95fcoe: don't install if there is no FCoE hostonly devices

ryncsn:
## Changes
When in hostonly mode, 95fcoe module will still be installed even there
is no FCoE hostonly device. So use the new block_is_fcoe helper to check
for hostonly device in hostonly mode, avoid installing unneccessary module.

## Checklist

### DIFF
--- a/modules.d/95fcoe-uefi/module-setup.sh
+++ b/modules.d/95fcoe-uefi/module-setup.sh
@@ -2,22 +2,15 @@
 
 # called by dracut
 check() {
-    local _fcoe_ctlr
-    [[ $hostonly ]] || [[ $mount_needs ]] && {
-        for c in /sys/bus/fcoe/devices/ctlr_* ; do
-            [ -L $c ] || continue
-            _fcoe_ctlr=$c
-        done
-        [ -z "$_fcoe_ctlr" ] && return 255
+    is_fcoe() {
+        block_is_fcoe $1 || return 1
     }
+
     [[ $hostonly ]] || [[ $mount_needs ]] && {
+        for_each_host_dev_and_slaves is_fcoe || return 255
         [ -d /sys/firmware/efi ] || return 255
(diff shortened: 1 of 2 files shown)


ryncsn:
Hi, @danimo , I saw in commit 480aa9695 you added this helper block_is_fcoe, but left a comment:
# block_is_fcoe <maj:min>
# Check whether $1 is an FCoE device
# Will not work for HBAs that hide the ethernet aspect
# completely and present a pure FC device

I'm not sure about the case that the ethernet aspect is hidden, in such cases, will the original FCoE detection logic in 95fcoe (check if "/sys/bus/fcoe/devices/ctlr_*" exists) work? And in that case, is the 95fcoe module still needed?

danimo:
> Hi, @danimo , I saw in commit [480aa96](https://github.com/dracutdevs/dracut/commit/480aa9695f8c2e2b30c8f41ae8483140020d23db) you added this helper block_is_fcoe, but left a comment:
> 
> block_is_fcoe maj:min
> Check whether $1 is an FCoE device
> Will not work for HBAs that hide the ethernet aspect
> completely and present a pure FC device
> 
> I'm not sure about the case that the ethernet aspect is hidden, in such cases, will the original FCoE detection logic in 95fcoe (check if "/sys/bus/fcoe/devices/ctlr_*" exists) work? And in that case, is the 95fcoe module still needed?

Our storage department told me that those adapters appear as if they were local block storage HBAs, so the FCoE aspect is completely hidden from the operating system and thus the 95fcoe module is not needed. If there are modules that have an inbetween behavior, the current logic should be good enough to detect whether or not to include the module.

danimo:
Based on that information, do you want me to merge the PR?

ryncsn:
Thanks for the info, I think it's OK to merge it.

PR #975, opened 2020-11-26
Drop 51-dracut-rescue-postinst.sh entirely

AdamWill:
It is only installed on RHEL 8+, but as noted in the first commit
that tried to do this - 0bb9a683 - we already have branches for
all RHEL releases, so there's no point keeping such a conditional
on the master branch. After 0bb9a683, 51-dracut-rescue-postinst.sh
was re-added because Fedora 30 needed it, with a FIXME saying to
remove it after F30 was released. But instead of that happening, it
got changed to a conditional to install it on Fedora <= 30 or RHEL
<= 8 in 9e68789d and the FIXME was removed, then in 9eb1d1ed that
conditional was simplified to only refer to RHEL, so we wound up
with this situation.

Note the last two commits were never ported to the RHEL-8 branch,
so it's broken, I will send a separate PR for that.

This pull request changes...

## Checklist

Fixes #


johannbg:
@AdamWill thanks for the patch @lnykryn Fedora/RHEL stuff otherwise LGTM...

PR #976, opened 2020-11-26
fix(rhel): drop 51-dracut-rescue-postinst.sh entirely

AdamWill:
It should not be installed in RHEL-8. See #975 and
https://bugzilla.redhat.com/show_bug.cgi?id=1677723 for further
detail; basically this should only ever be installed on RHEL < 8
and Fedora < 31, but it got pinged around a lot and the RHEL-8
branch happened to be forked off at a point where it gets
installed unconditionally. Later changes to make it only install
on RHEL < 8 were committed to master but never to the RHEL-8
branch, so never really did anything. Since this branch is for
RHEL-8 specifically, there's no sense fiddling with conditionals,
we can just ditch it entirely.

This pull request changes...

## Checklist

Fixes #


AdamWill:
This is still waiting for review from the maintainer.

lnykryn:
Could you please file rh bugzilla against rhel8 for this? And for bonus points could you add 
Resolves: #XXXXXXX
line with that bugzilla number to the commit messages?

haraldh (approved):

johannbg:
@lnykryn can you review this so we can get rid of this PR from the pipeline?

PR #982, opened 2020-12-04
fix(90kernel-modules): install generic crypto modules with hostonly unset

WOnder93:
Otherwise e.g. the xts(aes) implementation provided by the vmx_crypto
module (which does usually get included on ppc64le) fails to initialize
when xts is built as a module (CONFIG_CRYPTO_XTS=m), because it can't
instantiate the fallback generic xts(aes) implementation (needs the
generic xts module).

## Changes
(see commit message)

## Checklist
  - not sure if this can be easily tested, as it requires running a kernel with a specific config and some HW crypto driver that requires a fallback algorithm implementation

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -131,7 +131,8 @@ installkernel() {
         [[ $arch == x86_64 ]] && arch=x86
         [[ $arch == s390x ]] && arch=s390
         [[ $arch == aarch64 ]] && arch=arm64
-        instmods "=crypto" "=arch/$arch/crypto" "=drivers/crypto"
+        hostonly='' instmods "=crypto"
+        instmods "=arch/$arch/crypto" "=drivers/crypto"
     fi
     :
 }


johannbg (approved):
LGTM

WOnder93:
Yes, this is still an issue and is waiting for a review... :roll_eyes:

johannbg:
@haraldh, @danimo one stuck in the pipeline plz review....

danimo (approved):
LGTM

danimo:
Thanks for the patch and sorry our bots had a faster recovery from the festivities than us :wink:

johannbg:
@WOnder93 you need to fix your commit messages so it passes the commisery check in the pipeline

WOnder93:
@johannbg Right... should be fixed now.

haraldh (approved):

WOnder93:
Hm... the check seems to be checking out the wrong ref for some reason.

johannbg:
@WOnder93 out of curiosity what did you have to do to fix this?

WOnder93:
It seems the GH actions tooling created an additional merge commit internally, which got the commit header from the PR title. So fixing the PR title made the Commisery check pass. (I suppose you have merging by merge commit enabled by default and it's trying to simulate the merge?)

PR #990, opened 2020-12-15
test(conventional): add Conventional Commits PR github action

haraldh:
To ease the creation of the release notes, enforce some style on the git
commit message titles with `Commisery`.

https://github.com/marketplace/actions/commisery

Using https://github.com/clog-tool/clog-cli later on these commits will
generate a pretty nice base for the release notes.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,19 @@
+name: Commisery
+on:
+  pull_request:
+    types: [edited, opened, synchronize, reopened]
+
+jobs:
+  commit-message:
+    name: Conventional Commit Message Checker (Commisery)
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check-out the repo under $GITHUB_WORKSPACE
+      uses: actions/checkout@v2
+
+    - name: Run Commisery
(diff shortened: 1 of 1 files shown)


johannbg:
From the looks of it clog-cli upstream seams inactive/dead but it is as good start as any but I would recommend we continue looking for a tool with more activity since it's a high probability we might end up having to replace it down the road anyway.

johannbg (approved):

haraldh:
> From the looks of it clog-cli upstream seams inactive/dead but it is as good start as any but I would recommend we continue looking for a tool with more activity since it's a high probability we might end up having to replace it down the road anyway.

@johannbg oops... linked to the wrong github action in the commit/PR message.

https://github.com/KevinDeJong-TomTom/commisery-action
and
https://github.com/tomtom-international/commisery

are pretty active I guess.

haraldh:
> From the looks of it clog-cli upstream seams inactive/dead but it is as good start as any but I would recommend we continue looking for a tool with more activity since it's a high probability we might end up having to replace it down the road anyway.

Yeah, the log extractor does what we want, I guess and is easy to replace.

PR #1000, opened 2021-01-07
feat(kernel-modules): add drivers memory for ARM

kwizart:
External Memory Controller drivers can be built as a module.

On Tegra, this allows tegra_drm to load faster and without defered probes,
instead of waiting for the rootfs for the tegra*_emc dependency.

Tested on jetson-tk1 and ac100.

This pull request changes...

## Changes

Note: Building Tegra EMC drivers as a module is possible with mainline for 5.11+.

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -79,6 +79,7 @@ installkernel() {
                 "=drivers/hwmon" \
                 "=drivers/hwspinlock" \
                 "=drivers/i2c/busses" \
+                "=drivers/memory" \
                 "=drivers/mfd" \
                 "=drivers/mmc/core" \
                 "=drivers/phy" \


kwizart:
Actually, TEGRA124_EMC will be made modular with 5.12 (See also https://git.kernel.org/pub/scm/linux/kernel/git/krzk/linux-mem-ctrl.git/log/?h=for-v5.12/tegra-mc )
Other Tegra are already compatible...

danimo (approved):

danimo:
...ad before we can merge this, the commit request needs adhere to the styleguide layed out [here](https://github.com/dracutdevs/dracut/blob/master/HACKING.md#commit-messages).

kwizart:
Thanks for the comment, I believe to have changed for the new style.

johannbg:
@kwizart it's still broken. Change it to :point_right:  feat(kernel-modules): add driver memory

johannbg (approved):

johannbg:
@kwizart can you fix your commit msg so we can merge this?  The message it is failing on is "90km: arm: Add drivers/memory" changed the message to "feat(kernel-modules): add driver memory" without "" so it passes.

kwizart:
Okay, I've changed the merge title without changing the individual commit title

That's what I was missing...
Thanks for the pointer.

PR #1008, opened 2021-01-16
refactor: use /usr prefix

johannbg:
This change brings the project somewhat up to speed with the usr move which should be safe after a decade or so lol...

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,9 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
+# review when someone opens a pull request.
+*       @haraldh @danimo @johannbg
+
--- a/dracut.logrotate
+++ b/dracut.logrotate
@@ -1,6 +1,0 @@
-/var/log/dracut.log {
-    missingok
-    notifempty
-    size 30k
-    create 0600 root root
-}
--- a/modules.d/95zfcp_rules/module-setup.sh
+++ b/modules.d/95zfcp_rules/module-setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/bash
 
 # called by dracut
 cmdline() {
--- a/modules.d/95zfcp_rules/parse-zfcp.sh
+++ b/modules.d/95zfcp_rules/parse-zfcp.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/bash
 
 create_udev_rule() {
     local ccw=$1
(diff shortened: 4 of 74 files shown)


haraldh:
No... `/bin/sh` and `/bin/bash` should not be changed in my opinion. What is the benefit here?

behrmann:
>  secondly consistency with other building blocks of the OS like systemd which uses throughout the project #!/usr/bin/env bash , #!/usr/bin/env python3 

This seems a better solution and should skirt the potential outfall.

johannbg:
> At least the shell shebang can break some users, [we had this problem in mkosi](https://github.com/systemd/mkosi/issues/532). Also, Debian still supports split usr setups, although everything except for buildds default to merged usr and I'm not sure whether those use dracut.

If Debian can patch systemd for this ( which it must be doing, if this truly is a problem )  it can do so for dracut and the reality is that we cannot adhere to the whims of downstream trying to support multiple init systems or ship modules related to that since it generates a lot of additional work for the project, which resources are already stretched thin enough as is. ( Which should be quite obvious ). We need to reverse that evolution, clean that legacy cruff up and make the project being agile enough to move forward with the rest of the building blocks of the OS.

haraldh:
> > No... `/bin/sh` and `/bin/bash` should not be changed in my opinion. What is the benefit here?
> 
> Consistency.
> Let's start with the dialogue of always using bash instead of sh, which majority of the project uses ( or the other way around change it to sh ), secondly consistency with other building blocks of the OS like systemd which uses throughout the project #!/usr/bin/env bash , #!/usr/bin/env python3 ( which obviously means I have to re-do this but I just noticed this inconsistency when I was going to implement the spdx license stuff ).
> 
> What's the benefit of keeping it like is?

So, debian wants to use `dash` in the initramfs. Therefore the initramfs scripts use `/bin/sh` and must use posix compatible shell syntax. The initramfs creating part uses `/bin/bash`.

Although I made the usr-move in Fedora, I recommend not changing `/bin/sh` and `/bin/bash` as those are unix fundamentals. 

Also I don't recommend `env`, because dracut should use the system shell, instead of the one in /usr/local or $HOME/bin.

johannbg:
Most people that are posix standard advocates have no idea how posix came to existence or know it's short comings or aware of the reality that the current posix standard on its own, is not rich enough to form a basis for practical applications in the real world but people apparently care deeply about it as well as the unix *guideline* that was reflective of the time it was written, guidelines that somehow over time became viewed as a philosophy or word of god in some cases. I personally am neither with or against either of these things as long as they dont stand in the way of progress. 

Now what I personally care about is consistency, consistency in the code, consistency the documentation which leads to consistency in reviews etc. what that consistency is means less to me than it not being consistent et all so we just need to reach consensus what that is and create task to ensure we meet that consensus.

That said I'm curious to know why a single distribution ( Debian in this case ) has so much influence over the progress of the project? 

Not only are we bending with this decision but also not merging your rust implementation due to set back in Debian and I would like to know why? What makes it so special over the other parent distributions and derivatives? For the past years the Debian community has been at a stale mate burning it's maintainers out as a result of that. 

Maintainers there effectively cant move forward even thou the world around them is and that is mostly due to the e<fork> people being dominating vocal crowd there. A crowd that is resisting any kind of change in the FOSS ecosystem and will inevitable lead to the community demise if allowed to continue. A crowd which in turn can indirectly hold us back and lead to dracut demise if Debian is allow have some kind of influence over the progress of the project like that we cant make changes out of consideration for Debian.

PR #1009, opened 2021-01-16
fix: add a missing efi support

johannbg:
Let's make lsinitrd work on distributions that support efi and follow the boot loader specification shall we...

### DIFF
--- a/lsinitrd.sh
+++ b/lsinitrd.sh
@@ -97,7 +97,11 @@ if [[ $1 ]]; then
 else
     [[ -f /etc/machine-id ]] && read MACHINE_ID < /etc/machine-id
 
-    if [[ -d /boot/loader/entries || -L /boot/loader/entries ]] \
+    if [[ -d /efi/loader/entries || -L /efi/loader/entries ]] \
+        && [[ $MACHINE_ID ]] \
+        && [[ -d /efi/${MACHINE_ID} || -L /efi/${MACHINE_ID} ]] ; then
+        image="/efi/${MACHINE_ID}/${KERNEL_VERSION}/initrd"
+    elif [[ -d /boot/loader/entries || -L /boot/loader/entries ]] \
         && [[ $MACHINE_ID ]] \
         && [[ -d /boot/${MACHINE_ID} || -L /boot/${MACHINE_ID} ]] ; then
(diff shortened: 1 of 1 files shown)


lsinitrd.sh
```diff
@@ -97,7 +97,11 @@ if [[ $1 ]]; then
 else
     [[ -f /etc/machine-id ]] && read MACHINE_ID < /etc/machine-id
 
-    if [[ -d /boot/loader/entries || -L /boot/loader/entries ]] \
+    if [[ -d /efi/loader/entries || -L /efi/loader/entries ]] \
```
behrmann, line 100: More of a stylistic question: `-L` is also true if the symlink is broken, should that be guarded against?

behrmann (commented):
This seems straightforward enough and following the [spec](https://systemd.io/BOOT_LOADER_SPECIFICATION/). One could be generous and support `/boot/efi` as well, which according to the spec might be supported by implementations, but that's probably not necessary.

lsinitrd.sh
johannbg: > This seems straightforward enough and following the [spec](https://systemd.io/BOOT_LOADER_SPECIFICATION/). One could be generous and support `/boot/efi` as well, which according to the spec might be supported by implementations, but that's probably not necessary.

I dont think we want to support /boot/efi because people will inevitable getting the bright idea of having /boot/efi on a separate file system from /boot. The spec is wrong in this regard and should have prohibited as opposed to recommend against it from my pov. 

I leave any stylish question to @haraldh and the question if we want to change "|| -L $foo ]]" to something else, It's being used in few places so it's not too big of a scope to change it to something else if there is consensus what it should be.

haraldh (approved):

lsinitrd.sh
danimo: We are not consistently combining `-L` with `-e`  or `-x` across the codebase. Alas, a bit of scrutiny won't hurt.

johannbg: True but fixing it is a scope outside this pull request and would be a part of bigger cleanup

danimo: Granted.

danimo (approved):

PR #1013, opened 2021-01-18
chore(remove): eliminate bootchart module

johannbg:
This pull request removes the bootchart module since afaikt upstream for it is no longer active and has not been for many years. We can introduce systemd-bootchart if there is demand for it but there does not seem to be much activity going on there either. I honestly just think that people are using systemd-analyze instead of bootchart these days...


haraldh (approved):

danimo (approved):

danimo:
Commisery is not happy. Can you fix that?

johannbg:
Throw a guy a freaking :bone: here geez oh well let's try eliminate instead of remove...

PR #1015, opened 2021-01-19
Small fixes after 051

zboszor:
## Changes

3 fixes are included:
- check kernel config from dracutsysrootdir instead of the host root
- 99squash use kernel config instead of modprobe to check modules (the shell function check_kernel_module is left without callers after this)
- inst_libdir_file didn't always work with dracutsysrootdir set (e.g. not for 95nfs that wants to install a set of files using file globbing via inst_libdir_file)

## Checklist

Fixes #


johannbg (approved):
You need to change your commit message to "fix: your text all lower case" and "fix(99squash): your commit message" to pass the commisery check otherwise LGTM

haraldh (approved):

PR #1016, opened 2021-01-19
fix(1001): use efivars fs over the deprecated sysfs entries

### LINKED ISSUE #1001: /sys/firmware/efi/vars is deprecated and gone on some platforms
**Describe the bug**
`/sys/firmware/efi/vars` is deprecated since a long time and kernel 5.10 dropped it for non-x86 systems, such as aarch64.
`/sys/firmware/efi/efivars` should now be used.
Kernel commit: https://github.com/torvalds/linux/commit/963fabf37f6a94214a823df0a785e653cb8ad6ea

I spotted 2 occurrences:
* https://github.com/dracutdevs/dracut/blob/master/modules.d/95fcoe-uefi/parse-uefifcoe.sh#L31
* https://github.com/dracutdevs/dracut/blob/master/modules.d/99uefi-lib/uefi-lib.sh#L43

**Distribution used**
openSUSE Tumbleweed

**Dracut version**
051

**Init system**
systemd

**To Reproduce**
N/A

**Expected behavior**
dracut should be updated to use the new path layout.

**Additional context**

danimo:
sysfs efi vars has been dropped for non-x86 systems since kernel 5.10.

## Checklist
- [x] I have tested it locally (just verified path equivalence, no boot test)

Fixes #1001


johannbg (approved):
LGTM

haraldh (approved):

PR #1019, opened 2021-01-20
feat(dbus-broker): introducing dbus-broker module

johannbg:
This pull request introduces dbus-broker and should be considered "experimental" at this stage hence forth known as
 :bee:  honeybus :bee: :beers: . I wont know how the split the of the systemd integration into separated modules will affect it so it might change/evolve over time. 

The difference between this module and the already existing ( but newly introduced ) dbus is *intentional*.


dracut.conf.d/fedora.conf.example
```diff
@@ -22,7 +22,7 @@ dbusconfdir=/etc/dbus-1
 dbusinterfacesconfdir=/etc/dbus-1/interfaces
 dbusservicesconfdir=/etc/dbus-1/services
 dbussessionconfdir=/etc/dbus-1/session.d
-dbussystem=confdir/etc/dbus-1/system.d
+dbussystemconfdir/etc/dbus-1/system.d
```
haraldh, line 25: I doubt this is what you wanted.

haraldh (changes requested):

dracut.conf.d/fedora.conf.example
johannbg: Fixed along with another thing I noticed that I had missed.

danimo (approved):

modules.d/01dbus-broker/module-setup.sh
```diff
@@ -0,0 +1,94 @@
...
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o \
+        $dbusconfdir/session.conf \
+        $dbusconfdir/system.conf \
+        $sysusersconfdir/dbus.conf \
+        $systemdsystemconfdir/dbus.socket \
+        $systemdsystemconfdir/dbus.socket.d/*.conf \
+        $systemdsystemconfdir/dbus-broker.service \
+        $systemdsystemconfdir/dbus-broker.service.d/*.conf
+        ${NULL}
```
haraldh, line 88: needs one more indent

modules.d/01dbus-broker/module-setup.sh
```diff
@@ -0,0 +1,94 @@
...
+    inst_dir    $dbusservices
+    inst_dir    $dbussession
+    inst_dir    $dbussystem
+    inst_dir    $dbussystemservices
+    inst_dir    $dbusconfdir
+    inst_dir    $dbusinterfacesconfdir
+    inst_dir    $dbusservicesconfdir
+    inst_dir    $dbussessionconfdir
+    inst_dir    $dbussystemconfdir
+    inst_dir    $dbussystemservicesconfdir
```
haraldh, line 58: why so much space after `inst_dir` ?

haraldh (changes requested):
Your commits should be logical and with a meaningful message. Please squash them to logical units.

I know, I was doing bad commits too in the past... but I regret them.

modules.d/01dbus-broker/module-setup.sh
johannbg: Tab vs Space

johannbg:
Is not the logical unit the introduction of the module? ( and I now have squashed everything beneath it )

haraldh:
> Is not the logical unit the introduction of the module? ( and I now have squashed everything beneath it )

Fixing `dracut.conf.d/fedora.conf.example` would be one commit, or even one separate PR.

Introducing the dbus-broker module is the other commit.

modules.d/01dbus-broker/module-setup.sh
haraldh: `.editorconfig` has `indent_style = space`

johannbg:
@haraldh roger I'll make separated PR for spec file changes, global vars and the modules themselves in the future. ( these PR can have a mix of these three ).

modules.d/01dbus-broker/module-setup.sh
johannbg: I should have been more clearer so it would not be confused with TAB spaces when I ( or someone else ) presses tab it creates 4 intent spaces, which is what you are seeing there.
```
# Match config files, set indent to spaces with width of four.
[*.sh]
indent_style = space
indent_size = 4 <---
```

The reason for this being there is simply because I think it's cleaner and more readable than having a single space. I can most certainly make a single space between ```inst_$FOO BAR``` if that is preferred and more correct.

johannbg:
This was addressed in #1046 hence closing

PR #1020, opened 2021-01-21
fix(kernel-network-modules): also install modules from mdio subdirectory

martinezjavier:
## Changes

A recent change in Linux kernel v5.10 split the MDIO drivers from the PHY
subdirectory into a new MDIO subdirectory [0]. This caused MDIO drivers
to not be included in the initramfs.

Also include the modules of this new subdirectory to prevent this issue.

[0]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a9770eac511

## Checklist

Fixes #

### DIFF
--- a/modules.d/90kernel-network-modules/module-setup.sh
+++ b/modules.d/90kernel-network-modules/module-setup.sh
@@ -37,6 +37,7 @@ installkernel() {
     dracut_instmods -o -P ".*${_unwanted_drivers}.*" -s "$_net_symbols" "=drivers/net"
     #instmods() will take care of hostonly
     instmods \
+        =drivers/net/mdio \
         =drivers/net/phy \
         =drivers/net/team \
         =drivers/net/ethernet \


johannbg (approved):
You need to fix your commit msg to "fix(kernel-network-modules): also install modules from mdio subdirectory" to pass the commisery check otherwise LGTM

martinezjavier:
> You need to fix your commit msg to "fix(kernel-network-modules): also install modules from mdio subdirectory" to pass the commisery check otherwise LGTM

Done. I just use the same subject prefix that was in the previous commit touching this file and assumed that was correct.

johannbg:
@martinezjavier yep passes :beers:

fda77:
Thanks @martinezjavier for the quick bugfix!

fda77:
Thx @martinezjavier for the quick fix!

haraldh (approved):

PR #1021, opened 2021-01-21
WIP: feat(iwd): introducing iwd module

johannbg:
This pull request introduces iwd module which should be considered "experimental" at this point and depends on currently un-merged module. 

Later I might split this module into two parts since ethernet authentication daemon ( EAD ) should probably reside in it's own own module. 

This PR also adds two new global variables which I noticed we lacked.
$modulesload --> /usr/lib/modules-load.d
$modulesloadconfdir --> /etc/modules-load.d


danimo:
Please call the directory `35network-idw` for consistency.

danimo (approved):

PR #1022, opened 2021-01-21
feat(systemd-modules-load): introducing systemd-modules-load module

johannbg:
Introducing systemd-modules-load which is an early boot service
that loads kernel modules from a static list, which is required for
kernel modules that do for example not support automatic module loading
( like key type parsers ).

This is a standalone systemd module which means it depends on no other module than systemd.

### DIFF
--- a/modules.d/01systemd-modules-load/module-setup.sh
+++ b/modules.d/01systemd-modules-load/module-setup.sh
@@ -1,0 +1,49 @@
+#!/bin/sh
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed
+    require_binaries $systemdutildir/systemd-modules-load || return 1
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+
+}
+
+# Module dependency requirements.
+depends() {
+
+    # This module has external dependency on the systemd module.
+    echo systemd
+    # Return 0 to include the dependent systemd module in the initramfs.
+    return 0
+
+}
+
+# Install the required file(s) for the module in the initramfs.
+install() {
+
+        # Create systemd-modules-load related directories.
+        inst_dir    $modulesload
+        inst_dir    $modulesloadconfdir
+
+        # Install related files for systemd-modules-load
+        inst_multiple -o \
+            $systemdsystemunitdir/systemd-modules-load.service \
+            $systemdutildir/systemd-modules-load
+
+        # Install local user configurations if host only is enabled..
+        if [[ $hostonly ]]; then
+            inst_multiple -H -o \
+            $systemdsystemconfdir/systemd-modules-load.service \
+            $systemdsystemconfdir/systemd-systemd-modules-load.d/*.conf \
+            ${NULL}
+        fi
+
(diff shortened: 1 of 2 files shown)


danimo (approved):
Reminder: This is currently partly redundant with the systemd module

johannbg:
> Reminder: This is currently partly redundant with the systemd module

They will be until all modules are place that can replace the current 00systemd module and 00systemd turned into a meta module aimed at full blown systemd. We should also at that point be able to introduce something like 00systemd-core which absolute minimum required for systemd functionality then vendors/users can include the rest of the systemd bits that are "optional"

Note I was cleaning up the commit message as well resolving merge conflict, hence requesting re-review.

danimo (approved):

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -0,0 +1,59 @@
...
+        # Create systemd-modules-load related directories.
+        inst_dir    $modulesload
+        inst_dir    $modulesloadconfdir
+
+        # Install related files for systemd-modules-load
+        inst_multiple -o \
+        # Install the systemd type service unit for systemd-modules-load.
+        $systemdsystemunitdir/systemd-modules-load.service \
+        # Install the binary executable(s) for systemd-modules-load.
+        $systemdutildir/systemd-modules-load
```
haraldh, line 46: Line continuation cannot be broken by comments

haraldh (changes requested):

modules.d/01systemd-modules-load/module-setup.sh
johannbg: Fixed along with removing the module check(s)

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+        # Install related files for systemd-modules-load
+        inst_multiple -o \
+            $systemdsystemunitdir/systemd-modules-load.service \
+            $systemdutildir/systemd-modules-load
+
+        # Install local user configurations if host only is enabled..
+        if [[ $hostonly ]]; then
+            inst_multiple -H -o \
+            $systemdsystemconfdir/systemd-modules-load.service \
+            $systemdsystemconfdir/systemd-systemd-modules-load.d/*.conf
```
haraldh, line 43: backslash still missing

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -0,0 +1,50 @@
...
+            ${NULL}
+        fi
+
+        # Enable the systemd type service unit for systemd-modules-load.
+        systemctl -q --root "$initdir" enable systemd-modules-load.service
```
haraldh, line 48: recently we approved s/systemctl/$SYSTEMCTL/g IIRC

haraldh (changes requested):

modules.d/01systemd-modules-load/module-setup.sh
johannbg: Fixed

modules.d/01systemd-modules-load/module-setup.sh
johannbg: Fixed

haraldh (approved):

PR #1028, opened 2021-01-25
WIP: feat(systemd-cryptsetup): introducing systemd-cryptsetup module

johannbg:
This pull request introduces the systemd-cryptsetup module. 

This module includes the relevant section from the crypt module which may or may not be the wanted behaviour ( we might want to depend on it instead ).
 
Since I was unsure I decided to included it ( easier to remove than add ). 

Once consensus has been reached on how the systemd-cryptsetup module should look like I will add the newly introduced  systemd veritysetup as a module as well since cryptsetup module and veritysetup module should be pretty identical.


modules.d/01systemd-cryptsetup/crypt-lib.sh
```diff
@@ -0,0 +1,237 @@
+#!/bin/sh
```
haraldh, line 1: `crypt-lib.sh` should no be duplicated .. better depend on the `crypt` module

modules.d/01systemd-cryptsetup/crypt-run-generator.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
```
haraldh, line 1: depend on `90crypt`

modules.d/01systemd-cryptsetup/module-setup.sh
```diff
@@ -0,0 +1,198 @@
...
+
+    if ! dracut_module_included "systemd-ask-password"; then
+        derror "systemd-cryptsetup needs systemd-ask-password in the initramfs."
+        return 1
+    fi
+
+    if ! dracut_module_included "systemd-tmpfiles"; then
+        derror "systemd-cryptsetup needs tmpfiles in the initramfs."
+        return 1
+    fi
```
haraldh, line 52: better rely on `depends()`

modules.d/01systemd-cryptsetup/module-setup.sh
```diff
@@ -0,0 +1,198 @@
...
+
+    # This module has external dependency on the systemd module.
+    echo dm rootfs-block systemd systemd-ask-password systemd-tmpfiles
+    # Return 0 to include the dependent systemd module in the initramfs.
+    return 0
+
+}
+
+# Install kernel required file(s) for the module in the initramfs.
+installkernel() {
```
haraldh, line 70: depend on `crypt` and remove this

modules.d/01systemd-cryptsetup/module-setup.sh
```diff
@@ -0,0 +1,198 @@
...
+                    # in non-hostonly mode
+                    hostonly= instmods $(for k in "$@"; do echo "crypto-$k";done)
+                fi
+        done
+        }
+    return 0
+}
+
+# called by dracut
+cmdline() {
```
haraldh, line 101: depend on `crypt` and remove this

modules.d/01systemd-cryptsetup/module-setup.sh
```diff
@@ -0,0 +1,198 @@
...
+                break
+            done
+            )
+        [[ ${UUID} ]] || continue
+        printf "%s" " rd.luks.uuid=luks-${UUID}"
+    done
+}
+
+# Install the required file(s) for the module in the initramfs.
+install() {
```
haraldh, line 120: depend on `crypt` and remove most of this... Maybe remove the systemd stuff from `90crypt` then

modules.d/01systemd-cryptsetup/parse-crypt.sh
```diff
@@ -0,0 +1,193 @@
+#!/bin/sh
```
haraldh, line 1: depend on `crypt` and remove the non-systemd part... and remove the systemd part from `90crypt`

haraldh (changes requested):

modules.d/01systemd-cryptsetup/module-setup.sh
johannbg: Module adjusted to depend on crypt.

modules.d/01systemd-cryptsetup/module-setup.sh
johannbg: Module adjusted to depend on crypt and it removed

modules.d/01systemd-cryptsetup/parse-crypt.sh
johannbg: Module adjusted to depend on crypt and systemd parts related to 90crypt removed.

modules.d/90crypt/module-setup.sh
```diff
@@ -143,20 +143,5 @@ install() {
     inst_simple "$moddir/crypt-lib.sh" "/lib/dracut-crypt-lib.sh"
     inst_script "$moddir/crypt-run-generator.sh" "/sbin/crypt-run-generator"
```
haraldh, line 145: `90crypt/crypt-run-generator.sh` should be moved to systemd also and `parse-crypt.sh` should be split in a cryptsetup and systemd version

haraldh (changes requested):

modules.d/90crypt/module-setup.sh
```diff
@@ -4,7 +4,7 @@
 check() {
     local _rootdev
     # if cryptsetup is not installed, then we cannot support encrypted devices.
-    require_any_binary $systemdutildir/systemd-cryptsetup cryptsetup || return 1
+    require_binaries cryptsetup || return 1
```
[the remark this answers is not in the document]
johannbg: Sorry I dont follow previously we required both 

require_any_binary $systemdutildir/systemd-cryptsetup cryptsetup <---  

I just adjusted it to require only cryptsetup

systemd cryptsetup/veritysetup only require the crypsetup library afaikt

modules.d/90crypt/module-setup.sh
johannbg: Hmm does this generator even work or is it necessary? rd.luks.allow-discards <--??? ( where did this come from ) Is it not supposed to be enough that users put rd.luks.options=discard to the kernel command line and the initramfs image doesn't include a etc/crypttab and systemd takes care of rest? Or to rephrase this how much in those scripts are relevant today?

modules.d/90crypt/module-setup.sh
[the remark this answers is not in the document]
haraldh: `require_any_binary` == one of them... not *all* of them

`require_binaries` == *all* of them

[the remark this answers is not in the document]
johannbg: Ah that explains why this has been broken downstream. ( users ended up with no cryptsetup (library) installed )

PR #1031, opened 2021-01-26
feat(systemd-sysctl): introducing systemd-sysctl module 

johannbg:
Introducing systemd-sysctl module which is an early boot
service that configures sysctl.

The systemd-sysctl module has dependency's on systemd and
systemd-modules-load module.

### DIFF
--- a/modules.d/01systemd-sysctl/module-setup.sh
+++ b/modules.d/01systemd-sysctl/module-setup.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed
+    require_binaries $systemdutildir/systemd-sysctl || return 1
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+
+}
+
+# Module dependency requirements.
+depends() {
+
+    # This module has external dependency on other module(s).
+    echo systemd systemd-modules-load
+    # Return 0 to include the dependent module(s) in the initramfs.
+    return 0
+
+}
+
+# Install the required file(s) for the module in the initramfs.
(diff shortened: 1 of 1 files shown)


danimo (approved):

modules.d/01systemd-sysctl/module-setup.sh
```diff
@@ -0,0 +1,49 @@
...
+        $systemdsystemunitdir/sysinit.target.wants/systemd-sysctl.service \
+        $systemdutildir/systemd-sysctl
+
+    # Install the hosts local user configurations if enabled.
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o \
+            /etc/sysctl.conf \
+            $sysctldconfdir/*.conf \
+            $systemdsystemconfdir/systemd-sysctl.service \
+            $systemdsystemconfdir/systemd-sysctl.service.d/*.conf
```
haraldh, line 42: backslash missing

modules.d/01systemd-sysctl/module-setup.sh
```diff
@@ -0,0 +1,49 @@
...
+            ${NULL}
+    fi
+
+    # Enable the systemd type service unit for sysctl.
+    systemctl -q --root "$initdir" enable systemd-sysctl.service
```
haraldh, line 47: $SYSTEMCTL

haraldh (changes requested):

modules.d/01systemd-sysctl/module-setup.sh
johannbg: Fixed

modules.d/01systemd-sysctl/module-setup.sh
johannbg: Fixed

haraldh:
Depends on https://github.com/dracutdevs/dracut/pull/1022 .. waiting with review until it is merged

haraldh (approved):

PR #1039, opened 2021-01-28
fix(nbd): change units service type to oneshot

### LINKED ISSUE #1032: NBD disconnects during switch_root
**Describe the bug**
Once `switch_root` starts, a bunch of errors are thrown that indicate a read error on the device `/dev/nbd0`, further investigation shows that nbd seems to decide to stop talking the server (`kernel: block nbd0: shutting down sockets`). If I break before `switch_root`, `/sysroot` is mounted correctly, and I can view files in it. But once I exit the shell, everything just falls apart.

**Distribution used**
Arch Linux (with a custom kernel)

**Dracut version**
`051`

**Init system**
SystemD

**To Reproduce**

- `/etc/dracut.conf` 
  ```sh
  dracutmodules+=" nbd network-manager base shutdown ssh-client systemd systemd-initrd dracut-systemd "
  compress="lz4"
  add_drivers+=" hid usbhid hid-generic xhci_pci "
  nofscks="yes"
  install_items+=" /bin/btrfs "
  uefi_stub="/usr/lib/systemd/boot/efi/linuxx64.efi.stub"
  uefi_splash_image="/usr/share/systemd/bootctl/splash-office.bmp"
  kernel_cmdline=" ip=dhcp root=nbd:192.168.254.20:kiosk:btrfs:rw,noatime,compress=lzo rd.timeout=1 rd.retry=2 rd.shell=0  rd.emergency=reboot nowatchdog quiet i915.fastboot=1 vga=current loglevel=3 rd.systemd.show_status=auto  rd.udev.log_priority=3 vt.global_cursor_default=0 pci=noaer nmi_watchdog=0 "
  reproducible="yes"
  filesystems="btrfs"
  hostonly="no"
  hostonly_cmdline="no"
  ```

- `cmdline`
  ```
  ip=dhcp netroot=nbd:192.168.254.20:kiosk rootflags=rw,noatime,compress=lzo rd.shell rd.timeout=10 rd.retry=2
  ```

- `build command`
  ```sh
  dracut /linux.efi -v -f --no-hostonly-cmdline -N --uefi --kver 5.9.8-arch1-3-rockpix+
  ```

**Expected behavior**
Clean boot.

**Additional context**

[`rdsosreport.txt`](https://github.com/dracutdevs/dracut/files/5876165/report.txt)

johannbg:
Recently upstream changed the nbd-client from using the ioctl() interface to use the netlink interface to configure the NBD device, an upstream change which required changes the upstream systemd units from Type=forking to Type=oneshot, this PR reflects that change. Further information about this upstream change can be found here https://github.com/NetworkBlockDevice/nbd/issues/91. That change should fix #1032 

In addition to the above change this PR ( now ) also changes /bin/bash to sh, adds a missing license header, adds documentation to the code and a missing host only section which should be a much more complete implementation of the module.

### DIFF
--- a/modules.d/95nbd/module-setup.sh
+++ b/modules.d/95nbd/module-setup.sh
@@ -1,40 +1,69 @@
-#!/bin/bash
+#!/bin/sh
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
 
-# called by dracut
+# Prerequisite check(s) for module.
 check() {
     local _rootdev
 
-    # if an nbd device is not somewhere in the chain of devices root is
+    # If an nbd device is not somewhere in the chain of devices root is
     # mounted on, fail the hostonly check.
     [[ $hostonly ]] || [[ $mount_needs ]] && {
         _rootdev=$(find_root_block_device)
         [[ -b /dev/block/$_rootdev ]] || return 1
         check_block_and_slaves block_is_nbd "$_rootdev" || return 255
     }
+
+    # If the binary(s) requirements are not fulfilled
+    # return 1 to not include the binary.
     require_binaries nbd-client || return 1
 
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+
+}
+
+# Module dependency requirements.
+depends() {
+
+    # This module has external dependency on other module(s).
+    echo network rootfs-block
+    # Return 0 to include the dependent module(s) in the initramfs.
     return 0
 }
 
-# called by dracut
-depends() {
-    # We depend on network modules being loaded
-    echo network rootfs-block
-}
-
-# called by dracut
+# Install the required kernel module(s) for the module.
 installkernel() {
     instmods nbd
 }
 
-# called by dracut
+# Install the required file(s) for the module.
 install() {
-    inst nbd-client
-    inst_hook cmdline 90 "$moddir/parse-nbdroot.sh"
-    inst_script "$moddir/nbdroot.sh" "/sbin/nbdroot"
-    if dracut_module_included "systemd-initrd"; then
-        inst_script "$moddir/nbd-generator.sh" $systemdutildir/system-generators/dracut-nbd-generator
-    fi
-    dracut_need_initqueue
-}
 
+    # Install the required binary for module
+    inst nbd-client
+    # Install the required hook for module
+    inst_hook cmdline 90 "$moddir/parse-nbdroot.sh"
+    # Install the required script for module
+    inst_script "$moddir/nbdroot.sh" "/sbin/nbdroot"
+
+    if dracut_module_included "systemd"; then
+        # Install dracuts systemd generator for module
+        inst_script "$moddir/nbd-generator.sh" $systemdutildir/system-generators/dracut-nbd-generator
+        # Install the type service unit
+        $systemdsystemunitdir/nbd@.service
+
+        # Install the hosts local user configurations if enabled.
+        if [[ $hostonly ]]; then
+            inst_multiple -H -o \
+                /etc/nbdtab \
+                $systemdsystemconfdir/*.mount \
+                $systemdsystemconfdir/nbd@.service \
+                $systemdsystemconfdir/nbd@.service.d/*.conf
+                ${NULL}
+        fi
+    fi
+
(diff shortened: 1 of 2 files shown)


haraldh:
Please extend the commit message with an explanation *Why* and how your change fixes the problems.

johannbg:
@haraldh  better?

haraldh:
Both commits don't have any message body, telling us why this commit was necessary.

modules.d/95nbd/module-setup.sh
```diff
@@ -1,40 +1,69 @@
...
+
+    # Install the required binary for module
     inst nbd-client
+    # Install the required hook for module
     inst_hook cmdline 90 "$moddir/parse-nbdroot.sh"
+    # Install the required script for module
     inst_script "$moddir/nbdroot.sh" "/sbin/nbdroot"
-    if dracut_module_included "systemd-initrd"; then
+
+    if dracut_module_included "systemd"; then
```
haraldh, line 50: the `nbd-generator.sh` is only useful, if "systemd-initrd" is used.

modules.d/95nbd/module-setup.sh
```diff
@@ -1,40 +1,69 @@
...
+        # Install the type service unit
+        $systemdsystemunitdir/nbd@.service
+
+        # Install the hosts local user configurations if enabled.
+        if [[ $hostonly ]]; then
+            inst_multiple -H -o \
+                /etc/nbdtab \
+                $systemdsystemconfdir/*.mount \
+                $systemdsystemconfdir/nbd@.service \
+                $systemdsystemconfdir/nbd@.service.d/*.conf
```
haraldh, line 62: Err `$systemdsystemconfdir/*.mount` should only be included, if it is needed for the root device, or forced added as a dracut device.

modules.d/95nbd/nbdroot.sh
```diff
@@ -112,7 +112,7 @@ if [ "$nbdport" -gt 0 ] 2>/dev/null; then
     if [ -z "$DRACUT_SYSTEMD" ]; then
         nbd-client "$nbdserver" $nbdport /dev/nbd0 $preopts $opts || exit 1
     else
-        systemd-run --no-block --service-type=forking --quiet \
+        systemd-run --no-block --service-type=oneshot --quiet \
```
haraldh, line 115: Why do you think `oneshot` is the solution? This should also go into the commit message.

```
$ fgrep 'Type=' /usr/lib/systemd/system/nbd@.service
Type=forking
```

johannbg: We always follow upstream not never whatever is shipped in downstream right? ( https://github.com/NetworkBlockDevice/nbd/blob/master/systemd/nbd%40.service.sh.in ).

modules.d/95nbd/module-setup.sh
johannbg: We always make the assumption that administrator knows what he's doing when he enables hostonly mode right? So in a sense we expect that the administration follows directions like these [1], then adjusts it for dracut in the environment he plans on using it right or? 
https://libguestfs.org/nbdkit-client.1.html#Easy-mounting-at-boot-time

haraldh: `boot-time` != `initrd`

`hostonly` mode is automatically enabled for e.g. Fedora and RHEL.

Here are some points why we can't blindly copy all `*.mount` in the initrd, which the admin configured:
1. it would need including all necessary kernel modules, necessary for the device
2. it would need proper network configuration in the initrd for network devices
3. the initrd should only prepare the root filesystem
4. if the mount point includes a starting SAN, which takes a long time to boot and which is not needed for the root filesystem, this would slow down stuff a lot needlessly
5. ....

modules.d/95nbd/nbdroot.sh
haraldh: So, what I did in dracut, was to check the version of the software, so that we can at least support small changes in behavior. Maybe we can switch the option dynamically, if the newer version is detected?

Then fade out support for older versions over time.

johannbg:
I'm just closing this reports since it's gone completely out of wack, to messy to be merged and seemingly only partially fixed the reporters original issue.

PR #1040, opened 2021-01-29
fix: do not set cmdline for uefi images unless asked

### LINKED ISSUE #1012: `--no-hostonly-cmdline` is ignored when `--uefi` is set
**Describe the bug**
`--no-hostonly-cmdline` is ignored when `--uefi` is set

**Distribution used**
Gentoo

**Dracut version**
51-r1 (includes patch for https://bugs.gentoo.org/760249)

**Init system**
systemd

**To Reproduce**
Use the following dracut config, the kernel_cmdline should not be necessary and additionally should be ignored when `no_hostonly_commandline="yes"` is set.

```
add_dracutmodules+=" zfs "
force_drivers+=" nvme zfs "
compress="zstd"

uefi_secureboot_cert="/etc/efikeys/db.crt"
uefi_secureboot_key="/etc/efikeys/db.key"
uefi_stub="/usr/lib/systemd/boot/efi/linuxx64.efi.stub"
uefi="yes"

kernel_cmdline="zfs.zfs_arc_max=3221225472 root=zfs:AUTO init=/usr/lib/systemd/systemd log_buf_len=2M psmouse.synaptics_intertouch=1 acpi_backlight=none ro"
no_hostonly_commandline="yes"
omit_dracutmodules+=" btrfs lvm qemu multipath i18n qemu-net lunmask resume "
omit_dracutmodules+=" plymouth "
```

**Expected behavior**
dracut does not set command line parameters when `no_hostonly_commandline="yes"` is set.

**Additional context**
There is no check for no_hostonly_commandline in this section of code.
https://github.com/dracutdevs/dracut/blob/c93df5e0ea18dabeeaa1955c0c66e315b3f3964a/dracut.sh#L2010-L2024 (permalinked from master as of `Mon Jan 18 03:51:11 AM UTC 2021`).  I'd suggest wrapping that whole section in a check, probably with another check at line 2033 (where it objcopies the cmdline into the image).

Is there a reason why a command line is always defined?  I generate a uefi binary in order to get signed kernel (the only way of doing so), but I do use systemd-boot to load the kernel images and do not use the generated unifed efi binary directly (meaning systemd-boot can pass kernel command line options).

Passing kernel command line options like this overrides those passed via an external bootloader (possibily by design).

prometheanfire:
The kernel does not allow the cmdline to be overridden if one is defined
in the uefi image.  This means that bootloaders like systemd-boot or
grub cannot manage boot options (editing at boot for debug being a use
case).  This patchset allows for no cmdline to be set (instead of force
setting one if none is provided).

Closes: https://github.com/dracutdevs/dracut/issues/1012

This pull request changes...

## Checklist

Fixes #1012

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2163,13 +2163,16 @@ if [[ $uefi = yes ]]; then
             [ -e "$conf" ] || continue
             printf "%s " "$(< "$conf")" >> "$uefi_outdir/cmdline.txt"
         done
-    else
-        do_print_cmdline > "$uefi_outdir/cmdline.txt"
     fi
-    echo -ne "\x00" >> "$uefi_outdir/cmdline.txt"
 
-    dinfo "Using UEFI kernel cmdline:"
-    dinfo $(tr -d '\000' < "$uefi_outdir/cmdline.txt")
+    if [[ $kernel_cmdline ]] || [[ $hostonly_cmdline = yes && -d "$initdir/etc/cmdline.d" ]]; then
+        echo -ne "\x00" >> "$uefi_outdir/cmdline.txt"
+        dinfo "Using UEFI kernel cmdline:"
+        dinfo $(tr -d '\000' < "$uefi_outdir/cmdline.txt")
+        uefi_cmdline="--add-section .cmdline="${uefi_outdir}/cmdline.txt" --change-section-vma .cmdline=0x30000"
+    else
+        uefi_cmdline=""
+    fi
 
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

prometheanfire:
Is there anything I need to do to help this move forward?

johannbg:
@prometheanfire  You could update your branch which will help merge this once the other reviewers find time to review this.

haraldh (approved):

PR #1041, opened 2021-01-30
sysroot fixes

zboszor:
The file pattern passed to inst_multiple (if not quoted) is immediately substituted by the shell.
When dracutsysrootdir is set and the same package is not installed on the host then the files are obviously not found on the host system. dracut-install complains loudly in this case.

This can also create an issue if the package versions on the build host and the cross-compiled system differ in their file lists. In this case the file list is generated according to the host and not according to the sysroot.

Found during a test build on Yocto with wicked added for testing.

Handle file globbing in `dracut-install`'s `install_all()` function.

Fixed modules that were using file globbing with inst_multiple: 35network-wicked, 90kernel-modules, 90nvdimm, 95fcoe, 95udev-rules

Also fixed the dbus module so it uses dbus or messagebus UID/GID from dracutsysrootdir.

EDIT: Further fixes:
1. don't set PATH (valid for host) from NPATH (valid for sysroot) if dracutsysrootdir is set
2. use find_binary instead of type -P for modules' check and install functions in module-setup.sh
3. allow overriding the systemctl command. Yocto has its own Python script masquerading as systemctl but it lacks the `add-wants` subcommand that dracut needs.

## Checklist

Fixes #

### DIFF
--- a/dracut.8.asc
+++ b/dracut.8.asc
@@ -599,6 +599,9 @@ _SYSTEMD_VERSION_::
 _SYSTEMCTL_::
     overrides the systemctl binary. Used for **--sysroot**.
 
+_NM_VERSION_::
+    overrides the NetworkManager version. Used for **--sysroot**.
+
 _DRACUT_INSTALL_PATH_::
     overrides **PATH** environment for **dracut-install** to look for
     binaries relative to **--sysroot**. In a cross-compiled environment
--- a/dracut.8.asc
+++ b/dracut.8.asc
@@ -596,6 +596,9 @@ Default:
 _SYSTEMD_VERSION_::
     overrides systemd version. Used for **--sysroot**.
 
+_SYSTEMCTL_::
+    overrides the systemctl binary. Used for **--sysroot**.
+
 _DRACUT_INSTALL_PATH_::
     overrides **PATH** environment for **dracut-install** to look for
     binaries relative to **--sysroot**. In a cross-compiled environment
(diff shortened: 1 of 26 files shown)


haraldh:
Good point.. I would rather solve the problem with `dracut_install` accepting a file glob option and let `dracut_install` do the globbing with the sysroot.

zboszor:
Thanks. I will look into modifying dracut-install.

johannbg:
The scope of the commit message should be the module name (without numbers) so please remove that.

zboszor:
> The scope of the commit message should be the module name (without numbers) so please remove that.

Fixed.

johannbg (changes requested):
Could you squash the commits together

zboszor:
> Could you squash the commits together

Which ones? All of them into a single one or per-topic?

johannbg:
> > Could you squash the commits together
> 
> Which ones? All of them into a single one or per-topic?

One per topic.

johannbg (approved):
LGTM

danimo (approved):

danimo:
@zboszor Can you fix the conflict? `modules.d/06dbus/module-setup.sh` is gone and its replacement files do not need the patches.

zboszor:
> @zboszor Can you fix the conflict? `modules.d/06dbus/module-setup.sh` is gone and its replacement files do not need the patches.

Done.

zboszor:
> > @zboszor Can you fix the conflict? `modules.d/06dbus/module-setup.sh` is gone and its replacement files do not need the patches.
> 
> Done.

It's not quite true that the replacement module (06dbus-daemon) didn't need changes. It was still getting the dbus/messagebus UID/GID from the host instead of the sysroot.

PR #1042, opened 2021-02-01
feat(systemd-coredump): introducing systemd-coredump module

johannbg:
This pull request introduces the systemd-coredump module and adds a missing global variable ( systemdutilconfdir )

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1109,6 +1109,10 @@ if ! [[ -d "$dracutsysrootdir$systemdutildir" ]]; then
     [[ -e $dracutsysrootdir/usr/lib/systemd/systemd-udevd ]] && systemdutildir=/usr/lib/systemd
 fi
 
+[[ -d $dracutsysrootdir$systemdutilconfdir ]] \
+    || systemdutilconfdir=$(pkg-config systemd --variable=systemdutilconfdir 2>/dev/null)
+
+[[ -d "$dracutsysrootdir$systemdutilconfdir" ]] || systemdutilconfdir=/etc/systemd
 
 if [[ $no_kernel != yes ]] && [[ -d $srcmods ]]; then
     if ! [[ -f $srcmods/modules.dep ]]; then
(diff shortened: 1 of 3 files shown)


dracut.sh
```diff
@@ -1600,8 +1604,8 @@ export initdir dracutbasedir \
     dbusservices dbusservicesconfdir dbussession dbussessionconfdir \
     dbussystem dbussystemconfdir dbussystemservices dbussystemservicesconfdir \
     environment environmentconfdir sysctl sysctlconfdir sysusers sysusersconfdir \
-    systemdutildir systemdcatalog systemdntpunits systemdntpunitsconfdir \
-    systemdsystemunitdir systemdsystemconfdir \
+    systemdutildir systemdutilconfdirsystemdcatalog systemdntpunits \
+    systemdntpunitsconfdir systemdsystemunitdir systemdsystemconfdir \
```
haraldh, line 1608: are there some spaces missing?

haraldh (changes requested):

dracut.sh
johannbg: Yep well spotted, fixed.

modules.d/01systemd-coredump/module-setup.sh
```diff
@@ -0,0 +1,77 @@
...
+            inst_multiple -H -o \
+            $systemdutilconfdir/coredump.conf \
+            $systemdsystemconfdir/coredump.conf.d/*.conf
+            $systemdsystemconfdir/systemd-coredump.socket \
+            $systemdsystemconfdir/systemd-coredump.socket.d/*.conf \
+            $systemdsystemconfdir/systemd-coredump@.service \
+            $systemdsystemconfdir/systemd-coredump@.service.d/*.conf \
+            $systemdsystemconfdir/sockets.target.wants/systemd-coredump.socket
+            ${NULL}
+        fi
```
haraldh, line 75: nitpicking about the indention... is this mixed tabs and spaces? Or why is the `if` block not in line with `inst_multiple`?

johannbg:
There is no particular reason for this other than it looked good in the eyes of the beholder at the time of writing :smile: but this indeed is more cosmetically correct as you mention. Note that I will need to fix this cosmetic issue probably in the other modules as well.

haraldh:
would you mind squashing the fix commits to the commits introducing the errors?

danimo:
@haraldh can you acknowledge the changes?

haraldh:
the commit message is still bogus

johannbg:
@haraldh admittedly not perfect but good enough no?

haraldh (approved):

PR #1043, opened 2021-02-02
fix(watchdog-modules): numeric argument required

Peter-Levine:
Fix error `/usr/lib/dracut/modules.d/04watchdog/module-setup.sh: line 10: return: watchdog-modules: numeric argument required`.

### DIFF
--- a/modules.d/04watchdog/module-setup.sh
+++ b/modules.d/04watchdog/module-setup.sh
@@ -7,7 +7,8 @@ check() {
 
 # called by dracut
 depends() {
-    return "watchdog-modules"
+    echo watchdog-modules
+    return 0
 }
 
 # called by dracut


johannbg (changes requested):
You are missing a "return 0" to include the dependent systemd module in the initramfs. 
Also remove the "" surrounding the module.

A typical depend section should look like this.

```
depends() {
    # This module has external dependency on the <module-name> module.
    echo <module-name>
    # Return 0 to include the dependent <module-name> module in the initramfs.
    return 0
}
```

johannbg:
We also do not include the module number, just the module name in the commit message so please fix that.

johannbg:
@Peter-Levine you forgot to rebase your branch

johannbg (approved):

haraldh (approved):

danimo (approved):

PR #1044, opened 2021-02-03
fix(dbus): various fixes

### LINKED ISSUE #1025: dbus-broker and dbus modules should conflict
See https://github.com/dracutdevs/dracut/pull/1019

johannbg:
This pull request removes any reference to the dbus-broker since it's now it's own module and adds a hard conflict with it ( and dbus-broker is doing the same ) which should resolve #1025, this also replaces user/group grep with systemd-sysuser since dbus provides it thus introduces a dependency on systemd-sysusers module.

### DIFF
--- a/modules.d/06dbus/module-setup.sh
+++ b/modules.d/06dbus/module-setup.sh
@@ -1,70 +1,114 @@
-#!/bin/bash
+#!/bin/sh
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
 
-# called by dracut
+# Prerequisite check(s) for module.
 check() {
-  require_binaries dbus-daemon || return 1
 
-  return 255
+    # If the binary(s) requirements are not fulfilled
+    # return 1 to not include the binary.
+    require_binaries busctl || return 1
+    require_binaries dbus-daemon || return 1
+    require_binaries dbus-send || return 1
+
+    # If the module dependency requirements are not fulfilled
+    # return 1 to not include the required module(s).
+    if ! dracut_module_included "systemd"; then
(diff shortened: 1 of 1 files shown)


modules.d/06dbus/module-setup.sh
```diff
@@ -1,70 +1,114 @@
...
-  echo systemd
-  return 0
+    # If the module dependency requirements are not fulfilled
+    # return 1 to not include the required module(s).
+    if ! dracut_module_included "systemd"; then
+        derror "dbus needs systemd in the initramfs."
+        return 1
+    fi
+
+    if ! dracut_module_included "systemd-sysusers"; then
```
danimo, line 21: We cannot mandate sysusers here. Not all distros use it atm. If you want that, please add a logic like "if ! dracut_module_included "systemd-sysusers"; then add users manually"

modules.d/06dbus/module-setup.sh
```diff
@@ -1,70 +1,114 @@
...
+    if dracut_module_included "dbus-broker"; then
+        derror "dbus conflicts with dbus-broker in the initramfs."
+        exit 1
+    fi
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
 }
 
-adjust_dependencies() {
```
danimo, line 15: Why remove this? IIRC it's needed.

danimo (changes requested):

modules.d/06dbus/module-setup.sh
danimo: Even better: check how the distro injects system users...

PR #1045, opened 2021-02-03
fix(99squash): don't throw errors when requirements aren't met

ahesford:
Failing to meet requirements for for 99squash should trigger info messages, not errors that are printed even in quiet mode.

This pull request changes...

## Changes
- `derror` to `dinfo` in the `check` script for `99squash`

## Checklist

### DIFF
--- a/modules.d/99squash/module-setup.sh
+++ b/modules.d/99squash/module-setup.sh
@@ -2,18 +2,18 @@
 
 check() {
     if ! dracut_module_included "systemd-initrd"; then
-        derror "dracut-squash only supports systemd bases initramfs"
+        dinfo "dracut-squash module requires systemd-based initramfs"
         return 1
     fi
 
     if ! type -P mksquashfs >/dev/null || ! type -P unsquashfs >/dev/null ; then
-        derror "dracut-squash module requires squashfs-tools"
+        dinfo "dracut-squash module requires squashfs-tools"
         return 1
     fi
 
     for i in CONFIG_SQUASHFS CONFIG_BLK_DEV_LOOP CONFIG_OVERLAY_FS ; do
         if ! check_kernel_config $i; then
-            derror "dracut-squash module requires kernel configuration $i (y or m)"
+            dinfo "dracut-squash module requires kernel configuration $i (y or m)"
             return 1
         fi
     done


johannbg (commented):
Failing to meet requirements are errors not informative messages and should always be printed even in quiet mode. 

That said if we would wanted to change that, we will have to change that throughout the entire code not just change that behavior in a single module.

ahesford:
The vast majority of other modules are either silent in `check` when they return 1 or 255 or, in the case of `99memstrack`, provide informational messages when they return nonzero.

The place for a hard failure is in `install`, which is called when a module is to be installed, not `check`, which is called to determine eligibility and is run *even when the squash module is not explicitly requested in the configuration*.

An error is not appropriate in this use case.

ahesford:
On closer inspection, a lot of the "silent" `check` functions invoke `require_binaries` to determine eligibility, which prints a `dinfo`, not a `derror`, when the necessary components are missing. The change here is consistent with precedent.

johannbg:
It's should be consistent behavior throughout the code, which requires a larger scope to fix once consensus has been reached what that fix should be. @haraldh, @danimo your take on this.

ahesford:
It's not clear to me what "consensus" you mean. The consensus is currently reflected in the behavior of **every other** `module-setup.sh` shipped with dracut to either keep `check` totally silent or, mostly by virtual of relying on `require_binaries`, print via `dinfo` when requirements are not met, regardless of whether `check` returns 0, 1 or 255.

Aside from the behavior I recommend changing here, the only other `check`  function that prints `derror` is from `95ssh-client`, and the error is only printed when the user has *explicitly specified* an SSH key path but that path does not exist.

johannbg:
@ahesford the consensus just means how it should be, not what the current state is and then the code should be fixed to reflect that. For example it's a consensus that we should always be using "/bin/sh" instead of "/bin/bash" throughout the code, the reality is that there 154 instance in which we are using "/bin/bash" and that means there are 154 places in the code we need to fix, for the code to be consistent and reflect that consensus.

The reason why I'm using bash here as an example is because, there is a dependency on bash module in the squash module but no check() for it. which is a bug.

When a module does not meet the requirement that outlined for it to work, that's an error not an informative message otherwise it would not be a requirement in the first place. which how I view it and why I personally NACK this PR and argue that we should fix it in all the other places for the exact same reason but @haraldh, @danimo might agree with your point of view so we just wait for their input and then either merge this ( and fix this in all other places to reflect that ) or we dont merge this ( and fix this in all other places to reflect that ).

ahesford:
I think the relevant facts that are getting lost in this discussion are:
1. The `squash` module is neither required for operation nor automatically added to images.
2. My configuration, which does not change relevant defaults, does not request the `squash` module.

If all optional and unrequested modules are made to throw errors for failing to meet their requirements, dracut output in quiet mode will become so noisy the most likely effect will be to condition users to ignore all output and miss legitimate functional errors.

johannbg:
Interesting that's definitely a bug ( which needs fixing elsewhere regardless of the outcome here ) Is the error still displayed if you specifically omit the module?

ahesford:
`omit_dracutmodules+=" squash "` seems to circumvent `check`; no errors are printed. This seems reasonable; why bother checking module installability if the user expressly forbids a module?

I think the crux of the issue here is that `check` is trying to do double duty by answering both "Should I be installed?" and "Can I be installed?". Getting the right behavior either means rearchitecting everything to split `check` into two separate that each addresses one of these questions, or moving the "Can I be installed?" questions into `install` where hard failures are appropriate.

johannbg:
Modules should never decide if they are supposed to be installed or not, that's the architectural flaw anyway what is happening here is that dracut is parsing every module as it is configured with "return 0" when it fact it should only be parsing those that have explicitly declared "return 0" in the check() section + what ever module is listed to be included in configuration files. By default modules should be either treated as with "return 255" or not considered to have any default et all.

ryncsn:
That's an issue I never thought about... and right, this module should never be enabled automatically, in check it will return 255 even if all requirement  are met, it shouldn't be so noisy. This change looks good to me.

> Modules should never decide if they are supposed to be installed or not
This is a problem with this, dracut have a "hostonly" mode and in that mode, each module will check if it is needed, and that is performed in check().
The only problem with reusing check() so far seem to be check() being too noisy?

johannbg:
@ryncsn modules that fail to meet requirements should return error. This PR is just a workaround a (legacy) bug that buried somewhere in the code. Vendors( distribution) or users ( administrators ) should explicitly have to declare what should be installed. That said this is potentially such legacy bug suspect. 
https://github.com/dracutdevs/dracut/blob/4026cd3b012be7f06c7572bdf50dd54a18ecb684/dracut-init.sh#L619

ryncsn:
> @ryncsn modules that fail to meet requirements should return error. This PR is just a workaround a (legacy) bug that buried somewhere in the code. Vendors( distribution) or users ( administrators ) should explicitly have to declare what should be installed. That said this is potentially such legacy bug suspect.

Right, I think I get your point. btw I just reedited the comment above, I just quoted wrong context.

The problem with "Vendors or users should explicitly have to declare what should be installed" and "Modules should never decide if they are supposed to be installed" is that: 

Dracut have a feature to automatically detect if a module is required for the target initramfs (hostonly mode, and the "mount_needs" flag), many modules can detect if any device needs the module to do the mount work, and "return 0" in check() if needed. (return 255/1 if unsure or optional).

And it have been reusing check() to do this, so the check() call seems unavoidable unless we split it into different function, maybe splitting the detection is a good idea. But the only problem so far seem only check() being too noisy, I'm not sure if this worth a refactor.

johannbg:
This needs to be properly fixed not hacked around by changing an error msg for a failed requirement for a module to informative msg. 

The reason why dracut has arguably ended up like this in the first place is because endless hacks and or workaround like this PR have been piled on top of each other year after year for close to a decade now. 

People will need to learn to give themselves the time to look for and properly fix underlying bugs and sometimes that requires small changes ( which may be the case here, as in change the "default" to 255 instead of 0 and add "return 0" to the modules that are absolute core requirements,or remove bugs that blindly include everything ) and sometimes it requires re-factoring. 

"Dracut have a feature to automatically detect if a module is required for the target initramfs" 
Any auto-detection features should just work in the exact same manner as a vendor or and administrator would install,enable modules. 

Seriously we need to be able to introduce modules without them suddenly being installed and "enabled" on end system, wasting hosts resources being parsed through when those modules aren't even enable.

ahesford:
> Seriously we need to be able to introduce modules without them suddenly being installed and even treated as they are "enabled" on end system, wasting hosts resources being parsed through when those modules aren't even enable or used.

This argument makes no sense.

Modules aren't "suddenly being installed and even treated as they are 'enabled'". It **runs the check script to determine by return code whether it is needed**. This is used in almost every single module. Core modules return 0 when they are always needed. Every other module tends to return 255 to say "only when requested by the user" or 1 to say "no matter what, this is not installable". `squash` always wants to return 255, but now does some requirements checks to see if it *could* be installed **if the user wants it**.

**I don't ask for this module to be installed.** It is not a bug to call `check` to see if it's required anyway. However, it *is* a bug that it elevates these missing requirements to an error when their absence does not change the outcome. Even if they were satisfied, `squash` would not be built into my initramfs using my configuration.

If you want to redesign dracut to split `check` in two to separate the "is this required either by default" question from the "are its requirements met" question, then do so. In the meantime, with the current design constraints, the right behavior is to demote the significance of these messages. Anything else is closer to the "suddenly enabled" state to which you object.

johannbg:
"It is not a bug to call check to see if it's required" 
Yes it is, the ```check()``` should have never been used for this. 
An separated ```include()``` and or ```install()``` should have handled this logic while ```check()``` would only be used to checked if the prerequisites for the module to work had been met. 

What I'm arguing here is that the squash modules should not even be *parsed* in the first place unless any of the following conditions have been met...
a) Dracut automatically detected if the module was required for the target initramfs 
b) The module is being included (enabled) as dependency of another module
c) The module has been included (enabled) by the vendor.
d) The module has been by included(enabled) user/administrator.

If any of the above was true, the squash module would never have been parsed and you would have never experiencing the output of ```derror```. 

The ```derror``` in return 1, in ```check()``` section is correct and accurate behavior for it's intent and purpose and should not be demoted to ```dinfo``` since if the module is being *included* then it should trigger an error if any of it's *requirement* checks fails. 

Arguably it should even go as far as the script exiting if an requirement is not met.

johannbg:
On top of that, the squash module has an prerequisite on systemd-init which obviously requires systemd as an init system so it ( along with every other systemd based module ) will fail to meet it's requirement and trigger an error for any user on a non systemd based distribution if the downstream package maintainer there is not doing his or hers due diligence and excluding any systemd based dracut modules from that distribution. Meanwhile those that are on systemd based distribution will meet that requirement and never see or otherwise experience that error in the squash module.

ahesford:
> An separated include() and or install() should have handled this logic while check() would only be used to checked if the prerequisites for the module to work had been met.

That's what virtually every module currently uses `check` for, and nothing more. If you want to add more reliability, create a new interface, `verify_installable` and use that. Otherwise, put the hard errors in `install` and throw a hard error at the time installation is actively attempted.

Your proposal to break every module by changing the common use of `check` and then add that same functionality back under a different name is ludicrous. I'm all for more sanity checks during the install---dracut gets a lot wrong without catching obvious errors---but do it the sensible way, by adding that currently missing functionality *under a new name* rather than pointlessly shuffling everything.

> On top of that, the squash module has an prerequisite on systemd-init which obviously requires systemd as an init system so it ( along with every other systemd based module ) will fail to meet it's requirement and trigger an error for any user on a non systemd based distribution if the downstream package maintainer there is not doing his or hers due diligence and excluding any systemd based dracut modules from that distribution.

Nonsense. You've already pointed out that dracut couldn't care less about anything that fails to use systemd in other issues I've raised. That's of no consequence here. The unexpected error will also appear if squashfs utilities are not installed. Again, this belongs in `install` under the current implementation, not the auto-install check.

Trying to reconcile the dependency graph of dracut modules when packaging would be unnecessary if dracut did the right thing and silently inhibited automatic installation when it knows at runtime that the module is neither required nor requested, and only throw errors when installation is actually attempted with missing requirements. It also doesn't solve the problem for Gentoo, which provides alternatives but can also work with systemd. What is the right packaging behavior there?

johannbg:
> Nonsense. You've already pointed out that dracut couldn't care less about anything that fails to use systemd in other issues I've raised. That's of no consequence here. The unexpected error will also appear if squashfs utilities are not installed. Again, this belongs in `install` under the current implementation, not the auto-install check.

It does not matter if it's systemd or something else. I was just pointing out that under normal circumstance the squash module should not even be installed on non systemd based system, same thing applies with NetworkManager vs Wicked etc. it serves absolutely no point in installing modules that use those specific components ( or architecture) on a distribution that dont support/have those components (  or architecture ) etc. 

> 
> Trying to reconcile the dependency graph of dracut modules when packaging would be unnecessary if dracut did the right thing and silently inhibited automatic installation when it knows at runtime that the module is neither required nor requested, and only throw errors when installation is actually attempted with missing requirements. It also doesn't solve the problem for Gentoo, which provides alternatives but can also work with systemd. What is the right packaging behavior there?

Components should have proper and correct packaging and dependency always. I personally have had to clean up after sloppiness of package maintainers that had decided that the future was frozen in time so they did not even bother to depend on core component in the core OS stack, from the component they maintained.  They just somehow felt or otherwise expected that relevant component in question would always be there,that the evolution somehow did not apply to the linux ecosystem, that things would be frozen in time and then when I searched for dependency on that relevant component, those components of course did not have the any, thus they fell out of the scope and things started to break.

We already do support multiple init systems in some modules ( or rather check if systemd module is in use and take action based on that ) which is one approach but from my point of view ( if it must be supported ) that those alternative init systemd should get a ```00<their-init-name>``` module and those module supporting that init system should be separated into their own modules, and depend on that 00init module, which should help downstream packagers and us here to diagnose issues and dealing with bugs that are a result of a different behavior between init system and the code + It would give us something tangible to measure in times of cost vs usage but that's just my point of view.  

If the intent is to support multiple init system then it should be done properly, with proper isolation, test cases for each and every one of them, the whole shebang and if that cant be done, then there is no point in trying since it serves no good treating alternative init system as somekind of second class citizen. It does not benefit the project nor the consumers of said init system from my point of view.

And yes i'm no particular fan of supporting multiple init systems and I neither sugar coat nor hide that fact since I've seen plethora of components forked by the ```E<fork>``` people who then mutilate the component in question and then literally direct people upstream to solve issues related to the component they forked, costing upstreams countless of hours if not days, trying to debug issues in their component, only to later discover the person was not using their component but a fork ( that had in some cases, ripped out core functionality of said component).  

The most recent sample I noticed of such component that is directly, directing it's users base upstream is here [1]( bottom line, direct link to upstream iwd and that's his second ```E<fork>``` of iwd ) apparently iwd dbus support is root of all evil this time which begs the question will we see a fork and issues associated with that fork here upstream since we are about land a dbus support.

If people intend to fork a component regardless what it is ( which is fine and dandy ) they *also* carry the burden of having to maintain that fork from that point on, not just dump/redirect all the burden upstream and or expect downstream maintainers ( like in Debian) having to deal with it. 

On top of that, these ```E<fork>``` peoples behavior most certainly are doing the people who genuinely are trying to write an alternative init system solution to compete with systemd and advance the linux ecosystem in the process, a disservice with their actions. 

Dealing with these issues always cost time ( and money for those company that are sponsoring work on the project.) so I personally feel that our limited time is better spend focusing on single init system ( systemd along with those 215+ distribution that use it ) and try to support that well, along with systemd based modules where other upstream support systemd well. 
As thing stand now we cant be everything to everybody and I honestly dont see the resources available in the project for that. That's my opinion and assessment, other people might and do feel differently and have different assessments. 

Anyway the time we spend on fixing issues like this should be spent on a permanent solution, whatever that solution ends up being not be plastered or otherwise implement in some form of a workaround to buy some kind of an fictional time until it breaks again in a different way. We have to get the project out of that rabbit hole.

1. https://github.com/dylanaraps/eiwd

ericonr:
Hi! Full disclaimer, I am another Void dev.

On the specific matter that's being solved in this PR, I believe blocking it,
when it makes things better (in that it follows the current dracut style for
enabling a certain module or not), because it doesn't fit the ideal dracut
design (which isn't implemented yet), is not a good idea. There should
definitely be an open issue (maybe even pin it and tag it appropriately? helps
a lot with discovery) about the re-structuring of dracut that you believe is
necessary, but until that is ready, there isn't, in my view, a good reason to
carry the bug forward into the future.

This PR was made because ahesford noticed a new warning when he was testing the
dracut update for Void Linux, and thought it best to silence the warning
instead of having users worry about the dracut output when they updated their
systems. Instead of carrying multiple patches that distance us from what
upstream provides, he thought it best to open a PR.

Regarding the other issues you have brought up, I'd like to comment on a few of
them.

> It does not matter if it's systemd or something else. I was just pointing out that under normal circumstance the squash module should not even be installed on non systemd based system, same thing applies with NetworkManager vs Wicked etc. it serves absolutely no point in installing modules that use those specific components ( or architecture) on a distribution that dont support/have those components ( or architecture ) etc.

If you strongly believe this to be the case, it should be made extremely clear
in the dracut packaging documentation. As is, dracut modules take care of
determining if the dependencies for a certain module are installed or not, and
points them out to the user if they have requested a certain module but the
dependencies aren't available. If you believe this work should instead be done
by distributions, you are requiring that they either:

- split dracut into multiple sub packages, each one pulling in the dependencies for the dracut modules that they include (Void does do this for the network support inside dracut, but mostly as a convenience for users, so they don't have to manually find out what dependencies are required - other more obscure modules will be used by savvy users, who can figure out what they need). Users will have to research specifically which packages to install in before they can add a module to their dracut config.
- also split dracut into multiple sub packages, which are pulled in conditionally when a user has both dracut and the dependencies of the sub package installed. This obviously requires that they implement such a mechanism in their package manager, which, afaik, few (if any?) do.

To make life easier for distributions and to avoid duplicating work, ideally,
dracut would then document, in a clean manner, the requirements for each of
their modules and which ones go together, at which point, the work currently
being done in `check` for each module would be repeated, just elsewhere, but
with added complexity for packagers and with the disadvantage that
documentation will usually live in a separate place from the module code.

> Components should have proper and correct packaging and dependency always. I personally have had to clean up after sloppiness of package maintainers that had decided that the future was frozen in time so they did not even bother to depend on core component in the core OS stack, from the component they maintained. They just somehow felt or otherwise expected that relevant component in question would always be there,that the evolution somehow did not apply to the linux ecosystem, that things would be frozen in time and then when I searched for dependency on that relevant component, those components of course did not have the any, thus they fell out of the scope and things started to break.

I'm not sure what you mean here; do you have a specific example?

> We already do support multiple init systems in some modules ( or rather check if systemd module is in use and take action based on that ) which is one approach but from my point of view ( if it must be supported ) that those alternative init systemd should get a 00<their-init-name> module and those module supporting that init system should be separated into their own modules, and depend on that 00init module, which should help downstream packagers and us here to diagnose issues and dealing with bugs that are a result of a different behavior between init system and the code + It would give us something tangible to measure in times of cost vs usage but that's just my point of view.

We don't include any other init inside dracut, just generating an initramfs
without systemd in the system is enough for everything to work fine. If you
wish, I can look into adding a Void (or Alpine? I'm not sure they ship dracut)
target for CI to run on, but again, it would be nice to have open issues about
such matters, so we can see where help is needed and where dracut is aiming to
move. It should help people from outside the project to fit their expectations
better, and also avoid wasting your time on writing the same answers
everywhere.

> And yes i'm no particular fan of supporting multiple init systems and I neither sugar coat nor hide that fact since I've seen plethora of components forked by the E<fork> people who then mutilate the component in question and then literally direct people upstream to solve issues related to the component they forked, costing upstreams countless of hours if not days, trying to debug issues in their component, only to later discover the person was not using their component but a fork ( that had in some cases, ripped out core functionality of said component).

From this point forward, I believe you are strongly misrepresenting such
projects. Even in cases where systemd is not a possibility, such as for users
of musl or uclibc, some systemd components are still useful, which is why
projects such as elogind and eudev exist. In my experience dealing with them,
they are extremely clear on what they provide and where issues should be
reported (hint: not to systemd).

eiwd is a different project, forked from a different upstream, whose issues
with iwd stemmed from a dislike of dbus, not anything to do with a specific
init system; furthermore, such a project hasn't been packaged anywhere besides
the creator's own distro and gentoo's GURU overlay, which makes it clear, at
least to me, that it isn't representative of the other `e*` projects.  However,
I have no experience with this project, and the issue might just have been a
slightly more lost than usual user. While the situation and loss of time is
unfortunate, I'd hope your opinion of other `e*` projects wouldn't be based on
an incident with a user of an unnafiliated project.

johannbg:
> Hi! Full disclaimer, I am another Void dev.
> 
> On the specific matter that's being solved in this PR, I believe blocking it,
> when it makes things better (in that it follows the current dracut style for
> enabling a certain module or not), because it doesn't fit the ideal dracut
> design (which isn't implemented yet), is not a good idea. There should
> definitely be an open issue (maybe even pin it and tag it appropriately? helps
> a lot with discovery) about the re-structuring of dracut that you believe is
> necessary, but until that is ready, there isn't, in my view, a good reason to
> carry the bug forward into the future.

If we continue to delay issues like these or paper over them they never get fixed because there is always something more urgend around the corner which is why I'm of the opinion we have to start somewhere and now is as good time as any.

> This PR was made because ahesford noticed a new warning when he was testing the
> dracut update for Void Linux, and thought it best to silence the warning
> instead of having users worry about the dracut output when they updated their
> systems. Instead of carrying multiple patches that distance us from what
> upstream provides, he thought it best to open a PR.

Which is fine and appreciated.

> Regarding the other issues you have brought up, I'd like to comment on a few of
> them.
> 
> > It does not matter if it's systemd or something else. I was just pointing out that under normal circumstance the squash module should not even be installed on non systemd based system, same thing applies with NetworkManager vs Wicked etc. it serves absolutely no point in installing modules that use those specific components ( or architecture) on a distribution that dont support/have those components ( or architecture ) etc.
> 
> If you strongly believe this to be the case, it should be made extremely clear
> in the dracut packaging documentation. As is, dracut modules take care of
> determining if the dependencies for a certain module are installed or not, and
> points them out to the user if they have requested a certain module but the
> dependencies aren't available. If you believe this work should instead be done
> by distributions, you are requiring that they either:
> 
>     * split dracut into multiple sub packages, each one pulling in the dependencies for the dracut modules that they include (Void does do this for the network support inside dracut, but mostly as a convenience for users, so they don't have to manually find out what dependencies are required - other more obscure modules will be used by savvy users, who can figure out what they need). Users will have to research specifically which packages to install in before they can add a module to their dracut config.
> 
>     * also split dracut into multiple sub packages, which are pulled in conditionally when a user has both dracut and the dependencies of the sub package installed. This obviously requires that they implement such a mechanism in their package manager, which, afaik, few (if any?) do.
> 
> 
> To make life easier for distributions and to avoid duplicating work, ideally,
> dracut would then document, in a clean manner, the requirements for each of
> their modules and which ones go together, at which point, the work currently
> being done in `check` for each module would be repeated, just elsewhere, but
> with added complexity for packagers and with the disadvantage that
> documentation will usually live in a separate place from the module code.
> > Components should have proper and correct packaging and dependency always. I personally have had to clean up after sloppiness of package maintainers that had decided that the future was frozen in time so they did not even bother to depend on core component in the core OS stack, from the component they maintained. They just somehow felt or otherwise expected that relevant component in question would always be there,that the evolution somehow did not apply to the linux ecosystem, that things would be frozen in time and then when I searched for dependency on that relevant component, those components of course did not have the any, thus they fell out of the scope and things started to break.
> 
> I'm not sure what you mean here; do you have a specific example?

I'm referring to work I was doing in Fedora few years back. The reason I'm being so vague about it is because I have no interest in potentially putting RH employees in a tight spot for something that can ultimately being blamed on poor packaging guidelines in the project. If the packaging dependency would have been correct from the get go, it would have saved me a lot of ( my free contributed ) time ( we are talking weeks if not months give the scope of what 700+ components I was working on ) of my life  and end users from having to experiencing bugs as a result of that so I have I first hand experience what incorrect packaging dependencies can cause.
> 
> > We already do support multiple init systems in some modules ( or rather check if systemd module is in use and take action based on that ) which is one approach but from my point of view ( if it must be supported ) that those alternative init systemd should get a 00 module and those module supporting that init system should be separated into their own modules, and depend on that 00init module, which should help downstream packagers and us here to diagnose issues and dealing with bugs that are a result of a different behavior between init system and the code + It would give us something tangible to measure in times of cost vs usage but that's just my point of view.
> 
> We don't include any other init inside dracut, just generating an initramfs
> without systemd in the system is enough for everything to work fine. 
>If you
> wish, I can look into adding a Void (or Alpine? I'm not sure they ship dracut)
> target for CI to run on, but again, it would be nice to have open issues about
> such matters, so we can see where help is needed and where dracut is aiming to
> move. It should help people from outside the project to fit their expectations
> better, and also avoid wasting your time on writing the same answers
> everywhere.
> 

Realistic expectation for a contribution to opensource project is about 1 - 2 hours a day at best 5 days a week, for an individual that has 8 hour workday and a family to take care of ( which basically means she/he, is in two jobs at that point ) and has other hobbies. That equals to roughly 2 - 4 issues to go through, at best a day. ( that's excluding any other work ). 
We are *3* actively working on the project at the moment and I'm solely doing so purely out of my free time ( at the cost of other thing that I should be doing in my life but that's my choice ). 

It's not that there is not will to be everything to everybody but rather the lack of resources which forces us to being picking and choosing and people need to face that reality that something has to give, so the project can move forward. 

It should be quite evident that we need all the help we can get because the math simply does not add up for the work we need to do vs available resources in the project so if you can help, please help.  

Now I have snipped the rest because my response to the e<forks> and the above are related. 

Basically any fork that is diverging from it's original comes at a *price* and that price is always ( contributors ) time and it's always the same. 

When you fork a project, you need more resource then what the original had, available before you fork, yeah sure you could try to rely on you managing to convince some people to join you on your fork quest but you should not take that for granted. Now most forks are done out of "feelings" instead of technical merits or necessity so those resources are not readily available. 

Now what is happening directly or indirectly with the e<forks> is that those forks dont have the resources available themselves so they push their work onto someone else. In the case of eiwd fork, he pushes it directly upstream which will cause twice or more time from the available resource pool from the iwd community to deal with, and in the case for example of elogind where the downstream Debian NM maintainer [1] is stuck between a rock and a hard place, trying to follow upstream to reduce his maintenance load to free up some of his time ( or just struggling to manage it ) or add to it by following some ideology or the needs of the few. ( this friction will eventually burn him out and cause the Debian community to lose a maintainer, reducing it's contribution time by the factor that he's spending there ) .

The reality with the e<forks>, they will only diverge further from their original and become more costly to carry in the process *because* more and more upstreams are integrating themselves deeper with systemd to free up time in *their* available resource pool.

1. https://lists.debian.org/debian-ctte/2020/12/msg00027.html

haraldh:
So, without reading all of the above, my personal preference would be to have a `check()` specific `derror` version... like `dcheck_error`, which does:
* error - if the module was explicitly in the install list
* info - if the module was checking, if available to be included in the `include all modules possible` case

haraldh:
I am open for new functions in `module-setup.sh` like `is_optional()` and/or `check_for_install()` / `check_for_optional()` ... (names can be bike shedded).

ahesford:
I think a `dcheck_error` function, or a modification to `derror` that supports an optional second argument to change behavior, is a reasonable approach that prevents inappropriate errors when checking for installability of unrequested modules but throws errors when checking for installability of explicitly requested modules or their dependencies.

johannbg:
Just to keep everyone updated. 
After some back and fourth to day it was decide to... 
Remove all module checks for modules, which also exits in depends().  
Have no derror for optional (255) modules checking for dependencies 

The fixes that reflect that decision and fixes this module, along with others affected, is in PR #1077

Thank you for your contribution.

Issue #1026, opened 2021-01-25
systemd-modules-load: allow dbus or dbus-broker


danimo:
See https://github.com/dracutdevs/dracut/pull/1021/files#diff-cb62096d7dee784f8ef92441b48876562b3d23c64c5b63f28d7369c5bb3d754eR17.

johannbg:
This has been resolved.

PR #1046, opened 2021-02-03
feat(dbus,dbus-broker,dbus-daemon): Introducing dbus,dbus-broker,dbus-daemon.

johannbg:
This pull request introduces 00dbus a meta dbus module which is required to manage multiple dbus implementation under the same name (dbus).
It also introduces dbus-broker which aim is to provide high performance and reliability, while keeping compatibility to the D-Bus reference implementation as well as rework of the previous 06dbus implementations as a new module under 06dbus-daemon. 
These should complete the dbus implementation in dracut..
Closes #1044, #1019, #1025, #1026


haraldh:
could you add relevant changes in `dracut.spec` for test 99 to succeed?

johannbg:
@haraldh added. 

One point we probably should remove that test in the future and or somehow improve it since a) it is prone to cause merge conflicts ( which is why I stopped adding changes to the spec files to my module PR's ) and b) expecting contributors in any shape or form to be experts in downstream packaging solutions on top of the component they are contributing to, acts as an barrier of entry to contribution and or altogether prevents them from contributing.

haraldh (approved):

modules.d/00dbus/module-setup.sh
```diff
@@ -0,0 +1,20 @@
...
+    # We only want to return 255 since this is a bus meta module.
+    return 255
+}
+
+# Module dependency requirements.
+depends() {
+    # Add a dbus meta dependency based on the module in use.
+    for module in dbus-daemon dbus-broker; do
+        if dracut_module_included "$module" ; then
+            dbus="$module"
```
danimo, line 16: $dbus is not being `echo`'ed?

danimo (changes requested):

modules.d/00dbus/module-setup.sh
johannbg: Fixed

danimo (approved):

PR #1049, opened 2021-02-04
fix(spec): add new systemd-coredump module to spec

lnykryn:
939b7e11d introduced a new module, but it was not added to the spec file, which makes make rpm to fail.

## Checklist
- [ yep ] I have tested it locally
- [ nope ] I have reviewed and updated any documentation if relevant
- [ nope ] I am providing new code and test(s) for it


haraldh:
meh... although the CI failed and gave us an appropriate message:
https://github.com/dracutdevs/dracut/pull/1042/checks?check_run_id=1816532706#step:4:1595

haraldh (approved):

johannbg (approved):

PR #1050, opened 2021-02-04
fix(network-manager): run as a service if systemd module is present

lnykryn:
In the current state, services that depend on network need to
use dracut hooks, since nothing with pull in the network
targets into the transaction.

In the future, it would be nice to provide developers on systemd-only
systems the possibility to not use dracut hooks at all, but simply put
normal systemd services into the initrd.

Also, some modules even right now depend on systemd ordering, like
cryptsetup, so let's make sure, that the ordering inside systemd work
properly as well.

## Checklist
- [ partly ] I have tested it locally
- [ nope ] I have reviewed and updated any documentation if relevant
- [ nope ] I am providing new code and test(s) for it

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -34,7 +34,13 @@ install() {
     inst /usr/libexec/nm-initrd-generator
     inst_multiple -o teamd dhclient
     inst_hook cmdline 99 "$moddir/nm-config.sh"
-    inst_hook initqueue/settled 99 "$moddir/nm-run.sh"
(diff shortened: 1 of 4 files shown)


johannbg (approved):

johannbg:
@lnykryn I like this solution/approach, it's clean and elegant. One thing I think we should do as well and that is to label the type units with something like dracut (dracut-nm-run.service  ) to make them easily identifiable that the type units came from us so any issues with them end up here as opposed to the relevant upstream of the module.

lnykryn:
Hmm but the problem is that now we call online hooks and netroot.sh outside the initqueue.

PR #1052, opened 2021-02-04
fix(network-manager): run as a service if systemd module is present

lnykryn:
Sorry for the second pull request, but I no longer trust in local make check.

This is a slightly different approach, where we only start nm from the unit, but we keep other stuff in init queue.

In the current state, services that depend on network need to
use dracut hooks, since nothing with pull in the network
targets into the transaction.

In the future, it would be nice to provide developers on systemd-only
systems the possibility to not use dracut hooks at all, but simply put
normal systemd services into the initrd.

Also, some modules even right now depend on systemd ordering, like
cryptsetup, so let's make sure, that the ordering inside systemd work
properly as well.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -34,7 +34,13 @@ install() {
     inst /usr/libexec/nm-initrd-generator
     inst_multiple -o teamd dhclient
     inst_hook cmdline 99 "$moddir/nm-config.sh"
+    if dracut_module_included "systemd"; then
+        inst_simple "${moddir}/nm-run.service" "${systemdsystemunitdir}/nm-run.service"
+        $SYSTEMCTL -q --root "$initdir" enable nm-run.service
+    fi
+
     inst_hook initqueue/settled 99 "$moddir/nm-run.sh"
+
     inst_rules 85-nm-unmanaged.rules
     inst_libdir_file "NetworkManager/$_nm_version/libnm-device-plugin-team.so"
     inst_simple "$moddir/nm-lib.sh" "/lib/nm-lib.sh"
--- a/modules.d/35network-manager/nm-config.sh
+++ b/modules.d/35network-manager/nm-config.sh
@@ -6,4 +6,12 @@ if [ -n "$netroot" ] || [ -e /tmp/net.ifaces ]; then
     echo rd.neednet >> /etc/cmdline.d/35-neednet.conf
 fi
 
+if getargbool 0 rd.debug -d -y rdinitdebug -d -y rdnetdebug; then
+    mkdir -m 0755 -p /run/NetworkManager/conf.d
+    (
+        echo '[logging]'
+        echo 'level=TRACE'
+    ) > /run/NetworkManager/conf.d/initrd-logging.conf
+fi
+
 nm_generate_connections
--- a/modules.d/35network-manager/nm-run.service
+++ b/modules.d/35network-manager/nm-run.service
@@ -1,0 +1,28 @@
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+[Unit]
+#make sure all devices showed up
+Wants=systemd-udev-settle.service
+After=systemd-udev-settle.service
+
+#pull the network targets into transaction
+Wants=network.target network-online.target
+Before=network.target network-online.target
+
+#run before we try to mount anything from the dracut hooks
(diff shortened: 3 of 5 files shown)


fpokryvk:
I ran NM dracut tests and it works, but we lose NM debug logs with rd.debug cmdline option (which is crutial for us in case there is a bug in NM). 

I tried to fix that by running nm-run.sh in nm-run.service (installed to /usr/libexec/nm-run.sh with systemd detection disabled), but it breaks the boot for some reason. Is there a option to conditionaly enable NM debug  in systemd service? (pass `--configure-and-quit=initrd --debug --log-level=trace` to NM if rd.debug in cmdline options).

But, maybe this can be done by nm-initrd-generator, it can leave NM config with debug options if needed.

So, we will decide on Monday.

Thanks!

bengal:
> I ran NM dracut tests and it works, but we lose NM debug logs with rd.debug cmdline option

An easy solution could be that nm-config.sh parses `rd.debug` and writes `/run/NetworkManager/conf.d/initrd-logging.conf` containing
```
[logging]
level=TRACE
```

haraldh:
A useful PR is https://github.com/dracutdevs/dracut/pull/1061, so you can run the CI tests quickly with unprivileged podman

fpokryvk:
> A useful PR is #1061, so you can run the CI tests quickly with unprivileged podman

Thank you for the tip, but we have our own tests testing mainly NM behavior (and NM bugs) and run for each new NM version:

https://gitlab.freedesktop.org/NetworkManager/NetworkManager-ci/-/blob/master/nmcli/features/dracut.feature

The tests are based on dracut tests, but more focused to NM testing (that NM works after switchroot).

So, if there is a change to the network-manager dracut module, it is good to check whether our tests are fine with the changes.

johannbg (commented):
> So, if there is a change to the network-manager dracut module, it is good to check whether our tests are fine with the changes.

It's better that we duplicate those NM related tests here. We cant be expecting that contributors to dracut should somehow wonder off to some remote corner of the internet, to check if the changes they contributed to dracut, broke some CI checks over there...

modules.d/35network-manager/nm-run.service
```diff
@@ -0,0 +1,25 @@
+[Unit]
```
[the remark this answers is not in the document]
lnykryn: done

lnykryn:
> > I ran NM dracut tests and it works, but we lose NM debug logs with rd.debug cmdline option
> 
> An easy solution could be that nm-config.sh parses `rd.debug` and writes `/run/NetworkManager/conf.d/initrd-logging.conf` containing
> 
> ```
> [logging]
> level=TRACE
> ```

Done.

modules.d/35network-manager/nm-config.sh
```diff
@@ -6,4 +6,11 @@ if [ -n "$netroot" ] || [ -e /tmp/net.ifaces ]; then
     echo rd.neednet >> /etc/cmdline.d/35-neednet.conf
 fi
 
+if getargbool 0 rd.debug -d -y rdinitdebug -d -y rdnetdebug; then
+    (
+        echo '[logging]'
+        echo 'level=TRACE'
+    ) > /run/NetworkManager/conf.d/initrd-logging.conf
```
haraldh, line 13: Does the directory `/run/NetworkManager/conf.d` exist already?

modules.d/35network-manager/module-setup.sh
```diff
@@ -34,7 +34,13 @@ install() {
     inst /usr/libexec/nm-initrd-generator
     inst_multiple -o teamd dhclient
     inst_hook cmdline 99 "$moddir/nm-config.sh"
+    if dracut_module_included "systemd"; then
+        inst_simple "${moddir}/nm-run.service" "${systemdsystemunitdir}/nm-run.service"
+        systemctl -q --root "$initdir" enable nm-run.service
```
haraldh, line 39:
```suggestion
        $SYSTEMCTL -q --root "$initdir" enable nm-run.service
```

haraldh (changes requested):
Please also rebase to origin/master

modules.d/35network-manager/nm-config.sh
lnykryn: fixed

johannbg (approved):
LGTM

haraldh (approved):

PR #1063, opened 2021-02-07
Fix NFS mount failure for NFSv3

ryncsn:
commit 5840c466d tried to make it create the right directory according
to the directory layout on current host, but added "$initdir" prefix
to the if statement by accident, so the installation code have no
effect now. Just fix it.

This pull request changes...

## Checklist

Fixes #


johannbg (approved):
LGTM

danimo (approved):
Good catch. Thanks!

PR #1075, opened 2021-02-09
fix: correct the line continuation

haraldh:
Comments will break bash line continuation.
Also added some missing line continuation here and there.

### DIFF
--- a/modules.d/01systemd-repart/module-setup.sh
+++ b/modules.d/01systemd-repart/module-setup.sh
@@ -34,14 +34,13 @@ depends() {
 # Install the required file(s) for the module in the initramfs.
 install() {
 
-    # Install the required file(s.
+    # Install vendor repartition configurations
+    # Install the systemd type service unit for systemd repart.
+    # Install the binary executable(s) for systemd repart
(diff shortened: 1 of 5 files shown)


modules.d/01systemd-repart/module-setup.sh
```diff
@@ -34,14 +34,13 @@ depends() {
 # Install the required file(s) for the module in the initramfs.
 install() {
 
-    # Install the required file(s.
+    # Install vendor repartition configurations
+    # Install the systemd type service unit for systemd repart.
+    # Install the binary executable(s) for systemd repart
     inst_multiple -o \
-        # Install vendor repartition configurations
         $libdir/repart.d/*.conf
```
johannbg: The original missed an line break and so does this. It should be -->  $libdir/repart.d/*.conf \  <--

johannbg (approved):

danimo (approved):

modules.d/01systemd-repart/module-setup.sh
haraldh: fixed

PR #1076, opened 2021-02-09
fix(plymouth): install binaries with dependencies

zboszor:
If plymouth is used then its own installation script is preferred.
While plymouth's own variant of inst_binary resolves the library
dependencies properly, its inst_library shell function doesn't
actually install them in some cases properly.

Make sure to install both the required binaries and their
dependencies. In the worst case, it's a NOP.

Signed-off-by: Zoltán Böszörményi <zboszor@pr.hu>

This pull request changes...

## Checklist

Fixes #


haraldh (approved):

johannbg (approved):
LGTM

PR #1079, opened 2021-02-09
90kernel-modules: arm/arm64: Add reset controllers

vianpl:
Reset controllers might be needed by some of the devices used in the
initrd. Particularly on the Raspberry Pi 4, 'xhci-pci' depends on a
platform specific reset controller.

Signed-off-by: Nicolas Saenz Julienne <nsaenzjulienne@suse.de>

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -85,6 +85,7 @@ installkernel() {
                 "=drivers/phy" \
                 "=drivers/power" \
                 "=drivers/regulator" \
+                "=drivers/reset" \
                 "=drivers/rpmsg" \
                 "=drivers/rtc" \
                 "=drivers/soc" \


johannbg (approved):
LGTM

haraldh (approved):

vianpl:
The commit message checker doesn't approve me using `90kernel-modules` in the commit message, which I assume it's not really a problem.

On the other hand, it's not clear to me whether the rest test failures really apply to my patch. Should I dig deeper into them?

johannbg:
> The commit message checker doesn't approve me using `90kernel-modules` in the commit message, which I assume it's not really a problem.

Well it is, you need to change your commit msg to something like "fix(kernel-modules): add reset controllers for arm" ( I should have spotted this ) to pass the commisery test.

> 
> On the other hand, it's not clear to me whether the rest test failures really apply to my patch. Should I dig deeper into them?

Nope let us worry about that. 
On top of our test being somewhat flaky, github has been experiencing issues lately [1] which is why the tests have been timing out. 

1. https://www.githubstatus.com/incidents/5qdkkyg958vy

PR #1083, opened 2021-02-10
fix(shutdown): add timeout to umount calls

petrpavlu:
When terminating a system, the shutdown module attempts to unmount all file systems from under `/oldroot`. This reaps remaining file systems that systemd cannot unmount and detaches `/oldroot` itself.

In case that running umount for some file system repeatedly fails, the module reports this error and continues the processing in order to shutdown the system. This handles a condition when the umount command actually terminates but it can happen in some cases that it waits indefinitely.

An example with NFS mounts:
```
# mount -t nfs 192.168.0.1:/srv/nfs/dir /mnt/nfs
# mkdir /mnt/nfs/dir2
# mount -t nfs 192.168.0.1:/srv/nfs/dir2 /mnt/nfs/dir2
# touch /mnt/nfs/dir2/file
# systemd-run -pKillMode=none -pSendSIGKILL=no tail -f /mnt/nfs/dir2/file
Running as unit: run-r367825c967ca4d88a793ae4793c02f8b.service
# systemctl poweroff
```

The invoked tail command escapes normal termination by systemd and prevents stopping mnt-nfs.mount and mnt-nfs-dir2.mount as it makes the mounts busy. Systemd then again attempts to unmount these file systems in systemd-shutdown but this fails as well. The utility tries to unmount `/mnt/nfs/dir2` but the kernel waits indefinitely doing a path lookup for `/mnt/nfs` because network is no longer available at that point. The systemd-shutdown gives up after 90 seconds. Finally, the control is transferred to dracut which tries to unmount the file systems in the same way and ends up indefinitely waiting on umount to finish.

This situation causes that the system hangs during shutdown. The patch improves the shutdown module to add a timeout of 90 seconds for the umount operation and continue with the shutdown if it gets reached, similarly to what systemd-shutdown does.

## Checklist

### DIFF
--- a/modules.d/99shutdown/shutdown.sh
+++ b/modules.d/99shutdown/shutdown.sh
@@ -46,28 +46,56 @@ warn "Killing all remaining processes"
 
 killall_proc_mountpoint /oldroot || sleep 0.2
 
+# Timeout for umount calls. The value can be set to 0 to wait forever.
+_umount_timeout=$(getarg rd.shutdown.timeout.umount)
+_umount_timeout=${_umount_timeout:-90s}
+_timed_out_umounts=""
+
 umount_a() {
+    local _verbose="n"
+    if [ "$1" = "-v" ]; then
+        _verbose="y"; shift
+        exec 7>&2
+    else
+        exec 7>/dev/null
+    fi
+
     local _did_umount="n"
     while read a mp a || [ -n "$mp" ]; do
-        if strstr "$mp" oldroot; then
-            if umount "$mp"; then
-                _did_umount="y"
-                warn "Unmounted $mp."
-            fi
+        strstr "$mp" oldroot || continue
+        strstr "$_timed_out_umounts" " $mp " && continue
+
+        # Unmount the file system. The operation uses a timeout to avoid waiting
+        # indefinitely if this is e.g. a stuck NFS mount. The command is
+        # invoked in a subshell to silence also the "Killed" message that might
+        # be produced by the shell.
+        (set +m; timeout --signal=KILL "$_umount_timeout" umount "$mp") 2>&7
+        local ret=$?
+        if [ $ret -eq 0 ]; then
+            _did_umount="y"
+            warn "Unmounted $mp."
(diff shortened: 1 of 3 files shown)


johannbg (changes requested):
You need to fix your commit message. Change it to something like "fix(shutdown): add timeout to umount calls" see https://github.com/dracutdevs/dracut/blob/master/HACKING.md for more details.

petrpavlu:
> You need to fix your commit message. Change it to something like "fix(shutdown): add timeout to umount calls" see https://github.com/dracutdevs/dracut/blob/master/HACKING.md for more details.

Fixed in the updated commit, thanks.

modules.d/99shutdown/shutdown.sh
```diff
@@ -46,28 +46,56 @@ warn "Killing all remaining processes"
...
-                warn "Unmounted $mp."
-            fi
+        strstr "$mp" oldroot || continue
+        strstr "$_timed_out_umounts" " $mp " && continue
+
+        # Unmount the file system. The operation uses a timeout to avoid waiting
+        # indefinitely if this is e.g. a stuck NFS mount. The command is
+        # invoked in a subshell to silence also the "Killed" message that might
+        # be produced by the shell.
+        (set +m; timeout --signal=KILL "$_umount_timeout" umount "$mp") 2>&7
```
haraldh, line 72: I doubt, that the KILL signal will help in any way, because the task is most likely in the "D" state anyway.

haraldh (approved):

johannbg (approved):

modules.d/99shutdown/shutdown.sh
petrpavlu: The KILL signal can help for an uninterruptible umount task if its state is also killable (`TASK_WAKEKILL`). This was at least the case in my test with NFS.

Using the KILL signal importantly causes that this signal is also sent by the timeout process to itself when the time limit is reached. This guarantees the parent timeout process always terminates and the shutdown operation can proceed.

PR #1088, opened 2021-02-11
Refactor the 99squash module

ryncsn:
Note: this PR depends on https://github.com/systemd/systemd/pull/18124, so I'm not sure when this could be merged, if acceptable.

## Current problems
The aim of 99squash module is to reduce the memory usage of initramfs, but there are several problems with the 99squash module:
1. It try to reuse what is already installed in the initramfs, this make many files not squash-able, left outside of squash image. These files could be large.
2. The initramfs is doubled compressed, once by the mksquashfs, once upon generate the cpio (Many files are kept outside, so double compress is needed)
3. The code is complex, and the squash image setup layout is complex.
4. Performance and memory usage is not optimal due to above reasons.

## Changes
With this PR, the 99squash module is changed in following way:
1. Using busybox as the sole binary for setup the image, (almost) everything is packed in squash image.
2. Refactor image setup process, in a much simpler way. Now it setup a new root and chroot into instead of overlaying on multiple mount points on existing root. (depend on https://github.com/systemd/systemd/pull/18124)
3. Disable initramfs compress when 99squash is enabled. (everything is squashed, no need to compress twice)

Now the 99squash can be considered an alternative way of compressing the initramfs.
Instead of compress the initramfs image, and decompress everything to ram upon boot, 99squash make the intiamfs content compressed, and decompressed upon file access. This have some advantage on performance and memory usage.

Example initamfs content with the new 99squash (and also a lot of other module enabled for testing):
```
[root@localhost image]# lsinitrd /boot/initramfs-5.10.11-200.fc33.x86_64.img 
Image: /boot/initramfs-5.10.11-200.fc33.x86_64.img: 92M
========================================================================
Early CPIO image
========================================================================
drwxr-xr-x   1 root     root            0 Feb 11 15:51 .
-rw-r--r--   1 root     root            2 Feb 11 15:51 early_cpio
drwxr-xr-x   1 root     root            0 Feb 11 15:51 kernel
drwxr-xr-x   1 root     root            0 Feb 11 15:51 kernel/x86
drwxr-xr-x   1 root     root            0 Feb 11 15:51 kernel/x86/microcode
-rw-r--r--   1 root     root        30546 Feb 11 15:51 kernel/x86/microcode/AuthenticAMD.bin
-rw-r--r--   1 root     root      3620864 Feb 11 15:51 kernel/x86/microcode/GenuineIntel.bin
========================================================================
Version: dracut-051-76.git20210211.fc33

Arguments: --add 'squash' --force

dracut modules:
bash
systemd
systemd-initrd
modsign
nss-softokn
rdma
rngd
i18n
network-manager
network
ifcfg
drm
plymouth
bcache
btrfs
crypt
dm
dmraid
kernel-modules
kernel-modules-extra
kernel-network-modules
lvm
mdraid
nvdimm
qemu
qemu-net
cifs
iscsi
lunmask
nfs
resume
rootfs-block
terminfo
udev-rules
virtfs
dracut-systemd
usrmount
base
fs-lib
memstrack
shutdown
squash
========================================================================
drwxr-xr-x   1 root     root            0 Feb 11 15:51 .
-rwxr-xr-x   1 root     root         1281 Feb 11 15:51 init
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 lib -> usr/lib
drwxr-xr-x   1 root     root            0 Feb 11 15:51 squash
-rw-r--r--   1 root     root     90984448 Feb 11 15:51 squash-root.img
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/bin
-rwxr-xr-x   1 root     root      1293688 Jul 27  2020 usr/bin/busybox
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 usr/bin/cat -> busybox
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 usr/bin/echo -> busybox
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 usr/bin/mkdir -> busybox
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 usr/bin/modprobe -> busybox
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 usr/bin/mount -> busybox
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 usr/bin/rm -> busybox
lrwxrwxrwx   1 root     root            7 Feb 11 15:51 usr/bin/sh -> busybox
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/dracut
-rw-r--r--   1 root     root           23 Feb 11 15:51 usr/lib/dracut/build-parameter.txt
-rw-r--r--   1 root     root           31 Feb 11 15:51 usr/lib/dracut/dracut-051-76.git20210211.fc33
-rw-r--r--   1 root     root          358 Feb 11 15:51 usr/lib/dracut/modules.txt
-rw-r--r--   1 root     root            0 Feb 11 15:51 usr/lib/dracut/need-initqueue
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/drivers
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/drivers/block
-rw-r--r--   1 root     root        15760 Jan 28 04:41 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/drivers/block/loop.ko.xz
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/fs
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/fs/overlayfs
-rw-r--r--   1 root     root        53288 Jan 28 04:41 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/fs/overlayfs/overlay.ko.xz
drwxr-xr-x   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/fs/squashfs
-rw-r--r--   1 root     root        25476 Jan 28 04:41 usr/lib/modules/5.10.11-200.fc33.x86_64/kernel/fs/squashfs/squashfs.ko.xz
-rw-r--r--   1 root     root          185 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.alias
-rw-r--r--   1 root     root          198 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.alias.bin
-rw-r--r--   1 root     root         9865 Jan 28 04:35 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.builtin
-rw-r--r--   1 root     root            0 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.builtin.alias.bin
-rw-r--r--   1 root     root        12385 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.builtin.bin
-rw-r--r--   1 root     root          103 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.dep
-rw-r--r--   1 root     root          192 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.dep.bin
-rw-r--r--   1 root     root           78 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.devname
-rw-r--r--   1 root     root       151428 Jan 28 04:35 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.order
-rw-r--r--   1 root     root           55 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.softdep
-rw-r--r--   1 root     root          133 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.symbols
-rw-r--r--   1 root     root          110 Feb 11 15:51 usr/lib/modules/5.10.11-200.fc33.x86_64/modules.symbols.bin
========================================================================
```

Note all initramfs content is packed in squash-root.img, and consume 92M of RAM during boot (previously 180M).

This is mostly used for kdump, and the kdump test result is also improved compared to previous implementation.

## Checklist

Fixes #


johannbg (commented):
Given the scope of this change where you seem to be making the assumption that no one is using the squash module unless she/he is using buxybox as well. which I do believe is not the case here, it's probably best if these changes are introduced in a separated module ( squash-busybox ) and or separated in existing module depending on the busybox module being included or not.

ryncsn:
> Given the scope of this change where you seem to be making the assumption that no one is using the squash module unless she/he is using buxybox as well. which I do believe is not the case here, it's probably best if these changes are introduced in a separated module ( squash-busybox ) and or separated in existing module depending on the busybox module being included or not.

That's a good suggestion, I'm just not sure if it worth keep two version of 99squash, the new approach is superior in every way except it depends on busybox.

Anyway, I'll do a rebase as there are some conflicts, and try if I can introduce a squash-busybox, and still keep things clean.

haraldh:
Do we need a systemd version check for `SYSTEMD_IN_INITRD=lenient`, so it does not cause any data loss?

ryncsn:
> Do we need a systemd version check for `SYSTEMD_IN_INITRD=lenient`, so it does not cause any data loss?

I think that's better be done in the specfile, jump bump the required version, but have to wait until a new systemd release.

Without `SYSTEMD_IN_INITRD=lenient` support, this PR will not cause data loss, just fail to boot as systemd refuse to recognize the initrd and refuse to reach initrd.target.

ryncsn:
I've updated the PR, now busybox is a weak dependency, 99squash can work without it, and still keep things simple.

haraldh:
sorry, the latest shellcheck PR rendered this PR unmergeable.. Apologies, that you will have to rebase.

ryncsn:
Rebased and check the new change with shellcheck, now it looks good.

johannbg:
@ryncsn you need to fix your commit messages so they pass the commisery check. 
They should all start with ```fix(squash):```  ( we dont use number is modules since the numbers are subjected to change ) 
And don't use title case in the description. 

Good
```fix(squash): move the post install scripts into the module-setup.sh``` 
Bad
```fix(squash): Move the post install scripts into the module-setup.sh```

File: dracut.sh [hunk 1]
```diff
@@ -2118,16 +2080,28 @@ fi
...
 if dracut_module_included "squash"; then
     dinfo "*** Squashing the files inside the initramfs ***"
+    declare squash_compress_arg
+    if [[ $compress_l ]]; then
+        if ! mksquashfs /dev/null "$DRACUT_TMPDIR"/.squash-test.img -no-progress -comp $compress_l &>/dev/null; then
+            dwarn "mksquashfs doesn't support compressor '$compress_l', failing back to default compressor."
+        else
+            squash_compress_arg="-comp $compress_l"
+        fi
+    fi
```
haraldh, line 2090: I think you should use `$compress` rather then `$compress_l`, which is only the CLI parameter, if set and not the setting from any configuration file.

```suggestion
    if [[ $compress ]]; then
        if ! mksquashfs /dev/null "$DRACUT_TMPDIR"/.squash-test.img -no-progress -comp $compress &>/dev/null; then
            dwarn "mksquashfs doesn't support compressor '$compress', failing back to default compressor."
        else
            squash_compress_arg="-comp $compress"
        fi
    fi
```

dracut.sh
```diff
@@ -2118,16 +2080,28 @@ fi
 
 if dracut_module_included "squash"; then
     dinfo "*** Squashing the files inside the initramfs ***"
+    declare squash_compress_arg
+    if [[ $compress_l ]]; then
+        if ! mksquashfs /dev/null "$DRACUT_TMPDIR"/.squash-test.img -no-progress -comp $compress_l &>/dev/null; then
+            dwarn "mksquashfs doesn't support compressor '$compress_l', failing back to default compressor."
+        else
+            squash_compress_arg="-comp $compress_l"
```
haraldh, line 2088: Additionally, I would change this to 
```suggestion
            squash_compress_arg="$compress"
```
to use it later in the real call with `${squash_compress_arg:+-comp "$squash_compress_arg"}`

dracut.sh
```diff
@@ -2118,16 +2080,28 @@ fi
...
+
+    # shellcheck disable=SC2086
     if ! mksquashfs "$squash_dir" "$squash_img" \
-        -no-xattrs -no-exports -noappend -always-use-fragments \
-        -comp xz -Xdict-size 100% -no-progress \
-        > /dev/null; then
-        dfatal "dracut: Failed making squash image"
-        exit 1
+        -no-xattrs -no-exports -noappend -no-recovery -always-use-fragments \
+        -no-progress $squash_compress_arg 1> /dev/null; then
```
haraldh, line 2095:
```suggestion
        -no-progress ${squash_compress_arg:+-comp "$squash_compress_arg"} 1> /dev/null; then
```

dracut.spec
```diff
@@ -184,6 +184,7 @@ This package contains tools to assemble the local initrd and host configuration.
 Summary: dracut module to build an initramfs with most files in a squashfs image
 Requires: %{name} = %{version}-%{release}
 Requires: squashfs-tools
+Recommends: busybox
```
haraldh, line 187: I really don't like this, because otherwise we would `Recommend` any tools we ship for every module we ship, which could explode the installation.

Maybe a dracut subpackage would be a solution?

modules.d/99squash/module-setup.sh
```diff
@@ -5,7 +5,7 @@ check() {
 
     for i in CONFIG_SQUASHFS CONFIG_BLK_DEV_LOOP CONFIG_OVERLAY_FS ; do
         if ! check_kernel_config $i; then
-            dinfo "dracut-squash module requires kernel configuration $i (y or m)"
+            derror "dracut-squash module requires kernel configuration $i (y or m)"
```
haraldh, line 8: Please don't error here yet... We will introduce a special dcheck_error, which errors, if the modules was explicitly specified.

modules.d/99squash/init-squash.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/sh
```
haraldh, line 1:
```suggestion
#!/bin/sh
```

haraldh (changes requested):

dracut.spec
ryncsn: Actually this is already a "dracut-squash" sub package, do you want a dracut-squash-busybox? Or maybe I can just drop this and add a dinfo that busybox is not installed when squash is used.

modules.d/99squash/module-setup.sh
ryncsn: OK, will update .

modules.d/99squash/init-squash.sh
ryncsn: Thanks for the review, will fix it to /bin/sh. This part is being included in the initramfs, so I used /usr/bin/sh as the squash module setup will organize the initramfs that way and didn't create /bin and /sbin as symlink to /usr/xbin, that looks no so good after a second thought.

dracut.sh
ryncsn: I think I'll have to rework this part, $compress will be override by compressor detection code at line 924, and will contain invalid params for mksquash.

ryncsn:
> @ryncsn you need to fix your commit messages so they pass the commisery check.
> They should all start with `fix(squash):` ( we dont use number is modules since the numbers are subjected to change )
> And don't use title case in the description.
> 
> Good
> `fix(squash): move the post install scripts into the module-setup.sh`
> Bad
> `fix(squash): Move the post install scripts into the module-setup.sh`

Thanks, I'll fix these issues.

johannbg:
@ryncsn you still need to fix your commit message to pass the commisery check. 
Change this commit msg ( contains a title case in the description )
```refactor(squash): Move the post install scripts into the module-setup.sh```
to this
```refactor(squash): move the post install scripts into the module-setup.sh```
and this ( is repeating the type tag )
```refactor(squash): refactor in a cleaner way```
to this
```refactor(squash): structure in a cleaner way```
and it should pass the check.

File: dracut.spec (hunk 4 above)
Comment (johannbg) [reply dracut.spec:187]: @haraldh we probably end up packaging every module into a sub-package as dracut-module-<module> with proper dependency's etc. since that´s the proper way of doing things.

@ryncsn you could create this as an such sub-package as in dracut-module-squash-busybox

**Comment by ryncsn** [comment]
Yes, I just notices that. Is there anyway I can do the commisery check locally before pushing it?

**Comment by johannbg** [comment]
> Yes, I just notices that. Is there anyway I can do the commisery check locally before pushing it?

There are instructions here [1] but in most cases people ( including myself ) just have done the same mistakes you just did and once you get passed that + avoid including the module number you should be fine on your PR to dracut. 

The failure messages of the commisery check is pretty self explanatory, if you just read and fix based on that output of the error msg of test itself, you dont need to install/setup anything locally.

1. https://github.com/tomtom-international/commisery

File: modules.d/99squash/module-setup.sh (hunk 5 above)
Comment (haraldh) [reply modules.d/99squash/module-setup.sh:8]: still there

**Review by haraldh — CHANGES_REQUESTED** [review]

File: dracut.spec (hunk 4 above)
Comment (ryncsn) [reply dracut.spec:187]: @johannbg That will be an empty package, `squash` module itself will detect if busybox is installed and use it if available. How about I just drop this and add a dinfo if busybox is not used?

Comment (ryncsn) [reply dracut.spec:187]: Or I can just drop it silently, it easy to tell if busybox is used for an squashed initramfs with lsinitrd.

Comment (johannbg) [reply dracut.spec:187]: @ryncsn yeah just drop the recommend, I have already create issue for a sub-packaging scope in which we will be dealing with packaging of the modules so you dont have to worry about that as in we will take care of it.

File: modules.d/99squash/module-setup.sh (hunk 5 above)
Comment (johannbg) [reply modules.d/99squash/module-setup.sh:8]: @ryncsn you forgot this

Comment (ryncsn) [reply modules.d/99squash/module-setup.sh:8]: silly me, will fix it.

**Review by johannbg — APPROVED** [review]
LGTM

File: dracut.sh [hunk 7]
```diff
@@ -2117,16 +2032,29 @@ fi
 
 if dracut_module_included "squash"; then
     dinfo "*** Squashing the files inside the initramfs ***"
+    declare squash_compress_arg
+    # shellcheck disable=SC2086
+    if [[ $compress ]]; then
+        if ! mksquashfs /dev/null "$DRACUT_TMPDIR"/.squash-test.img -no-progress -comp $compress &>/dev/null; then
```
haraldh, line 2038:
```suggestion
        if ! mksquashfs /dev/null "$DRACUT_TMPDIR"/.squash-test.img -no-progress -comp "$compress" &>/dev/null; then
```

dracut.sh
```diff
@@ -2117,16 +2032,29 @@ fi
...
+
+    # shellcheck disable=SC2086
     if ! mksquashfs "$squash_dir" "$squash_img" \
-        -no-xattrs -no-exports -noappend -always-use-fragments \
-        -comp xz -Xdict-size 100% -no-progress \
-        > /dev/null; then
-        dfatal "dracut: Failed making squash image"
-        exit 1
+        -no-xattrs -no-exports -noappend -no-recovery -always-use-fragments \
+        -no-progress ${squash_compress_arg:+-comp $squash_compress_arg} 1> /dev/null; then
```
haraldh, line 2048:
```suggestion
        -no-progress ${squash_compress_arg:+-comp "$squash_compress_arg"} 1> /dev/null; then
```

haraldh (changes requested):

dracut.sh
johannbg: @ryncsn ping :point_up:

dracut.sh
johannbg: @ryncsn ping :point_up:

ryncsn: Hi, actually I left it unquoted intentionally, from dracut man page:
```
--compress <compressor>
    ... If you pass it just the name of a compression program, it
           will call that program with known-working arguments. If you pass a
           quoted string with arguments, it will be called with exactly those
           arguments. 
```
So I also added `# shellcheck disable=SC2086` above.

haraldh (approved):

PR #1096, opened 2021-02-13
fix: skip empty modalias files in drm module setup

evgeni:
## Changes

on my system (ThinkPad X201s) I have several modalias entries that are
empty:

    /sys/bus/platform/devices/dock.0/modalias
    /sys/bus/platform/devices/dock.1/modalias
    /sys/bus/platform/devices/dock.2/modalias

executing the drm module setup thus results in the following errors

    dracut-install: No SOURCE argument given

when calling dracut_instmods and hostonly mode is enabled.

Skip those entries, as there are no modules to load in this case
anyways.

Note: it is not sufficient to use [['s -s test (file size is greater
than zero), as all those files are 4096B if you stat them.

## Checklist

Fixes https://bugs.debian.org/982146

### DIFF
--- a/modules.d/50drm/module-setup.sh
+++ b/modules.d/50drm/module-setup.sh
@@ -33,6 +33,7 @@ installkernel() {
     if [[ $hostonly ]]; then
         for i in /sys/bus/{pci/devices,platform/devices,virtio/devices,soc/devices/soc?}/*/modalias; do
             [[ -e $i ]] || continue
+            [[ -n $(<"$i") ]] || continue
             if hostonly="" dracut_instmods --silent -s "drm_crtc_init|drm_dev_register|drm_encoder_init" -S "iw_handler_get_spy" $(<"$i"); then
                 if strstr "$(modinfo -F filename $(<"$i") 2>/dev/null)" radeon.ko; then
                     hostonly='' instmods amdkfd


modules.d/50drm/module-setup.sh
```diff
@@ -32,7 +32,7 @@ installkernel() {
     # loading of the driver if needed
     if [[ $hostonly ]]; then
         for i in /sys/bus/{pci/devices,platform/devices,virtio/devices,soc/devices/soc?}/*/modalias; do
-            [[ -e $i ]] || continue
+            [[ -e $i && -n $(<"$i") ]] || continue
```
evgeni, line 35: technically, the `-e` is not required, as the `for i in <wildcard>` will already return only existing things, and `-e` doesn't test the file for anything else.

johannbg (approved):

modules.d/50drm/module-setup.sh
haraldh, line 35:
```suggestion
            [[ -e $i ]] || continue
            [[ -n $(<"$i") ]] || continue
```

haraldh (changes requested):

modules.d/50drm/module-setup.sh
evgeni: done!

haraldh (approved):

PR #1111, opened 2021-02-17
chore: update authors in .mailmap

tpgxyz:
This pull request removes duplicated my email in AUTHORS files

## Checklist
- [] I am providing new code and test(s) for it

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -34,3 +34,4 @@ Xunlei Pang <xlpang@redhat.com> <xpang@redhat.com>
 Daniel Molkentin <daniel.molkentin@suse.com> <dmolkentin@suse.com>
 Thomas Blume <thomas.blume@suse.com> <Thomas.Blume@suse.com>
 Brian C. Lane <bcl@redhat.com> <bcl@brianlane.com>
+Tomasz Paweł Gajc <tpgxyz@gmail.com> <tpgxyz@gmail.com>


haraldh:
Please modify `.mailmap` instead, as `AUTHORS` is automatically generated from the git commits

tpgxyz:
@haraldh done

johannbg (changes requested):
You need to fix your commit message to pass the commisery check. Change it to "fix: update authors in .mailmap" . Drop the changes you made to the Author files as @haraldh has already pointed out that the file is automatically generated from the git commits, which means any manual changes to the file will get overwritten...

haraldh:
it's more like 'chore:' than 'fix:' as it does not fix dracut

johannbg:
@haraldh he's fixing his email addresses and he does so by adding himself to the authors in .mailmap file, that's a fix not a chore If he had already existed in the .mailmap file and then update it, it would have been a chore: instead of a fix: I would say. I also argue that fixes cover the entire tree not just dracut but either way it beats "remove duplicate me" and passes the commisery check so meh.

haraldh:
Ok, then more like `docs: update authors in .mailmap`

johannbg:
mailmap files aren't documentation so that makes no sense but if we break this action down all we are effectively doing there, is modifying the visual output of commands like git shortlog or git log --use-mailmap as in we are not rewriting commit history or prevent commits with varying names and/or email addresses and that action can be a done as a part of a chore as you pointed out or part of a fix ( which is the case what he arguably is doing or it could be a fix of incorrect entries in the file itself ). 

Anyway @tpgxyz just use chore since @haraldh feels strongly about it, we can run a script that detects and updates this as a part of a chore: which makes it more future proof and inline with what he proposes I guess.

tpgxyz:
@haraldh @johannbg please forgive the inconvenience that I made. Hope it is fixed for good.

johannbg:
@tpgxyz actually you have an extra whitespace between "in  .mailmap" so it's not "fixed" I mean chored :smile:  also change it to chore: which is better suited than docs + this no inconvenience, we just recently adopted and introduced commisery checks so we are still hashing out the details among ourselves.

johannbg:
@tpgxyz you forgot to remove the extra whitespace as in "chore: update authors in<you have an extra whitespace here> .mailmap" which is triggering the commisery check to fail.

johannbg (approved):

haraldh (approved):

PR #1113, opened 2021-02-18
fix(fcoe): rd.nofcoe=0 should disable fcoe

lnykryn:
8446c8f9 Changed the default behavior, but also flipped meaning of 0/1.
Right now rd.nofcoe=0 enables fcoe, which is the opposite what manpage
says.

### DIFF
--- a/modules.d/95fcoe/lldpad.sh
+++ b/modules.d/95fcoe/lldpad.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if getargbool 0 rd.nofcoe ; then
+if ! getargbool 1 rd.nofcoe ; then
 	info "rd.nofcoe=0: skipping lldpad activation"
 	return 0
 fi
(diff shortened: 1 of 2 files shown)


haraldh:
The man page is wrong... 
```diff
diff --git a/dracut.cmdline.7.asc b/dracut.cmdline.7.asc
index b4cbbbcd..6275f0d6 100644
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -868,7 +868,7 @@ iscsistart -b --param node.session.timeo.replacement_timeout=30
 
 FCoE
 ~~~~
-**rd.nofcoe=0**::
+**rd.nofcoe=1**::
     disable FCoE and lldpad
 
 **fcoe=**__<edd|interface|MAC>__:__{dcb|nodcb}__:__{fabric|vn2vn}__::
```

PR #1114, opened 2021-02-18
fcoe: rename rd.nofcoe to rd.fcoe

lnykryn:
fcoe: rename rd.nofcoe to rd.fcoe
    
    The current name of this bool is kinda stupid. Based on the manpage
    setting it to 0 turns off fcoe, which means that nofcoe=1 should mean
    that it is on.
    
    Let's just do the same thing as with rd.lvm=0, rd.luks=0,....

### DIFF
--- a/modules.d/95fcoe/lldpad.sh
+++ b/modules.d/95fcoe/lldpad.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if getargbool 0 rd.nofcoe ; then
+if ! getargbool 1 rd.nofcoe ; then
 	info "rd.nofcoe=0: skipping lldpad activation"
 	return 0
 fi
--- a/modules.d/95fcoe/lldpad.sh
+++ b/modules.d/95fcoe/lldpad.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-if ! getargbool 1 rd.nofcoe ; then
-	info "rd.nofcoe=0: skipping lldpad activation"
+if ! getargbool 1 rd.fcoe -d -n rd.nofcoe ; then
+	info "rd.fcoe=0: skipping lldpad activation"
 	return 0
 fi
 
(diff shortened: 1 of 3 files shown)


modules.d/95fcoe/lldpad.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-if getargbool 0 rd.nofcoe ; then
-	info "rd.nofcoe=0: skipping lldpad activation"
+if ! getargbool 1 rd.fcoe -d rd.nofcoe ; then
```
haraldh, line 3: Shall we follow the wrong documentation with the deprecated option or the correct one with:

```suggestion
if ! getargbool 1 rd.fcoe -d -n rd.nofcoe ; then
```

haraldh (approved):

modules.d/95fcoe/parse-fcoe.sh
```diff
@@ -13,8 +13,8 @@
 # fcoe=eth0:nodcb:vn2vn
 # fcoe=4a:3f:4c:04:f8:d7:nodcb:fabric
 
-if getargbool 0 rd.nofcoe ; then
-	info "rd.nofcoe=0: skipping fcoe"
+if ! getargbool 1 rd.fcoe -d rd.nofcoe ; then
```
haraldh, line 16: Shall we follow the wrong documentation with the deprecated option:

```suggestion
if ! getargbool 1 rd.fcoe -d -y rd.nofcoe ; then
```

or the correct one with:

```suggestion
if ! getargbool 1 rd.fcoe -d -n rd.nofcoe ; then
```

haraldh (changes requested):

modules.d/95fcoe/parse-fcoe.sh
johannbg: The correct one I would say.

johannbg:
@lnykryn you need to fix your commit message to pass the commisery check, Change it to fix(fcoe): rename rd.nofcoe to rd.fcoe and it should pass the check.

modules.d/95fcoe/parse-fcoe.sh
lnykryn: I would be for that one as well.

johannbg (approved):

haraldh (approved):

PR #1115, opened 2021-02-18
fix:  proper return code for inst_multiple in dracut-init.sh

ahesford:
A test of the form

    if ! command; then
        _ret=$?
        ...
        return _ret
    fi

does not capture the return code of `command`, but the negation of the return code, leaving _ret == 0. The test of this form in inst_multiple has been refactored to capture and return the right value.

## Checklist

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -258,12 +258,13 @@ inst_symlink() {
 
 inst_multiple() {
     local _ret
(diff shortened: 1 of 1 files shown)


haraldh:
Thank you!

haraldh (approved):

johannbg (approved):
LGTM

PR #1121, opened 2021-02-19
fix(dracut.sh): unfreeze /boot on exit

tsipa:
If SIGWHATEVER will be processed after fsfreeze -f, but before fsfreeze
-u we will end up with /boot/ never unfrozen, let's try to minimize risk of this.

This pull request changes...

## Changes
Adds cleanup of freeze on exit.

## Checklist
- [ V ] I have tested it locally
Yes, i added `sleep 999` between `fsfreeze -f` and `fsfreeze -u` and did `kill`. Prior to fix result was - boot is frozen, after the fix - boot is fine.
- [ X ] I have reviewed and updated any documentation if relevant
I don't think it's applicalbe

- [ X ] I am providing new code and test(s) for it
I think it's not applicable.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1009,6 +1009,9 @@ readonly DRACUT_TMPDIR="$(mktemp -p "$TMPDIR/" -d -t dracut.XXXXXX)"
 trap '
     ret=$?;
     [[ $keep ]] && echo "Not removing $DRACUT_TMPDIR." >&2 || { [[ $DRACUT_TMPDIR ]] && rm -rf -- "$DRACUT_TMPDIR"; };
+    if [[ ${FSFROZEN} ]]; then
+      fsfreeze -u "${FSFROZEN}"
+    fi
     exit $ret;
     ' EXIT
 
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
You need fix your commit message so it passes the commisery check. Change it to  
```fix(dracut.sh): unfreeze /boot on exit``` and it should pass the check.

Note this probably wont get a full review until sometime after the weekend.

File: dracut.sh [hunk 1]
```diff
@@ -1009,6 +1009,9 @@ readonly DRACUT_TMPDIR="$(mktemp -p "$TMPDIR/" -d -t dracut.XXXXXX)"
 trap '
     ret=$?;
     [[ $keep ]] && echo "Not removing $DRACUT_TMPDIR." >&2 || { [[ $DRACUT_TMPDIR ]] && rm -rf -- "$DRACUT_TMPDIR"; };
+    if [ "x${FSFROZEN}" != "x" ]; then
```
haraldh, line 1012: could this be?

```suggestion
    if [[ ${FSFROZEN} ]]; then
```

haraldh (changes requested):

dracut.sh
tsipa: Done.

johannbg (approved):
LGTM

haraldh (approved):

PR #1127, opened 2021-02-24
Add a CI lint for formatting using indent and shfmt

haraldh:
* feat: customize .editorconfig according to shfmt
* chore: add .shellcheckrc
* fix: shfmt all the scripts
* fix: shfmt -s all the scripts
* fix: `make indent` for C
* ci: add lint for C and shell format


johannbg (commented):
Just a side note since we are adding a linter, there is an EditorConfig linter which might be worth considering adding as well. 
https://gitlab.com/greut/eclint
https://github.com/marketplace/actions/editorconfig-lint
Sample conf...
``` 
on: [...]
       ....
jobs:
 lint:
   runs-on: ubuntu-latest
   name: EditorConfig linter
   steps:
   — name: checkout
     uses: actions/checkout@v2
   — name: eclint-action
     uses: greut/eclint-action@v0
  ```

**Review by johannbg — CHANGES_REQUESTED** [review]
.

File: .editorconfig [hunk 1]
```diff
@@ -12,6 +12,12 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 charset = utf-8
+indent_style = space
```
johannbg, line 15: You forgot to add those same changes to .dir-local.el and .vimrc.

The things is when we make changes to .dir-locals.el, .vimrc and .editorconfig we have to propagate the changes we make to all three files since we cannot make the assumption that contributors have installed and or are using ( in this case ) EditorConfig plugins for Emacs/VIM.

configure
```diff
@@ -86,23 +86,23 @@ EOF
...
-		found=yes
-		break;
-	fi
+    # shellcheck disable=SC2086
+    ${CC} $CFLAGS $LDFLAGS conftest.c -Wl,$lib > /dev/null 2>&1
+    ret=$?
+    if test $ret -eq 0; then
+        FTS_LIBS="$lib"
+        found=yes
+        break
```
johannbg, line 95: Is the removal of the semicolon here ```break``` from the previous statement ```break;``` intentional or an oversight?

dracut-bash-completion.sh
```diff
@@ -15,83 +15,92 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-__contains_word () {
-        local word="$1"; shift
-        for w in "$@"; do [[ $w = "$word" ]] && return 0; done
-        return 1
+__contains_word() {
+    local word="$1"
```
johannbg, line 19: Is the removal of the semicolon here ```word="$1"``` from the previous statement ```word="$1";``` intentional or an oversight?

configure
haraldh: a semicolon at the end of a line is redundant in shell

dracut-bash-completion.sh
haraldh: a semicolon at the end of a line is redundant in shell

.editorconfig
haraldh: done

haraldh:
> Just a side note since we are adding a linter, there is an EditorConfig linter which might be worth considering adding as well.
> https://gitlab.com/greut/eclint
> https://github.com/marketplace/actions/editorconfig-lint
> Sample conf...
> 
> ```
> on: [...]
>        ....
> jobs:
>  lint:
>    runs-on: ubuntu-latest
>    name: EditorConfig linter
>    steps:
>    — name: checkout
>      uses: actions/checkout@v2
>    — name: eclint-action
>      uses: greut/eclint-action@v0
> ```

phew... @johannbg that opens another can of worms.. did you try to run `eclint`?

johannbg:
> phew... @johannbg that opens another can of worms.. did you try to run `eclint`?

Nope did not try it, just came across it. It sounded cool at that time but if the experience is that we would be walking through a minefield it is obviously out.

johannbg (approved):

haraldh:
@danimo ?

danimo (approved):
Ok, but see comments.

install/hashmap.c
```diff
@@ -101,7 +107,8 @@ int hashmap_ensure_allocated(Hashmap **h, hash_func_t hash_func, compare_func_t
         return 0;
 }
 
-static void link_entry(Hashmap *h, struct hashmap_entry *e, unsigned hash) {
+static void link_entry(Hashmap * h, struct hashmap_entry *e, unsigned hash)
```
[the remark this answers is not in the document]
haraldh: right... maybe we can switch to `astyle`

PR #1135, opened 2021-02-27
feat(nbd): support ipv6 link local nbds

TheJJ:
For https://github.com/SFTtech/stiefelsystem we use NBD over IPv6-linklocal. This was not supported in dracut.

## Changes
* Add v6 linklocal support for network module (already upstream in network manager)
* Add v6 support for NBD
* Some minor fixes discovered along the way

## Checklist

### DIFF
--- a/dracut.cmdline.7.asc
+++ b/dracut.cmdline.7.asc
@@ -575,7 +575,7 @@ USB Android phone::
 * enp0s29u1u2
 =====================
 
-**ip=**__{dhcp|on|any|dhcp6|auto6|either6|single-dhcp}__::
+**ip=**__{dhcp|on|any|dhcp6|auto6|either6|link6|single-dhcp}__::
     dhcp|on|any::: get ip from dhcp server from all interfaces. If root=dhcp,
     loop sequentially through all interfaces (eth0, eth1, ...) and use the first
     with a valid DHCP root-path.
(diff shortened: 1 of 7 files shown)


TheJJ:
One thing I'd like to also fix is the root switch using systemd when / is on NBD. Currently this somehow disconnects the nbd and then results in IO errors.

johannbg:
> One thing I'd like to also fix is the root switch using systemd when / is on NBD. Currently this somehow disconnects the nbd and then results in IO errors.

If you haven't already you might want to go through [1][2] as as starting point for that fix.

1. https://systemd.io/ROOT_STORAGE_DAEMONS/
2. https://systemd.io/INITRD_INTERFACE/

TheJJ:
Hm that's weird, I use `nbd-client -systemd-mark`, which should do exactly this and set `argv[0][0] = '@'`. But it seems to be killed anyway.

johannbg:
> Hm that's weird, I use `nbd-client -systemd-mark`, which should do exactly this and set `argv[0][0] = '@'`. But it seems to be killed anyway.

Noticed that you have removed one of the ```-``` as in in your PR you change ```--systemd-mark``` to ```-systemd-mark``` could that be the root cause?

johannbg:
@TheJJ also see https://github.com/NetworkBlockDevice/nbd/issues/91 upstream recently change their systemd type units.

johannbg:
> > Hm that's weird, I use `nbd-client -systemd-mark`, which should do exactly this and set `argv[0][0] = '@'`. But it seems to be killed anyway.
> 
> Noticed that you have removed one of the `-` as in in your PR you change `--systemd-mark` to `-systemd-mark` could that be the root cause?

Just went upstream and through the code and ```-systemd-mark``` is correct ```--systemd-mark``` was incorrect so ignore this.

johannbg:
@TheJJ maybe we should move this out of draft, review and merge and deal with the switch root issue in a separate PR or is there something else keeping this in draft mode?

TheJJ:
Indeed we can review/merge it, I'll remove the changed systemd-parts for now.

johannbg (approved):
LGTM

haraldh (approved):

PR #1143, opened 2021-03-05
fix: correctly handle kernel parameters

haraldh:
The kernel has an odd way to handle `"` surrounded parameters.
To handle the parameters as the kernel would do, no simple shell script
suffices, so a new utility `dracut-util` is introduced. Written in "C"
it handles `dracut-getarg` and `dracut-getargs` as the old shell script
functions `_dogetarg` and `_dogetargs` would.

### DIFF
--- a/test/TEST-98-GETARG/test.sh
+++ b/test/TEST-98-GETARG/test.sh
@@ -1,0 +1,152 @@
+#!/bin/bash
+
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+TEST_DESCRIPTION="dracut getarg command"
+
+test_check() {
+    return 0
+}
+
+test_setup() {
+    make -C "$basedir" dracut-util
+    ln -sfnr "$basedir"/dracut-util "$TESTDIR"/dracut-getarg
+    ln -sfnr "$basedir"/dracut-util "$TESTDIR"/dracut-getargs
+    ln -sfnr "$basedir"/modules.d/99base/dracut-lib.sh "$TESTDIR"/dracut-lib.sh
+    return 0
+}
+
+test_run() {
+    set -x
+    (
+        cd "$TESTDIR"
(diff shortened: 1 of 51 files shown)


test/TEST-99-GETARG/test.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
```
johannbg, line 2: Minor nitpick add proper header for new files ( even if it is an test ) it's missing
```
# This file is part of dracut.
# SPDX-License-Identifier: GPL-2.0-or-later
```

util/util.c
```diff
@@ -0,0 +1,317 @@
+// SPDX-License-Identifier: GPL-2.0
+
```
johannbg, line 2: Should this be GPL-2.0-or-later so a full header would look like this  
```
// This file is part of dracut.
// SPDX-License-Identifier: GPL-2.0-or-later
```
or is the C code GPL-2.0 only?

nl6720:
`make install` doesn't install `dracut-util`.

util/util.c
haraldh: yeah, the argument parsing code is copied from the kernel

test/TEST-99-GETARG/test.sh
haraldh: will add

haraldh:
> `make install` doesn't install `dracut-util`.

@nl6720 right.. fixed

test/TEST-98-GETARG/test.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
```
johannbg, line 2: Missed.
```
# This file is part of dracut.
# SPDX-License-Identifier: GPL-2.0-or-later
```

haraldh: done.. this time for real :)

johannbg (approved):

danimo (approved):

PR #1154, opened 2021-03-16
dracut-initramfs-restore: detect Arch Linux initramfs

gportay:
This pull request detects the Arch Linux initramfs for restoration.

## Changes

Arch Linux creates initramfs with the names `initramfs-linux{,-lts,-hardened,-zen-whatevervariant}.img`.

The name of the variant is stored in the file `/usr/lib/modules/$KERNEL_VERSION/pkgbase`:

```
$ cat /usr/lib/modules/*/pkgbase
linux-lts
linux
```

## Checklist
- [ ] ~~I am providing new code and test(s) for it~~

### DIFF
--- a/dracut-initramfs-restore.sh
+++ b/dracut-initramfs-restore.sh
@@ -13,6 +13,7 @@ SKIP="$dracutbasedir/skipcpio"
 [[ -x $SKIP ]] || SKIP=cat
 
 [[ -f /etc/machine-id ]] && read MACHINE_ID < /etc/machine-id
+[[ -f "/lib/modules/${KERNEL_VERSION}/pkgbase" ]] && read PKGBASE < "/lib/modules/${KERNEL_VERSION}/pkgbase"
 
 mount -o ro /boot &>/dev/null || true
 
@@ -24,6 +25,8 @@ elif [[ -d /boot/loader/entries || -L /boot/loader/entries ]] \
     && [[ $MACHINE_ID ]] \
     && [[ -d /boot/${MACHINE_ID} || -L /boot/${MACHINE_ID} ]] ; then
     IMG="/boot/${MACHINE_ID}/${KERNEL_VERSION}/initrd"
+elif [[ $PKGBASE ]] && [[ -f "/boot/initramfs-${PKGBASE}.img" ]]; then
+    IMG="/boot/initramfs-${PKGBASE}.img"
 else
     IMG="/boot/initramfs-${KERNEL_VERSION}.img"
 fi


johannbg:
Why does Arch Linux create the initramfs with all these variant names? What's the usecase for this?

johannbg (changes requested):
You need to fix your commit message so it passes the commisery check. 
Change it to... 
```
fix(dracut-initramfs-restore): detect Arch Linux initramfs
```

gportay:
> Why does Arch Linux create the initramfs with all these variant names? What's the usecase for this?

I am no an Arch maintainer, so I do not know honestly.

gportay:
> You need to fix your commit message so it passes the commisery check.
> Change it to...
> 
> ```
> fix(dracut-initramfs-restore): detect Arch Linux initramfs
> ```

Thanks, done

dracut-initramfs-restore.sh
```diff
@@ -13,6 +13,7 @@ SKIP="$dracutbasedir/skipcpio"
 [[ -x $SKIP ]] || SKIP=cat
 
 [[ -f /etc/machine-id ]] && read MACHINE_ID < /etc/machine-id
+[[ -f "/lib/modules/${KERNEL_VERSION}/pkgbase" ]] && read PKGBASE < "/lib/modules/${KERNEL_VERSION}/pkgbase"
```
codicodi, line 16: This isn't gonna work after a kernel update as `/lib/modules/$(uname -r)/pkgbase` won't exist

gportay: right, good catch; so I guess there is no simple way to get it after an update.

t-8ch:
@johannbg 
> Why does Arch Linux create the initramfs with all these variant names? What's the usecase for this?

The initramfs names match the kernel variant (package) they belong to.
For example: https://archlinux.org/packages/core/x86_64/linux/ , https://archlinux.org/packages/extra/x86_64/linux-hardened/
Each installed kernel variant only creates two initramfs images, one with machine-specific modules and one with all modules.

johannbg:
@t-8ch you did not answer my question I'm not asking for the how I'm asking for the why as in why does does Arch create the initramfs with all these variant names? What's the usecase for doing it this way? What problem does it solve for Arch? 
Why does Arch not use the standard `$ESP/initramfs-${KERNEL_VERSION}.img` or `$ESP/<machine-id>/<kernel-version>/initrd` 

Once we understand why Arch deviates from the rest of the distributions then we can see if we want to support that deviations here upstream.

t-8ch:
(Not an ArchLinux developer, so I can't provide authoritative answers)

I'm not sure where these locations have been defined as standard (at least when mkinitcpio has been written in 2006).
Both initramfs-tools from Debian and mkinitcpio from Arch use different locations than dracut.

Historically bootloaders have been configured manually on Arch, so having to adapt those configurations on kernel updates would have been an issue.
Why this was not worked around using symlinks, I can't answer.

gportay:
I do not know the exact reason but having the same filename for the initramfs through all upgrades of the Linux package may simplify things.

For example, you do not need to reconfigure the bootloader at every upgrade of the kernel (`grub-mkconfig -o /boot/grub/grub.cfg`). You run it once after you have picked up and install your bootloader and that is all till the end of time. However, the package manager continues to regenerate the initramfs after each upgrade of the package.

On Arch Linux, you install the package `linux`, or/and `linux-lts`, or/and any other alternatives. This package is not a meta-package that points to the latest version of Linux; it is the latest version of the alternative. Therefore, you always have a single version of the kernel.

This approach has drawbacks as pointed out by @codicodi, after a kernel upgrade, you lost the modules installed by the previous versions (there is an alternative for that with https://aur.archlinux.org/packages/kernel-modules-hook/).

gportay:
>  Why this was not worked around using symlinks, I can't answer.

Guess because some filesystems like FAT do not support symlinks.

gportay:
> Once we understand why Arch deviates from the rest of the distributions then we can see if we want to support that deviation here upstream.

Of course, there is an alternative solution if this one is too invasive as `/lib/modules/*/pkgbase` is an Arch specific file.

For example, dracut can call a script that echoes the initramfs path to stdout if none of the "standard" initrd filenames are found.

This way, it hides the specificity of distributions that will provide their own scripting to tell dracut the path to the initramfs image. Gentoo has probably its own way of naming its initramfs files.

johannbg:
The thing is it's like Arch is not appending it's variant name to the kernel version, at the kernel build time `( --append-to-version=-lts or -zen or -whatever )` for those variants and has resorted into somekind of hackish workarounds to fetch the kernel variant from a pkgbase file instead ( probably all over the place as in various install hooks and what not ) o_O 

What does the output from `uname -r`look on those variants? Does it contain the variant name?

gportay:
> What does the output from `uname -r`look on those variants? Does it contain the variant name?

```
[gportay@archlinux ~]$ uname -r
5.10.25-1-lts
```

gportay:
The `uname -r` reports the package release version (`-1`).

https://github.com/archlinux/svntogit-packages/blob/packages/linux-lts/trunk/PKGBUILD#L38-L41?

johannbg:
So Arch appends it's variant names to it's kernel builds which is good which brings us back to the question why is Arch creating `initramfs-linux{,-lts,-hardened,-zen-whatevervariant}.img` in it's naming scheme as opposed to `$ESP/initramfs-${KERNEL_VERSION}.img which ends up being (`initramfs-5.10.25-1-lts.img` ) , `( $ESP/<machine-id>/5.10.25-1-lts/initrd )`

`initramfs-linux` <-- this makes no sense unless you are also creating initramfs images on non linux systems which afaik Arch does not do so this could be removed.

`initramfs-linux-lts.img` <-- this wont work if you have multiple LTS kernels installed right thus you need the actual kernel release to distinguish between multiple installed kernels and their relevant initrd images. 

So it awfully looks like the proper place to fix this is in Arch not workaround and or this naming scheme supported in dracut unless there is some valid reason/usescase for the initrd image naming scheme as it's currently being done in Arch.

@grazzolini can you shed some lights why the initramfs images are being named as they are in Arch and weigh in on how you see or would like to see this being handled for Arch.

dracut-initramfs-restore.sh
grazzolini: pkgbase should always exist on the official Arch kernels

grazzolini:
Due to its rolling release nature, Arch doesn't have versioned kernels, therefore, it doesn't need to have versioned initramfs images either. Having said that, nothing prevents one from masking mknitcpio hook that copies the kernel to /boot and use something like kernel-install or something else entirely. Even copy the kernels versioned, if one wishes so.

Also, since we have 4 official kernels, and one can have all 4 installed simultaneously, the naming convention is to use the pkgbase to name both the kernel binary as well as the initramfs.

dracut-initramfs-restore.sh
gportay: indeed, but after an update, the parent directory (`unamre -r`) does not exist anymore.

haraldh:
With this PR https://github.com/dracutdevs/dracut/pull/1434 you can create your initrd to `/lib/modules/$KERNEL_VERSION/initrd` and sync it to /boot and restore will work as expected.

PR #1162, opened 2021-03-24
fix: Added _libexec_prefix variable to NM module-setup.sh for distros w/o libexec

### LINKED ISSUE #1161: network-manager has files missing on Arch Linux (+ other distros without /usr/libexec)
**Describe the bug**
Including the network-manager dracut module on Arch Linux doesn't work fully; as Arch doesn't use libexec, both of /usr/lib/nm-dhcp-helper and /usr/lib/nm-initrd-generator are left out of the initrd.

**Distribution used**
Arch Linux

**Dracut version**
dracut-053

**Init system**
systemd

**To Reproduce**
Add `add_dracutmodules+=" network-manager"` to dracut.conf/dracut.conf.d on an Arch system (or on any system missing /usr/libexec) and generate an initramfs.

**Expected behavior**
Including the network-manager dracut module and then generating the initramfs works just fine, with no missing files.

**Additional context**
This is easily worked around by a quick symlink from libexec to lib.

rosieeyes:
Added _libexec_prefix variable to NM module-setup.sh for distros (e.g. Arch Linux) that don't use /usr/libexec

This pull request changes...

## Changes

When added to the initramfs, the network-manager module will miss the nm-dhcp-helper and nm-initrd-generator files on distros such as Arch Linux that install files which would otherwise go into /usr/libexec, to /usr/lib.

This changes module-setup.sh to use /usr/lib if /usr/libexec doesn't exist.

This assumes that if the two NM files can't be found under /usr/libexec, that they'll be under /usr/lib instead, which I think is fair - it works for Arch Linux, at least.

I'm definitely welcome to hear better variable names than "_libexec_prefix".

## Checklist

Fixes #1161

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1736,6 +1736,11 @@ if ! [[ -d $dracutsysrootdir$tmpfilesdir ]]; then
     [[ -d $dracutsysrootdir/usr/lib/tmpfiles.d ]] && tmpfilesdir=/usr/lib/tmpfiles.d
 fi
 
+[[ -d $dracutsysrootdir$libexecdir ]] \
+    || libexecdir=$(pkg-config libexecdir --variable=libexecdir 2> /dev/null)
+
+[[ -d $dracutsysrootdir$libexecdir ]] || libexecdir=/usr/libexec
+
 export initdir dracutbasedir \
     dracutmodules force_add_dracutmodules add_dracutmodules omit_dracutmodules \
     mods_to_load \
@@ -1752,7 +1757,7 @@ export initdir dracutbasedir \
     systemdutildir systemdutilconfdir systemdcatalog systemdntpunits \
     systemdntpunitsconfdir systemdsystemunitdir systemdsystemconfdir \
     hostonly_cmdline loginstall \
-    tmpfilesdir
+    tmpfilesdir libexecdir
 
 mods_to_load=""
 # check all our modules to see if they should be sourced.
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -31,7 +31,7 @@ install() {
     inst_multiple ip sed grep
 
     inst NetworkManager
-    inst /usr/libexec/nm-initrd-generator
+    inst $libexecdir/nm-initrd-generator
     inst_multiple -o teamd dhclient
     inst_hook cmdline 99 "$moddir/nm-config.sh"
     if dracut_module_included "systemd"; then
@@ -46,7 +46,7 @@ install() {
     inst_simple "$moddir/nm-lib.sh" "/lib/nm-lib.sh"
 
     if [[ -x "$initdir/usr/sbin/dhclient" ]]; then
-        inst /usr/libexec/nm-dhcp-helper
+        inst $libexecdir/nm-dhcp-helper
     elif ! [[ -e "$initdir/etc/machine-id" ]]; then
         # The internal DHCP client silently fails if we
         # have no machine-id
--- a/modules.d/35network-manager/nm-lib.sh
+++ b/modules.d/35network-manager/nm-lib.sh
@@ -4,7 +4,7 @@ type getcmdline > /dev/null 2>&1 || . /lib/dracut-lib.sh
 
 nm_generate_connections() {
     rm -f /run/NetworkManager/system-connections/*
-    /usr/libexec/nm-initrd-generator -- $(getcmdline)
+    $libexecdir/nm-initrd-generator -- $(getcmdline)
 
     if getargbool 0 rd.neednet; then
         for i in /usr/lib/NetworkManager/system-connections/* \
--- a/modules.d/80cms/cmsifup.sh
+++ b/modules.d/80cms/cmsifup.sh
@@ -47,7 +47,7 @@ fi
 IFACES="$IFACES $DEVICE"
 echo "$IFACES" >> /tmp/net.ifaces
 
-if [ -x /usr/libexec/nm-initrd-generator ]; then
+if [ -x $libexecdir/nm-initrd-generator ]; then
     type nm_generate_connections > /dev/null 2>&1 || . /lib/nm-lib.sh
     nm_generate_connections
 else
--- a/modules.d/91crypt-gpg/module-setup.sh
+++ b/modules.d/91crypt-gpg/module-setup.sh
@@ -29,7 +29,7 @@ install() {
     if sc_requested; then
         inst_multiple gpg-agent
         inst_multiple gpg-connect-agent
-        inst_multiple /usr/libexec/scdaemon
+        inst_multiple $libexecdir/scdaemon
         cp "$dracutsysrootdir$(sc_public_key)" "${initdir}/root/"
     fi
 }
@@ -45,8 +45,8 @@ sc_supported() {
     if [[ ${gpgMajor} -gt 2 || ${gpgMajor} -eq 2 && ${gpgMinor} -ge 1 ]] \
         && require_binaries gpg-agent \
         && require_binaries gpg-connect-agent \
-        && require_binaries /usr/libexec/scdaemon \
-        && ($DRACUT_LDD "$dracutsysrootdir"/usr/libexec/scdaemon | grep libusb > /dev/null); then
+        && require_binaries $libexecdir/scdaemon \
+        && ($DRACUT_LDD "$dracutsysrootdir$libexecdir"/scdaemon | grep libusb > /dev/null); then
         return 0
     else
         return 1
--- a/modules.d/35network-wicked/module-setup.sh
+++ b/modules.d/35network-wicked/module-setup.sh
@@ -41,7 +41,7 @@ install() {
     inst_multiple "/etc/dbus-1/system.d/org.opensuse.Network*"
     inst_multiple "/usr/share/wicked/schema/*"
     inst_multiple "/usr/lib/wicked/bin/*"
-    inst_multiple "/usr/libexec/wicked/bin/*"
+    inst_multiple "$libexecdir/wicked/bin/*"
     inst_multiple "/usr/sbin/wicked*"
 
     wicked_units="
--- a/modules.d/40network/module-setup.sh
+++ b/modules.d/40network/module-setup.sh
@@ -19,7 +19,7 @@ depends() {
     if [ -z "$network_handler" ]; then
         if find_binary wicked &> /dev/null; then
             network_handler="network-wicked"
-        elif [[ -x $dracutsysrootdir/usr/libexec/nm-initrd-generator ]]; then
+        elif [[ -x $dracutsysrootdir$libexecdir/nm-initrd-generator ]]; then
             network_handler="network-manager"
         else
             network_handler="network-legacy"
--- a/modules.d/50plymouth/module-setup.sh
+++ b/modules.d/50plymouth/module-setup.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 pkglib_dir() {
-    local _dirs="/usr/lib/plymouth /usr/libexec/plymouth/"
+    local _dirs="/usr/lib/plymouth $libexecdir/plymouth/"
     if find_binary dpkg-architecture &> /dev/null; then
         _dirs+=" /usr/lib/$(dpkg-architecture -qDEB_HOST_MULTIARCH)/plymouth"
     fi


johannbg (changes requested):
You need to introduce this change as an global libexecdir variable not per module fix and then each module that uses /usr/libexec be fixed to use that global libexecdir variable

At that point each distro can just configure $libexecdir=<it's distro path> in it's dracut distribution configuration file

Basically the scope of this change is dracut.sh, 35network-manager, 40network, 80cms,91crypt-gpg, 35network-wicked,50plymouth modules along with a downstream patch to Arch default dracut.conf

rosieeyes:
Thanks for the input! I definitely agree that a global config option is best, but there are some modules that already handle both /usr/lib and /usr/libexec where necessary. Is it worth changing these instances when they're already working fine? Context below...

* [35network-manager](https://github.com/dracutdevs/dracut/blob/cc914eaa8b292e846a3a794abe9ef40db6eab7b5/modules.d/35network-manager/nm-lib.sh#L7) has another instance in which I'll change as discussed
* [35network-wicked](https://github.com/dracutdevs/dracut/blob/cc914eaa8b292e846a3a794abe9ef40db6eab7b5/modules.d/35network-wicked/module-setup.sh#L43) installs from both /usr/lib and /usr/libexec
* [40network](https://github.com/dracutdevs/dracut/blob/cc914eaa8b292e846a3a794abe9ef40db6eab7b5/modules.d/40network/module-setup.sh#L22) needs changing much like 35network-manager
* [50plymouth](https://github.com/dracutdevs/dracut/blob/cc914eaa8b292e846a3a794abe9ef40db6eab7b5/modules.d/50plymouth/module-setup.sh#L4) looks in both of /usr/lib and /usr/libexec already
* 80cms; I couldn't find any references to /usr/libexec there
* [91crypt-gpg](https://github.com/dracutdevs/dracut/blob/cc914eaa8b292e846a3a794abe9ef40db6eab7b5/modules.d/91crypt-gpg/module-setup.sh#L48) just uses /usr/libexec, so I'll change this

I'm also not keen on testing each module where I modify $libexecdir; I hope that just testing 35network-manager will suffice.

johannbg:
> Thanks for the input! I definitely agree that a global config option is best, but there are some modules that already handle both /usr/lib and /usr/libexec where necessary. Is it worth changing these instances when they're already working fine?

When introducing global variable we should replace all the existing instances to use that global variable instead.  

>     * 80cms; I couldn't find any references to /usr/libexec there

It's in cmsifup.sh
```
if [ -x /usr/libexec/nm-initrd-generator ]; then
...
```
Note when you add the global variable in dracut.sh which will be something along the lines of ... 

```
[[ -d $dracutsysrootdir$libexecdir ]] \
    || libexecdir=$(pkg-config libexecdir --variable=libexecdir 2> /dev/null)
[[ -d $dracutsysrootdir$libexecdir ]] || libexecdir=/usr/libexec
```

Dont forget to add your global variable to the export line
```
export initdir dracutbasedir \
....
tmpfilesdir <your global variable>
```

You also need to fix your commit messages to pass the commisery check which will be something along the lines of ... 

```
feat(dracut.sh): introducing <your global variable> 
....
```
for the dracut .sh file and for modules we dont include the module numbers since those are subjected to changes so it would look like something like... 

```
fix(network-manager): use <your global variable > instead of path
...
```
and 
```
fix(network-wicked): use <your global variable > instead of path
...
```
etc.

rosieeyes:
That's mostly everything taken care of, though I'm struggling to find where to add the necessary downstream changes to change libexecdir. I think it's best to add it as a part of ./configure, though I don't see where to make changes to facilitate that... it seems as though there's already been some work in the Makefile and configure files, but I don't see where to go next with it. I'll keep working on it, but feel free to direct me if you wish. :)

I also squashed everything together in order to fix the Commisery issue, but it seems that doesn't work. Does each file change need to be its own commit?

johannbg:
> That's mostly everything taken care of, though I'm struggling to find where to add the necessary downstream changes to change libexecdir from its default. I think it's best to add it as a part of ./configure, though I don't see where to make changes to facilitate that... it seems as though there's already been some work in the Makefile and configure files, but I don't see where to go next with it. I'll keep working on it, but feel free to direct me if you wish. :)

I simply suggest you wait a bit with that one until this is merged since we might end up actually using the systemd file-hierarchy which Arch actually defaults to so you might end up having to do nothing downstream. 

> 
> I also squashed everything together + amended the commit message in order to fix the Commisery issue, but it seems that doesn't work. Does each file change need to be its own commit?

Either that or better yet in two commits one for the feature ```feat: introducing libexecdir global variable``` and another squashed commits for all the fixes  under ```fix: use libexecdir global variable``` 

I suggest you read how to write a good commit message which can be found here [1] before fixing the commit message since you seem to be somewhat rushing it. 

This wont be merged until it has passed the second reviewer and he's not available until sometime tomorrow so there's no rush so just take your time.  

https://chris.beams.io/posts/git-commit/

rosieeyes:
> > That's mostly everything taken care of, though I'm struggling to find where to add the necessary downstream changes to change libexecdir from its default. I think it's best to add it as a part of ./configure, though I don't see where to make changes to facilitate that... it seems as though there's already been some work in the Makefile and configure files, but I don't see where to go next with it. I'll keep working on it, but feel free to direct me if you wish. :)
> 
> I simply suggest you wait a bit with that one until this is merged since we might end up actually using the systemd file-hierarchy which Arch actually defaults to so you might end up having to do nothing downstream.

Alrighty, is there a PR or issue for that? It'd be useful to keep tabs on and see how that progresses.

> Either that or better yet in two commits one for the feature `feat: introducing libexecdir global variable` and another squashed commits for all the fixes under `fix: use libexecdir global variable`
> 
> I suggest you read how to write a good commit message which can be found here [1] before fixing the commit message since you seem to be somewhat rushing it.
> 
> This wont be merged until it has passed the second reviewer and he's not available until sometime tomorrow so there's no rush so just take your time.
> 
> https://chris.beams.io/posts/git-commit/

Hah, you could tell I was rushing... I just thought this would be a quick fix I could push out on a night and have done with it, but here we are. I've never had to have my commit messages pass linting as much my code does; thanks for the reference.

johannbg:
> > I simply suggest you wait a bit with that one until this is merged since we might end up actually using the systemd file-hierarchy which Arch actually defaults to so you might end up having to do nothing downstream.
> 
> Alrighty, is there a PR or issue for that? It'd be useful to keep tabs on and see how that progresses.

Well it's in your PR the change would happen and it would be made by you. 

It's something we need to discuss among ourselves how it affects Fedora/OpenSuse if we do, so adjustment there can be made accordingly at the same time.

> Hah, you could tell I was rushing... 

Well I have been uhum guilty of doing it myself couple of times in the past so I know all the signs :smile:

modules.d/35network-manager/nm-lib.sh
```diff
@@ -4,7 +4,7 @@ type getcmdline > /dev/null 2>&1 || . /lib/dracut-lib.sh
 
 nm_generate_connections() {
     rm -f /run/NetworkManager/system-connections/*
-    /usr/libexec/nm-initrd-generator -- $(getcmdline)
+    $libexecdir/nm-initrd-generator -- $(getcmdline)
```
haraldh, line 7:
```suggestion
    "$libexecdir"/nm-initrd-generator -- $(getcmdline)
```

modules.d/35network-manager/module-setup.sh
```diff
@@ -46,7 +46,7 @@ install() {
     inst_simple "$moddir/nm-lib.sh" "/lib/nm-lib.sh"
 
     if [[ -x "$initdir/usr/sbin/dhclient" ]]; then
-        inst /usr/libexec/nm-dhcp-helper
+        inst $libexecdir/nm-dhcp-helper
```
haraldh, line 49:
```suggestion
        inst "$libexecdir"/nm-dhcp-helper
```

modules.d/35network-manager/module-setup.sh
```diff
@@ -31,7 +31,7 @@ install() {
     inst_multiple ip sed grep
 
     inst NetworkManager
-    inst /usr/libexec/nm-initrd-generator
+    inst $libexecdir/nm-initrd-generator
```
haraldh, line 34:
```suggestion
    inst "$libexecdir"/nm-initrd-generator
```

modules.d/80cms/cmsifup.sh
```diff
@@ -47,7 +47,7 @@ fi
 IFACES="$IFACES $DEVICE"
 echo "$IFACES" >> /tmp/net.ifaces
 
-if [ -x /usr/libexec/nm-initrd-generator ]; then
+if [ -x $libexecdir/nm-initrd-generator ]; then
```
haraldh, line 50: there is no `$libexecdir` in the initramfs root... yet

modules.d/91crypt-gpg/module-setup.sh
```diff
@@ -29,7 +29,7 @@ install() {
     if sc_requested; then
         inst_multiple gpg-agent
         inst_multiple gpg-connect-agent
-        inst_multiple /usr/libexec/scdaemon
+        inst_multiple $libexecdir/scdaemon
```
haraldh, line 32:
```suggestion
        inst_multiple "$libexecdir"/scdaemon
```

modules.d/91crypt-gpg/module-setup.sh
```diff
@@ -45,8 +45,8 @@ sc_supported() {
     if [[ ${gpgMajor} -gt 2 || ${gpgMajor} -eq 2 && ${gpgMinor} -ge 1 ]] \
         && require_binaries gpg-agent \
         && require_binaries gpg-connect-agent \
-        && require_binaries /usr/libexec/scdaemon \
-        && ($DRACUT_LDD "$dracutsysrootdir"/usr/libexec/scdaemon | grep libusb > /dev/null); then
+        && require_binaries $libexecdir/scdaemon \
```
haraldh, line 48:
```suggestion
        && require_binaries "$libexecdir"/scdaemon \
```

modules.d/35network-manager/nm-lib.sh
haraldh: .. but... there is no $libexecdir in the initramfs root... yet

haraldh (changes requested):
I would prefer to install all `$libexedir` to a standard location like `/usr/bin`, so that we don't have to distinguish distro specific stuff in the initramfs runtime...

E.g. install `nm-initrd-generator` to `/usr/bin` and call it from there, instead of introducing a `libexecdir` global variable in the initramfs to be read at runtime.

johannbg:
> I would prefer to install all `$libexedir` to a standard location like `/usr/bin`, so that we don't have to distinguish distro specific stuff in the initramfs runtime...

Given the needless fragmentation on the core/baseOS layer between distribution I dont think we want to end up with a situation where paths in the initramfs differ from the vendor/distribution so we have to provide means to accommodate for it, which is best done using global variables as much as possible in modules. 

> 
> E.g. install `nm-initrd-generator` to `/usr/bin` and call it from there, instead of introducing a `libexecdir` global variable in the initramfs to be read at runtime.

The purpose of `/usr/libexec` was to host internal binaries that are not intended to be executed directly by users or shell scripts so they dont belong in /usr/bin 

FHS and SD-FHS refer to `/usr/lib` should be used for that purpose ( FHS states that `/usr/libexec` can also "optionally" be used for that purpose  ) so we should be defaulting to use `/usr/lib` for this not `/usr/bin` to be following the FHS's.

Conan-Kudo:
Both openSUSE and Debian have since migrated to `/usr/libexec`, so I'm a little surprised Arch hasn't made that change too.

dracut.sh
```diff
@@ -1736,6 +1736,11 @@ if ! [[ -d $dracutsysrootdir$tmpfilesdir ]]; then
     [[ -d $dracutsysrootdir/usr/lib/tmpfiles.d ]] && tmpfilesdir=/usr/lib/tmpfiles.d
 fi
 
+[[ -d $dracutsysrootdir$libexecdir ]] \
+    || libexecdir=$(pkg-config libexecdir --variable=libexecdir 2> /dev/null)
+
+[[ -d $dracutsysrootdir$libexecdir ]] || libexecdir=/usr/libexec
```
haraldh, line 1742: There is no such thing as a `libexecdir` pkg-config...

```suggestion
if ! [[ -d $dracutsysrootdir$libexecdir ]]; then
    if [[ -d $dracutsysrootdir/usr/libexec ]]; then
        libexecdir=/usr/libexec
    else
        libexecdir=/usr/lib
    fi
fi    
```

haraldh (changes requested):

haraldh:
Fixed with https://github.com/dracutdevs/dracut/pull/1480 and https://github.com/dracutdevs/dracut/pull/1473

PR #1163, opened 2021-03-26
[RFC] Add kdump tests to keep dracut change from breaking kexec-tools

coiby:
kexec-tools heavily depends on dracut. However, the change of dracut occasionally breaks. For example, recently it's found that
 - commit 5840c466dcfe1c4322142980cac6606870bd586f  ("95nfs: /var/lib/nfs/statd/sm is /var/lib/nfs/sm on SUSE")  caused the failure of dumping vmcore to nfs (v3) target
 - commit 828feae4f1814a915b2f7f362a5920322e0d6fcc ("dracut.spec: remove unnecessary dependencies") caused the dumping failure when a NIC is configured a static IP
 - [/bin/bash is lost when calling dracut to re-generate initramfs of kdump.](https://bugzilla.redhat.com/show_bug.cgi?id=1936781)

## Changes

This new test will,
 - Build the dracut rpm package 
 - Clone kexec-tools 
 - Pass the build rpm to kexec-tools' testcases

## Checklist

### DIFF
--- a/.github/workflows/fedora-33.yml
+++ b/.github/workflows/fedora-33.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: quay.io/haraldh/dracut-fedora:33
-      options: "-v /dev/kvm:/dev/kvm"
+      options: "-v /dev/kvm:/dev/kvm -v/dev:/dev --privileged -v/lib/modules:/lib/modules:ro"
     timeout-minutes: 45
     strategy:
       matrix:
@@ -37,6 +37,7 @@ jobs:
           "36",
           "40",
           "41",
+          "96",
           "98",
         ]
       fail-fast: false
--- a/fedora-test-github.sh
+++ b/fedora-test-github.sh
@@ -23,7 +23,7 @@ if ! [[ $TESTS ]]; then
     [[ -d .git ]] && git fetch --tags && git describe --tags
     make -j "$NCPU" all syncheck rpm logtee
 else
-    if [[ $TESTS == "99" ]]; then
+    if [[ $TESTS == "99" || $TESTS == "96" ]]; then
         [[ -d .git ]] && git fetch --tags && git describe --tags
         make_docs=yes
     else
--- a/fedora-test.sh
+++ b/fedora-test.sh
@@ -14,7 +14,7 @@ NCPU=$(getconf _NPROCESSORS_ONLN)
 if ! [[ $TESTS ]]; then
     make -j$NCPU all syncheck rpm logtee
 else
-    [[ $TESTS == "99" ]] && make_docs=yes || make_docs=no
+    [[ $TESTS == "99" || $TESTS == "96" ]] && make_docs=yes || make_docs=no
     make -j$NCPU enable_documentation=$make_docs all logtee
 
     cd test
--- a/test/TEST-96-KEXEC-TOOLS/Makefile
+++ b/test/TEST-96-KEXEC-TOOLS/Makefile
@@ -1,0 +1,1 @@
+-include ../Makefile.testdir
--- a/test/TEST-96-KEXEC-TOOLS/test.sh
+++ b/test/TEST-96-KEXEC-TOOLS/test.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+TEST_DESCRIPTION="kexec-tools test"
+
+KVERSION=${KVERSION-$(uname -r)}
+
+KEXEC_TOOLS_SRC="kexec-tools"
+test_run() {
+    # kexec-tools also use TESTDIR variable, recovert it for kexec-tools
+    KEXEC_TOOLS_TESTDIR="$TESTDIR/$KEXEC_TOOLS_SRC/tests"
+    make -C "$KEXEC_TOOLS_TESTDIR" TESTDIR="$KEXEC_TOOLS_TESTDIR" EXTRA_RPMS=$TESTDIR/*.rpm test-run V=$V || return 1
+    return 0
+}
+
+test_setup() {
+    # Buidling dracut rpm requires rpm-build
+    dnf -y rpm-build
+    make -C "$basedir" DESTDIR="$TESTDIR/" rpm V=$V || return 1
+    # dnf would bail out when asked to install a source rpm package during kexec-tools testing
+    rm -f "$TESTDIR"/*.src.rpm
+    git clone --branch rawhide https://src.fedoraproject.org/rpms/kexec-tools.git "$TESTDIR/$KEXEC_TOOLS_SRC" || return 1
+    # kexec-tools needs the following packages
+    dnf -y install qemu wget qemu-img fedpkg dnf-utils || return 1
+    yum-builddep kexec-tools || return 1
+    return 0
+}
+
+test_cleanup() {
+    rm -fr -- "$TESTDIR"/*.rpm
+    return 0
+}
+
+. $testdir/test-functions


johannbg (changes requested):
Arguably this test belongs with kdump since kdump depends on dracut not the other way around and or should be done downstream in Fedora ( maybe triggered via packit? ). 

Anyway you need to fix your commit message. 
Change this
`
test: add kexec-tools tests
`
To this
`
ci: add kexec-tools tests
`
And this 
`
Enable TEST-96-KEXEC-TOOLS
`
To this
`
ci: enable TEST-96-KEXEC-TOOLS
`

coiby:
> Arguably this test belongs with kdump since kdump depends on dracut not the other way around and or should be done downstream in Fedora ( maybe triggered via packit? ).
>

Thanks for your sharing your thoughts! Yes, it's a reverse-gating test. One goal my team wants to achive to catch the PRs that  break kexec-tools before they gets merged. I don't know much about packit. Does packit support monitoring PRs and testing them? 
 
> Anyway you need to fix your commit message.
> Change this
> `test: add kexec-tools tests`
> To this
> `ci: add kexec-tools tests`
> And this
> `Enable TEST-96-KEXEC-TOOLS`
> To this
> `ci: enable TEST-96-KEXEC-TOOLS`

Thanks for provding the fixes!

haraldh:
Looking at https://src.fedoraproject.org/rpms/kexec-tools/blob/rawhide/f/tests/scripts you could have integrated some of  these tests in dracut upstream and even help to refactor dracuts test suite.

I think the basics they rely on, should just be included in dracut upstream in the test suite so we can catch the dracut regressions immediately. And without getting development test failures of kdump upstream.

coiby:
> Looking at https://src.fedoraproject.org/rpms/kexec-tools/blob/rawhide/f/tests/scripts you could have integrated some of these tests in dracut upstream and even help to refactor dracuts test suite.
>  
Can you elaborate on which test cases need to be refactored?

> I think the basics they rely on, should just be included in dracut upstream in the test suite so we can catch the dracut regressions immediately. And without getting development test failures of kdump upstream.

Thanks for suggesting integrating some of the tests in dracut upstream. In the first I did try this approach. The reasons I don't adopt it or defer it are as follows,
1. kexec-tools also run these kdump tests, it's better not to duplicate the tests in dracut in order to reduce maintenance efforts.
2. kexec-tools' test suite share some of the code with dracut's but not much. The kdump tests relies on Fedora Cloud Base Image whereas dracut's tests (e.g. TEST-20-NFS) build the disk image on the fly. So the code related to installing a package or a file is not shared. 
3. This PR is the fastest way to achieve the goal of preventing dracut changes from breaking kdump. Personally, I love dracut's style of building a minimum disk image and I'd like to try if this approach works for kdump tests. But I don't know how much efforts are needed so I want to do this later.

coiby:
> > Arguably this test belongs with kdump since kdump depends on dracut not the other way around and or should be done downstream in Fedora ( maybe triggered via packit? ).
> 
> Thanks for your sharing your thoughts! Yes, it's a reverse-gating test. One goal my team wants to achive to catch the PRs that break kexec-tools before they gets merged. I don't know much about packit. Does packit support monitoring PRs and testing them?
> 
I find dracut CI has already included packit to run tests when there are PRs:)

johannbg:
> I find dracut CI has already included packit to run tests when there are PRs:)

Yes we already use packit do some basic build tests on all Fedora releases ( as does Anaconda and systemd which is ensuring that the installer + initrd + init is covered ) which we could extend to make use of the "testing-farm" ( run tests in the test directory ) [1] but you need to somewhat set that up, downstream in the Fedora repo for the component.

1. https://packit.dev/docs/testing-farm/#enable-testing

coiby:
> > I find dracut CI has already included packit to run tests when there are PRs:)
> 
> Yes we already use packit do some basic build tests on all Fedora releases ( as does Anaconda and systemd which is ensuring that the installer + initrd + init is covered ) which we could extend to make use of the "testing-farm" ( run tests in the test directory ) [1] but you need to somewhat set that up, downstream in the Fedora repo for the component.
> 
> 1. https://packit.dev/docs/testing-farm/#enable-testing

Correct me if I am wrong, but hasn't dracut enabled testing-farm? In order to catch the dracut PRs that break kdump before they get merged, I think the tests in the "testing-farm" should be added to dracut.

coiby:
> > I find dracut CI has already included packit to run tests when there are PRs:)
> 
> Yes we already use packit do some basic build tests on all Fedora releases ( as does Anaconda and systemd which is ensuring that the installer + initrd + init is covered ) which we could extend to make use of the "testing-farm" ( run tests in the test directory ) [1] but you need to somewhat set that up, downstream in the Fedora repo for the component.
> 
> 1. https://packit.dev/docs/testing-farm/#enable-testing

@johannbg I got confirmation from Packit&tmt developers [2] that tmt lacks some features to be able to run these dump tests,
 - "Currently, for Fedora CI and Packit, we use AWS EC2 instances, which do not support nested virtualization"
 - Launching multiple VM instances is not supported yet 

If the second feature gets implemented, the kdump tests could run without AWS EC2 supporting nested virtualization. I will follow up with tmt developers to know how and when the second feature gets implemented. 

[2] https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/thread/NCHTFDUF2X6P553OBYTHYELSZ7U2RUFR/

coiby:
> > Looking at https://src.fedoraproject.org/rpms/kexec-tools/blob/rawhide/f/tests/scripts you could have integrated some of these tests in dracut upstream and even help to refactor dracuts test suite.
> 
> Can you elaborate on which test cases need to be refactored?
> 
> > I think the basics they rely on, should just be included in dracut upstream in the test suite so we can catch the dracut regressions immediately. And without getting development test failures of kdump upstream.
> 
> Thanks for suggesting integrating some of the tests in dracut upstream. In the first I did try this approach. The reasons I don't adopt it or defer it are as follows,
> 
> 1. kexec-tools also run these kdump tests, it's better not to duplicate the tests in dracut in order to reduce maintenance efforts.
> 2. kexec-tools' test suite share some of the code with dracut's but not much. The kdump tests relies on Fedora Cloud Base Image whereas dracut's tests (e.g. TEST-20-NFS) build the disk image on the fly. So the code related to installing a package or a file is not shared.
> 3. This PR is the fastest way to achieve the goal of preventing dracut changes from breaking kdump. Personally, I love dracut's style of building a minimum disk image and I'd like to try if this approach works for kdump tests. But I don't know how much efforts are needed so I want to do this later.

@haraldh How you think about running the tests with Packit? One great benefit of using Packit is I can run the sames tests for both upstream and downstream kexec-tools and dracut (commits backported from upstream dracut have also broken download stream kexec-tools). So instead of integrating some functions of https://src.fedoraproject.org/rpms/kexec-tools/blob/rawhide/f/tests/scripts into dracut's testing framework, another way is to integrate them into Packit&tmt .

haraldh:
> @haraldh How you think about running the tests with Packit? One great benefit of using Packit is I can run the sames tests for both upstream and downstream kexec-tools and dracut (commits backported from upstream dracut have also broken download stream kexec-tools). So instead of integrating some functions of https://src.fedoraproject.org/rpms/kexec-tools/blob/rawhide/f/tests/scripts into dracut's testing framework, another way is to integrate them into Packit&tmt .

Yes, packit is fine

coiby:
@haraldh @johannbg 

Currently Packit lackes [the feature of multihost tests](https://github.com/psss/tmt/issues/726) to run the kdump tests and isn't going to be implemented soon. Red Hat has an internal idm-ci to support mulithost tests but I don't know how easily can it be integrated into Packit or Github Action. Considering kexec-tools needs also to run the kdump tests, another idea is to make a RPM subpackage for the kdump test framework which supports maminupating the disk images and launching multiple QEMU VM instances to run a test. So the kdump test framework can be shared by both dracut and kexec-tools. What do you think of this idea?

coiby:
@haraldh After having a discussion with my team members, we think we can also move the kdump testing frame into dracut instead since kexec-tools can also make use of dracut's updated test framework to run kdump tests. If you are in favor of this idea, I'm going to implement it.

LaszloGombos:
> Yes, packit is fine

So, packit has been now removed
https://github.com/dracutdevs/dracut/commit/8fd37d20f4b7cc08ee0970e0249aac4cd5b47a4e

> Considering kexec-tools needs also to run the kdump tests, another idea is to make a RPM subpackage

Would it be possible to install kexec-tools into all the supported test containers - https://github.com/dracutdevs/dracut/tree/master/test/container (using the disto's native package manager), instead of relying on rpm/dnf for this ? 

Perhaps just start with the the "local-kdump" test first and than as a second PR include more advanced test cases ?

johannbg:
Yup in the end of the day it turned out that packit is not worth using.
So @coiby what's the current status with this, does the team still want to move forward with this ( if so then you and @LaszloGombos who owns the test suite need to work together for a mutual solution ) or do you want to close this PR?

johannbg:
I'm going to go ahead and close this PR it can always be re-open if the intent is to move on with it.

PR #1164, opened 2021-03-26
fix(base): shellcheck for modules.d/99base

haraldh:
Also remove some functions, which are not used and broken anyway.

### DIFF
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -3,9 +3,10 @@
 export DRACUT_SYSTEMD
 export NEWROOT
 if [ -n "$NEWROOT" ]; then
-    [ -d $NEWROOT ] || mkdir -p -m 0755 $NEWROOT
+    [ -d "$NEWROOT" ] || mkdir -p -m 0755 "$NEWROOT"
 fi
 
+# shellcheck disable=SC2153
 if [ -z "$PREFIX" ]; then
     if ! [ -d /run/initramfs ]; then
         mkdir -p -m 0755 /run/initramfs/log
@@ -71,8 +72,9 @@ if [ -z "$DRACUT_SYSTEMD" ]; then
     info() {
         check_quiet
         echo "<30>dracut: $*" > /dev/kmsg
-        [ "$DRACUT_QUIET" != "yes" ] \
-            && echo "dracut: $*" >&2 || :
+        if [ "$DRACUT_QUIET" != "yes" ]; then
+            echo "dracut: $*" >&2
+        fi
     }
 
 else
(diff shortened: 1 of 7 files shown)


modules.d/99base/dracut-lib.sh
```diff
@@ -844,11 +836,11 @@ add_mount_point() {
 wait_for_mount() {
     local _name
     _name="$(str_replace "$1" '/' '\\x2f')"
-    printf '. /lib/dracut-lib.sh\nismounted "%s"\n' $1 \
+    printf '. /lib/dracut-lib.sh\nismounted "%s"\n' "$1" \
         >> "$hookdir/initqueue/finished/ismounted-${_name}.sh"
     {
-        printf 'ismounted "%s" || ' $1
-        printf 'warn "\"%s\" is not mounted"\n' $1
+        printf 'ismounted "%s" || ' "$1"
```
johannbg, line 842: Should not this be

`
printf ' "\"%s\" is mounted || ' "$1"
`
or
`
printf ' "\"%s\" already mounted || ' "$1"
`
to be inline with the following
`
printf 'warn "\"%s\" is not mounted"\n' "$1"
`
?

modules.d/99base/init.sh
```diff
@@ -16,20 +16,18 @@ PATH=/usr/sbin:/usr/bin:/sbin:/bin
...
 # mount some important things
-[ ! -d /proc/self ] \
-    && mount -t proc -o nosuid,noexec,nodev proc /proc > /dev/null
-
-if [ "$?" != "0" ]; then
-    echo "Cannot mount proc on /proc! Compile the kernel with CONFIG_PROC_FS!"
-    exit 1
+if [ ! -d /proc/self ]; then
+    if ! mount -t proc -o nosuid,noexec,nodev proc /proc > /dev/null; then
+        echo "Cannot mount proc on /proc! Compile the kernel with CONFIG_PROC_FS!"
```
johannbg, line 21: Probably a better wording here is 
`
echo "Cannot mount sysfs on /sys! Perhaps the kernel is not compiled with CONFIG_SYSFS?"
`
A subtle hint in a form of suggestion is better perceived than orders right.

johannbg (approved):

modules.d/99base/dracut-lib.sh
haraldh: no.. inside `'` we don't have to quote `"`

modules.d/99base/init.sh
haraldh: created an issue: https://github.com/dracutdevs/dracut/issues/1260

PR #1110, opened 2021-02-16
feat: simplify global vars

haraldh:
WIP for the simplification of global variables.

PR #1174, opened 2021-03-26
fix(systemd-modules-load): shellcheck for modules.d/01systemd-modules…

haraldh:
…-load

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-modules-load/module-setup.sh
+++ b/modules.d/01systemd-modules-load/module-setup.sh
@@ -6,7 +6,7 @@
 check() {
 
     # If the binary(s) requirements are not fulfilled the module can't be installed
-    require_binaries $systemdutildir/systemd-modules-load || return 1
+    require_binaries "$systemdutildir"/systemd-modules-load || return 1
 
     # Return 255 to only include the module, if another module requires it.
     return 255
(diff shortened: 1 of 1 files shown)


johannbg (approved):

johannbg (approved):

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -27,20 +27,19 @@ depends() {
 install() {
 
     # Create systemd-modules-load related directories.
-    inst_dir $modulesload
-    inst_dir $modulesloadconfdir
+    inst_dir "$modulesload"
+    inst_dir "$modulesloadconfdir"
```
haraldh, line 31: These global variables don't exist.

danimo: Any idea where this comes from then? Otherwise let's :fire: it.

johannbg: @danimo it's a needed global variable which is missing since that global variable ended up as part of another sd-module work. Work which I put on ice until @haraldh system wide fixes, which touch a large chunk of the code base settles down. ( we are just signing ourselves up for merge conflicts otherwise ) which hopefully is done now for the time being.
@haraldh has created issues for this and assigned it to me to clean up ( issues #1262 #1263 ) and that work kinda depends on @haralds draft PR #1110 being finished as in it's better that fix it there after we merged that work.

danimo: @johannbg So we keep this as-is for now?

johannbg: @danimo if you mean "keep this as-is" as being "approve @haraldh PR and merge as is" then the answer is yes.

haraldh: Hmm, it was not working before and not working after this patch... So, I don't know.

danimo: Create an issue and I'll approve. Let's not change too much at once in a single commit.

johannbg: @danimo @haraldh already created an issue for missing global variables in SD modules and assigned it to me.

PR #1190, opened 2021-03-26
fix(network-legacy): shellcheck for modules.d/35network-legacy

haraldh:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-legacy/parse-team.sh
+++ b/modules.d/35network-legacy/parse-team.sh
@@ -52,11 +52,13 @@ for team in $(getargs team); do
 
     parseteam "$team" || continue
 
-    echo "teammaster=$teammaster" > /tmp/team.${teammaster}.info
-    echo "teamslaves=\"$teamslaves\"" >> /tmp/team.${teammaster}.info
-    echo "teamrunner=\"$teamrunner\"" >> /tmp/team.${teammaster}.info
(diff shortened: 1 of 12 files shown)


modules.d/35network-legacy/parse-team.sh
```diff
@@ -52,11 +52,11 @@ for team in $(getargs team); do
 
     parseteam "$team" || continue
 
-    echo "teammaster=$teammaster" > /tmp/team.${teammaster}.info
-    echo "teamslaves=\"$teamslaves\"" >> /tmp/team.${teammaster}.info
-    echo "teamrunner=\"$teamrunner\"" >> /tmp/team.${teammaster}.info
+    echo "teammaster=$teammaster" > /tmp/team."${teammaster}".info
```
danimo, line 55: Why not `{} >` group here as well?

danimo (changes requested):

modules.d/35network-legacy/parse-team.sh
haraldh: good point! will do

haraldh: done

johannbg (approved):

danimo (approved):

PR #1203, opened 2021-03-26
fix(crypt): shellcheck for modules.d/90crypt

haraldh:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -2,9 +2,9 @@
 
 # called by dracut
 check() {
-    local _rootdev
+    local fs
     # if cryptsetup is not installed, then we cannot support encrypted devices.
-    require_any_binary $systemdutildir/systemd-cryptsetup cryptsetup || return 1
+    require_any_binary "$systemdutildir"/systemd-cryptsetup cryptsetup || return 1
 
     [[ $hostonly ]] || [[ $mount_needs ]] && {
         for fs in "${host_fs_types[@]}"; do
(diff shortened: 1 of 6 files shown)


modules.d/90crypt/module-setup.sh
```diff
@@ -33,20 +33,18 @@ installkernel() {
     [[ $hostonly ]] || [[ $mount_needs ]] && {
         # dmsetup returns s.th. like
         # cryptvol: 0 2064384 crypt aes-xts-plain64 :64:logon:cryptsetup:....
-        dmsetup table | while read name _ _ is_crypt cipher _; do
-            [[ $is_crypt != "crypt" ]] && continue
+        dmsetup table | while read -r name _ _ is_crypt cipher _; do
+            [[ $is_crypt == "crypt" ]] || continue
```
johannbg, line 37: We might want to document this preferred? evaluation method someplace ( coding style )

`[[ $is_crypt != "crypt" ]] && continue` 
vs
 `[[ $is_crypt == "crypt" ]] || continue`

johannbg (approved):

modules.d/90crypt/module-setup.sh
haraldh: yeah.. positiv testing is preferred here

```
 [[ $is_crypt == "crypt" ]] || continue
```
as it catches the error case here.
`$?` is zero for the following commands

or
```
if [[ $is_crypt != "crypt" ]]; then
    continue
fi
```

danimo (approved):

PR #1204, opened 2021-03-26
fix(dm): shellcheck for modules.d/90dm

haraldh:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/90dm/dm-shutdown.sh
+++ b/modules.d/90dm/dm-shutdown.sh
@@ -5,19 +5,19 @@ _remove_dm() {
     local s
     local devname
 
-    for s in /sys/block/${dev}/holders/dm-*; do
-        [ -e ${s} ] || continue
-        _remove_dm ${s##*/}
+    for s in /sys/block/"${dev}"/holders/dm-*; do
+        [ -e "${s}" ] || continue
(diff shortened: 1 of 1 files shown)


johannbg (approved):

modules.d/90dm/dm-shutdown.sh
```diff
@@ -47,7 +47,7 @@ _do_dm_shutdown() {
 
 if command -v dmsetup > /dev/null \
     && [ "x$(dmsetup status)" != "xNo devices found" ]; then
-    _do_dm_shutdown $1
+    _do_dm_shutdown "${1-$final}"
```
danimo, line 50: This changes behavior. Document in commit msg / Split out into separate commit?

danimo (changes requested):

modules.d/90dm/dm-shutdown.sh
haraldh: yeah, let me revert this .. it was made under false assumptions

haraldh: reverted

johannbg (approved):

danimo (approved):

PR #1228, opened 2021-03-26
fix(fcoe): shellcheck for modules.d/95fcoe

haraldh:
This pull request changes...

## Checklist

Fixes #


johannbg (approved):

danimo (changes requested):
The removal of fcoe-genrules is more than satisfying shellcheck, right? If so, the commit message should be adjusted.

haraldh:
> The removal of fcoe-genrules is more than satisfying shellcheck, right? If so, the commit message should be adjusted.

True.. I just inserted the contents right in the calling source and since it's not called anywhere else...

danimo (approved):

haraldh:
hmpf... auto-merge hit

PR #1231, opened 2021-03-26
fix(iscsi): shellcheck for modules.d/95iscsi

haraldh:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/95iscsi/parse-iscsiroot.sh
+++ b/modules.d/95iscsi/parse-iscsiroot.sh
@@ -84,7 +84,7 @@ if [ -n "$iscsi_firmware" ]; then
     modprobe -b -q iscsi_boot_sysfs 2> /dev/null
     modprobe -b -q iscsi_ibft
     # if no ip= is given, but firmware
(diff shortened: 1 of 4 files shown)


johannbg (approved):

modules.d/95iscsi/parse-iscsiroot.sh
```diff
@@ -47,13 +47,6 @@ if [ "${root%%:*}" = "iscsi" ]; then
     write_fs_tab /dev/root
 fi
 
-# If it's not empty or iscsi we don't continue
```
danimo, line 50: Is this change (with the check in 137) functionally equivalent or does it change behavior?

danimo (changes requested):
See comment. Good otherwise.

modules.d/95iscsi/parse-iscsiroot.sh
haraldh: I will double check again, after clearing my head with all the shellcheck

haraldh: reverted the logic checks

haraldh: opened https://github.com/dracutdevs/dracut/issues/1279

danimo (approved):

PR #1265, opened 2021-03-26
docs: add docs/BASH.md

haraldh:
Add a little document providing help in common bash problems.


johannbg (approved):
Nice! 

Good bash base for in the future CODINGSTYLE.md

PR #1280, opened 2021-03-29
dbus-broker: enable the service

bengal:
dbus-broker.service has a 'dbus.service' alias which is installed when the service gets enabled.

If the alias is not present in the initrd image, services requiring D-Bus in the initrd fail to start because they depend on dbus.service, which doesn't exist.

Therefore, enable the service to create the alias.

## Checklist


johannbg (changes requested):
You need to fix your commit message to pass the commisery check.

Change it from
`dbus-broker: enable the service`
to
`fix(dbus-broker): enable the service`

bengal:
> You need to fix your commit message to pass the commisery check.

Done, thanks.

haraldh (approved):

johannbg (approved):

PR #1281, opened 2021-03-29
network-manager: run as daemon with D-Bus

bengal:
This is version 2 of #927. I did only some basic testing on it; I plan to do more tests in the next days. In the meantime, comments are welcome.

This commit changes how NM is started inside the initrd. Instead of running NM in the special `--configure-and-quit=initrd` mode, which sets up network and quits, start it as a daemon.

This has multiple advantages. First, we no longer need to run NM in a special mode that requires additional code and maintenance. NetworkManager works exactly as in the real root.

One problem of the current configure-and-quit approach is that once NM has quit, dynamic addresses can expire if the initrd setup takes longer than the DHCP lease interval or than the IPv6 address lifetime. Running NM as a service
solves this problem.

Now NM runs with D-Bus support and therefore its API can be used by other modules. This open the possibility, for example, to integrate nm-cloud-setup to automatically configure networking based on cloud metadata.

Use the NetworkManager-wait-online.service, ordered before dracut-initqueue.service, to delay the initqueue until NM has terminated its configuration.

### DIFF
--- a/modules.d/35network-manager/initrd-no-auto-default.conf
+++ b/modules.d/35network-manager/initrd-no-auto-default.conf
@@ -1,0 +1,5 @@
+[.config]
+enable=env:initrd
+
+[main]
+no-auto-default=*
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -10,6 +10,7 @@ check() {
 
 # called by dracut
 depends() {
+    echo dbus
     return 0
 }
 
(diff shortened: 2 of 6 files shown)


johannbg (changes requested):
This is a nack in it's current form, you are literally introducing a regression by removing the nm-run.service unit. 

Re-introduce you PR's as two new separated NM modules. 

The network part in one module and nm-wait-online in another and have them include a full type unit services ( nm-initrd.service, nm-initrd-wait-online.service ) which gets included ( like nm-run.service is being done ) as in no echo's or sed's to create or alter downstream/upstream existing NM component type units.

bengal:
> This is a nack in it's current form, you are literally introducing a regression by removing the nm-run.service unit.

The point of this PR is to get rid of nm-run.service because the way in which NM is started currently (as a one-shot service) breaks some scenarios. If NM quits after configuring the network, it can't renew DHCP leases or extend IPv6 addresses, and we have some cases where boot fails due to this problem.

So it is intended to remove nm-run.service and replace it with something which hopefully works better.

bengal:
> Re-introduce you PR's as two new separated NM modules.

Note that the two services are not independent. The NM service configures the network; since we want to order the initqueue service after NM has finished the configuration we can't just add a "Before=dracut-initqueue.service" to NM service because the service succeeds immediately after it registers on D-Bus.

A mechanism to delay the initqueue is needed. NM-wait-online is used for that.

Both services are needed and I think they should be in the same module, and that should be the network-manager module.

bengal:
> The network part in one module and nm-wait-online in another and have them include a full type unit services ( nm-initrd.service, nm-initrd-wait-online.service ) which gets included ( like nm-run.service is being done ) as in no echo's or sed's to create or alter downstream/upstream existing NM component type units.

Okay, I will create new units instead of copying and modifying them.

johannbg:
@bengal could you fix the conflict

johannbg:
@bengal I have made a PR for systemd-hostnamed #1354 ( which you seemed to have been interested in the past ). 

Now what's might be of particular interest to you is the inst_simple section which is a drop-in section to allow networkd to request hostname changes via DBUS without elevating its privileges since we dont have a polkit module and probably never will.

bengal:
> @bengal I have made a PR for systemd-hostnamed #1354 ( which you seemed to have been interested in the past ).

Yeah, in the previous PR NM service ran in the initrd with the same `CapabilityBoundingSet=` of the service in real root. Since CAP_SYS_ADMIN was not among the capabilities, NM had to set the hostname via hostnamed.

In this new PR the NM service has all capabilities, so hostnamed is not needed. It would be nice to have it to drop again capabilities, though.

haraldh:
Needs a rebase. If you insist to use DBUS.. fine with me..

haraldh (approved):

johannbg (approved):

PR #1284, opened 2021-03-30
fix(usrmount): shellcheck for modules.d/98usrmount

haraldh:


modules.d/98usrmount/mount-usr.sh
```diff
@@ -6,12 +6,13 @@ type fsck_single > /dev/null 2>&1 || . /lib/fs-lib.sh
 filtersubvol() {
     local _oldifs
     _oldifs="$IFS"
-    IFS=","
-    set $*
+    local IFS=","
```
danimo, line 9: Everywhere else, you separated the  local declaration from the assignment. Why isn't shellcheck complaining here?

danimo (changes requested):

modules.d/98usrmount/mount-usr.sh
haraldh: It's only complaining, if a command is executed in the assignment.

johannbg (approved):

danimo (approved):

PR #1342, opened 2021-03-31
fix(dracut.sh): omission is an addition to other omission in conf files

### LINKED ISSUE #1341: dracut command line `--omit` flag is not cumulative in relation to `/etc/dracut.conf.d/*.conf` config files
**Describe the bug**
When using the `--omit "module"` flag on the dracut command line it overrides any omit module statement that was already present in `/etc/dracut.conf.d/*`. I did not expect this behavior since `omit_dracutmodules` is also cumulative over all config files.

Probably this is a feature instead of a bug, however at the moment I can't really see why it is desirable the current behavior in contrast to a cumulative behavior. I find it really not intuitive and after reading the man pages one has the impression that `--add` and  `--omit` work in an equivalent way, cumulative since both can be specified multiple times and both can also be specified as cumulative in config files.

Current behavior forces me to review all config files before making use of the `--omit` flag in the command line if my purpose was simply omitting an specific module.

**Distribution used**
openSUSE Tumbleweed

**Dracut version**
dracut-053

**Expected behavior**
I expected that making use of `--omit <module>` flag on the command line would instruct `dracut` to create an image omitting the provided module and **also** omit any other module that is listed as part of `omit_dracutmodules` variable any `/etc/dracut.conf.d/*.conf`.

**Additional context**
This has been originally reported in https://github.com/OSInside/kiwi/issues/1751. KIWI is an appliance builder and we make use of `--omit` flag to omit, for instance, the `multipath` module. This helps us on making sure some firstboot deployment specific modules are not interfered with `multipath`, however we do not aim to completely prevent `multipath` module as a persistent config in `/etc/dracut.conf.d` in the resulting appliance image.

johannbg:
When omitting a module from the command line via -o or --omit
it's expected that it behaves in the same manner as adding a module from the
command line as in it does not overwrite existing omissions of other modules in
configuration file(s).

This should fix #1341

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -936,6 +936,7 @@ export SYSTEMCTL=${SYSTEMCTL:-systemctl}
 
 # these options add to the stuff in the config file
 ((${#add_dracutmodules_l[@]})) && add_dracutmodules+=" ${add_dracutmodules_l[*]} "
+((${#omit_dracutmodules_l[@]})) && omit_dracutmodules+=" ${omit_dracutmodules_l[*]} "
 ((${#force_add_dracutmodules_l[@]})) && force_add_dracutmodules+=" ${force_add_dracutmodules_l[*]} "
 ((${#fscks_l[@]})) && fscks+=" ${fscks_l[*]} "
 ((${#add_fstab_l[@]})) && add_fstab+=" ${add_fstab_l[*]} "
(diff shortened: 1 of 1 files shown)


johannbg:
If we want to keep the existing omit behavior as an option we should probably introduce a separated switch `--omit-clear` ? that does that.

danimo (approved):

haraldh (approved):

PR #1344, opened 2021-04-12
fix(TEST ISCSI): try to debug md error 524

haraldh:
While creating the striped md raid0 sometimes this error occures:

```
mdadm: Defaulting to version 1.2 metadata
mdadm: RUN_ARRAY failed: Unknown error 524
  Failed to clear hint file.
  Device /dev/md0 not found.
Powering off.
```

Add debug output with the size of the disks.

### DIFF
--- a/test/TEST-30-ISCSI/create-client-root.sh
+++ b/test/TEST-30-ISCSI/create-client-root.sh
@@ -7,6 +7,9 @@ rm -f -- /etc/lvm/lvm.conf
 udevadm control --reload
 udevadm settle
 
+echo "Size of /dev/sdc and /dev/sdd"
+blockdev --getsize64 /dev/sdc /dev/sdd
+
 mkfs.ext3 -j -L singleroot -F /dev/sda \
     && mkdir -p /sysroot \
     && mount /dev/sda /sysroot \
--- a/test/TEST-30-ISCSI/test.sh
+++ b/test/TEST-30-ISCSI/test.sh
@@ -129,9 +129,12 @@ test_setup() {
     fi
 
     # Create the blank file to use as a root filesystem
-    dd if=/dev/zero of="$TESTDIR"/root.ext3 bs=1M count=100
-    dd if=/dev/zero of="$TESTDIR"/iscsidisk2.img bs=1M count=50
-    dd if=/dev/zero of="$TESTDIR"/iscsidisk3.img bs=1M count=50
+    rm -f "$TESTDIR"/root.ext3
+    dd if=/dev/zero of="$TESTDIR"/root.ext3 bs=4096 count=$((200 * 256))
+    rm -f "$TESTDIR"/iscsidisk2.img
+    dd if=/dev/zero of="$TESTDIR"/iscsidisk2.img bs=4096 count=$((100 * 256))
+    rm -f "$TESTDIR"/iscsidisk3.img
+    dd if=/dev/zero of="$TESTDIR"/iscsidisk3.img bs=4096 count=$((100 * 256))
 
     kernel=$KVERSION
     # Create what will eventually be our root filesystem onto an overlay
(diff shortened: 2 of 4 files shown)


johannbg:
I think this warrants some further investigation since this error msg should only appear if the devices are not the same size which afaikt we arent doing ( it whould be interesting to see the output from blockdev --getsize64 /dev/sdc /dev/sdd when this gets trigger) 

Which indicates something else is going on either with `# Invoke KVM and/or QEMU to actually create the target filesystem` section, the `client.img` creation and or the mkfs.ext3 command which probably should look more like this `mkfs.ext3 -v -L sysroot -b 4096 -E stride=128,stripe-width=256 /dev/dracut/root` so we dont accidentally use some built in configuration defaults for ext3 ( why are we using ext3? ).

Plus it's probably just simpler to add `raid0.default_layout=2` to the kernel command line in relevant `append` sections instead of using

`
if [ -e /sys/module/raid0/parameters/default_layout ]; then
    echo 2 > /sys/module/raid0/parameters/default_layout
fi
`

haraldh:
> I think this warrants some further investigation since this error msg should only appear if the devices are not the same size which afaikt we arent doing ( it whould be interesting to see the output from blockdev --getsize64 /dev/sdc /dev/sdd when this gets trigger)

Yeah, I don't know.. some strange things happen in the CI, which never appear on bare metal.

> 
> Which indicates something else is going on either with `# Invoke KVM and/or QEMU to actually create the target filesystem` section, the `client.img` creation and or the mkfs.ext3 command which probably should look more like this `mkfs.ext3 -v -L sysroot -b 4096 -E stride=128,stripe-width=256 /dev/dracut/root` so we dont accidentally use some built in configuration defaults for ext3 ( why are we using ext3? ).

It errors before `mkfs.ext3 -j -L sysroot /dev/dracut/root` ... The creation of `/dev/md0` fails. So, no need to tackle `mkfs.ext3`.

> 
> Plus it's probably just simpler to add `raid0.default_layout=2` to the kernel command line in relevant `append` sections instead of using
> 
> `if [ -e /sys/module/raid0/parameters/default_layout ]; then echo 2 > /sys/module/raid0/parameters/default_layout fi`

Hmm...

johannbg:
> > I think this warrants some further investigation since this error msg should only appear if the devices are not the same size which afaikt we arent doing ( it whould be interesting to see the output from blockdev --getsize64 /dev/sdc /dev/sdd when this gets trigger)
> 
> Yeah, I don't know.. some strange things happen in the CI, which never appear on bare metal.

Which should narrow this down to QEMU/KVM sections

haraldh:
Ok, changed this to report the size of the disks and create the disks with multiple of 4096. Let's see, if the error occurs again.

johannbg (approved):

johannbg:
Encase we cant figure out what's causing this, these are the options at our disposal ( afaikt ).

To force the kernel to use the pre-3.14 layout from the kernel command line
`raid0.default_layout=1`

To force it to use the 3.14-and-later layout from the kernel command line
`raid0.default_layout=2`

To assemble an old array on a new kernel without using the module parameter
`
--update=layout-original
`
`
--update=layout-alternate
`

When create a new array from unused devices
`
--layout=original
`
`
--layout=alternate
`
`
--layout=dangerous ( if the running kernel doesn't support RAID0 layouts for whatever reason )
`

Echo options 
unset
`
echo 0 > /sys/module/raid0/parameters/default_layout
`
`
mdadm --stop /dev/md0 ( if running ) 
`
`
mdadm --assemble /dev/md0 /dev/sdc /dev/sdd
`

original
`
echo 1 > /sys/module/raid0/parameters/default_layout
`
`
mdadm --stop /dev/md0 ( if running ) 
`
`
mdadm --assemble /dev/md0 /dev/sdc /dev/sdd
`

alternate
`
echo 2 > /sys/module/raid0/parameters/default_layout
`
`
mdadm --stop /dev/md0 ( if running ) 
`
`
mdadm --assemble /dev/md0 /dev/sdc /dev/sdd
`

johannbg:
@danimo ping

PR #1347, opened 2021-04-13
fix(network-manager): only run NetworkManager if rd.neednet=1 

dustymabe:
```
commit 0152fa88c9a155ac3e34c575da6c7ed2134b32f0
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Tue Apr 13 11:45:35 2021 -0400

    network-manager: use /run/NetworkManager/initrd/neednet in initqueue
    
    We don't want to start NetworkManager if networking is not needed.
    Right now nm-config.sh lays down /usr/lib/dracut/hooks/initqueue/finished/nm.sh
    which will cause the initqueue to run. If nothing exists in
    /usr/lib/dracut/hooks/initqueue/finished/ then it will short circuit and
    the initqueue won't run anything. But what if something else needed
    something to run in the initqueue? nm-run.sh would still get started,
    even though /usr/lib/dracut/hooks/initqueue/finished/nm.sh didn't exist.
    In this case let's just trigger off of /run/NetworkManager/initrd/neednet
    like we are doing in the systemd unit (nm-run.service).

commit 127fc8028a47a44c3a73c09b00bea269fa1d70b3
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Tue Apr 13 11:36:21 2021 -0400

    network-manager: only run NetworkManager if rd.neednet=1
    
    Don't run the new systemd unit (nm-run.service) if rd.neednet=1
    isn't set. nm-initrd-generator will generate configuration even
    without rd.neednet=1 so determining if we should start based on
    just if connection profiles exist isn't enough. We need some other
    indicator. In this case we lay down a /run/NetworkManager/initrd/neednet
    if rd.neednet=1, which is used by nm-run.service to determine the
    need to run.
```

### DIFF
--- a/modules.d/35network-manager/nm-lib.sh
+++ b/modules.d/35network-manager/nm-lib.sh
@@ -14,6 +14,7 @@ nm_generate_connections() {
             /etc/sysconfig/network-scripts/ifcfg-*; do
             [ -f "$i" ] || continue
             echo '[ -f /tmp/nm.done ]' > "$hookdir"/initqueue/finished/nm.sh
+            : > /run/NetworkManager/initrd/neednet # activate nm-run.service
             break
         done
     fi
--- a/modules.d/35network-manager/nm-run.service
+++ b/modules.d/35network-manager/nm-run.service
@@ -15,6 +15,9 @@ Before=network.target network-online.target
 #run before we try to mount anything from the dracut hooks
 Before=dracut-initqueue.service
 
+#do not run if networking not needed
+ConditionPathExists=/run/NetworkManager/initrd/neednet
+
 #do not run, if there is no configuration
 ConditionPathExistsGlob=|/usr/lib/NetworkManager/system-connections/*
 ConditionPathExistsGlob=|/run/NetworkManager/system-connections/*
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -6,15 +6,19 @@ if [ -e /tmp/nm.done ]; then
     return
 fi
 
-[ -z "$DRACUT_SYSTEMD" ] \
-    && for i in /usr/lib/NetworkManager/system-connections/* \
-        /run/NetworkManager/system-connections/* \
-        /etc/NetworkManager/system-connections/* \
-        /etc/sysconfig/network-scripts/ifcfg-*; do
-        [ -f "$i" ] || continue
-        /usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
-        break
-    done
+if [ -z "$DRACUT_SYSTEMD" ]; then
+    # Only start NM if networking is needed
+    if [ -e /run/NetworkManager/initrd/neednet ]; then
+        for i in /usr/lib/NetworkManager/system-connections/* \
+            /run/NetworkManager/system-connections/* \
+            /etc/NetworkManager/system-connections/* \
+            /etc/sysconfig/network-scripts/ifcfg-*; do
+            [ -f "$i" ] || continue
+            /usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
+            break
+        done
+    fi
+fi
 
 if [ -s /run/NetworkManager/initrd/hostname ]; then
     cat /run/NetworkManager/initrd/hostname > /proc/sys/kernel/hostname


dustymabe:
## Changes

This pull request changes things so NetworkManager doesn't start if not needed. It also changes
NetworkManager in the initqueue (non-systemd run) to only start NetworkManager if it's needed,
otherwise it will be started if any other module placed files in /usr/lib/dracut/hooks/initqueue/finished/
and takes some time to complete.

## Checklist
- [✅] I have tested it locally
- [⚠️] I have reviewed and updated any documentation if relevant

    I don't see any documentation on this to update.
    
- [❌] I am providing new code and test(s) for it

    No new modules here. Just existing module fixes.

johannbg:
Hmm you cannot have an on off switch of modules from the kernel command line since you end up breaking any other modules that depends on the module being turned off and you cannot start all the modules that require it when it's turned on. 

Even in a perfect pony world where the entire initramfs image was made up of modules using systemd type units which correctly can express dependency for other type units in other modules and start them if required, when you would implement such an on/off kernel command line switch, the requires/wants section of the other modules type units would overwrite it and simply start the service/module if they required it so the only option is to omit the module(s) in question.

So switches like `rd.neednet=` dont work in practice unless they are implemented somehow globally which every module ends up having to parse and be turned on/off also ( the scripts/type units in the modules that depend on networking in this case). 

Which brings us to the question what use case does the switch solve in the first place. For example when would you not want networking to be started as soon as possible? Why would you not omit the NM module ( and any other module that has dependency on the NM module or networking ) if networking is not needed when generating the initramfs in the first place in those use cases?

johannbg (changes requested):
You need to fix your commit messages to pass the commisery check. 
Change it from
`
network-manager: only run NetworkManager if rd.neednet=1 
`
to
`
fix(network-manager): only run NetworkManager if rd.neednet=1
`
and squash the other commit message.

modules.d/35network-manager/nm-lib.sh
```diff
@@ -14,6 +14,7 @@ nm_generate_connections() {
             /etc/sysconfig/network-scripts/ifcfg-*; do
             [ -f "$i" ] || continue
             echo '[ -f /tmp/nm.done ]' > "$hookdir"/initqueue/finished/nm.sh
+            > /run/NetworkManager/initrd/neednet # activate nm-run.service
```
johannbg, line 17: You need to change this line to 

`: > /run/NetworkManager/initrd/neednet`

johannbg (changes requested):

dustymabe:
> Which brings us to the question what use case does the switch solve in the first place. For example when would you not want networking to be started as soon as possible? Why would you not omit the NM module ( and any other module that has dependency on the NM module or networking ) if networking is not needed when generating the initramfs in the first place in those use cases?

Sometimes the initrd is generated when images (or other higher order OS artifacts) are built and are meant to satisfy a wide range of use cases. The team/project I work on (Fedora CoreOS/Red Hat CoreOS) is one such case. We use RPM-OSTree technology which generates the initramfs when the OSTree is composed in release engineering. That same initramfs is used by all clients, except in some very rare cases. So we're not using a "hostonly" initramfs, but something more generic. We pay a penalty in initramfs size for consistency.

modules.d/35network-manager/nm-lib.sh
dustymabe: done :)

dustymabe:
> You need to fix your commit messages to pass the commisery check.
> Change it from
> `network-manager: only run NetworkManager if rd.neednet=1 `
> to
> `fix(network-manager): only run NetworkManager if rd.neednet=1`
> and squash the other commit message.

Thanks for the tip. I've updated the commit messages and pushed up another revision.

I'm chasing down another bug in this space related to network device teaming so I'll probably give this code some more testing tomorrow. I'd also like to get a look from @thom311 and maybe @bengal if they have some time.

johannbg:
> > 
> 
> Sometimes the initrd is generated when images (or other higher order OS artifacts) are built and are meant to satisfy a wide range of use cases. The team/project I work on (Fedora CoreOS/Red Hat CoreOS) is one such case. We use RPM-OSTree technology which generates the initramfs when the OSTree is composed in release engineering. That same initramfs is used by all clients, except in some very rare cases. So we're not using a "hostonly" initramfs, but something more generic. We pay a penalty in initramfs size for consistency.

That a generic initramfs is created and used among multiple products does not explain why you would need an kernel command line on and off switch for networking which again is a switch that does not work in practice unless nothing depends on the module in question ( which most certainly is not the case for networking ) and based on your answer all seemingly because you dont want to create product specific image for the use case which makes the "generic" image stop being "generic" ( which probably is the underlying problem here as in CoreOS usecases are not generic enough to be used with Fedora's traditional distribution model ).

Using the kernel command line to solve a release engineering problem once an generic image stops being generic makes no sense and will only make things more messier throughout the core/baseOS stack for Fedora/RHEL on top of that your PR will force *any* distribution that uses NM to set "rd.neednet=1" on the kernel command line...

dustymabe:
> That a generic initramfs is created and used among multiple products 

Quick clarification. It's not a generic initramfs among multiple products (or projects). It's per project. So Fedora CoreOS builds one, Red Hat CoreOS builds one. Fedora IoT builds one. etc.. 

But each product serves different users. Some might need networking in the initramfs, some might not.

We also have this thing called [Ignition](https://github.com/coreos/ignition/). It provisions a machine on first boot. Often times the Ignition config comes from a remote resource (but sometimes not). So often our machines will bring up networking in the initramfs on first boot to grab the Ignition configuration, but not on any subsequent boot. 


> does not explain why you would need an kernel command line on and off switch for networking which again is a switch that does not work in practice unless nothing depends on the module in question ( which most certainly is not the case for networking ) 


I'm not sure I follow. We definitely want the 35network-manager module included in the initramfs. We just want networking activated based on conditions.

In our case we're not network booting (i.e. netroot), so maybe that's where some of the confusion lies.


> and based on your answer all seemingly because you dont want to create product specific image for the use case which makes the "generic" image stop being "generic" ( which probably is the underlying problem here as in CoreOS usecases are not generic enough to be used with Fedora's traditional distribution model ).

We do create product/project specific images (see above).

> 
> Using the kernel command line to solve a release engineering problem once an generic image stops being generic makes no sense and will only make things more messier throughout the core/baseOS stack for Fedora/RHEL on top of that your PR will force _any_ distribution that uses NM to set "rd.neednet=1" on the kernel command line...

Before c17c5b7 `rd.neednet=1` would need to be provided (or implied) in order to [get the initqueue to wait for NM to bring up networking](https://github.com/dracutdevs/dracut/blob/651fe01e7937d86bbd471d9621581bed44f23dfa/modules.d/35network-manager/nm-lib.sh#L10-L16). Mostly what I'm trying to do here is get the same functionality back as to what we had before (see https://github.com/coreos/fedora-coreos-tracker/issues/782).

jlebon:
> So switches like `rd.neednet=` dont work in practice unless they are implemented somehow globally which every module ends up having to parse and be turned on/off also ( the scripts/type units in the modules that depend on networking in this case).

Not sure I follow this. `rd.neednet=` works well in practice from what we've seen. It's already set up as an on/off switch for networking. Notice how `nm-lib.sh` is already currently gating the networking initqueue hook on it: https://github.com/dracutdevs/dracut/blob/651fe01e7937d86bbd471d9621581bed44f23dfa/modules.d/35network-manager/nm-lib.sh#L10-L19

And it knows to automatically turn on networking using this switch if e.g. netroot is enabled: https://github.com/dracutdevs/dracut/blob/651fe01e7937d86bbd471d9621581bed44f23dfa/modules.d/35network-manager/nm-config.sh#L5-L7

All this PR is doing is strengthening the semantics that already exist.

bengal:
Just to check that I understand correctly: the first patch ensures that the NM service is not started unless rd.neednet=1. Before the patch NM was always started in presence of a connection, even if rd.neednet wasn't set.

The second patch does the same for the non-systemd case. The only difference is that before the patch, when rd.neednet wasn't set NM could run or not depending on whether something else added a finish hook.

By the way, I always find confusing that if the kernel command line contains an ip= argument, then no configuration happens unless there is a netroot or the command line contains rd.neednet=1. If a user explicitly passes ip=dhcp, there shouldn't be the need to also say "yes, I really want to do DHCP" via rd.neednet=1.

Anyway, since it works in this way, I agree that NM should not be started without need and so patches look correct.

The only small concern I have is that in the non-systemd case there could be an case where some other module relies on network being configured without rd.neednet=1, because the module itself ensures that the initqueue runs. But that would be a bug of the module, which should instead set rd.neednet=1 during the config phase.

We have some dracut tests in the NM CI:

https://gitlab.freedesktop.org/NetworkManager/NetworkManager-ci/-/blob/master/nmcli/features/dracut.feature

Perhaps it would be good to have the changes also tested there. I guess @fpokryvk can help with that.

johannbg:
> The only small concern I have is that in the non-systemd case there could be an case where some other module relies on network being configured without rd.neednet=1, because the module itself ensures that the initqueue runs. But that would be a bug of the module, which should instead set rd.neednet=1 during the config phase.

If you are speaking about some kind of config phase for the NM module itself it might make sense but it's unrealistic to expect modules that depend on a meta module such as networking to set option specifically tailored to other modules provided by the meta module. 

To give you an example of this, modules that have been configured to depend on the network meta module ( as all modules that require network should  ) means they can be using IWD ( in the future ) NM, systemd-networkd ( once it's on par with other providers ), wicked + legacy and whatever else might emerge in the future and get added to that list.

jlebon:
> If you are speaking about some kind of config phase for the NM module itself it might make sense but it's unrealistic to expect modules that depend on a meta module such as networking to set option specifically tailored to other modules provided by the meta module.

The modules would just do what NM itself does, which is e.g. `echo rd.neednet >> /etc/cmdline.d/...`. This is then the generic API which network-implementing modules would key off of.

For example, we already use this API in our own dracut modules in FCOS/RHCOS: https://github.com/coreos/fedora-coreos-config/blob/d29a62f3357ce5ea5a038978ce276b0bf4a9bb72/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-enable-network.sh#L18

dustymabe:
> Just to check that I understand correctly: the first patch ensures that the NM service is not started unless rd.neednet=1. Before the patch NM was always started in presence of a connection, even if rd.neednet wasn't set.

Yes, the systemd service would always start if connection files were present.
> 
> The second patch does the same for the non-systemd case. The only difference is that before the patch, when rd.neednet wasn't set NM could run or not depending on whether something else added a finish hook.

Correct.

> By the way, I always find confusing that if the kernel command line contains an ip= argument, then no configuration happens unless there is a netroot or the command line contains rd.neednet=1. If a user explicitly passes ip=dhcp, there shouldn't be the need to also say "yes, I really want to do DHCP" via rd.neednet=1.

I kind of agree with you about it being non-intuitive.

> 
> Anyway, since it works in this way, I agree that NM should not be started without need and so patches look correct.

👍

> 
> The only small concern I have is that in the non-systemd case there could be an case where some other module relies on network being configured without rd.neednet=1, because the module itself ensures that the initqueue runs. But that would be a bug of the module, which should instead set rd.neednet=1 during the config phase.

Yeah. The problem there is (I think) that the initqueue could finish before NM was fully started because it wasn't configured to wait until /tmp/nm.done existed. So yes, either the other module would need to be changed or we'd need to change the logic in https://github.com/dracutdevs/dracut/blob/651fe01e7937d86bbd471d9621581bed44f23dfa/modules.d/35network-manager/nm-lib.sh#L10 

> We have some dracut tests in the NM CI:
> 
> https://gitlab.freedesktop.org/NetworkManager/NetworkManager-ci/-/blob/master/nmcli/features/dracut.feature
> 
> Perhaps it would be good to have the changes also tested there. I guess @fpokryvk can help with that.

👍

haraldh (approved):

johannbg (approved):

PR #1349, opened 2021-04-14
fix(network-manager): nm-run.service: don't kill forked processes

dustymabe:
If teaming is set up via NetworkManager we don't want systemd to take
down the userspace teamd process when NetworkManager quits. `KillMode=process`
will allow it to leave those processes behind.

This is fallout from the change to run NetworkManager via systemd (c17c5b7).

With `KillMode=process` we get something like:

```
sh-5.1# journalctl -u nm-run -o cat | tail
<info>  [1618411262.7030] quitting now that startup is complete
<info>  [1618411262.7030] device (team0): carrier: link connected
<info>  [1618411262.7033] device (team0): team port ens2 was released
<info>  [1618411262.7033] device (team0): team port ens3 was released
<info>  [1618411262.7033] manager: NetworkManager state is now CONNECTED_SITE
<info>  [1618411262.7034] exiting (success)
nm-run.service: Deactivated successfully.
nm-run.service: Unit process 476 (teamd) remains running after unit stopped.
Finished nm-run.service.
```

### DIFF
--- a/modules.d/35network-manager/nm-run.service
+++ b/modules.d/35network-manager/nm-run.service
@@ -25,6 +25,8 @@ ConditionPathExistsGlob=|/etc/sysconfig/network-scripts/ifcfg-*
 #run the script and wait before it finishes
 Type=oneshot
 ExecStart=/usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
+#don't kill forked off processes (for example: teamd for teaming)
+KillMode=process
 
 [Install]
 WantedBy=initrd.target


dustymabe:
## Changes

This pull request changes the NetworkManager systemd unit to not clean up forked
processes such as teamd. Killing those processes means that teaming in the initramfs
will not work since it utilizes userspace processes to back the team.

## Checklist
- [✅] I have tested it locally
    - Tested by updating a development tree for Fedora 34
- [⚠️] I have reviewed and updated any documentation if relevant
    - I don't see any documentation on this to update.
- [❌] I am providing new code and test(s) for it
    - No new modules here. Just existing module fixes.

dustymabe:
cc @bengal @thom311

johannbg (approved):
LGTM

thom311:
is this necessary?

With team, you have the user space `teamd` daemon, but the interface is supposed to continue working while teamd gets (re)started.

So, when switching the real-root, NetworkManager and teamd get killed, but afterwards, NetworkManager will start a new instance of teamd, and all is good. Or why not?

dustymabe:
> is this necessary?

Yes.

> With team, you have the user space `teamd` daemon, but the interface is supposed to continue working while teamd gets (re)started.

My understanding is that if you need networking in the initramfs then what we have now with `nm-run.service` isn't sufficient. From my tests teaming will only be up while `/usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon` is running. After it exits, systemd will tear down `teamd` and the team will lose connection.

The reason I found this issue (bug?), is because I have a test that specifically tests bringing up teaming in the initramfs and verifies remote connectivity.

> 
> So, when switching the real-root, NetworkManager and teamd get killed, but afterwards, NetworkManager will start a new instance of teamd, and all is good. Or why not?

Yeah, once you're in the real root all is good, but we need it working in the initramfs.

dustymabe:
I just noticed #1281 - which I think would also fix this problem. But since that is a large change in behavior I think it's fair to go ahead and merge this change while that one gets review/testing.

bengal:
> With team, you have the user space `teamd` daemon, but the interface is supposed to continue working while teamd gets (re)started.

I think most of the logic is implemented in the user space daemon, so if it gets killed the interface is not able to react to link changes. Also, when teamd is killed it leaves the team interface up (good) because NM passes the -N option, but it removes the ports breaking connectivity.

So, the change looks good to me.

haraldh (approved):

PR #1352, opened 2021-04-15
feat(systemd-initrd): add initrd-usr-fs.target

johannbg:
Adding a new systemd initrd-usr-fs.target to the systemd-initrd module

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-initrd/module-setup.sh
+++ b/modules.d/01systemd-initrd/module-setup.sh
@@ -23,6 +23,7 @@ install() {
         "$systemdsystemunitdir"/initrd-fs.target \
         "$systemdsystemunitdir"/initrd-root-device.target \
         "$systemdsystemunitdir"/initrd-root-fs.target \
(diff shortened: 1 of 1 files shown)


haraldh:
```
❯ ls /usr/lib/systemd/system/initrd-usr-fs.target
ls: cannot access '/usr/lib/systemd/system/initrd-usr-fs.target': No such file or directory
```

haraldh:
see also: https://github.com/systemd/systemd/blob/b5a80aa9d89cd82e53181cfd3288bba694622786/units/meson.build#L37

johannbg:
@haraldh depends on systemd/systemd#19234 which I probably should have mentioned

johannbg:
@haraldh the target has been merged upstream. 

See
https://github.com/systemd/systemd/blob/main/units/meson.build

And this for the whole PR that did it. 

https://github.com/systemd/systemd/commit/29a24ab28e9790680348b1ffab653a321fa49a67

PR #1359, opened 2021-04-16
fix(bluetooth): adding missing qoutes

johannbg:
The bluetooth PR predated the linter and is now tripping it after it got automerged


johannbg:
@haraldh it would be nice if you could force merge this as soon as you see this since or revert the bluetooth merge...

PR #1366, opened 2021-04-19
feat(network): install hosts and nss file in hostonly mode

ryncsn:
## Changes
Install hosts and nss file in hostonly initramfs can make it easier to
maintain the config and allow using hostname in initramfs stage.

## Checklist


johannbg:
Meta modules should not contain hostonly sections so either this needs to be added to every networking module and or I can update the systemd-resolved module PR to include this.

ryncsn:
I think it will be a good idea to include this in systemd-resolved.

johannbg:
I might actually end up introducing nss module for the nss part of this

haraldh:
Merged with   https://github.com/dracutdevs/dracut/pull/1378

PR #1370, opened 2021-04-20
WIP: refactor(systemd): use systemd modules

johannbg:
This pull request refactors the systemd module to start using systemd modules

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -11,9 +11,14 @@ check() {
     return 255
 }
 
-# called by dracut
+# Module dependency requirements.
 depends() {
+
+    # This module has external dependency on other module(s).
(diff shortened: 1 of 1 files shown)


johannbg:
@aafeijoo-suse already done those changes locally ( but not update the pr ) but thanks for the offer. 
These somewhat invasive changes so we need to make them in one sweep between releases as in make a release then merge the changes, ( which includes replacing udev with systemd-udev and deprecate the systemd code in udev module, ) as well as double check what's been happening in upstream systemd, the unit file for systemd-resolved has been change so you can look at that and make a PR for it if you have the time. 

We also need to include our own type unit files in our modules as in stop sourcing them from the distribution/vendor/host to make sure they actually work since downstream is playing mix and match of components releases on the core/baseOS level ( for example latest dracut with old systemd etc. ) and often are shipping outdated unit files or simply unit files not meant for initrd

PR #1372, opened 2021-04-20
ci: WIP use /dev/disk/by-id

haraldh:
Due to parallel probing of the linux kernel `/dev/sd*` can't be used to
reliably address a hard disk. This can be seen by the many spurious
failures of the dracut CI, where `mdadm` failed with error 524 or tests
failed due to the success marker message written to the wrong disk.

* don't rely on `/dev/sd*` but use disk ids and `/dev/disk/by-id/ata-disk_<name>`

* specify the exact qemu machine architecture `-M q35` needed for the
  disk ids. A later patch will move this to `run-qemu`, when all tests are converted

* due to `-M q35` the interface names have changed from
  `ens2` -> `enp0s1` and `ens3` -> `enp0s2`

### DIFF
--- a/test/TEST-01-BASIC/create-root.sh
+++ b/test/TEST-01-BASIC/create-root.sh
@@ -1,24 +1,22 @@
 #!/bin/sh
+
+trap 'poweroff -f' EXIT
+
 # don't let udev and this script step on eachother's toes
 for x in 64-lvm.rules 70-mdadm.rules 99-mount-rules; do
     : > "/etc/udev/rules.d/$x"
 done
 rm -f -- /etc/lvm/lvm.conf
 udevadm control --reload
-set -e
-# save a partition at the beginning for future flagging purposes
-sfdisk /dev/sda << EOF
-,1M
-,
-EOF
-
 udevadm settle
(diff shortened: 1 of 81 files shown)


test/TEST-01-BASIC/create-root.sh
```diff
@@ -1,24 +1,22 @@
...
-,1M
-,
-EOF
-
 udevadm settle
-mkfs.ext3 -L '  rdinit=/bin/sh' /dev/sda2
+
+set -ex
+
+mkfs.ext3 -L '  rdinit=/bin/sh' /dev/disk/by-id/ata-disk_root
```
johannbg, line 15: is this the double space here  `'  rdinit=/bin/sh'` supposed be there in other words why do you need double space instead of single?

johannbg (approved):
Other than a bit of curiosity regarding the double space LGTM

test/TEST-01-BASIC/create-root.sh
haraldh: part of the test of the kernel command line parsing

PR #1373, opened 2021-04-20
fix(systemd-modules-load): misc repairs

johannbg:
- Removing dependency on the systemd meta module since it causes circular dependency,
  prevents the existance of multiple systemd based meta modules and is redunant.

- Removing directory creation for modules-load since it should not be
  necessary and if it turns out that it is, it will be re-introduced in the
  form of a systemd tmpfile instead of inst_dir.

- Add a missing inclusion of configuration files placed in /usr/lib/modules-load.d
  and /etc/modules-load.d ( hostonly ) directories.

- Fix a spelling error in the systemd type unit file conf directory overwrite
  in the hostonly section.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-modules-load/module-setup.sh
+++ b/modules.d/01systemd-modules-load/module-setup.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # This file is part of dracut.
 # SPDX-License-Identifier: GPL-2.0-or-later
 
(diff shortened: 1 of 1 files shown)


modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -16,33 +16,30 @@ check() {
 # Module dependency requirements.
 depends() {
 
-    # This module has external dependency on the systemd module.
-    echo systemd
+    # This module has external dependency on other module(s).
```
haraldh, line 19: remove the comment as well

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -16,33 +16,30 @@ check() {
...
     # Return 0 to include the dependent systemd module in the initramfs.
```
haraldh, line 20: remove the comment as well

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -16,33 +16,30 @@ check() {
...
+# Install the required file(s) and directories for the module in the initramfs.
 install() {
 
-    # Create systemd-modules-load related directories.
-    inst_dir "$modulesload"
-    inst_dir "$modulesloadconfdir"
-
-    # Install related files for systemd-modules-load
     inst_multiple -o \
+        "$modulesload"/*.conf \
```
haraldh, line 29:
```suggestion
        "$modulesload/*.conf" \
```

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -16,33 +16,30 @@ check() {
...
+        "$systemdsystemunitdir"/sysinit.target.wants/systemd-modules-load.service
+
+    # Enable systemd type unit(s)
+    $SYSTEMCTL -q --root "$initdir" enable systemd-modules-load.service
 
-    # Install local user configurations if host only is enabled..
+    # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$modulesloadconfdir"/*.conf \
```
haraldh, line 40:
```suggestion
            "$modulesloadconfdir/*.conf" \
```

modules.d/01systemd-modules-load/module-setup.sh
```diff
@@ -16,33 +16,30 @@ check() {
...
             "$systemdsystemconfdir"/systemd-modules-load.service \
-            "$systemdsystemconfdir"/systemd-systemd-modules-load.d/*.conf
+            "$systemdsystemconfdir"/systemd-modules-load.service.d/*.conf
```
haraldh, line 42:
```suggestion
            "$systemdsystemconfdir/systemd-modules-load.service.d/*.conf"
```

haraldh (changes requested):

modules.d/01systemd-modules-load/module-setup.sh
johannbg: I rephrased it to ` # Return 0 to include the dependent module(s) in the initramfs.`  since we should explain why we have a return 0 in the depends() section.

It should be a general rule of thumb that we leave a comment like this since it should be possible for someone else to learn how to use the program or to use a function in a library by reading the comments (and self-help, if provided) without reading the code.

modules.d/01systemd-modules-load/module-setup.sh
johannbg: Done

PR #1374, opened 2021-04-20
fix(systemd-modules): remove dependency on systemd meta module

johannbg:
Depending on systemd meta module causes circular dependency,
prevents the existance of multiple systemd based meta modules
and is redunant.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-ask-password/module-setup.sh
+++ b/modules.d/01systemd-ask-password/module-setup.sh
@@ -17,9 +17,7 @@ check() {
 # Module dependency requirements.
 depends() {
 
-    # This module has external dependency on the systemd module.
-    echo systemd
-    # Return 0 to include the dependent systemd module in the initramfs.
+    # Return 0 to include the dependent module(s) in the initramfs.
     return 0
 
 }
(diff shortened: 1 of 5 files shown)


modules.d/01systemd-sysusers/module-setup.sh
```diff
@@ -18,7 +18,6 @@ check() {
 depends() {
 
     # This module has external dependency on the systemd module.
```
haraldh, line 20: remove the comment as well

modules.d/01systemd-repart/module-setup.sh
```diff
@@ -17,7 +17,6 @@ check() {
 depends() {
 
     # This module has external dependency on the systemd module.
```
haraldh, line 19: remove the comment as well

modules.d/01systemd-ask-password/module-setup.sh
```diff
@@ -18,7 +18,6 @@ check() {
 depends() {
 
     # This module has external dependency on the systemd module.
```
haraldh, line 20: remove the comment as well

haraldh (changes requested):

PR #1377, opened 2021-04-21
90kernel-modules: add watchdog drivers for generic initrd

nullr0ute:
The watchdog module pulls in the device specific watchdog if that
module is enabled, but in the case where we need a generic initrd
we don't get all watchdog drivers which means if we have a watchdog
enabled for that usecase it may get kicked too late in the boot
process so we need the drivers in the initrd for the generic case too.

This pull request changes...

## Checklist

Fixes #


johannbg (approved):
LGTM

haraldh (approved):

PR #1382, opened 2021-04-22
fix(network-manager): set timeout via command line option

haraldh:
With Fedora 32, only the command line option works for `nm-online`.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-manager/nm-wait-online-initrd.service
+++ b/modules.d/35network-manager/nm-wait-online-initrd.service
@@ -8,9 +8,8 @@ ConditionPathExists=/run/NetworkManager/initrd/neednet
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/nm-online -s -q
+ExecStart=/usr/bin/nm-online -s -q -t 3600
 RemainAfterExit=yes
-Environment=NM_ONLINE_TIMEOUT=3600
 
 [Install]
 WantedBy=initrd.target


haraldh:
CC @dustymabe

johannbg (changes requested):
Presumably this is an ongoing work since there is nothing that parsing the cmd and puts a value there ( you are setting a fixed value of 3600 )  + nm-online defaults to 30s timeout ( which is resonably sane hence a fixed timeout value should not be needed there in the first place). 

Not sure why there was an hour in that Environment line ( 3600/60 = 60 minutes ) so it must be somekind of a mistake. 

Defaulting to 300 ( 5 minutes ) would be a more of an industry "standard" entry for an upper limit and I'm actually quite interesting to hear why an hour is needed.

haraldh:
I didn't change the value. I just move it to the command line. Changing the value could be another PR.

johannbg:
> I didn't change the value. I just move it to the command line. Changing the value could be another PR.

The environmental line should just be remove since the value in it makes no sense but let's hear what @dustymabe has to say.

dustymabe:
cc @bengal 

I agree 60 minutes seems a little excessive. 

Otherwise the change LGTM.

bengal:
The service delays the initqueue until NM has finished activating (successfully or not) devices. Currently each device has a default timeout of 90 seconds for DHCP, but the user can change that value via kernel command line (`rd.net.timeout.dhcp=`).

If the nm-online timeout is too short, it could trigger before the user-supplied DHCP timeout is reached. Ideally the nm-online timeout should be infinity, so that it's never reached and the configured timeout for devices takes effect. However, I considered 1 hour long enough for any reasonable use case. Yeah, perhaps it can be lowered further.

bengal:
The PR looks good to me.

johannbg:
@bengal as @haraldh has already pointed out it makes no sense to try to increase the timeout for nm-online longer then what is set by systemd since it will have no effect. If you plan on doing that you have to increase the system wide default timeouts ( via configuration snippet ) so that glorious hour that has been set there is useless.

haraldh:
> @bengal as @haraldh has already pointed out it makes no sense to try to increase the timeout for nm-online longer then what is set by systemd since it will have no effect. If you plan on doing that you have to increase the system wide default timeouts ( via configuration snippet ). That glorious hour that has been set there is useless. 

@johannbg I thought you can the timeout for individual units as well

johannbg:
@haraldh hmm that can circumvent the system wide defaults without having to increase it first? I will have to double check on that.

PR #1384, opened 2021-04-22
fix(systemd-hostnamed): unmatching quotes

ryncsn:
Just fix a simple syntax issue.

## Checklist

### DIFF
--- a/modules.d/01systemd-hostnamed/module-setup.sh
+++ b/modules.d/01systemd-hostnamed/module-setup.sh
@@ -44,6 +44,6 @@ install() {
         inst_multiple -H -o \
             /etc/hostname \
             "$systemdsystemconfdir"/systemd-hostnamed.service \
-            "$systemdsystemconfdir"/systemd-hostnamed.service.d/*.conf"
+            "$systemdsystemconfdir"/systemd-hostnamed.service.d/*.conf
     fi
 }


modules.d/01systemd-hostnamed/module-setup.sh
```diff
@@ -44,6 +44,6 @@ install() {
         inst_multiple -H -o \
             /etc/hostname \
             "$systemdsystemconfdir"/systemd-hostnamed.service \
-            "$systemdsystemconfdir"/systemd-hostnamed.service.d/*.conf"
+            "$systemdsystemconfdir"/systemd-hostnamed.service.d/*.conf
```
johannbg, line 47: Interesting I fixed this last night but apparently forgot to push it, anyway fix it to look like this 
` "$systemdsystemconfdir/systemd-hostnamed.service.d/*.conf" `
since the `dir"/systemd` quote is the wrong one

johannbg (changes requested):

modules.d/01systemd-hostnamed/module-setup.sh
haraldh, line 47:
```suggestion
            "$systemdsystemconfdir/systemd-hostnamed.service.d/*.conf"
```

haraldh (changes requested):

haraldh:
Fixed with https://github.com/dracutdevs/dracut/pull/1387
Thanks for the PR!

modules.d/01systemd-hostnamed/module-setup.sh
ryncsn: Thanks for the fix.

PR #1392, opened 2021-04-22
fix: quotes in systemd modules

johannbg:
Fixing quotes around *.conf entries in systemd modules

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-repart/module-setup.sh
+++ b/modules.d/01systemd-repart/module-setup.sh
@@ -5,7 +5,7 @@
 # Prerequisite check(s) for module.
 check() {
 
-    # If the binary(s) requirements are not fulfilled the module can't be installed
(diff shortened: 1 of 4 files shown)


modules.d/01systemd-repart/module-setup.sh
```diff
@@ -35,6 +35,6 @@ install() {
         inst_multiple -H -o \
             /etc/repart.d/*.conf \
```
haraldh, line 36: what about this one?

haraldh (changes requested):

modules.d/01systemd-repart/module-setup.sh
johannbg: There is always that one that gets away :smile:  
Fixed.

haraldh (approved):

PR #1393, opened 2021-04-23
WIP: feat(iwd): introducing the iwd module

johannbg:
Introducing the iwd module

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35iwd/module-setup.sh
+++ b/modules.d/35iwd/module-setup.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries \
+        iwmon \
+        iwctl \
+        || return 1
+
+    require_any_binary \
+        /usr/lib/iwd \
+        /usr/libexec/iwd \
+        || return 1
+
+    # Return 255 to only include the module, if another module requires it.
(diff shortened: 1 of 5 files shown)


johannbg:
Needs a generator to configure the SSID and password for iwd from the kernel command line

johannbg:
@dustymabe @bengal it would be nice to get some feedback how on the NM module will pick up and use/manage iwd in the initrd. This is not a problem with networkd since we are just using upstream type unit files however the NM module is somewhat special ( as opposed to just use the exact same type unit as is used in real root ) and I'm unsure if it's enough to just use the iwd_backend.conf file as you would do with NM 1.30+ and works fine on F34

modules.d/35iwd/module-setup.sh
```diff
@@ -0,0 +1,82 @@
...
+
+installkernel() {
+    instmods '=drivers/net/wireless'
+}
+
+# Install the required file(s) and directories for the module in the initramfs.
+install() {
+
+    inst_rules "$moddir/99-iwd.rules"
+    inst_simple "$moddir/iwd.service" "$systemdsystemunitdir"/iwd.service
```
bengal, line 42: This file is not included in the commit.

bengal:
> I'm unsure if it's enough to just use the iwd_backend.conf file as you would do with NM 1.30+ and works fine on F34

Yes, it should be enough to enable IWD in NM.

polarathene:
I think you need to have the stale bot ignore PR / issues by using an explicit label to opt-out, then you won't have this open/close battle :sweat_smile:

PR #1396, opened 2021-04-25
[WIP] Understand dbus-daemon.service

mikhailnov:
ROSA Linux uses dbus daemon, not dbus broker, and it has:

- dbus-daemon.service
- dbus.socket

Let's understand both variants.

dbus.service was not found.

### DIFF
--- a/modules.d/06dbus-daemon/module-setup.sh
+++ b/modules.d/06dbus-daemon/module-setup.sh
@@ -51,9 +51,30 @@ install() {
         "$dbusservicesconfdir"/org.freedesktop.systemd1.service \
         "$dbussystemservices"/org.freedesktop.systemd1.service \
         "$systemdsystemunitdir"/dbus.service \
+        "$systemdsystemunitdir"/dbus-daemon.service \
         "$systemdsystemunitdir"/dbus.socket \
+        "$systemdsystemunitdir"/dbus-daemon.socket \
         "$systemdsystemunitdir"/dbus.target.wants \
         busctl dbus-send dbus-daemon
+
+    local dbus_service=""
+    if [ -e "${initdir}${systemdsystemunitdir}/dbus-daemon.service" ]
+        then dbus_service="dbus-daemon.service"
+        else dbus_service="dbus.service"
+    fi
+    local dbus_socket=""
+    if [ -e "${initdir}${systemdsystemunitdir}/dbus-daemon.socket" ]
+        then dbus_socket="dbus-daemon.socket"
+        else dbus_socket="dbus.socket"
+    fi
+    if [ ! -e "${initdir}${systemdsystemunitdir}/${dbus_service}" ]; then
+        derror "dbus systemd service file not found!"
+        return 1
+    fi
+    if [ ! -e "${initdir}${systemdsystemunitdir}/${dbus_socket}" ]; then
+        derror "dbus systemd socket file not found!"
+        return 1
+    fi
 
     # Adjusting dependencies for initramfs in the dbus service unit.
     # shellcheck disable=SC1004
@@ -61,7 +82,7 @@ install() {
         '/^\[Unit\]/aDefaultDependencies=no\
         Conflicts=shutdown.target\
         Before=shutdown.target' \
-        "$initdir$systemdsystemunitdir/dbus.service"
+        "$initdir$systemdsystemunitdir/${dbus_service}"
 
     # Adjusting dependencies for initramfs in the dbus socket unit.
     # shellcheck disable=SC1004
@@ -70,7 +91,7 @@ install() {
         Conflicts=shutdown.target\
         Before=shutdown.target
         /^\[Socket\]/aRemoveOnStop=yes' \
-        "$initdir$systemdsystemunitdir/dbus.socket"
+        "$initdir$systemdsystemunitdir/${dbus_socket}"
 
     # Adding the user and group for dbus
     grep '^\(d\|message\)bus:' "$dracutsysrootdir"/etc/passwd >> "$initdir/etc/passwd"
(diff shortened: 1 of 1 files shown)


johannbg:
@mikhailnov hmm this should not be necessary. 
Basically dracut has 3 modules dbus meta module, dbus-broker and dbus-daemon . 
The dbus meta module pulls in either the dbus-broker module or dbus depending on what is installed on the system.

johannbg:
@mikhailnov I think understand what the problem is. Unlike upstream dbus-broker,  dbus-daemon upstream does not seem to be shipping it's own type unit files which has resulted in downstream distribution start naming dbus-daemon units as they see fit ( typical non technical downstream nonsense on the core/baseOS layer like apache in Debian vs httpd in Fedora ) which we most certainly wont support here upstream since that can end up in gazillion variants and if statements ( there are atleast 220 systemd based distributions in the wild that I'm aware of which all could differ from one another ) or I simply made a mistake when I made these. 

Anyway that means we have to pick one type unit names and support that for dbus-daemon and downstream will either have to a) adopt that name or b) patch it downstream should the distribution use a different type unit name. 

Now the name the name that will be chosen will have to be based on what majority of the 5 primary distributions ( Arch,Debian,Fedora,Gentoo,OpenSuse ) use so you probably end up simply having to reduce your pr and replace dbus.service and dbus.socket with dbus-daemon.service and dbus-daemon.socket ( or dbus.socket if that's what's most commonly used ).

mikhailnov:
My PR is only about detecting dbus-daemon.service when necessary, what do you suggest to reduce?

mikhailnov:
dbus-daemon.service has `Alias=dbus.service`, so it can be found like this:

```
$ systemctl cat dbus.service | head -n 1 | awk '{print $NF}'
/lib/systemd/system/dbus-daemon.service
```

(an a bit better code with additional checks is required)

johannbg:
Just shrink your entire PR down to changing the name from 'dbus.service` to dbus-daemon.service` .

We dont need the if statements there and if it comes down to it we just add another include which will silently fail if the files dont exist.

Also note that we are probably eventually going to remove all these sections
`
sed -i -e \
        '/^\[Unit\]/aDefaultDependencies=no\
        Conflicts=shutdown.target\
        Before=shutdown.target' \
`
And ship our own type units instead.

johannbg:
Also change your commit messasge to `fix(dbus-daemon): correcting type unit names`

mikhailnov:
So you are sure that all distros using dbus-daemon name it as dbus-daemon.service, not dbus.service, and so there is no need to support both variants?

mikhailnov:
It is `dbus.service` in Ubuntu:

```
[user@rosa2019 tmp]$ ls 0/lib/systemd/system/dbus*
0/lib/systemd/system/dbus-org.freedesktop.hostname1.service
0/lib/systemd/system/dbus-org.freedesktop.import1.service
0/lib/systemd/system/dbus-org.freedesktop.locale1.service
0/lib/systemd/system/dbus-org.freedesktop.login1.service
0/lib/systemd/system/dbus-org.freedesktop.machine1.service
0/lib/systemd/system/dbus-org.freedesktop.portable1.service
0/lib/systemd/system/dbus-org.freedesktop.timedate1.service
0/lib/systemd/system/dbus.service
0/lib/systemd/system/dbus.socket
```

and it is not dbus broker

johannbg:
Ubuntu is not part of the primary's ( Debian is ) so whatever Ubuntu uses is irrelevant and as I mentioned earlier we are not going to fall into the trap of start taking into account whatever downstream randomly decides to name their type units or other files and directories on the core/base OS level and support every variant of it.

What decides will always have to be upstream naming scheme or if it does not exist, what is used by the majority of the primary's ( Arch,Debian, Fedora,Gentoo, OpenSuse ) which means used in three of them or more.

mikhailnov:
Do you know names of unit service of the classical dbus daemon used in Arch,Debian, Fedora,Gentoo, OpenSuse?

johannbg:
Arch --> dbus.service
Debian --> dbus.service
Fedora --> dbus-daemon.service
Gentoo --> dbus.service
OpenSuse --> dbus.service

So basically it's only Fedora but Fedora uses dbus-broker not dbus daemon ( so it uses the dbus-broker module ) so it should not be required for it's derivatives.

The other primaries including their derivatives will be using dbus.service up to the point they decide to default using dbus-broker then most likely they will rename the type unit file from dbus.service to dbus-daemon.service anyway from my point of view I see no incentive of supporting the dbus-daemon for it's service name in the dbus-daemon module since all the primaries are using the dbus for it's type unit name and the one that is not is not using dbus-daemon et all.

haraldh:
```
❯ cat /usr/lib/systemd/system/dbus-broker.service
[…]
[Install]
Alias=dbus.service
```

haraldh:
```
❯ ls -l /etc/systemd/system/dbus.service
lrwxrwxrwx. 1 root root 43 23. Apr 10:04 /etc/systemd/system/dbus.service -> /usr/lib/systemd/system/dbus-broker.service
```

mikhailnov:
This pull request is not full actually, so, I did not notice it, for some reasons dbus.socket inside initrd tries to find dbus.service, but cannot find it and fails to launch inside initrd, that it probably because a symlink for the alias is not installed.

@haraldh - I also have this symlink in ROSA:
```
$ ls /etc/systemd/system/dbus* -la
lrwxrwxrwx 1 root root 40 апр 23 14:46 /etc/systemd/system/dbus-org.freedesktop.Avahi.service -> /lib/systemd/system/avahi-daemon.service
lrwxrwxrwx 1 root root 44 дек 24 07:21 /etc/systemd/system/dbus-org.freedesktop.network1.service -> /lib/systemd/system/systemd-networkd.service
lrwxrwxrwx 1 root root 45 дек 25 15:17 /etc/systemd/system/dbus-org.freedesktop.timesync1.service -> /lib/systemd/system/systemd-timesyncd.service
lrwxrwxrwx 1 root root 39 дек 24 07:21 /etc/systemd/system/dbus.service -> /lib/systemd/system/dbus-daemon.service
```

But in theory dbus service may be not enabled when building initrd, so this symlink will not exist... Maybe we can try reading it if it exists, but some systemd in theory may use /run/systemd instead of /etc/systemd, trying to follow all possible cases will make dracut code too complex...

mikhailnov:
Maybe we can run `systemctl --root="$dracutsysrootdir" dbus.service` ?

mikhailnov:
But ROSA's dbus.socket is the following:

```
$ systemctl cat dbus.socket
# /lib/systemd/system/dbus.socket
[Unit]
Description=D-Bus System Message Bus Socket

[Socket]
ListenStream=/run/dbus/system_bus_socket

[Install]
WantedBy=sockets.target
```

A simpler solution will be adding `Service=dbus-daemon.service` into it.

but running systemctl on dracut's root looks like an interesting idea. I hope it does not require running dbus on the host when building initrd.

johannbg:
This is a nack hence closing, further discussion can be had in #1487

PR #1405, opened 2021-04-27
fix(network): use wicked unit instead of find_binary

johannbg:
dracut-install triggers "ERROR: installing '/usr/lib/wicked/bin/*'"
when find_binary is being used to determine if wicked is installed
so let's just check for the wicked type service unit instead.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/40network/module-setup.sh
+++ b/modules.d/40network/module-setup.sh
@@ -17,7 +17,7 @@ depends() {
     done
 
     if [ -z "$network_handler" ]; then
-        if find_binary wicked &> /dev/null; then
+        if [[ -x $dracutsysrootdir$systemdsystemunitdir/wicked.service ]]; then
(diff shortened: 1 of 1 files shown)


modules.d/40network/module-setup.sh
```diff
@@ -17,7 +17,7 @@ depends() {
     done
 
     if [ -z "$network_handler" ]; then
-        if find_binary wicked &> /dev/null; then
+        if [[ -x $dracutsysrootdir$systemdsystemunitdir/wicked.service ]]; then
```
Vogtinator, line 20: This is broken and always fails now: The service file is never executable.

Vogtinator: (CC @danimo)

PR #1408, opened 2021-04-27
fix(squash): only load modules if they are not already loaded

johannbg:
Let's only load modules if they are not already loaded

This pull request changes...

## Checklist

Fixes #


haraldh:
I wouldn't do that... it's not that time critical and it does not cover module aliases

PR #1409, opened 2021-04-28
fix(udev-rules): remove sourcing of network link files

johannbg:
Removing sourcing of network link files from the udev-rules module.
What always should be included should be placed in the systemd-network
module and other modules that provide/require spesific link files
should add them, themselves via their relevant include section.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-legacy/module-setup.sh
+++ b/modules.d/35network-legacy/module-setup.sh
@@ -21,6 +21,13 @@ installkernel() {
 # called by dracut
 install() {
     local _arch
+
+    #Adding default link
+    if dracut_module_included "systemd"; then
+        inst_multiple -o "${systemdutildir}/network/99-default.link"
+        [[ $hostonly ]] && inst_multiple -H -o "${systemdsystemconfdir}/network/*.link"
+    fi
+
     inst_multiple ip dhclient sed awk grep pgrep tr
 
     inst_multiple -o arping arping2
--- a/test/TEST-30-ISCSI/test.sh
+++ b/test/TEST-30-ISCSI/test.sh
@@ -242,8 +242,7 @@ test_setup() {
         )
         inst /etc/passwd /etc/passwd
         inst_multiple sh ls shutdown poweroff stty cat ps ln ip \
-            dmesg mkdir cp ping \
-            modprobe tcpdump setsid \
+            dmesg mkdir cp ping modprobe tcpdump setsid \
             /etc/services sleep mount chmod pidof
         inst_multiple tgtd tgtadm
         for _terminfodir in /lib/terminfo /etc/terminfo /usr/share/terminfo; do
@@ -314,7 +313,6 @@ test_setup() {
         inst_multiple poweroff shutdown
         inst_hook shutdown-emergency 000 ./hard-off.sh
         inst_hook emergency 000 ./hard-off.sh
-        inst_simple ./99-default.link /etc/systemd/network/99-default.link
     )
 
     # Make server's dracut image
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -45,6 +45,10 @@ install() {
         inst_simple "$moddir"/nm-initrd.service "$systemdsystemunitdir"/nm-initrd.service
         inst_simple "$moddir"/nm-wait-online-initrd.service "$systemdsystemunitdir"/nm-wait-online-initrd.service
 
+        # Adding default link
+         inst_multiple -o "${systemdutildir}/network/99-default.link"
+         [[ $hostonly ]] && inst_multiple -H -o "${systemdsystemconfdir}/network/*.link"
+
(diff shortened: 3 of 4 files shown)


haraldh:
so, yeah, please add them to the relevant modules in the same commit

johannbg:
> so, yeah, please add them to the relevant modules in the same commit

Sorry I dont follow the code is blindly including everything from the link directory `"${systemdutildir}/network/*.link"` <-- so there is no way of telling what which modules needs + the systemd-networkd module already includes the upstream link files so a) if it provides the .link file in question those modules should depend on it or b) depend on the module that provides required .link file and that module have dependency on systemd-networkd module.

haraldh:
Well, in the case of non-systemd PID 1, we still need the network interface renaming policy. So, the network-legacy module should install the /lib/systemd/network/*.link files at least, which would make the TEST-30 patching obsolete, I guess.

test/TEST-30-ISCSI/test.sh
```diff
@@ -260,7 +257,6 @@ test_setup() {
         inst ./dhcpd.conf /etc/dhcpd.conf
         inst_multiple /etc/nsswitch.conf /etc/rpc /etc/protocols
         inst /etc/group /etc/group
-
         cp -a /etc/ld.so.conf* "$initdir"/etc
         ldconfig -r "$initdir"
         dracut_kernel_post
```
haraldh, line 262: Hmm, you didn't change anything here, besides whitespace

test/TEST-35-ISCSI-MULTI/test.sh
```diff
@@ -317,6 +317,7 @@ test_setup() {
         inst_multiple poweroff shutdown
         inst_hook shutdown-emergency 000 ./hard-off.sh
         inst_hook emergency 000 ./hard-off.sh
+        inst_simple ./99-default.link /etc/systemd/network/99-default.link
```
haraldh, line 320: Is this still needed with the `modules.d/35network-legacy/module-setup.sh` change?

modules.d/35network-legacy/99-default.link
```diff
@@ -0,0 +1,8 @@
+# This file is part of dracut network-legacy module.
```
haraldh, line 1: Why don't we take the systemd shipped one, so we leverage all the potential new options?

test/TEST-40-NBD/test.sh
```diff
@@ -225,7 +225,7 @@ make_encrypted_root() {
         inst_binary "${basedir}/dracut-util" "/usr/bin/dracut-util"
         ln -s dracut-util "${initdir}/usr/bin/dracut-getarg"
         ln -s dracut-util "${initdir}/usr/bin/dracut-getargs"
-
+        inst_simple ./99-default.link /etc/systemd/network/99-default.link
```
haraldh, line 228: Is this still needed with the modules.d/35network-legacy/module-setup.sh change?

Same for the rest of the changes.

modules.d/35network-legacy/module-setup.sh
```diff
@@ -32,6 +32,7 @@ install() {
     inst_script "$moddir/ifup.sh" "/sbin/ifup"
     inst_script "$moddir/dhcp-multi.sh" "/sbin/dhcp-multi.sh"
     inst_script "$moddir/dhclient-script.sh" "/sbin/dhclient-script"
+    inst_script "$moddir/99-default.link" "/usr/lib/systemd/network/99-default.link"
```
haraldh, line 35: Hmm, this is missing in the `35network-manager` module also

haraldh (changes requested):

modules.d/35network-legacy/module-setup.sh
johannbg: Hmm none of the nm test failed after adding this to legacy but I'll add it and throw a test at test-30 to see if it fails without the include line in the process.

test/TEST-40-NBD/test.sh
johannbg: Yup this was still needed for that test after the network-legacy change. ( did not revert the changes in the other test to check if they worked or not ).

test/TEST-30-ISCSI/test.sh
johannbg: At one point it included an include line for the link I can add the whitespace back

modules.d/35network-legacy/99-default.link
johannbg: Because you cannot guarantee that systemd is installed on those alternative init systems

johannbg: For the legacy network stuff it probably should be a udev rule not an .link file so this works on alternative init systems ( silly to be placing it under /usr/lib/systemd/network/ since systemd does not exist on those systems )

haraldh (approved):

PR #1412, opened 2021-04-29
feat(tpm2-tss): introducing the tpm2-tss module

johannbg:
Introducing the tpm2-tss module. 

Note this depends on the udev global var PR so dont merge until after we have merged it.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/dracut.spec
+++ b/dracut.spec
@@ -367,6 +367,7 @@ echo 'dracut_rescue_image="yes"' > $RPM_BUILD_ROOT%{dracutlibdir}/dracut.conf.d/
 %{dracutlibdir}/modules.d/90qemu
 %{dracutlibdir}/modules.d/91crypt-gpg
 %{dracutlibdir}/modules.d/91crypt-loop
+%{dracutlibdir}/modules.d/91tpm2-tss
 %{dracutlibdir}/modules.d/95debug
 %{dracutlibdir}/modules.d/95fstab-sys
 %{dracutlibdir}/modules.d/95lunmask
--- a/modules.d/91tpm2-tss/module-setup.sh
+++ b/modules.d/91tpm2-tss/module-setup.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+
+}
+
+# Module dependency requirements.
+depends() {
+
+    # This module has external dependency on other module(s).
+    echo systemd-sysusers systemd-udev
+    # Return 0 to include the dependent module(s) in the initramfs.
+    return 0
+
+}
+
+# Install kernel module(s).
+installkernel() {
+    instmods '=drivers/char/tpm'
+}
+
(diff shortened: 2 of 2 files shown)


haraldh:
please add the following binaries:
```
            tpm2_pcrread tpm2_pcrextend tpm2_createprimary tpm2_createpolicy \
            tpm2_create tpm2_load tpm2_unseal 
```

haraldh (changes requested):
please add the following binaries:
```
            tpm2_pcrread tpm2_pcrextend tpm2_createprimary tpm2_createpolicy \
            tpm2_create tpm2_load tpm2_unseal 
```

haraldh:
Want to extend the `check()`??

johannbg:
> Want to extend the `check()`??

Do we want to do that I mean those binaries are not part of upstream tpm2-tss module ( which is why I did not add them in the first place ) nor Fedora's tpm2-tss.  Now by adding them without adding them to check() section we get a "soft" dependcy as in those binaries are included if available but if not, the module can in sense just be used as an tpm2-tss "library" package.

But if you want them as a strict requirement ( as in they are always needed ) I can add them.

johannbg:
@nullr0ute given that you are somewhat involved with the maintainership of tpm2 in Fedora, can you explain why the json files are being placed in /etc also the dbus files for the tpm2-abrmd component ( as opposed to under /usr ) ?

nullr0ute:
> @nullr0ute given that you are somewhat involved with the maintainership of tpm2 in Fedora, can you explain why the json files are being placed in /etc also the dbus files for the tpm2-abrmd component ( as opposed to under /usr ) ?

No, I can't, TBH I've not looked at the tpm2-abrmd closely since it was dumped on me from the Intel maintainer, and generally abrmd isn't needed as generally these days you'll want to use the kernel resource manager at /dev/tpmrm0 rather than the userspace resource manager. I would definitely not be pulling abrmd into the initrd.

nullr0ute:
> Introducing the tpm2-tss module.

What does this actually achieve/what's it used for?

nullr0ute:
> Thanks for the heads up regarding abrmd I was wondering if I should create a module for that. What about those json file can we just skip them or are they being used by something in Fedora and or in the wild? ( this module would then just contain the libraries + tpm2 and it's links if available )

Any chance you could file a RHBZ and I'll take a look, don't have the cycles right now to dig into it. Generally none of the stuff I do with TPM2 in Fedora uses abrmd, it all uses the kernel resource manager.

johannbg:
> > Introducing the tpm2-tss module.
> 
> What does this actually achieve/what's it used for?

Cryptsetups

johannbg:
> > Thanks for the heads up regarding abrmd I was wondering if I should create a module for that. What about those json file can we just skip them or are they being used by something in Fedora and or in the wild? ( this module would then just contain the libraries + tpm2 and it's links if available )
> 
> Any chance you could file a RHBZ and I'll take a look, don't have the cycles right now to dig into it. Generally none of the stuff I do with TPM2 in Fedora uses abrmd, it all uses the kernel resource manager.

Dont waste your time digging into it. We will simply exclude the files and add them if and then when an RFE arrives for them.

haraldh (approved):

PR #1416, opened 2021-04-30
fix(integrity): Properly set up EVM when using an x509 cert

stefanberger:
The current EVM script does not handle the EVM setup properly when X509
certificates are involved. In this patch we extend the setup and add
the necessary flags for support of EVM activation that include
x509 certificates, possibly in conjunction with an HMAC key. We also
first try activating EVM for x509 certificates using
EVM_ALLOW_METADATA_WRITES for newer kernels, then without it for older
ones that did not support this flag.

We add support for additional EVM activation bits to be set, such
as EVM_SETUP_COMPLETE (0x80000000) via the config file and
EVM_ACTIVATION_BITS variable.

To avoid error messages related to unloading the HMAC key if none is
used, only attempt to unload the HMAC key if one was actually set.

We add documentation about the variables that can be set in the EVM
config file.

Signed-off-by: Stefan Berger <stefanb@linux.ibm.com>
Reviewed-by: Mimi Zohar <zohar@linux.ibm.com>
Cc: Roberto Sassu <roberto.sassu@huawei.com>

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/98integrity/evm-enable.sh
+++ b/modules.d/98integrity/evm-enable.sh
@@ -11,6 +11,15 @@ EVMCONFIG="${NEWROOT}/etc/sysconfig/evm"
 EVMKEYDESC="evm-key"
 EVMKEYTYPE="encrypted"
 EVMKEYID=""
+EVM_ACTIVATION_BITS=0
+
+# The following variables can be set in /etc/sysconfig/evm:
+# EVMKEY: path to the symmetric key; defaults to /etc/keys/evm-trusted.blob
+# EVMKEYDESC: Description of the symmetric key; default is 'evm-key'
+# EVMKEYTYPE: Type of the symmetric key; default is 'encrypted'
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
Minor nitpick we don't include module numbers in commit messages since those can be subjected to changes so change your commit msg from 
`fix(98integrity): Properly set up EVM when using an x509 cert` 
to  
`fix(integrity): Properly set up EVM when using an x509 cert` 
other than that LGTM.

stefanberger:
> Minor nitpick we don't include module numbers in commit messages since those can be subjected to changes so change your commit msg from
> `fix(98integrity): Properly set up EVM when using an x509 cert`
> to
> `fix(integrity): Properly set up EVM when using an x509 cert`
> other than that LGTM.

Fixed and pushed again to entertain the ci/cd a bit more.

johannbg (approved):

haraldh (approved):

Issue #1404, opened 2021-04-26
add option to make network-legacy be the default (?)


mikhailnov:
Currently the logics of including a network module is the following on a typical Linux desktop without wicked:

- try network-manager, if ok, put it into the initrd
- try systemd-networkd, if ok, put it into the initrd
- use network-legacy otherwize

So, NetworkManager is chosen in most cases on Linux desktops and even servers. The problems that I see are the following:

- big size of initrd, longer compression time (not a very big problem)
- if NetworkManager and dbus are updated, initrd should be regenerated to ensure smooth transition of network setup state from initrd to the main system

I do not see any big advantages that NM gives so that rebuilding initrd would be worth spending time and resources. To my mind, may be I am missing something, configuring network in initrd is not required in by far the vast majority of cases, and, when it is needed, a simple dhclient from network-legacy fits most of the needs. A typical desktop or server system does not require network in initrd; when it is necessary if, for example, a PXE boot: anaconda-dracut downloads a kickstart file, a rootfs image etc. In this case, I think, it is not so important to transfer state of the network setup from initrd to the main system.

Are there usecases when NM+dbus is a really great solution? Are those cases ot rare so that it makes sense to make NM be the default choice?

I do not like a so much overcomplicated initrd by default and I am thinking of making network-legacy ot systemd-networkd the default choice in the distro (ROSA 2019.1+). I can explicitly list the needed module in the default config like this:
`add_modules+=" network-legacy "`
but how should a user who want another network module overwrite this default by a config in /etc/dracut without touching the default config in /usr/lb/dracut? Probably there is no way to do it cleanly, so I am thinking about adding a special option like:
`preferred_network_module=network-legacy`
or something else like that.

Here are my thoughts. I would be happy to hear arguements for and against NetworkManager as the default choice and thoughts about such option.

johannbg:
It's up to distributions/vendors/users to decide what they considered default and configure it as such. 
The network module is only a meta module which serves as the role for other modules that require networking to depend upon.

If there ever would be a default it would end up being systemd networkd since the legacy network stack should be considered obsolete at this point and replaced with more modern solutions iwd,networkd,NM, Wicked etc.

[...]

PR #1421, opened 2021-05-01
network: add option to specify preferred network module

mikhailnov:
Most desktop Linux systems have NetworkManager, many of them have systemd-networkd inside the systemd package and many of them have dhclient installed.

Current code oft he network meta module, if no specific network module is included, pulls NetworkManager and its dependency dbus into the initramfs.

I want to avoid such complex systems inside initrd and want to avoid necessity to rebuild initrds when NetworkManager and dbus packages are updated to keep them in the initrd and on the host in sync, and also want to standartize initrds by having the same network module in most of them.

If I add a specific network module into the distribution (ROSA) default config of dracut, e.g. let it be systemd-networkd, dracut will try to include it even on systems where systemd-networkd is not installed and fill fail with error.

So adding a config option which allows to change the _preferred_ default if no specific network module is configured to be added into the initrd.

For now I will set it in ROSA to network-legacy to avoid pulling dbus, maybe it will have to be changed to systemd-networkd later when this module matures in dracut or when network-legacy is dropped.

Fixes: https://github.com/dracutdevs/dracut/issues/1404

### DIFF
--- a/dracut.conf.5.asc
+++ b/dracut.conf.5.asc
@@ -106,6 +106,11 @@ Configuration files must have the extension .conf; other extensions are ignored.
 *tmpdir=*"__<temporary directory>__"::
     Specify temporary directory to use.
 
+*preferred_network_handler=*"__{network-wicked|network-manager|network-legacy|systemd-networkd}__"::
+    Multiple modules which can setup network network are available.
+    By default one of them is chosen.
+    You can specify a preferred one which will be used if it is available.
+
 [WARNING]
 ====
 If chrooted to another root other than the real root device, use --fstab and
--- a/modules.d/40network/module-setup.sh
+++ b/modules.d/40network/module-setup.sh
@@ -9,12 +9,23 @@ check() {
 depends() {
     is_qemu_virtualized && echo -n "qemu-net "
 
-    for module in network-wicked network-manager network-legacy systemd-networkd; do
-        if dracut_module_included "$module"; then
-            network_handler="$module"
-            break
-        fi
-    done
+    preferred_network_handler="${preferred_network_handler:-none}"
+    case "$preferred_network_handler" in
+		none )
+			for module in network-wicked network-manager network-legacy systemd-networkd; do
+				if dracut_module_included "$module"; then
+					network_handler="$module"
+					break
+				fi
+			done
+		;;
+		network-wicked | network-manager | network-legacy | systemd-networkd )
+			network_handler="$preferred_network_handler"
+		;;
+		* )
+			dwarn "Preferred network module $preferred_network_handler is not known, ignoring."
+		;;
+    esac
 
     if [ -z "$network_handler" ]; then
         if [[ -x $dracutsysrootdir$systemdsystemunitdir/wicked.service ]]; then


johannbg:
Well this is a nack from me and should not be carried here upstream since you are effectively trying to bring a downstream policy/integration problem here upstream and expect it to be worked around/solved here ( as opposed on a packaging level ). Just carry this patch downstream.

mikhailnov:
No, I am trying to avoid making downstream hacks to workaround the default upstream policy being tot suitable for downstream. Other distros will meet with the same problem. I have described why solutions like including a specific network module by default and avoiding usage of the meta module by this are not suitable. The only solution instead of this one is patching the meta module "network" and just reorder modules in the code. It is a hack. Proposed patch, to my mind, is useful for many downstreams who use dracut for different purposes, not build a special-purpose embeded system, that is why I have brought it here to upstream. There is no sense to try to carry it downstream without upstreamizing because it is a generic solution which will be useful for many users and not a downstream hack. Most downstreams including me are interested to make upstream meet their needs instead of patching it by hack in different places.

mikhailnov:
As I understood from your comments in https://github.com/dracutdevs/dracut/issues/1404 , NetworkManager as a preffered default choice is a near to random choice (probably there was no systemd-network module at that moment), so it makes sense to make to allow users to make it less random. Otherwise the meta module does not make sense at all, you may tell downstreams to just specify a specific network module explicitly, than it it not clear what the meta module exists for.

johannbg:
As I tried to explain to you there that there is no order or precedence in meta modules, it's entirely dependent on what downstream decides to include and default to and support and there will be more meta modules like this like an upcoming udev one, even init system one, that is if we dont decide to split out modules for that too keep them small easy to read,review, work on and debug and for downstream to exclude depending on what they use ( systemd vs non systemd ). 

And we wont start placing these kinds of workaround running after downstream decision makings ( or lack there of ) in them since it will only create a mess for `us` to deal with.

mikhailnov:
I don't understand which problems you are trying to solve with meta modules.  The netrwork meta module is doing automatic magic and you do not want to provide parameters to tune this automagic. In my case I, as downstream, do not _decide_ what to include, I can just tune the default variant suggested to the end user of dracut packaged by me.

> it's entirely dependent on what downstream decides to include and default to and support

In practice you, as upstream, decide what will be chosen in most cases, and it will be NetworkManager in most cases on RHEL, CentOS, Fedora, SUSE etc. You do decide and do not want to allow to provide a config parameter. I do not understand logics behind this. It is the same as completely remove the meta module and tell downstreams to deal with what the meta module does by themselves. I think it is not what you intend to do.

I, as an upstream for end users who use my RPM package of dracut, and as downstream for you -- the upstream of dracut -- simply want the metamodule to do its job and do not force my end users to use a specific network module, just tune which module will be included by default, and do it in such a way that will be useful for other distros (I believe so).

I of course can carry this patch in downstream or even make my own meta module for network (which will effectively the same as carrying this patch), but then I do not understand how you see what the upstream meta module exists for. Just as a not usable example?

mikhailnov:
You see, I am not a simply end user, I am a distro packager and my aim is to both provide a useable default and to not force users to follow my choice. And in this patch I am trying to adapt the meta module for this need.

mikhailnov:
I think that comparing this situation with e.g. systemd vs non-systemd in dracut should not be done, because distro MUST make a choice of systemd vs non-systemd in initrd, because only one case will actually work, but DOES NOT HAVE TO make a choice of the network module, because multiple cases will work.

johannbg:
> I think that comparing this situation with e.g. systemd vs non-systemd in dracut should not be done, because distro MUST make a choice of systemd vs non-systemd in initrd, because only one case will actually work, but DOES NOT HAVE TO make a choice of the network module, because multiple cases will work.

Well comparing it is relevant in the sense that it is an option that distributions have to make a choice for ( and do ) for the consumer and there are distributions out there that would disagree and ship multiple init system and even multiple initramfs generators which all is madness from my pov, madness they should carry all the maintenance burden for downstream but then again I dont think the core base/OS is about choice as in you have to pick a lane and stick with it, from supported architecture,compile options all the way to the final product which includes choosing a network management solution inside the initrd and outside it ( which should normally be the same ) with dbus or without it etc.

mikhailnov:
>
>Well comparing it is relevant in the sense that it is an option that
>distributions have to make a choice for ( and do ) for the consumer and

Most ditributions ship default choices and allow consumers to change them if needed and where it makes sense to support multiple choices. Supporting 2 init systemd does not make sense from my point of view, but supporting both systemd-networkd and networkmanager does make sense, especially when the distro supports both desktop and server usecases.

>there are distributions out there that would disagree and ship multiple
>init system and even multiple initramfs generators which all is madness
>from my pov, madness they should carry all the maintenance burden for
>downstream but then again I dont think the core base/OS is about choice
>as in you have to pick a lane and stick with it, from supported
>architecture,compile options all the way to the final product which
>includes choosing a network management solution inside the initrd and
>outside it ( which should normally be the same ) with dbus or without
>it etc. 

Most consumers of a generic-purpose linux OS do not need to setup network inside initrd. Those who need it usually have very custom configs and understan what they are doing, I am not trying to support all such possible usecases out of the box, the only usecase that is not rare and requires network inside initrd is PXE boot, and a simple dhcp client (network-legacy) is enough there in most cases. I believe that we should not overcomplicate things by trying to support extremely rare usecases automatically, where consumers have enough qualification to libe without automatic help, if that automatic help spoils lifes of 99% oter users.

NetworkManager and dbus inside initrd look cool, but pulling them into most initrds and thus making more users use them and thus trying to support even more usecases and dealing with more bugs looks like a real maintainance burden.

Rebuilding initrds on every update of dbus and NM is an update burden for consumers.

From my point of view, maintainance burden is caused by code that supports usecases and conditions that cannot be tested out side of a very specific environment.

My another pull request, which adds support of dbus-daemon.service, can formally be classified as adding additional maintainance burden, yes, there I can agree with you. In practice that code is very simple and must not cause problems, but formally it does add additional maintainance burden.

But this cannot be said about the code in this pull request. It just makes the meta module more useful for different consumers in different distros, does not add anything distro-specific and does not add codepaths that cannot be tested in supported environments (and here we return to the question what the meta module must do and which problems it must solve, I see it as a helper to make choices, and this helper can be a bit cleverer than very stupid).

PR #1428, opened 2021-05-03
chore(tree): move tools into their own directory

johannbg:
Moving tools into their own directory

This pull request changes...

## Checklist

Fixes #


tools/fedora-test.sh
```diff
@@ -2,7 +2,7 @@
 
 set -ex
 
-[[ -d ${0%/*} ]] && cd "${0%/*}"
+[[ -d ${0%../*} ]] && cd "${0%../*}"
```
haraldh, line 5: `${0%/*}` is now the `tools` directory. We want `tools/..` to get to the dracut source root directory.
```suggestion
[[ -d ${0%/*} ]] && cd "${0%/*}"/../
```

tools/fedora-test-github.sh
```diff
@@ -2,7 +2,7 @@
 
 set -ex
 
-[[ -d ${0%/*} ]] && cd "${0%/*}"
+[[ -d ${0%../*} ]] && cd "${0%../*}"
```
haraldh, line 5: `${0%/*}` is now the `tools` directory. We want `tools/..` to get to the dracut source root directory.
```suggestion
[[ -d ${0%/*} ]] && cd "${0%/*}"/../
```

haraldh (changes requested):

tools/fedora-test.sh
johannbg: Done

PR #1431, opened 2021-05-04
test: use `uname -m` to determine qemu-system

haraldh:
E.g. `uname -i` does not work in Arch Linux containers.

### DIFF
--- a/test/run-qemu
+++ b/test/run-qemu
@@ -1,6 +1,7 @@
 #!/bin/bash
 # Check which virtualization technology to use
 # We prefer kvm, kqemu, userspace in that order.
+
 export PATH=/sbin:/bin:/usr/sbin:/usr/bin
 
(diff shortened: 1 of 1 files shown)


johannbg (approved):
What's the reason for it not working on Arch

haraldh:
> What's the reason for it not working on Arch

Arch:
```console
# /usr/bin/uname --version
uname (GNU coreutils) 8.32
# /usr/bin/uname -i
unknown
```

Fedora:
```console
# /usr/bin/uname --version
uname (GNU coreutils) 8.32
# /usr/bin/uname -i
x86_64
```

`strace` does not reveal why. I don't want to go into the source code.... 

`-m` sounds like we want to use it anyway:

```
  -m, --machine            print the machine hardware name
  -i, --hardware-platform  print the hardware platform (non-portable)
```

PR #1434, opened 2021-05-04
feat: also restore the initramfs from /lib/modules

haraldh:
Fallback to /lib/modules/$(uname -r)/initrd, if present and all other
files don't exist.

This pull request changes...

## Checklist

Fixes #


johannbg (approved):
btw dont we need one for the efi path as well

PR #1437, opened 2021-05-04
fix: install all depmod relevant configuration files

haraldh:
This is important for running `depmod` in the initrd.

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -128,6 +128,11 @@ installkernel() {
         hostonly='' instmods "=crypto"
         instmods "=arch/$arch/crypto" "=drivers/crypto"
     fi
+
+    inst_multiple -o "$depmodd/*.conf"
+    if [[ $hostonly ]]; then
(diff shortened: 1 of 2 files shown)


modules.d/90kernel-modules/module-setup.sh
```diff
@@ -128,6 +128,11 @@ installkernel() {
         hostonly='' instmods "=crypto"
         instmods "=arch/$arch/crypto" "=drivers/crypto"
     fi
+
+    inst_multiple -o "$depmodd/*.conf"
```
johannbg, line 132: You need to change this to only include configuration files that we ship since dont blindly include everything in directories *except* for host only setups in which the administrator is expected to know what he or she is doing and what is placed there.

johannbg (changes requested):

modules.d/90kernel-modules/module-setup.sh
haraldh: If some distro is shipping something in `/usr/lib/depmod.d` , then it has a reason and we better use it.

PR #1442, opened 2021-05-04
Support ZSTD-compressed kernel modules

mikhailnov:
Modern Linux kernels support zstd-compressed modules, which was added
by commit 73f3d1b48f50 ("lib: Add zstd modules").

Commit c3d7ef377eb ("kbuild: add support for zstd compressed modules")
added support of compressing modules with zstd to kernel Makefiles.

libkmod >= 28 built with libzstd is also required.

P.S. Support of decompressing zstd-compressed /lib/firmware seems to be missing in dracut-install, I did not try.

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -625,6 +625,7 @@ get_decompress_cmd() {
         *.gz) echo 'gzip -f -d' ;;
         *.bz2) echo 'bzip2 -d' ;;
         *.xz) echo 'xz -f -d' ;;
+        *.zst) echo 'zstd -f -d ' ;;
     esac
 }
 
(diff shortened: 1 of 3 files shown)


johannbg (changes requested):
Fix your comment msg
from 
`
Support ZSTD-compressed kernel modules
`
to
`
feat(dracut.inits.sh): support ZSTD-compressed kernel modules
`

mikhailnov:
Done. what does "feat(dracut.inits.sh)" mean?

mikhailnov:
ah, feat probably means adding a new feature?

johannbg (approved):

haraldh:
Hmm, please change `feat(dracut.inits.sh)` to `feat`, because:
1. it has typos
2. it's more a general feature for `dracut`

haraldh (approved):

mikhailnov:
Done, plus fixed name and hash of the commit which added support of zstd-compressed kernel modules.

mikhailnov:
(If someone is interested, I made measures of xz vs zstd compression of kernel modules, the results are here: https://abf.io/import/kernel-5.10/commit/ccf7aa3acbf44d8874cb010687e79bbdd8975acf , in general I did not find much practical sense in switching from xz to zstd, but zstd is a good choice)

PR #1458, opened 2021-05-07
fix: make testsuite pass on OpenSuse

haraldh:
This pull request changes...

## Checklist

Fixes #


test/TEST-20-NFS/server-99-default.link
```diff
@@ -0,0 +1,6 @@
+[Match]
```
johannbg, line 1: We should just call this server.link since it's not the default file nor does having a priority ( 99 ) make any sense since it is not being picked up and thirdly this is the only link file being included.

test/TEST-20-NFS/test.sh
```diff
@@ -406,20 +407,30 @@ test_setup() {
...
         --no-hostonly-cmdline -N \
         -f "$TESTDIR"/initramfs.testing "$KVERSION" || return 1
+
+    (
+        # shellcheck disable=SC2031
+        export initdir="$TESTDIR"/overlay
+        # shellcheck disable=SC1090
+        . "$basedir"/dracut-init.sh
+        rm "$initdir"/etc/systemd/network/99-default.link
+        inst_simple ./server-99-default.link /etc/systemd/network/99-default.link
```
johannbg, line 424: Let's just call the link file server.link to prevent any kind of confusion that this is the 99-default link file.

test/TEST-30-ISCSI/99-default.link
```diff
@@ -0,0 +1,6 @@
+[Match]
```
johannbg, line 1: Same here as previously noted no longer the default and presumably this is the "client" link file so let's just call it client.link

test/TEST-50-MULTINIC/test.sh
```diff
@@ -219,28 +219,30 @@ test_setup() {
...
 
         inst_multiple rpc.idmapd /etc/idmapd.conf
 
         inst_libdir_file 'libnfsidmap_nsswitch.so*'
         inst_libdir_file 'libnfsidmap/*.so*'
         inst_libdir_file 'libnfsidmap*.so*'
 
-        _nsslibs=$(sed -e '/^#/d' -e 's/^.*://' -e 's/\[NOTFOUND=return\]//' /etc/nsswitch.conf \
-            | tr -s '[:space:]' '\n' | sort -u | tr -s '[:space:]' '|')
+        _nsslibs=$(
```
johannbg, line 236: It's always better ( and resource cheaper ) to include our own file as opposed to be sed-ing existing ones and it does not scale since one distribution does one thing and another distribution does another.

johannbg (changes requested):
It goes without saying we should replace the link file names to client.link and server.link where applicable for every test case.

test/TEST-50-MULTINIC/test.sh
haraldh: Including more libraries in a test root is always better. We don't want failing stuff, because a library was missing, which needs to be dlopen'ed. This does no harm, other than disk space. It's not changing any config.

test/TEST-20-NFS/server-99-default.link
haraldh: true

test/TEST-20-NFS/test.sh
haraldh: yes

test/TEST-30-ISCSI/99-default.link
haraldh: yes

test/TEST-50-MULTINIC/test.sh
johannbg: Just a thought on that note maybe we can be a bit smarter about library inclusion in the future and create something like an binaries() section in which we list the binaries ( which always know since we usually add them as a requirement or in the install() section  ) which then would be parsed with something along the lines of

```
	for file in $(the-listed-binaries-in-binaries()); do \
		cp -fL $$file initramfs/$$file || exit 1;\
		if file $$file | grep -q ELF; then \
			for lddfile in $$(ldd $$file | awk '{print $$3}' | grep "^/"); do \
				file_path=$$(dirname $$lddfile); \
				cp -Ln $$lddfile initramfs/$$file_path/; \
			done \
		fi \
	done
```
And or passes the output to to inst_libdir_file instead.

Conan-Kudo:
Note that in a couple of weeks or so, openSUSE will complete UsrMerge, so that's probably important to keep in mind. Otherwise this looks remarkably good.

Conan-Kudo (approved):
Modulo existing comments, this looks fine.

test/TEST-50-MULTINIC/test.sh
haraldh: dracut-install does the `ldd` thing already. It's just for `dlopen`'ed libraries, which we can't catch.

PR #1459, opened 2021-05-07
dracut-install,dracut: fix ldd output parsing

Dzejrou:
The ldd output can contain the variable $LIB, which is a documented feature of ldd. In a previous commit [0], dracut-install received support for this variable, but that was later reverted [1] due to issues [2][3] on Gentoo ARM64.

The part before '=>' does not necessarily refer to an existing file (e.g. due to the usage of $LIB) and thus [1] could be seen as a regression to anyone that uses this ldd feature. This PR combines both cases together and whenever it find a '$' character (i.e. a variable) on the left side of the '=>' symbol, it uses the right hand path (and thus uses evaluation done by ldd), otherwise falls back to the behavior set by [1].

Reproducer that was presented to me:

```
# grep "ibz.so" /etc/ld.so.preload || cat << EOF >> /etc/ld.so.preload
/\$LIB/libz.so.1.2.11
EOF
# mkdir -p /var/tmp/dracut.xitk6p/initramfs
# strace /usr/lib/dracut/dracut-install -D /var/tmp/dracut.xitk6p/initramfs -l /bin/bash 2>&1|grep ibz
# rm -rf /var/tmp/dracut.xitk6p/
```

[0] https://github.com/dracutdevs/dracut/commit/45404a2ad884edac2646f5f86045247f635b5673
[1] https://github.com/dracutdevs/dracut/commit/6d886bb74d1608e4565d926aa259ea5afc9df7b9
[2] https://github.com/dracutdevs/dracut/pull/471
[3] https://bugs.gentoo.org/667752


haraldh:
Looks good.

Please:
* fix the commit title to: `fix(install): handle $LIB in ldd output parsing`
* add your detailed description in the PR to the commit message
* fix your C indention by running `make indent-c` which requires `astyle`

danimo (approved):
LGTM

PR #1462, opened 2021-05-10
fix(dracut-init): add src path for install

johannbg:
We probably end up having to add more if statements like this one for install for everything that we moved into the src directory so dracut src files themselves can be moved there and will work in test cases...

## Checklist

Fixes #


haraldh:
Hmm, I think this is not necessary, because the `Makefile` contains:
```Makefile
dracut-install: src/install/dracut-install
    ln -fs $< $@
```

johannbg:
Normally that should be enough but I was weirdly bitten by this when moving dracut's own sh files into it's own src directory and bunch of other seemingly path issues throughout dracut sh files and our test cases in which the test cases either found our files or they didn't and the test cases would fail with "command not found" all related to the "$basedir"/ entries and how Makefile.testdir is setup and the test cases are constructed. 

Anyways we need to re-think our test cases setup when we move dracut src files into it's own src directory.

PR #1466, opened 2021-05-12
ci: run the integration tests on more distributions

haraldh:
Add OpenSuse and Arch Linux.
Only run extended tests on `Fedora:latest` to reduce the workload.

Also remove the `hung_task_panic=1` on the kernel command line, as qemu
tends to be really slow in non-kvm mode.


.github/workflows/integration.yml
```diff
@@ -0,0 +1,83 @@
...
+        runs-on: ubuntu-latest
+        timeout-minutes: 45
+        concurrency:
+            group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.container }}-${{ matrix.test }}
+            cancel-in-progress: true
+        strategy:
+            matrix:
+                container: [
+                        "fedora:latest",
+                        "opensuse:latest",
```
danimo, line 47: `opensuse:latest` has been deprecated for quite a while now noted in the description. I am not even sure what state this is. We should be using `opensuse/tumbleweed:latest`  and probably add `opensuse/leap:latest`, too.

danimo (changes requested):

.github/workflows/integration.yml
haraldh: it is in fact opensuse/tumbleweed:latest

haraldh: https://github.com/dracutdevs/dracut/blob/5a04bd2dd3a091af333c534f4534c87c5175fe13/test/container/Dockerfile-OpenSuse-latest#L1

danimo: my bad, thanks!

danimo (approved):

johannbg (approved):

PR #1481, opened 2021-05-12
fix(url-lib): fix passing args

mikhailnov:
Fixes: 8e84fa726 ("fix(url-lib): shellcheck for modules.d/45url-lib")

Behaviour introduced by that commit made the following to be run:
`curl "--globoff --location --retry 3 --fail --show-error" http://192.168.1.173:8000/test.ks`
instead of:
`curl --globoff --location --retry 3 --fail --show-error http://192.168.1.173:8000/test.ks`

This broke downloading kickstart file in anaconda-dracut.

### DIFF
--- a/modules.d/45url-lib/url-lib.sh
+++ b/modules.d/45url-lib/url-lib.sh
@@ -66,13 +66,15 @@ curl_fetch_url() {
     local url="$1" outloc="$2"
     echo "$url" > /proc/self/fd/0
     if [ -n "$outloc" ]; then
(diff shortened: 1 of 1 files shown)


haraldh (approved):
you might want to fix the commit message as proposed in the failed test output

haraldh:
Thanks for the fix!!

johannbg (changes requested):
Change your commit message 
from.
`
fix(url-lib): fix passing args
`
to
`
fix(url-lib): passing args
`
Other than that LGTM

PR #1482, opened 2021-05-12
fix: adding crc32c for ext3

johannbg:
Noticed that ext3 is still being used in the wild so let's add it to the crc32c list
so downstream does not have to carry a patch for it.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/99fs-lib/module-setup.sh
+++ b/modules.d/99fs-lib/module-setup.sh
@@ -37,7 +37,7 @@ echo_fs_helper() {
 include_fs_helper_modules() {
     local fs=$2
     case "$fs" in
-        xfs | btrfs | ext4)
+        xfs | btrfs | ext4 | ext3)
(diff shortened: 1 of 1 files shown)


haraldh (approved):

marcosfrm:
Btrfs since 5.5 supports additional hash functions:

```
# modinfo btrfs
filename:       /lib/modules/5.12.1-arch1-1/kernel/fs/btrfs/btrfs.ko.xz
softdep:        pre: blake2b-256
softdep:        pre: sha256
softdep:        pre: xxhash64
softdep:        pre: crc32c
```

But the softdeps are there since 5.5

https://github.com/torvalds/linux/blob/v5.5/fs/btrfs/super.c#L2467

hence no need to specify they here right?

haraldh:
> hence no need to specify they here right?

right

PR #1484, opened 2021-05-13
fix(squash): don't mount the mount points if already mounted

ryncsn:
It is possible that user setup some early mount in the rootfs even
earlier, so just be more robust and cover that case too.

## Checklist

Fixes #

### DIFF
--- a/modules.d/99squash/init-squash.sh
+++ b/modules.d/99squash/init-squash.sh
@@ -1,12 +1,17 @@
 #!/bin/sh
 PATH=/bin:/sbin
 
-# Basic mounts for mounting a squash image
-mkdir /proc /sys /dev /run
-mount -t proc -o nosuid,noexec,nodev proc /proc
-mount -t sysfs -o nosuid,noexec,nodev sysfs /sys
-mount -t devtmpfs -o mode=755,noexec,nosuid,strictatime devtmpfs /dev
-mount -t tmpfs -o mode=755,nodev,nosuid,strictatime tmpfs /run
+[ -e /proc/self/mounts ] \
+    || (mkdir -p /proc && mount -t proc -o nosuid,noexec,nodev proc /proc)
+
+grep -q '^sysfs /sys sysfs' /proc/self/mounts \
+    || (mkdir -p /sys && mount -t sysfs -o nosuid,noexec,nodev sysfs /sys)
+
+grep -q '^devtmpfs /dev devtmpfs' /proc/self/mounts \
+    || (mkdir -p /dev && mount -t devtmpfs -o mode=755,noexec,nosuid,strictatime devtmpfs /dev)
+
+grep -q '^tmpfs /run tmpfs' /proc/self/mounts \
+    || (mkdir -p /run && mount -t tmpfs -o mode=755,noexec,nosuid,strictatime tmpfs /run)
 
 # Load required modules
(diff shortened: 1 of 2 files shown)


modules.d/99squash/init-squash.sh
```diff
@@ -1,12 +1,17 @@
...
-mkdir /proc /sys /dev /run
-mount -t proc -o nosuid,noexec,nodev proc /proc
-mount -t sysfs -o nosuid,noexec,nodev sysfs /sys
-mount -t devtmpfs -o mode=755,noexec,nosuid,strictatime devtmpfs /dev
-mount -t tmpfs -o mode=755,nodev,nosuid,strictatime tmpfs /run
+[ -e /proc/self/mounts ] || \
+( mkdir -p /proc && mount -t proc -o nosuid,noexec,nodev proc /proc )
+
+grep '^sysfs /sys sysfs' &>/dev/null < /proc/self/mounts || \
+( midir -p /sys && mount -t proc -o nosuid,noexec,nodev proc /proc )
```
johannbg, line 8: Spelling error here for the command.

johannbg (changes requested):
You need to fix the spelling of the mkdir command, other than that looks good to me.

modules.d/99squash/init-squash.sh
dtardon, line 8: Copy-paste error: proc -> sys

johannbg:
There are also formatting errors that you need to fix.

modules.d/99squash/init-squash.sh
```diff
@@ -1,12 +1,17 @@
 #!/bin/sh
 PATH=/bin:/sbin
 
-# Basic mounts for mounting a squash image
-mkdir /proc /sys /dev /run
-mount -t proc -o nosuid,noexec,nodev proc /proc
-mount -t sysfs -o nosuid,noexec,nodev sysfs /sys
-mount -t devtmpfs -o mode=755,noexec,nosuid,strictatime devtmpfs /dev
-mount -t tmpfs -o mode=755,nodev,nosuid,strictatime tmpfs /run
+[ -e /proc/self/mounts ] || \
```
dtardon, line 4: Use mountpoint(1) for the tests?

ryncsn: This script is for the dracut squash early init process, available binaries are limited,so have to use basic commands like grep and bash test to check for mount.

ryncsn:
Thank you so much for finding out these issues, I'm surprised I can make so many mistakes in a single commit :), I tested it and seems my workflow doesn't really depend on some of the mount points so it just worked. I'll fix the issues.

johannbg:
@ryncsn we are planning on releasing today so you need to act fast to have this included.

ryncsn:
Sorry for the late response, seems I missed the release, this is for some rare use cases which are still WIP, so should be OK.

johannbg:
Note assuming you are running F34 you can fix your formatting errors by running from the dracut git clone directory
`sudo dnf -y install shfmt`
`shfmt -s -w modules.d/99squash/init-squash.sh`

ryncsn:
Thanks for the info, that's very helpful!

johannbg (approved):
LGTM

danimo (approved):

PR #1495, opened 2021-05-14
fix(base): silence dracut-getarg command not found

johannbg:
Fixes 1494

## Checklist

Fixes #


haraldh:
https://github.com/dracutdevs/dracut/pull/1496 should be a better fix

Issue #1494, opened 2021-05-14
dracut-054: dracut-getarg: command not found


Polynomial-C:
**Describe the bug**
While creating an initrd the following message pops up:
```
/usr/lib/dracut/modules.d/99base/dracut-lib.sh: line 217: dracut-getarg: command not found
```

**Distribution used**
Gentoo Linux

**Dracut version**
054

**Init system**
sysvinit + openrc

**To Reproduce**
`dracut -f /boot/initrd-5.12.4_01 5.12.4`

**Expected behavior**
No error mesage :smile: 

**Additional context**
```
notebook:~ # dracut -f /boot/initrd-5.12.4_01 5.12.4
dracut: Executing: /usr/bin/dracut -f /boot/initrd-5.12.4_01 5.12.4
dracut: *** Including module: bash ***
dracut: *** Including module: dash ***
dracut: *** Including module: caps ***
dracut: *** Including module: i18n ***
dracut: *** Including module: btrfs ***
dracut: *** Including module: crypt ***
dracut: *** Including module: dm ***
dracut: Skipping udev rule: 64-device-mapper.rules
dracut: Skipping udev rule: 60-persistent-storage-dm.rules
dracut: Skipping udev rule: 55-dm.rules
dracut: *** Including module: kernel-modules ***
dracut: *** Including module: resume ***
dracut: *** Including module: rootfs-block ***
dracut: *** Including module: terminfo ***
dracut: *** Including module: udev-rules ***
dracut: Skipping udev rule: 40-redhat.rules
dracut: Skipping udev rule: 50-firmware.rules
dracut: Skipping udev rule: 50-udev.rules
dracut: Skipping udev rule: 91-permissions.rules
dracut: Skipping udev rule: 80-drivers-modprobe.rules
dracut: Skipping udev rule: 70-persistent-net.rules
dracut: *** Including module: base ***
/usr/lib/dracut/modules.d/99base/dracut-lib.sh: line 217: dracut-getarg: command not found
dracut: *** Including module: fs-lib ***
dracut: *** Including module: shutdown ***
dracut: *** Including modules done ***
dracut: *** Installing kernel module dependencies ***
dracut: *** Installing kernel module dependencies done ***
dracut: *** Resolving executable dependencies ***
dracut: *** Resolving executable dependencies done ***
dracut: *** Hardlinking files ***
Mode:           real
Files:          190
Linked:         2 files
Compared:       0 xattrs
Compared:       5 files
Saved:          24 B
[...]

PR #1496, opened 2021-05-17
fix(base): split out `dracut-dev-lib.sh`

haraldh:
To share the `wait_for_dev` function and use it without side effects on
install time, split out the needed functions in an extra library.

Fixes: https://github.com/dracutdevs/dracut/issues/1494


johannbg (approved):
LGTM

PR #1497, opened 2021-05-17
Introduce --no-nm-initrd-generator option

coiby:
With this option, kexec-tools can copy the network configurations files in
/etc/NetworkManager/system-connections/ or /etc/sysconfig/network-scripts/ifcfg
to the kdump initramfs directly. Thus it would be much easier for kexec-tools
to set up network to dump vmcore to a remote fs.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -178,6 +178,9 @@ Creates initial ramdisk images for preloading modules
   --no-hostonly-i18n    Install all keyboard and font files available.
   --hostonly-nics [LIST]
                         Only enable listed NICs in the initramfs.
+  --no-nm-initrd-generator
+                        Don't call nm-initrd-generator to generate network
+                        configuration files
   --persistent-policy [POLICY]
                         Use [POLICY] to address disks and partitions.
                         POLICY can be any directory name found in /dev/disk.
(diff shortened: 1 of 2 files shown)


haraldh:
Wouldn't `--kernel-cmdline 'rd.no_nm_initrd_generator=1'` work also?

haraldh:
Also, I don't like `rd.no_foo=1` anymore and would prefer `rd.foo=0`.

johannbg:
This should be a generic switch not a module specific since you want the same behavior regardless if legacy,networkd,nm,iwd and wicked is in use right.

coiby:
> Wouldn't `--kernel-cmdline 'rd.no_nm_initrd_generator=1'` work also?

> Also, I don't like rd.no_foo=1 anymore and would prefer rd.foo=0.

Thanks! I've opened @1605 to address these two issues since this PR has been closed.

coiby:
> This should be a generic switch not a module specific since you want the same behavior regardless if legacy,networkd,nm,iwd and wicked is in use right.

Thanks for the suggestion! But after doing some homework, I decide to only enable it for NetworkManager,

- Although for nm and wicked, to implement this switch is very easy, much effor is needed for  legacy network which doesn't have a network configruation files generator like nm and wicked. And for systemd-networkd, I failed to configure it to bring up the network connection and to find the switch to turn on/off  network configruation files generation.
-  The name "legacy" implies support for legacy network could be dropped.
-  Even I implement this feature for other network managers, some effort is needed by this application to copy the relevant network configurations files to initrd. So if no application needs it, it could be a waste of effort.

PR #1501, opened 2021-05-18
fix(warpclock): minor cleanups

johannbg:
- Removing shell mode lines
- Remove uncessary file requirements
- Remove hard coded path to binary
- Add missing license header
- Document sections
- refactor install section

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/00warpclock/module-setup.sh
+++ b/modules.d/00warpclock/module-setup.sh
@@ -1,29 +1,39 @@
 #!/bin/bash
-# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
-# ex: ts=8 sw=4 sts=4 et filetype=sh
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
 
-# called by dracut
+# Prerequisite check(s) for module.
 check() {
+
     # hwclock does not exist on S390(x), bail out silently then
     local _arch=${DRACUT_ARCH:-$(uname -m)}
     [ "$_arch" = "s390" -o "$_arch" = "s390x" ] && return 1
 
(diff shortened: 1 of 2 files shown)


modules.d/00warpclock/module-setup.sh
```diff
@@ -1,29 +1,39 @@
...
 
-# called by dracut
+# Prerequisite check(s) for module.
 check() {
+
     # hwclock does not exist on S390(x), bail out silently then
     local _arch=${DRACUT_ARCH:-$(uname -m)}
     [ "$_arch" = "s390" -o "$_arch" = "s390x" ] && return 1
 
-    [ -e /etc/localtime -a -e /etc/adjtime ] || return 1
```
danimo, line 11: Why remove this?

haraldh: What's the point of installing this module, if it is never used?

haraldh: In fact, we should even test for `hostonly` _and_ that `adjtime` is set to `LOCAL`

haraldh (changes requested):

modules.d/00warpclock/module-setup.sh
johannbg: So warpclock.sh never has been not sufficient ?

johannbg: > Why remove this?

Because we include those files anyway ( which will fail if they dont exist ) so we should never need to have hard requirements on files, just the binaries. ( which often have built-ins conf themselves or otherwise gracefully handle such scenarios )

haraldh: Ok, I am fine with removing this check, because with `return 255` the user manually added the module by purpose.

haraldh (approved):

PR #1506, opened 2021-05-19
WIP: feat(lvm-systemd): introducing the lvm-systemd module

johannbg:
Let's get the ball rolling on this one...

Introducing the lvm-systemd module which is basically somewhat sanitized version of those upstream changes here [1]. 

I went with the lvm-systemd naming since we need to split modules into systemd based and none systemd based and those upstream changes bring a hard dependency on systemd.

I made a hard requirement of the existence of the relevant udev script which should be included in the distribution package along with the 64-lvm.rules as in those should normal not be included with the module but handled on a distribution level. 

I cleaned up the entire host section of the shebang flare of what's being done upstream and simply included the relevant hostonly files which should be done anyway since administrators own /etc and we expect them to make changes there as opposed to in other parts of the filesystem.
  
Afaikt that "rework" should not make any functional change.

I left the command line parser script untouched even thou the whole rm foo probably wont work in practice.

1. https://sourceware.org/git/?p=lvm2.git;a=tree;f=dracut;hb=c53bb2b5ea5e5a985e32235515ad21fff62e3b5b

### DIFF
--- a/modules.d/90lvm-systemd/module-setup.sh
+++ b/modules.d/90lvm-systemd/module-setup.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    [[ $hostonly ]] || [[ $mount_needs ]] && {
+        for fs in "${host_fs_types[@]}"; do
+            [[ $fs == LVM*_member ]] && return 0
+        done
+        return 255
+    }
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries \
+        lvm \
+        "$udevdir"/pvscan-udev-initrd.sh \
+        || return 1
+
+    # Return 0 to include the module.
+    return 0
+
+}
+
+# Module dependency requirements.
+depends() {
+
+    # This module has external dependency on other module(s).
+    echo rootfs-block dm systemd-udevd
+    # Return 0 to include the dependent module(s) in the initramfs.
+    return 0
+
+}
+
+# Commandline.
+cmdline() {
+    local _activated
+    declare -A _activated
+
+    for dev in "${!host_fs_types[@]}"; do
+        [ -e /sys/block/"${dev#/dev/}"/dm/name ] || continue
+        [ -e /sys/block/"${dev#/dev/}"/dm/uuid ] || continue
+        uuid=$(< /sys/block/"${dev#/dev/}"/dm/uuid)
+        [[ ${uuid#LVM-} == "$uuid" ]] && continue
+        dev=$(< /sys/block/"${dev#/dev/}"/dm/name)
+        eval "$(dmsetup splitname --nameprefixes --noheadings --rows "$dev" 2> /dev/null)"
+        [[ ${DM_VG_NAME} ]] && [[ ${DM_LV_NAME} ]] || return 1
+        if ! [[ ${_activated[DM_VG_NAME / DM_LV_NAME]} ]]; then
+            printf " rd.lvm.lv=%s " "${DM_VG_NAME}/${DM_LV_NAME} "
+            _activated["${DM_VG_NAME}/${DM_LV_NAME}"]=1
+        fi
+    done
+}
+
+# Install kernel module(s).
+installkernel() {
+    hostonly='' instmods dm-snapshot
+}
+
+# Install the required file(s) and directories for the module in the initramfs.
+install() {
+
+    if [[ $hostonly_cmdline == "yes" ]]; then
+        # shellcheck disable=SC2155
+        local _lvmconf=$(cmdline)
+        [[ $_lvmconf ]] && printf "%s\n" "$_lvmconf" >> "${initdir}/etc/cmdline.d/90lvm.conf"
+    fi
+
+    inst_hook cmdline 30 "$moddir/parse-lvm.sh"
+
+    inst_multiple -o \
+        "$udevdir"/pvscan-udev-initrd.sh \
+        "$udevrulesdir"/11-dm-lvm.rules \
+        "$udevrulesdir"/64-lvm.rules \
+        cache_dump cache_restore cache_check cache_repair \
+        era_check era_dump era_invalidate era_restore lvm \
+        thin_dump thin_restore thin_check thin_repair
+
+    # Install required libraries.
+    _arch=${DRACUT_ARCH:-$(uname -m)}
+    inst_libdir_file {"tls/$_arch/",tls/,"$_arch/",}"libdevmapper-event-lvm*.so"
+
+    # Install the hosts local user configurations if enabled.
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o \
+            /etc/lvm/lvm.conf \
+            /etc/lvm/lvmlocal.conf
+    fi
+
+}
--- a/modules.d/90lvm-systemd/parse-lvm.sh
+++ b/modules.d/90lvm-systemd/parse-lvm.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+# This file is part of dracut lvm-systemd module.
+
+if [ -e /etc/lvm/lvm.conf ] && ! getargbool 1 rd.lvm.conf -d -n rd_NO_LVMCONF; then
+    rm -f -- /etc/lvm/lvm.conf
+fi
+
+LV_DEVS="$(getargs rd.lvm.vg -d rd_LVM_VG=) $(getargs rd.lvm.lv -d rd_LVM_LV=)"
+
+# shellcheck disable=SC2030
+# shellcheck disable=SC2031
+if ! getargbool 1 rd.lvm -d -n rd_NO_LVM \
+    || ([ -z "$LV_DEVS" ] && ! getargbool 0 rd.auto); then
+    info "rd.lvm=0: removing LVM activation"
+    rm -f -- /etc/udev/rules.d/64-lvm*.rules
+else
+    for dev in $LV_DEVS; do
+        wait_for_dev -n "/dev/$dev"
+    done
+fi
--- a/pkgbuild/dracut.spec
+++ b/pkgbuild/dracut.spec
@@ -365,6 +365,7 @@ echo 'dracut_rescue_image="yes"' > $RPM_BUILD_ROOT%{dracutlibdir}/dracut.conf.d/
 %{dracutlibdir}/modules.d/90kernel-modules
 %{dracutlibdir}/modules.d/90kernel-modules-extra
 %{dracutlibdir}/modules.d/90lvm
+%{dracutlibdir}/modules.d/90lvm-systemd
 %{dracutlibdir}/modules.d/90mdraid
 %{dracutlibdir}/modules.d/90multipath
 %{dracutlibdir}/modules.d/90nvdimm


coiby:
Hi,

This PR breaks kdump (I run a test robot to monitor dracut PRs), please don't merge it before I find out why.

johannbg:
> Hi,
> 
> This PR breaks kdump (I run a test robot to monitor dracut PRs), please don't merge it before I find out why.

It's somewhat of an unreasonable request that we halt our development cycle until your home crown test suit gets fixed whenever you get around to it o_O. 

The best way to prevent us from breaking kdump is for kdump dracut modules being hosted, maintained and tested here upstream then any breakage we introduce will be instantly on our radar ( not that I'm seeing how on earth these changes somehow broke kdump  )

coiby:
> > Hi,
> > This PR breaks kdump (I run a test robot to monitor dracut PRs), please don't merge it before I find out why.
> 
> It's somewhat of an unreasonable request that we halt our development cycle until your home crown test suit gets fixed whenever you get around to it o_O.

Sorry, it's never my intention to halt your development cycle. In fact, I prioritize today's work on figuring out why the test failed.  Please ignore this request if it halts your development cycle. 

> 
> The best way to prevent us from breaking kdump is for kdump dracut modules being hosted, maintained and tested here upstream then any breakage we introduce will be instantly on our radar ( not that I'm seeing how on earth these changes somehow broke kdump )

Thanks for the suggestion! As for hosting dracut kdump modules, I'll evaluate this idea and have a discussion with my colleagues.  As for bringing kdump tests to dracut, I'm still working on #1163 by following your suggestion [1]. 

[1] https://github.com/dracutdevs/dracut/pull/1163#issuecomment-820017748

johannbg:
Note that lvm implementation will be in a state of flux in the coming days. 

This module depends on a unreleased lvm code ( so it will fail if installed/enabled unless you do a built of the upstream development branch ) and those upstream changes force us to create a meta module out of the existing lvm module to be able to support the "old" lvm and the "new" lvm depending on which lvm release is installed on the host.

coiby:
> Note that lvm implementation will be in a state of flux in the coming days.
> 
> This module depends on a unreleased lvm code ( so it will fail if installed/enabled unless you do a built of the upstream development branch ) and those upstream changes force us to create a meta module out of the existing lvm module to be able to support the "old" lvm and the "new" lvm depending on which lvm release is installed on the host.

Thanks for the note! Eventually, I did a git bisect and find it's #1450 that breaks kdump.  Sorry for putting the blame on this PR. The test robot needs to be improved because it had been failing to catch #1450 until this PR.

teigland:
Hi, thanks for looking at this.  Could you confirm which files you'd want the lvm package should install and where?  Would lvm install pvscan-udev-initrd.sh and 64-lvm.rules to /lib/dracut/modules.d/90lvm-systemd?  To be clear, 64-lvm.rules is only used in dracut (it's a bit different from 69-dm-lvm.rules.)

johannbg:
@teigland lvm would install the `64-lvm.rules` rule into `/usr/lib/udev/rules.d` directory with correct lexical order to other udev rules lvm ships and the `pvscan-udev-initrd.sh` to `/usr/lib/udev/` directory.  

The module picks those files up via `"$udevrulesdir"/64-lvm.rules` and `"$udevdir"/pvscan-udev-initrd.sh` and installs them under the exact same path in the initrd. 

I should also probably mention that this `rm -f -- /etc/lvm/lvm.conf` and this `rm -f -- /etc/udev/rules.d/64-lvm*.rules` in `parse-lvm.sh` wont work in modern OS design since /etc can be non existent or read only and I would think lvm want's to support such setups.

teigland:
When running in the root fs, /lib/udev/rules.d/69-dm-lvm.rules is used.  64-lvm.rules is used only in initrd, so we'll need some further thinking on the location of that since it would obviously conflict with 69-dm-lvm.

I'll have a look at parse-lvm, I'd not really looked at that part of the puzzle yet.

johannbg:
> When running in the root fs, /lib/udev/rules.d/69-dm-lvm.rules is used. 64-lvm.rules is used only in initrd, so we'll need some further thinking on the location of that since it would obviously conflict with 69-dm-lvm.

Hmm you have both the lexical ordering as well as the `:=` operator to prevent rule overwrite if required and if neither or both aren't sufficient then arguably something is wrong in the overall approach of solving whatever you are trying to solve which begs the question what usecase are you solving with this approach and why are you doing it this way?

We can of course ship the rules if absolutely necessary with the module but usually it raises an red flag of the overall design if that is required since typically these days you should just need a command line generator, followed by systemd type units, which contain the relevant conditions in them ( for the command line it would be `ConditionKernelCommandLine=` to start the type unit or `ConditionKernelCommandLine=!` to prevent the type unit from being started ) , with correct enablement for the type units in the relevant initrd related targets and the module should just be able to include the generator, the type unit files and udev rules ( if applicable ) and those should not conflict with anything used in the real root.

modules.d/90lvm-systemd/module-setup.sh
```diff
@@ -0,0 +1,92 @@
...
+    inst_libdir_file {"tls/$_arch/",tls/,"$_arch/",}"libdevmapper-event-lvm*.so"
+
+    # Install the hosts local user configurations if enabled.
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o \
+            /etc/lvm/lvm.conf \
+            /etc/lvm/lvmlocal.conf
+    fi
+
+    dracut_need_initqueue
```
lnykryn, line 91: Does it really need initqueue?

johannbg: @lnykryn well that's what upstream is doing here [1] @teigland ? 
We kinda need the scope of the change and why ( I could not find any discussion on the mailinglist about this upstream change ) 

1. https://sourceware.org/git/?p=lvm2.git;a=blob;f=dracut/module-setup.sh;h=51d9cd33a21686c666694c4235d049b1af8bb530;hb=c53bb2b5ea5e5a985e32235515ad21fff62e3b5b

teigland: > Does it really need initqueue?

Probably not, I'm largely unaware of what most of this code is doing.  I took the previous code, changed the bits related to lvm that I did understand, and didn't touch most of the rest related to dracut that I didn't understand.  

> We kinda need the scope of the change and why

What do you mean, could you be more specific?

johannbg: > > We kinda need the scope of the change and why
> 
> What do you mean, could you be more specific?

Just an explanation as to what usecase(s) are you solving and why? 
Your upstream commit msg explains what you are doing but not why you have to do it, why the old module no longer suffices etc. at least that part is somewhat unclear to me. 

If we have the bigger picture of the change we can maybe help to improve it ( and maybe just improve the existing module so we dont have to go the three module route as in a meta module + lvm old module and lvm new module ).
 
For example I would rather we had a good generator and we called systemd type units from udev ( which we can put various conditions in the type units themselves, and include the the type unit in the initrd. target  )  as opposed to calling `RUN+="/usr/bin/systemd-run --no-block --property DefaultDependencies=no` foo and try to come up with udev ruleset that does not conflict with each other. `+` we need to rename the 64-lvm.rules to something else so it wont get confused with the existing 64-lvm.rules since those files are not identical.

teigland: > Just an explanation as to what usecase(s) are you solving and why?

Activating the LVs listed by rd.lvm.lv (or VGs listed by rd.lvm.vg which is less common.)

> Your upstream commit msg explains what you are doing but not why you have to do it, why the old module no longer suffices etc. at least that part is somewhat unclear to me.

The old module does suffice, but it is inefficient and prone to failure with many devices.  It attempts to activate LVs after every lvm device appears rather than waiting for all the necessary devices to appear and then activating LVs.   lvm now knows when all the necessary devices are present, and we can use that intelligence to only run activation when it will succeed.

Repeated trying and failing to activate LVs (when there are multiple devices) is inefficient because each of those commands scans all devices.  As the number of devices grows, the number of failing activation commands becomes excessive, disk scanning becomes excessive and slow, and all the unnecessary overhead can lead to a boot timeout (emergency mode).

The new lvm capabilities allow us to scan each lvm device once, and then usually run a single activation command, which will then scan the disks a second time.

> For example I would rather we had a good generator and we called systemd type units from udev ( which we can put various conditions in the type units themselves, and include the type unit in the initrd. target ) as opposed to calling `RUN+="/usr/bin/systemd-run --no-block --property DefaultDependencies=no` foo 

I'm not a systemd expert so I don't know what advantages some other method might have.

teigland:
In addition to the new uevent-based activation, it will probably be good to have a new timer-based systemd service that will attempt to activate degraded raid LVs after say 30 seconds.  If the root LV uses raid and is missing a device, it won't be activated by the uevent-based mechanism which requires LVs to be complete.  But, the timer-based service would run lvchange -ay --activationmode degraded which would activate the LV with a missing device.  It would be best for the lvchange command to specifically activate the LVs named by rd.lvm.lv (or rd.lvm.vg).  What is a good way for a service file to access those rd.lvm values and use them as an lvchange command arg?  I'd appreciate any help or pointers on this.

The current initqueue-based lvm module does something similar, it uses --partial which is an old/incorrect option for doing this.

johannbg:
>  degraded raid LV

You will need to list all the items that tells that raid LV is in degraded state so I know what we can work with.

teigland:
> > degraded raid LV
> 
> You will need to list all the items that tells that raid LV is in degraded state so I know what we can work with.

lvm doesn't give us very good information around this.  When any LV is missing a PV, lvs -o health_status will say "partial", so we could attempt degraded activation on those.  But it's probably better to just attempt the degraded activation on all LVs named by rd.lvm.lv and hope it helps.

johannbg:
> > > degraded raid LV
> > 
> > 
> > You will need to list all the items that tells that raid LV is in degraded state so I know what we can work with.
> 
> lvm doesn't give us very good information around this. When any LV is missing a PV, lvs -o health_status will say "partial", so we could attempt degraded activation on those. But it's probably better to just attempt the degraded activation on all LVs named by rd.lvm.lv and hope it helps.

There must be something on the filesystem that tells us that something has become degraded? 
What does lvs -o health_status do to figure this out?

teigland:
> > lvm doesn't give us very good information around this. When any LV is missing a PV, lvs -o health_status will say "partial", so we could attempt degraded activation on those. But it's probably better to just attempt the degraded activation on all LVs named by rd.lvm.lv and hope it helps.
> 
> There must be something on the filesystem that tells us that something has become degraded?

An lvm command is the only thing that will tell you if an LV is missing devices.  When you say "become degraded" it sounds like you might be talking about an already active LV that has lost a device, which isn't relevant to this case.

> What does lvs -o health_status do to figure this out?

lvm looks at the devices that an LV needs, compares that to the devices that are present on the system, and if any of the necessary devices are missing then health is "partial".  lvm determines which devices are present by reading all the devices on the system.

teigland:
Reviewing issues and questions from the discussion above:

- The location of new 64-lvm.rules and pvscan-udev-initrd.sh files.  I think the dracut rpm should own and install these files in /usr/lib/dracut/modules.d/90lvm-systemd/.  dracut install() would need to check the current version of lvm to decide whether to build the initrd using files from 90lvm or 90lvm-systemd.  Perhaps check the lvm version by checking if pvscan --help shows the --listlvs option?

- I'd suggest renaming 90lvm-systemd/64-lvm.rules to 64-lvm-systemd.rules, to avoid confusion with the existing 90lvm/64-lvm.rules.

- parse-lvm: rm /etc/lvm/lvm.conf (for rd.lvm.conf=0).  Setting the env var LVM_SYSTEM_DIR="" is equivalent to removing that file.  I'm not sure if we can simply set LVM_SYSTEM_DIR in parse-lvm or if there are hoops we need to jump through so it applies to the lvm commands being run.

- parse-lvm: rm /etc/udev/rules.d/64-lvm*.rules (for rd.lvm=0).  I've looked at what other modules are doing, and a number still remove the rules, but I may not be finding the best examples to follow.  90mdraid parse does "udevproperty rd_NO_MD=1" and the udev rule quits if ENV{rd_NO_MD}.  We could do the same for lvm.

- degraded activation after timeout: perhaps the timer-based systemd unit could run a script, where the script takes rd.lvm.lv values from /proc/cmdline and then runs lvchange with them.

teigland:
I'm not familiar with using github, so let me know if there's something different I should be doing to move this along.  I've added a commit to this one, adding the two files mentioned and making a small change to install().

https://github.com/teigland/dracut/tree/lvmsd-dct1

Of the issues listed above, I'd like to know how dracut wants to decide to install 90lvm-systemd vs 90lvm.  Is there a method for this I should copy from another module?  Thanks

johannbg:
@teigland with regards to github I need to make changes to my PR ( as in incorporate your changes ) or you submit a new one The main show stopper is lack of my personal time. Basically I think there is a better way to deal with this than how you have decided to approach this but that requires me to do some research, get up to speed on lvm and set up a test system to try the solution out. For the old vs new module we need to introduce a meta module to deal with that.

PR #1508, opened 2021-05-21
feat(dracut): add --preferred-init option

hbathini:
This pull request introduces 'preferred-init' option to replace the default
init program with a customized one, while moving the default /init to
/init.dracut

    Usage: dracut --preferred-init /path-to/customized-init.sh

The init program may need customizations when multiple boot scenarios
are possible on a system. Along with this option, --include could be
used to include squash'ed image(s) built for the other possible boot
scenario(s). And the customized init program can be leveraged to exec
into the default init sequence (exec /init.dracut) in case of regular
boot scenario and to activate the squash'ed image in case of the other
possible boot scenario(s).

An example usecase is firmware-assisted dump (fadump) on powerpc. If a
powerpc system configured with fadump crashes, f/w boots the system
back with memory preserved, along with an additional device-tree
attribute to indicate an active system crash. So, the boot after a
crash is expected to offload this active crash dump to the disk while
the regular boot is expected to bring up the production environment.
For such cases, this option can be leveraged to replace default init
with one that activates the squash'ed image (built to offload active
crash dump) for a boot after crash while exec'ing into the default
boot sequence (to bring up the production environment) for regular
boot scenario.

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -195,6 +195,9 @@ Creates initial ramdisk images for preloading modules
                          Target directory in the final initramfs.
                         If SOURCE is a file, it will be installed to TARGET
                          in the final initramfs.
+  --preferred-init [FILE]
+                        Install customized script [FILE] as /init program while
+                         moving the default init program to /init.dracut.
   -I, --install [LIST]  Install the space separated list of files into the
                          initramfs.
   --install-optional [LIST]  Install the space separated list of files into the
@@ -352,6 +355,7 @@ rearrange_params() {
             --long omit: \
             --long drivers: \
             --long filesystems: \
+            --long preferred-init: \
             --long install: \
             --long install-optional: \
             --long fwdir: \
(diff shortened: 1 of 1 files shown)


haraldh:
The can be with a simple dracut module after `99base`, say `99zzMyInit` with `module-setup.sh` :

```bash
install() {
    mv -f "$initdir/init" "$initdir/init.dracut"
    inst_simple "$moddir/myinit" /init
    chmod 0755 "$initdir/init"
}    
```

johannbg (changes requested):
Please update the relevant man pages and shell-completion when adding a new switch.

hbathini:
> The can be done with a simple dracut module after `99base`, say `99zzMyInit` with `module-setup.sh` :
> 
> ```shell
> install() {
>     mv -f "$initdir/init" "$initdir/init.dracut"
>     inst_simple "$moddir/myinit" /init
>     chmod 0755 "$initdir/init"
> }    
> ```

Thanks, Harald. An out-of-tree module as suggested seems like a better choice for my use case..
Will try that out..

PR #1509, opened 2021-05-21
fix: read whitespaces

johannbg:
Fixes rhbz #1962975

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/45url-lib/module-setup.sh
+++ b/modules.d/45url-lib/module-setup.sh
@@ -32,9 +32,9 @@ install() {
         for _lib in "$dracutsysrootdir$_dir"/libcurl.so.*; do
             [[ -e $_lib ]] || continue
             if ! [[ $_nssckbi ]]; then
-                read -r -d '' _nssckbi < <(grep -F --binary-files=text -z libnssckbi "$_lib")
+                read -r -d '' _nssckbi <<< "(grep -F --binary-files=text -z libnssckbi $_lib)"
             fi
-            read -r -d '' _crt < <(grep -F --binary-files=text -z .crt "$_lib")
+            read -r -d '' _crt <<< "(grep -F --binary-files=text -z .crt $_lib)"
             [[ $_crt ]] || continue
             [[ $_crt == /*/* ]] || continue
             if ! inst "${_crt#$dracutsysrootdir}"; then
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -165,7 +165,7 @@ install() {
         done
     }
 
-    mapfile -t _mods < <(modules_load_get /usr/lib/modules-load.d)
+    mapfile -t _mods <<< "$(modules_load_get /usr/lib/modules-load.d)"
     [[ ${#_mods[@]} -gt 0 ]] && hostonly='' instmods "${_mods[@]}"
 
     if [[ $hostonly ]]; then
--- a/modules.d/90kernel-modules-extra/module-setup.sh
+++ b/modules.d/90kernel-modules-extra/module-setup.sh
@@ -74,7 +74,7 @@ installkernel() {
     done
 
     if ((${#cfgs[@]} > 0)); then
-        mapfile -t cfgs < <(printf '%s\n' "${cfgs[@]}" | LANG=C sort -u -k1,1 -t '/' | cut -f 2- -d '/')
+        mapfile -t cfgs <<< "$(printf '%s\n' "${cfgs[@]}" | LANG=C sort -u -k1,1 -t '/' | cut -f 2- -d '/')"
     fi
 
     ## Parse configurations


haraldh (changes requested):
Your patch does not work at all..

```bash
$ mapfile -t tt < <(echo foo bar); declare -p tt
declare -a tt=([0]="foo bar")

$ mapfile -t tt <<< "(echo foo bar)"; declare -p tt
declare -a tt=([0]="(echo foo bar)")
```

haraldh:
See also https://bugzilla.redhat.com/show_bug.cgi?id=1962975#c2
The chroot environment setup for dracut to run in is just wrong.

haraldh:
Basically I refuse to make workarounds just for incomplete chroot environments.

haraldh:
On a side note you forgot the `$`:
```bash
$ mapfile -t tt <<< "$(echo foo; echo bar)"; declare -p tt
declare -a tt=([0]="foo" [1]="bar")
```

johannbg:
> On a side note you forgot the `$`:
> 
> ```shell
> $ mapfile -t tt <<< "$(echo foo; echo bar)"; declare -p tt
> declare -a tt=([0]="foo" [1]="bar")
> ```

> Basically I refuse to make workarounds just for incomplete chroot environments.

The core of this issue is that the certs aren't being pulled which triggers failure in composes and my patch most certainly fixes that.
lsinitrd initramfs-5.12.0-198.fc35.x86_64.img | grep pki
```
drwxr-xr-x   3 root     root            0 May 18 08:36 etc/pki
drwxr-xr-x   3 root     root            0 May 18 08:36 etc/pki/ca-trust
drwxr-xr-x   5 root     root            0 May 18 08:36 etc/pki/ca-trust/source
drwxr-xr-x   2 root     root            0 Jan 26 02:08 etc/pki/ca-trust/source/anchors
drwxr-xr-x   2 root     root            0 Jan 26 02:08 etc/pki/ca-trust/source/blacklist
drwxr-xr-x   2 root     root            0 Jan 26 02:08 etc/pki/ca-trust/source/blocklist
-rw-r--r--   1 root     root            0 Jan 26 02:04 etc/pki/ca-trust/source/ca-bundle.legacy.crt
-rw-r--r--   1 root     root          932 Jan 26 02:04 etc/pki/ca-trust/source/README
-rwxr-xr-x   1 root     root        32768 May 12 07:42 usr/bin/pkill
drwxr-xr-x   3 root     root            0 May 18 08:36 usr/share/pki
drwxr-xr-x   5 root     root            0 May 18 08:36 usr/share/pki/ca-trust-source
drwxr-xr-x   2 root     root            0 Jan 26 02:08 usr/share/pki/ca-trust-source/anchors
drwxr-xr-x   2 root     root            0 Jan 26 02:08 usr/share/pki/ca-trust-source/blacklist
drwxr-xr-x   2 root     root            0 Jan 26 02:08 usr/share/pki/ca-trust-source/blocklist
-rw-r--r--   1 root     root       943903 Jan 26 02:04 usr/share/pki/ca-trust-source/ca-bundle.trust.p11-kit
-rw-r--r--   1 root     root          937 Jan 26 02:04 usr/share/pki/ca-trust-source/README
```
The rest of the seemingly whites paces error in systemd,kernel-module-extra, is something Adam piled on the cert issue.

haraldh:
Anyway, the real fix would be to fix the chroot setup, where the symlink `/dev/fd -> /proc/self/fd` is missing.

modules.d/45url-lib/module-setup.sh
```diff
@@ -32,9 +32,9 @@ install() {
         for _lib in "$dracutsysrootdir$_dir"/libcurl.so.*; do
             [[ -e $_lib ]] || continue
             if ! [[ $_nssckbi ]]; then
-                read -r -d '' _nssckbi < <(grep -F --binary-files=text -z libnssckbi "$_lib")
+                read -r -d '' _nssckbi <<< "(grep -F --binary-files=text -z libnssckbi $_lib)"
             fi
-            read -r -d '' _crt < <(grep -F --binary-files=text -z .crt "$_lib")
+            read -r -d '' _crt <<< "(grep -F --binary-files=text -z .crt $_lib)"
```
haraldh, line 37: missing `$`

modules.d/45url-lib/module-setup.sh
```diff
@@ -32,9 +32,9 @@ install() {
         for _lib in "$dracutsysrootdir$_dir"/libcurl.so.*; do
             [[ -e $_lib ]] || continue
             if ! [[ $_nssckbi ]]; then
-                read -r -d '' _nssckbi < <(grep -F --binary-files=text -z libnssckbi "$_lib")
+                read -r -d '' _nssckbi <<< "(grep -F --binary-files=text -z libnssckbi $_lib)"
```
haraldh, line 35: missing `$`

haraldh (changes requested):

modules.d/45url-lib/module-setup.sh
haraldh, line 35:
```suggestion
                read -r -d '' _nssckbi <<< "$(grep -F --binary-files=text -z libnssckbi "$_lib")"
```

modules.d/45url-lib/module-setup.sh
haraldh, line 37:
```suggestion
            read -r -d '' _crt <<< "$(grep -F --binary-files=text -z .crt "$_lib")"
```

haraldh (changes requested):

modules.d/45url-lib/module-setup.sh
johannbg: Are you sure? I mean this works just fine without it and adding it triggers warning: command substitution: ignored null byte in input due to grep -z and at that point we just down a rabbit hole that my brain cant cope with at the moment :)

johannbg: Adding your suggested change triggers...
```
/45url-lib/module-setup.sh: line 37: warning: command substitution: ignored null byte in input
/usr/lib/dracut/modules.d/45url-lib/module-setup.sh: line 37: warning: command substitution: ignored null byte in input
/usr/lib/dracut/modules.d/45url-lib/module-setup.sh: line 37: warning: command substitution: ignored null byte in input
/usr/lib/dracut/modules.d/45url-lib/module-setup.sh: line 37: warning: command substitution: ignored null byte in input
```

Anyways it's probably best we either merge this as is since it's working and you then do a follow if necessary or we nack this since my brain is not fully working...

haraldh: Yes, without `$` the grep command is never executed:
```bash
$ read -r -d '' _crt <<< "(grep -F --binary-files=text -z .crt $_lib)"; declare -p _crt
declare -- _crt="(grep -F --binary-files=text -z .crt )"
```

And if the `grep` command is executed, the string will contain null bytes. This is why we don't convert it to a string in the first place and use `read` with `< <()`

haraldh: NACK NACK NACK, your patch is incorrect.

haraldh: The `ignored null byte in input` is expected and could be worked around with `tr`... which is also done in our code base before... but this all papers over the bad chroot setup.

johannbg: papering over it with tr did not work I tried...

PR #1511, opened 2021-05-21
fix(dracut-functions): get_maj_min without get_maj_min_cache_file set

haraldh:
This pull request changes...

## Changes

If `get_maj_min_cache_file` is unset `get_maj_min()` would error out.

Fix it to work without a cache file set.

## Checklist

Fixes #

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -234,13 +234,20 @@ get_fs_env() {
 # 8:2
 get_maj_min() {
     local _majmin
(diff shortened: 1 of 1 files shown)


haraldh:
CC @coiby

johannbg (approved):

coiby:
> CC @coiby

Thanks! I can confirm with this PR, kdump no longer breaks because of #1450.

PR #1517, opened 2021-05-24
fix(install): configure logging earlier

marcosfrm:
There are log_error() and log_debug() calls in parse_argv().

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -1164,6 +1164,10 @@ static int parse_argv(int argc, char *argv[])
                 }
         }
 
+        if (arg_loglevel >= 0) {
+                log_set_max_level(arg_loglevel);
+        }
+
         if (!kerneldir) {
                 struct utsname buf;
                 uname(&buf);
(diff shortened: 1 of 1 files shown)


src/install/dracut-install.c
```diff
@@ -1164,6 +1164,10 @@ static int parse_argv(int argc, char *argv[])
                 }
         }
 
+        if (arg_loglevel >= 0) {
+                log_set_max_level(arg_loglevel);
```
haraldh, line 1168: `arg_loglevel == -1` at this point of execution.

haraldh (changes requested):

haraldh:
> There are log_error() and log_debug() calls in parse_argv().

yes, with:
```
static LogTarget log_target = LOG_TARGET_CONSOLE;
static int log_max_level = LOG_WARNING;
static int log_facility = LOG_DAEMON;
```

src/install/dracut-install.c
marcosfrm: Not if `-v --verbose` or `--debug` are specified.

haraldh: Ah, you moved it to `parse_argv()` .... sorry, didn't parse the patch correctly in the first place.

haraldh (approved):

johannbg (approved):

PR #1519, opened 2021-05-25
feat(iscsi): Skip iscsi parsing on systems without iSCSI boot

anjalidk:
This fix gives a boot improvement of about 4-5secs on non-iSCSI boot systems

For a system which did not boot via iSCSI, there is no ipxe and hence no
ipxe added iBFT ACPI table - /sys/firmware/acpi/tables/iBFT.

So we check for the precence of this table to determine if we can skip
iscsi parsing and then subsequently skip running iscsiroot.sh script.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/95iscsi/parse-iscsiroot.sh
+++ b/modules.d/95iscsi/parse-iscsiroot.sh
@@ -11,6 +11,19 @@
 # root= takes precedence over netroot= if root=iscsi[...]
 #
 
+# For a PV VM, there is no ipxe and hence no ipxe added iBFT ACPI table.
+# So we check for the precence of this table to determine if we are
+# a PV or iSCSI VM/BM.
+
+acpi_ibft="/sys/firmware/acpi/tables/iBFT"
+
+if [ ! -f $acpi_ibft ]; then
+    echo "No $acpi_ibft, assuming PV boot"
+    return 1
+else
+    echo "Found $acpi_ibft, assuming iSCSI boot"
+fi
+
 # This script is sourced, so root should be set. But let's be paranoid
 [ -z "$root" ] && root=$(getarg root=)
 if [ -z "$netroot" ]; then


haraldh:
Machines without an iBFT ACPI table can very well boot from an iSCSI volume, so the `return 1` at this point in the code does not only prevent the code path of `rd.iscsi.ibft=1`, but all the other iSCSI boot options.

haraldh (changes requested):
Machines without an iBFT ACPI table can very well boot from an iSCSI volume, so the return 1 at this point in the code does not only prevent the code path of rd.iscsi.ibft=1, but all the other iSCSI boot options.

konradwilk:
> Machines without an iBFT ACPI table can very well boot from an iSCSI volume, so the return 1 at this point in the code does not only prevent the code path of rd.iscsi.ibft=1, but all the other iSCSI boot options.

Hey Harold, thank you for the quick review!

Would you be OK expanding on this? That is if you use Intel NICs with built in iSCSI initiator it creates an ACPI iBFT table. Ditto for Broadcom. If you use EDK2 built-in iSCSI initiator it creates an iBFT table. If you use iPXE it creates ACPI iBFT table too. 

What is the scenario in which the iBFT table would be *not* created and still boot from iSCSI?

anjalidk:
Hi Harold,
I am not sure I got your concern right, but I wanted to mention that even with iSCSI parameters in boot like: netroot=iscsi:, we can still boot the VM (via iSCSI) with this fix. So iBFT table is always created on iSCSI boot, irrespective of whether ibft is actually used as the initiator, that’s our observation.
Anjali

On May 26, 2021, at 12:07 AM, Harald Hoyer ***@***.******@***.***>> wrote:



@haraldh requested changes on this pull request.

Machines without an iBFT ACPI table can very well boot from an iSCSI volume, so the return 1 at this point in the code does not only prevent the code path of rd.iscsi.ibft=1, but all the other iSCSI boot options.

PR #1525, opened 2021-06-01
fix(multipath): get config. dir from configuration

dtardon:
Include additional multipath configuration files even if they are placed in a non-default location.

## Checklist

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -63,6 +63,7 @@ installkernel() {
 # called by dracut
 install() {
     local -A _allow
+    local config_dir
 
     add_hostonly_mpath_conf() {
         if is_mpath "$1"; then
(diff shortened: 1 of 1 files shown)


johannbg:
Hmm why should we be supporting configuration files being placed in a non-default location?
And why are people/projects placing those configuration files in a non-default location to begin with? 
What usecases are they solving by doing that?

johannbg (approved):

dtardon:
> Hmm why should we be supporting configuration files being placed in a non-default location?

Because multipathd allows it and there apparently are people who do it. A separate question is why multipathd allows it, but that's not for me to answer...

> And why are people/projects placing those configuration files in a non-default location to begin with?
> What usecases are they solving by doing that?

Beats me. I cannot think of any sane reason either. All I know is it's being done.

haraldh (approved):

PR #1526, opened 2021-06-01
tpm2-tss module: fix typo in depends()

mberndt123:
There is no systemd-udev module to depend on, because the module is called systemd-udevd.

### DIFF
--- a/modules.d/91tpm2-tss/module-setup.sh
+++ b/modules.d/91tpm2-tss/module-setup.sh
@@ -17,7 +17,7 @@ check() {
 depends() {
 
     # This module has external dependency on other module(s).
-    echo systemd-sysusers systemd-udev
+    echo systemd-sysusers systemd-udevd
     # Return 0 to include the dependent module(s) in the initramfs.
     return 0
 


johannbg (changes requested):
Change it 
from 
`tpm2-tss module: fix typo in depends()`
to 
`fix(tpm2-tss): typo in depends()`

mberndt123:
yeah, whatever.

johannbg:
> yeah, whatever.

Does this respond mean you plan on fixing it or does it mean you plan not to fix it?

mberndt123:
> Does this respond mean you plan on fixing it or does it mean you plan not to fix it?

I means I've already fixed it.

johannbg (approved):

haraldh (approved):

SoumyadipYT-OSS:
In my Garuda Linux:
darcut[E]: Module 'systemd-pcrphase' depends on 'tpm2-tss' which can't be installed. During installation my nvidia version on arch linux distro(Garuda Linux Xfce). Can anybody solve this problem.

TheOverpassArsonist:
I'm also on Garuda and having the same issue as Soumyadlip but I'm not sure if it's related to this issue/merge at all since this appears to have been resolved years ago and has a different (though quite similar) error.

PR #1529, opened 2021-06-08
Remove quotes from cryptsetupopts in 90crypt module

### LINKED ISSUE #1528: hang at boot due to cryptsetup error when using module 90crypt with rd.luks.key (regression)
**Describe the bug**
cryptsetup error when using module 90crypt with rd.luks.key due to incorrectly quoted command, leading to hang at boot. It displays a password prompt, but does not accept the correct password for the key file.

The bug is triggered if `$cryptsetupopts` in `dracut/modules.d/90crypt/cryptroot-ask.sh` contains multiple options separated by spaces, which happens for example if allow discards is specified. The two offending lines are:

```
| cryptsetup -d - "$cryptsetupopts" luksOpen "$device" "$luksname"; then
```
and
```
| cryptsetup -d - "$cryptsetupopts" luksOpen "$device" "$luksname" \
```
In both `"$cryptsetupopts"` should be changed to `$cryptsetupopts` (without quotation marks).

**Distribution used**
Fedora 34

**Dracut version**
dracut 054-6.git20210518.fc34

**Init system**
systemd omitted from dracutmodules

**To Reproduce**
See downstream bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1963424 (only original bug report and not comments, as they seem to be off-topic).

**Expected behavior**
The root partition should be unlocked automatically.

**Additional context**
The bug was first introduced in version 054 commit 682b297207330d6abe18dcc61f55bacf7a8724c1.

lapseofreason:
Fixes a regression that prevents unlocking of a luks volume with a key file using the 90crypt module.

The reason for the regression is the incorrectly quoted variable $cryptsetupoptions that can contain multiple arguments separated by spaces.

## Changes
* Remove quotes from $cryptsetupopts in 90crypt module.

Fixes #1528

### DIFF
--- a/modules.d/90crypt/cryptroot-ask.sh
+++ b/modules.d/90crypt/cryptroot-ask.sh
@@ -138,8 +138,9 @@ unset allowdiscards
 ask_passphrase=1
 
 if [ -n "$luksfile" -a "$luksfile" != "none" -a -e "$luksfile" ]; then
+    # shellcheck disable=SC2086
     if readkey "$luksfile" / "$device" \
-        | cryptsetup -d - "$cryptsetupopts" luksOpen "$device" "$luksname"; then
+        | cryptsetup -d - $cryptsetupopts luksOpen "$device" "$luksname"; then
         ask_passphrase=0
     fi
 elif [ "$is_keysource" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


haraldh (approved):

johannbg (changes requested):
Your pr is failing both the commisery test and the shell linter test

To fix the commisery check change your commit msg
from
`Remove quotes from cryptsetupopts in 90crypt module`
to
`fix(crypt): remove quotes from cryptsetupopts`

You will need to disable the failing linter shell check (  Double quote to prevent globbing and word splitting) which get triggered when you remove the quotes for your change. 

You can do so by adding the following line to the if/readkey statements

`# shellcheck disable=SC2086`

lapseofreason:
@johannbg Thanks for the pointers, I made the changes.

johannbg (approved):

PR #1531, opened 2021-06-08
feat(cpio): improve initramfs image performance and efficiency via cpio reflinks

ddiss:
## Changes

This patchset attempts to speed up initramfs generation for some common (Btrfs / XFS) setups by having Dracut make heavier use of reflinks (AKA copy-on-write clones) during initramfs generation. A good portion of an uncompressed+unstripped initramfs image is duplicate data, which really shouldn't need to be shuffled around when on the same COW clone capable FS.

This is a rework of my https://github.com/dracutdevs/dracut/pull/1148 feature submission. Instead for relying on the [GNU cpio patchset](https://lists.gnu.org/archive/html/bug-cpio/2021-04/msg00004.html) for copy-on-write FS optimized reflink I/O, a new `dracut-cpio` tool is provided for cpio archive creation.
My motivating factors for dropping GNU cpio in favour of `dracut-cpio` are:
- GNU cpio appears to receive very little attention upstream
- relative simplicity of implementation
  - `dracut-cpio` only needs to provide support for cpio `newc` archive creation for extraction by the kernel
  - rust's standard library supports `copy_file_range()` natively

The new `dracut-cpio` functionality is disabled by default. It can be explicitly enabled by building with the `--enable-dracut-cpio` configure option and then calling `dracut` with `--enhanced-cpio`.

### Performance

Preliminary benchmarks indicate a significant improvement in initramfs creation and kernel extraction times with reflinks, as shown via the *Dracut runtime* and *QEMU boot   time* values respectively.
Storage utilization is also improved with initramfs reflinks, as shown by the significant reduction in *exclusive* extents. *Shared* extents represent space reclaimed        through deduplication of initramfs source file and cpio data segments. 

|   **Btrfs**                      |   Dracut `xz`  |    Dracut `zstd`   |    Dracut reflink   |
|:--------------------------------:|:--------------:|:------------------:|:-------------------:|
| Dracut runtime                   |  16.809s (ref) |  12.935s (-23.05%) |   10.292s (-38.59%) |
| QEMU boot time: kernel to /init  |   2361ms (ref) |  951.0ms (-59.71%) |   857.8ms (-63.66%) |
| **initramfs `fiemap` data**      |                |                    |                     |
|                      total bytes | 14647296 (ref) | 16191488 (+10.54%) |  34447360 (+135.2%) |
|            shared (deduplicated) |        0 (ref) |        0 (+-0.00%) | 31309824 (**note**) |
|                        exclusive | 14647296 (ref) | 16191488 (+10.54%) |  3137536 (**note**) |

**note**: Btrfs `fiemap` results varied significantly across multiple runs, therefore `shared` and `exclusive` values above should only be seen as a guide. In contrast, XFS  results were consistent across all runs: 

|   **XFS**                        |    Dracut `xz` |    Dracut `zstd`   |   Dracut reflink   |
|:--------------------------------:|:--------------:|:------------------:|:------------------:|
| Dracut runtime                   |  14.180s (ref) |  10.437s (-26.40%) |  9.7687s (-31.11%) |
| QEMU boot time: kernel to /init  |   2373ms (ref) |  950.9ms (-59.92%) |  859.5ms (-63.78%) |
| **initramfs `fiemap` data**      |                |                    |                    |
|                      total bytes | 14675968 (ref) | 16158720 (+10.10%) | 34689024 (+136.4%) |
|            shared (deduplicated) |        0 (ref) |        0 (+-0.00%) |           26525696 |
|                        exclusive | 14675968 (ref) | 16158720 (+10.10%) |  8163328 (-44.37%) |

The *Dracut `xz`* case corresponds to SUSE Dracut version 055, patched with dracut-cpio reflink support, but configured to run with current SUSE defaults:
GNU cpio archiving, `xz -0 --check=crc32 --memlimit-compress=50%` compression and `strip` symbol discard.
The *Dracut `zstd`* case matches *Dracut `xz`* except for compression which is configured to use `zstd -3 -T0`.
The *Dracut reflink* case has dracut-cpio alignment and reflinks enabled via `enhanced_cpio=yes`. To ensure successful extent sharing, initramfs compression and symbol       discard are disabled, alongside use of a reflink friendly Dracut staging area (`tmpdir=/boot`).

These benchmarks were performed on Tumbleweed 20210924 (5.14.6 kernel) virtual machines assigned 2 vCPUs and 8GiB RAM. The QEMU/KVM hypervisor host was running the same OS   and stored the raw VM disk images in memory (`tmpfs`).
Importantly, bootloader initramfs image read times were not evaluated due to a lack of time and existing instrumentation.

Memory backed storage is significantly less vulnerable to the fragmentation effects of CoW reflinks compared to HDDs. Further benchmarking on SSDs and HDDs is necessary, at least before considering `dracut-cpio` as a complete replacement for GNU cpio archive creation. The benchmarks were done on a completely different machine and kernel to https://github.com/dracutdevs/dracut/pull/1148 , so shouldn't be used for comparison with those numbers.

Special thanks to @firstyear for helping me get my rust changes in shape for upstream submission.

## Checklist


rapid-cpio/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "dracut-cpio"
+description = "cpio archive generator for Dracut"
+authors = ["David Disseldorp"]
+license = "GPL-2.0"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+crosvm = { path = "third_party/crosvm" }
```
Firstyear, line 10: An alternate approach is to use `crosvm = "0.1"`, then run "cargo vendor" and add the vendor subdir. Additionally you add .cargo/config with the content that cargo vendor displays and it will automatically find the location during builds. This way you can cargo vendor in the future, and it will update the crosvm sources without needing more manual interactions.

ddiss: Thanks, I wasn't aware of cargo vendor. I like that it's currently easy to see from Cargo.toml that the dependency is carried in the tree, but I suppose it makes sense to use proper vendoring. I'll make this change alongside the `lint-shell` fixes.

configure
```diff
@@ -110,6 +112,7 @@ sysconfdir ?= ${sysconfdir:-${prefix}/etc}
 sbindir ?= ${sbindir:-${prefix}/sbin}
 mandir ?= ${mandir:-${prefix}/share/man}
 enable_documentation ?= ${enable_documentation:-yes}
+enable_dracut_cpio ?= ${enable_dracut_cpio}
```
tpgxyz, line 115: I'd really appreaciate to add tests for rust and its friends on early stage, rather that bail out later.

man/dracut.8.asc
```diff
@@ -530,6 +530,15 @@ will not be able to boot.
     Specifies the kernel image, which to include in the UEFI executable. The default is
     _/lib/modules/<KERNEL-VERSION>/vmlinuz_ or _/boot/vmlinuz-<KERNEL-VERSION>_
 
+**--cpio-reflink**::
+    Attempt to use the dracut-cpio binary, which optimizes archive creation for
+    copy-on-write filesystems by using the copy_file_range(2) syscall via Rust's
```
tpgxyz, line 535: I'd suggest to mention that copy_file_range(2) is available with kernel-5.3 and glibc-2.27

dracut.sh
```diff
@@ -1151,6 +1154,17 @@ trap 'exit 1;' SIGINT
 readonly initdir="${DRACUT_TMPDIR}/initramfs"
 mkdir -p "$initdir"
 
+if [[ $cpio_reflink == "yes" ]]; then
+    dracut_cpio="$dracutbasedir/dracut-cpio"
```
tpgxyz, line 1158: Please add checks if kernel version is sufficent and libc does support copy_file_range(2).
Not everybody are using glibc :)

configure
ddiss: Fair enough. I'll add a check for cargo presence in the `--enable-dracut-cpio` configure path.

man/dracut.8.asc
ddiss: The `copy_file_range()` syscall first appeared with linux 4.5 and the rust standard library invokes it via libc::syscall, so glibc version needn't be mentioned IMO. More important is the whether or not the rust standard library carries https://github.com/rust-lang/rust/pull/75272 , i.e. rust 1.50+.

I think including the term "reflink" in the feature could be a little misleading, as it's very filesystem specific. I think it might make more sense to use `--rapid-cpio`, `--accelerated-cpio` or similar.

johannbg:
@ddiss note that @haraldh is on vacation until end of June so expect late review from him

johannbg:
@ddiss hows the progress going on @Firstyear  and @tpgxyz pointers ? @haraldh  ping

ddiss:
> @ddiss hows the progress going on @Firstyear and @tpgxyz pointers ? @haraldh ping

I'm back from leave so will return to this in the coming days. Thanks for the ping.

the8472:
> rust's standard library supports copy_file_range() natively

Please keep in mind that in `std` it is considered an optimization and not part of our API guarantees. We had to disable copy_file_range before due to regressions. So if you want to rely on that it you may want to write a test to have a canary. E.g. by tracing syscalls for a particular test or (assuming a suitable filesystem) checking that extents really are shared.

Other options would be using a crate that makes stronger guarantees or documenting that it's best-effort.

ddiss:
> > rust's standard library supports copy_file_range() natively
> 
> Please keep in mind that in `std` it is considered an optimization and not part of our API guarantees. We had to disable copy_file_range before due to regressions. So if you want to rely on that it you may want to write a test to have a canary. E.g. by tracing syscalls for a particular test or (assuming a suitable filesystem) checking that extents really are shared.
> 
> Other options would be using a crate that makes stronger guarantees or documenting that it's best-effort.

Ack, understood. Best-effort is fine, as that's what we already get from the kernel with regard to performing COW reflink or splice fallback. Thanks for the clarification.

dracut.sh
ddiss: The new functionality is disabled by default at build and runtime, so my preference would be to document the requirements and leave version checks, etc. up to the user.

ddiss:
Changes since last version:
- perform configure time check for Cargo presence
- fix lint-shell warnings

rapid-cpio/Cargo.toml
ddiss: I had a play with cargo vendor, but wasn't having much luck using it with `cargo --offline`, so kept things as is for now.

man/dracut.8.asc
ddiss: I've toyed with adding the following as a `configure` time check:
```
    rustc -V | awk -F '[ .]' \
        '$1 !~ /rustc/ {
             exit 1
         }
         $2 * 1000 + $3 <= 1050 {
             printf("Warning: %s < 1.50 does not support the copy_file_range() syscall\n", $1)
         }'
    uname -r | awk -F '.' \
        '$1 * 1000 + $2 < 4050 {
             print "Warning: Linux kernel < 4.5 does not support copy_file_range()"
         }'
```

I think it's a bit cumbersome, so will go with documentation for now.

ddiss:
Changes since last version:
- rename `--cpio-reflink` parameter to `--enhanced-cpio` to reflect the fact that it won't necessarily result in reflinked data
- support compression alongside `dracut-cpio`
- move `dinfo` call after `dracut-init.sh` is loaded (thanks @mwilck)
- fix cmdline vs config parameter precedence (thanks @mwilck)
- minor debug output and variable rewording

man/dracut.8.asc
ddiss: I've renamed the parameter `--enhanced-cpio`, to reflect that it won't necessarily result in reflinked cpio file data. The man page documents requirements for successful reflinks via `copy_file_range()`.

ddiss:
I've tacked on one extra commit which adds test coverage for the `dracut.sh --enhanced-cpio` code-path, as well as kernel cpio archive extraction.

The `TEST-62-CPIO` test should probably now be renamed `TEST-62-SKIPCPIO`... I'll squash that in with the next rebase.

ddiss:
Changes since last version:
- rebase atop f9c7dea73e2eac6a35edff278ccae7ca1548fe35
- rename `TEST-62-CPIO` to `TEST-62-SKIPCPIO`, so that it's distinct from `dracut-cpio` tests

ddiss:
Ping - is there anything I can do to move this forwards?

ddiss:
I've updated the cover letter to include more recent benchmark results, which include XFS and Dracut zstd numbers for comparison.

ddiss:
One other note is that we encountered what appears to be a bug in GRUB's Btrfs driver when reading initramfs images with a large number of shared extents: https://bugzilla.opensuse.org/show_bug.cgi?id=1190982 . A Btrfs developer is investigating the issue.
The GRUB issue was not seen with reflinked initramfs images on XFS.

ddiss:
> @haraldh ping - I'd really appreciate some feedback on whether this can be merged. Just to make it clear, there's no change to default behaviour here, the new functionality is only enabled if built with `configure --enable-dracut-cpio` and run with `dracut --enhanced-cpio`.

Ping again... sorry to be a pain.

johannbg (approved):
I dont see anything at this point why this cant be merged
@haraldh could you review this and add your feedback.

haraldh (approved):

PR #1532, opened 2021-06-10
fix(squash): create relative symlinks

ryncsn:
Don't use absolute path, or it may resolve to wrong files after unpack
the initramfs to a subpath on a running system.

## Checklist

### DIFF
--- a/modules.d/99squash/module-setup.sh
+++ b/modules.d/99squash/module-setup.sh
@@ -53,8 +53,8 @@ installpost() {
     dracut_kernel_post
 
     # Install squash image init script.
-    ln -sfn /usr/bin "$initdir/bin"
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

ryncsn:
Hmm, I just found this will result in some wired result with busybox, will update the commit shortly.

ryncsn:
The `ln_r` dracut helper can handle things well so updated the PR to just use the helper.

ryncsn:
@haraldh @danimo @johannbg ping? I've tested this multiple times, should be OK to merge.

johannbg:
@ryncsn Harald is on vacation. I think he comes back next week and at that point it should be merged once he has review this.

haraldh (approved):

PR #1534, opened 2021-06-11
fix(mksh): requires printf

### LINKED ISSUE #1530: printf not installed
**Describe the bug**
Multiple errors about missing printf are logged during boot when `/bin/sh` points to `mksh`

**Distribution used**
PLD

**Dracut version**
055

**Init system**
systemd

**To Reproduce**
Generate image with `/bin/sh -> mksh`

**Expected behavior**
All messages are logged correctly.

**Additional context**
`printf` is not a builtin in `mksh` (at least does not seem to be by default) and `printf` binary is not installed.

tpgxyz:
This pull request changes requirements on printf for mksh module

## Checklist

Fixes #1530

### DIFF
--- a/modules.d/00mksh/module-setup.sh
+++ b/modules.d/00mksh/module-setup.sh
@@ -7,6 +7,7 @@ check() {
 
     # If the binary(s) requirements are not fulfilled the module can't be installed.
     require_binaries mksh || return 1
+    require_binaries printf || return 1
 
     # Return 255 to only include the module, if another module requires it.
     return 255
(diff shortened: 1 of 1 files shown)


jpalus:
I believe this PR covers case when `mksh` module is explicitly included, however it does not handle scenario in which system `/bin/sh` points to `mksh` and `/bin/sh -> mksh` is installed implicitly when processing shebang of installed script.

tpgxyz:
Yes, this PR fixes specific bug, which you has filled.
Speaking of detecting real /bin/sh is something that needs to be done

jpalus:
Note that my bug report was triggered by the implicit case (`/bin/sh` points to `mksh` in PLD).

haraldh (approved):

johannbg (approved):

PR #1535, opened 2021-06-11
fix(dracut.sh): handle symlinks and files beginning with '.' appropriately while using '-i' option

hbathini:
While including a directory using '--include' option, the file and
subdirectory names that begin with '.' are not included. Also, dracut
throws a warning message when a subdirectory is empty or only has
files or subdirectories that begin with '.'.

For example, while trying to include /tmpdata directory with the
below tree:

```  # tree -a /tmpdata
  /tmpdata
  ├── .anothertestdir
  ├── testdir
  │   └── .testsubdir
  └── .testfile
```
dracut throws the below warning message:

```
  # dracut --include /tmpdata /root
  cp: cannot stat '/tmpdata/testdir/*': No such file or directory
  #
```

and this is how the included /tmpdata directory tree looks:

```
  # tree -a root
  root
  └── testdir
```

No file or directory beginning with '.' is included & also, copying
/tmpdata/testdir reported "No such file or directory" warning. Using
'.' instead of '*' in the below command will fix the warning whether
the directory being copied is empty or only has files or directories
that begin with dot:

```  $DRACUT_CP -t "$object_destdir" "$dracutsysrootdir$objectname"/*```

Also, enable 'dotglob' temporarily to include files and directories
beginning with a `.' in the results of pathname expansion of source
directory being included.

With the above mentioned changes files and directories beginning
with '.' are also included successfully. The included /tmpdata looks
like below with the patch:

```  # tree -a root
  root
  ├── .anothertestdir
  ├── testdir
  │   └── .testsubdir
  └── .testfile
```

 Also, "No such file or directory" warning is not seen anymore for an
empty subdirectory after the patch:

```
  # dracut --include /tmpdata /root
  #
```

Also, if the first level sub-directory is a symlink, a directory is created instead and
all the files in the directory pointed to by the symlink are copied. Ensure a symlink
is created instead by checking for it..

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2071,7 +2071,7 @@ for ((i = 0; i < ${#include_src[@]}; i++)); do
             shopt -q -s dotglob
             for objectname in "$src"/*; do
                 [[ -e $objectname || -L $objectname ]] || continue
-                if [[ -d $objectname ]]; then
+                if [[ -d $objectname ]] && [[ ! -L $objectname ]]; then
                     # objectname is a directory, let's compute the final directory name
                     object_destdir=${destdir}/${objectname#$src/}
                     if ! [[ -e $object_destdir ]]; then
--- a/dracut.sh
+++ b/dracut.sh
@@ -2067,6 +2067,8 @@ for ((i = 0; i < ${#include_src[@]}; i++)); do
             # check for preexisting symlinks, so we can cope with the
             # symlinks to $prefix
             # Objectname is a file or a directory
+            reset_dotglob="$(shopt -p dotglob)"
+            shopt -q -s dotglob
             for objectname in "$src"/*; do
                 [[ -e $objectname || -L $objectname ]] || continue
                 if [[ -d $objectname ]]; then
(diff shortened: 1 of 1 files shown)


ryncsn:
Hi Hari, thanks for the fix.

I just came up with another approach of fixing this issue, but having created a PR yet since I think I need more test:
https://github.com/ryncsn/dracut/commit/0eae3240376bf1222d877121f2cccfe3d6e99fdd

I'm fine with either way, let see how others think about it.

ryncsn:
Hi Hari,

Turns out there is one more issue to be fixed here, please see:
https://github.com/ryncsn/dracut/commit/e8da43aae22dc42e1143bc5fda3360ef8fb258b1

I was seeing redundant filed being include in the initramfs with --include while testing the patch, this should fix it.

hbathini:
> Hi Hari,
> 
> Turns out there is one more issue to be fixed here, please see:
> [ryncsn@e8da43a](https://github.com/ryncsn/dracut/commit/e8da43aae22dc42e1143bc5fda3360ef8fb258b1)
> 
> I was seeing redundant filed being include in the initramfs with --include while testing the patch, this should fix it.

Thanks, Kairui.
I have added this patch to the PR...

johannbg:
Note that you need to fix your commit msg since `fix(foo): fix` is redundant in other word rephrase your commit msg excluding the fix word after `fix(dracut.sh):`

hbathini:
> Note that you need to fix your commit msg since `fix(foo): fix` is redundant in other word rephrase your commit msg excluding the fix word after `fix(dracut.sh):`

Sure @johannbg 

@danimo @haraldh @johannbg Let me know, if you prefer the approach taken by @ryncsn over this.
I mean, commit [607d4a94](https://github.com/hbathini/dracut/commit/607d4a94691ab36c0f418362de93080d2661d861) or [0eae3240](https://github.com/ryncsn/dracut/commit/0eae3240376bf1222d877121f2cccfe3d6e99fdd)

haraldh (approved):

ryncsn:
@johannbg @danimo 
Hi, can you help approve this?

johannbg (approved):

PR #1538, opened 2021-06-16
docs: introduce a 'ib-pkey=' command line argument

bengal:
An Infiniband network can be segmented in different partitions, similar to VLANs. Each partition is identified by a partition key (P_Key).

On Linux, a new partition for a IP-over-Infiniband device can be created via e.g.:
```
echo 0x8001 > /sys/class/net/ib0/create_child
```
The resulting interface will be `ib0.8001` with P_Key 0x8001.

Document a new command line argument to create the partition.

NetworkManager is going to implement the new option in [1].

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/884

### DIFF
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -575,6 +575,11 @@ USB Android phone::
 * enp0s29u1u2
 =====================
 
+The following options are supported by the 'network-legacy' dracut
+module. Other network modules might support a slightly different set of
+options; refer to the documentation of the specific network module in use. For
+NetworkManager, see *nm-initrd-generator*(8).
+
 **ip=**__{dhcp|on|any|dhcp6|auto6|either6|link6|single-dhcp}__::
     dhcp|on|any::: get ip from dhcp server from all interfaces. If netroot=dhcp,
     loop sequentially through all interfaces (eth0, eth1, ...) and use the first


johannbg:
@bengal  Hmm aren't we missing what actually parses that command line argument ( from the nm module? ) in this PR? 
Is this still an WIP?

tyll:
> @bengal Hmm aren't we missing what actually parses that command line argument ( from the nm module? ) in this PR?
> Is this still an WIP?

The NetworkManager part is here: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/884

bengal:
> @bengal Hmm aren't we missing what actually parses that command line argument ( from the nm module? )

Right, this PR only adds documentation for the new command line argument.

For the network-manager module, the new argument doesn't need to be implemented in the dracut repo but in NM itself. There is already a PR ready to be merged, as soon as this gets accepted. Perhaps the man page should document the NM version in which the feature is added?

tyll:
@johannbg in case it is not clear: The NM implementation is blocked by getting approval for the naming of the command line argument requested here. Therefore it would be great to some feedback whether the name is fine or not. Thank you.

johannbg:
@tyll well I think the tradition would be to have it `ib.pkey=` also note the decided command line argument might need to be excluded from systemd's network generator but I leave it up to @haraldh to ack or nack the command line argument as is, he however is on vacation and will return I think next week so this is stuck until he's back.

johannbg:
@haraldh ping since you seem to be back from vacation :smile:

haraldh:
Shall we document kernel command line options, which dracut does not parse in any way?

A better way is to mention additional man pages, where useful kernel command line options are documented.

haraldh:
This way, the dracut documentation is rarely out of sync.

tyll:
So if we implement this as ib.pkey it would be fine by you? We can document this at our end, too. It would be great if we can get some  go ahead for this so we can merge the implementation for NM.

bengal:
> A better way is to mention additional man pages, where useful kernel command line options are documented.

@haraldh @johannbg , what do you think about the new commit? My only concern with this approach is that it could cause more incompatibility, with different modules implementing different options (or the same option with different names).

Another question: is it better to namespace new options with the `rd.` prefix (e.g. `rd.ib.pkey`) or is something like `ib.pkey` fine?

johannbg:
@bengal yeah it will end up causing more confusion and incompatibility regarding the rd. prefix @haraldh is more familiar with existing naming conventions but I would say just keep it ib.pkey

haraldh (approved):

PR #1539, opened 2021-06-17
fix(fips): handle s390x OSTree systems

jlebon:
On s390x, the `BOOT_IMAGE` karg injected by the bootloader is not a path
to the kernel image, but rather an integer describing the index of the
menu entry selected. Because of the way the s390x bootloader works,
there is no information retained about e.g. the path of the kernel that
was loaded.

This causes issues for the FIPS code which assumes that `BOOT_IMAGE` is
a path to the kernel image to derive the HMAC path. In non-OSTree
systems, this ends up working anyway, because the kernel is located at
the root of the boot partition.  In OSTree systems, this is not the
case. However, OSTree systems use BLS configs, and they are named in
reverse order of precedence (i.e. menu ordering). So from the
`BOOT_IMAGE` integer, we can figure out which BLS entry was selected.

Add some code to do just this on s390x. This isn't completely foolproof,
because it presumes that (1) BLS configs were used to populate the
bootloader (and that they were exactly in the same state they currently
are when `zipl` was run), and (2) there are no other menu entries
originating from outside the BLS configs. However, if these assumptions
are wrong we would simply fail the boot, which is currently what is
happening anyway.

See also:
https://github.com/openshift/os/pull/546
https://github.com/ibm-s390-linux/s390-tools/issues/78

Tested-by: Muhammad Adeel <muhammad.adeel@ibm.com>

### DIFF
--- a/modules.d/01fips/fips.sh
+++ b/modules.d/01fips/fips.sh
@@ -124,6 +124,21 @@ do_fips() {
     else
         BOOT_IMAGE="$(getarg BOOT_IMAGE)"
 
+        # On s390x, BOOT_IMAGE isn't a path but an integer representing the
+        # entry number selected. Let's try the root of /boot first, and
+        # otherwise fallback to trying to parse the BLS entries if it's a
+        # BLS-based system.
+        if [ "$(uname -m)" = s390x ]; then
+            if [ -e "/boot/vmlinuz-${KERNEL}" ]; then
+                BOOT_IMAGE="vmlinuz-${KERNEL}"
+            elif [ -d /boot/loader/entries ]; then
+                bls=$(find /boot/loader/entries -name '*.conf' | sort -rV | sed -n "$((BOOT_IMAGE + 1))p")
+                if [ -e "${bls}" ]; then
+                    BOOT_IMAGE=$(grep ^linux "${bls}" | cut -d' ' -f2)
+                fi
+            fi
+        fi
+
         # Trim off any leading GRUB boot device (e.g. ($root) )
         BOOT_IMAGE="$(echo "${BOOT_IMAGE}" | sed 's/^(.*)//')"
 
--- a/modules.d/01fips/module-setup.sh
+++ b/modules.d/01fips/module-setup.sh
@@ -67,7 +67,7 @@ install() {
     inst_hook pre-udev 01 "$moddir/fips-load-crypto.sh"
     inst_script "$moddir/fips.sh" /sbin/fips.sh
 
-    inst_multiple sha512hmac rmmod insmod mount uname umount
+    inst_multiple sha512hmac rmmod insmod mount uname umount grep sed cut find sort
 
     inst_simple /etc/system-fips
     [ -c "${initdir}"/dev/random ] || mknod "${initdir}"/dev/random c 1 8 \


jlebon:
> - [ ] I have tested it locally

I do not have an s390x machine on hand, but it was tested by someone who does.

> - [ ] I have reviewed and updated any documentation if relevant

I think there's nothing to update there. Let me know if not!

> - [ ] I am providing new code and test(s) for it

I don't see existing tests for FIPS in the tree. I guess coverage for that happens more downstream?

jlebon:
> This isn't completely foolproof,
because it presumes that (1) BLS configs were used to populate the
bootloader (and that they were exactly in the same state they currently
are when zipl was run), and (2) there are no other menu entries
originating from outside the BLS configs. However, if these assumptions
are wrong we would simply fail the boot, which is currently what is
happening anyway.

One thing we could also do is gate this on `/run/ostree-booted`. This isn't really OSTree-specific code: any system which puts its kernel images in a subdir will hit this. But if we'd like to reduce potential fallout, we can start off with just applying it to OSTree systems for now.

jlebon:
TBH, it seems to me like this part of the FIPS verification process on s390x is just broken because there simply isn't enough information to definitively know which image file corresponds to the loaded kernel (s390x SMEs, please chime in if this is incorrect!).

johannbg:
Hmm I would rather avoid we implement fixes that rely on us shaking a magic 8 ball to see if it works so it would be preferable if we actually had someone that had an s390x machine on hand or access to one to chime in on this. Given that have not used my connection to Nordic IBM and the s390 team for a couple of years, those sources have run dry @danimo does Suse have one on hand or in it's basement or an active connection?  :)

madeelibm:
> 
> 
> TBH, it seems to me like this part of the FIPS verification process on s390x is just broken because there simply isn't enough information to definitively know which image file corresponds to the loaded kernel (s390x SMEs, please chime in if this is incorrect!).

@jlebon I think in this context we know which Kernel is loaded when we did zipl from rhcos-fips.sh:
`
            for f in "${tmpsysroot}"/boot/loader/entries/*.conf; do
                  for line in title version linux initrd options; do
                      echo $(grep $line $f) &gt;&gt; $tmpfile
                  done
              done
              grep options $tmpfile | cut -c 9- &gt; $optfile
       
              zipl --verbose \
                         --target "${tmpsysroot}/boot" \
                         --image $tmpsysroot/boot/"$(grep linux $tmpfile | cut -d' ' -f2)" \
                         --ramdisk $tmpsysroot/boot/"$(grep initrd $tmpfile | cut -d' ' -f2)" \
                         --parmfile $optfile
`
and if Kernel image is not found on location as mentioned in this PR, the FIPS verification will fail which fails anyway without this PR.

jlebon:
> > TBH, it seems to me like this part of the FIPS verification process on s390x is just broken because there simply isn't enough information to definitively know which image file corresponds to the loaded kernel (s390x SMEs, please chime in if this is incorrect!).
> 
> @jlebon I think in this context we know which Kernel is loaded when we did zipl from rhcos-fips.sh:

The problem is that at boot time, there's no way to tell the kernel path trusting only the bootloader and kernel arguments.

> Hmm I would rather avoid we implement fixes that rely on us shaking a magic 8 ball to see if it works so it would be preferable if we actually had someone that had an s390x machine on hand or access to one to chime in on this.

To be clear, this PR was successfully tested on s390x (by @madeelibm). See https://github.com/openshift/os/pull/546#issuecomment-863330340.

I see this as a compromise until we have s390x match the behaviour of other arches (that's https://github.com/ibm-s390-linux/s390-tools/issues/78).

jlebon:
Updated to appease ShellCheck! :arrow_up: 
@madeelibm Can you retest this new version?

nikita-dubrovskii:
> Updated to appease ShellCheck! arrow_up
> @madeelibm Can you retest this new version?

did test today with rhcos-4.8 + dracut*.rpm overrides - it works

jlebon:
Gentle ping; can we have someone take a look at this?

johannbg (approved):
LGTM

haraldh (approved):

PR #1541, opened 2021-06-18
fix(nvmf): validate_ip_conn

charles-rose:
1. ifname should be the interface on which $traddr can be reached. It currently gets set to "lo", which is incorrect. This patch identifies the correct interface that would be used for the nvme-of-tcp connection.
2. 'ip l show' and 'ip route get' both succeed if interface and route are present. Current 'if' condition (wrongly) produces a warning when the commands succed. This patch fixes that.

Signed-off-by: Charles Rose <charles.rose@dell.com>

This pull request changes...

## Changes
- fix ifname and route check for nvmf tcp connections.

## Checklist

Fixes #

### DIFF
--- a/modules.d/95nvmf/parse-nvmf-boot-connections.sh
+++ b/modules.d/95nvmf/parse-nvmf-boot-connections.sh
@@ -40,15 +40,15 @@ validate_ip_conn() {
         return 1
     fi
 
-    ifname=$(ip -o route get to "$local_address" | sed -n 's/.*dev \([^ ]*\).*/\1/p')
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
Change it 
from 
`nvmf: fix validate_ip_conn`
to 
`fix(nvmf): validate_ip_conn`

johannbg:
@charles-rose you need to fix the commit message itself not the title of the pull request.

haraldh (approved):

charles-rose:
> 
> 
> @charles-rose you need to fix the commit message itself not the title of the pull request.
Thank you for the review @johannbg
Should I create a new pull request with the amended commit?

haraldh:
No, just force push a corrected version to your PR branch.

PR #1544, opened 2021-06-21
fix(crypt-gpg): Execute --card-status on each try

LinkTed:
## Changes
Sometimes I forget to insert the gpg card before the first password prompt is printed. 

If the gpg card is not inserted before the `--card-status` command is executed then the public key is not linked with the card. Therefore, the LUKS partition cannot be decrypted. To solve this, the `--card--status` command is executed on each try.

## Checklist

Fixes #


haraldh (approved):

LinkTed:
I fix the commit message.

johannbg (approved):

PR #1547, opened 2021-06-24
fix(network-manager): support teaming under NM+systemd

dustymabe:
Previously when NM was run without dbus then teaming would come
up appropriately [1], but now that dbus exists we also need to
include some supporting infrastructure to allow for it to work
again.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/d689380cfc5734a29b1302d68027190e1a606265

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -38,6 +38,11 @@ install() {
         inst "$dbussystem"/org.freedesktop.NetworkManager.conf
         inst_multiple nmcli nm-online
 
+        # teaming support under systemd+dbus
+        inst_multiple -o \
+            "$dbussystem"/teamd.conf \
+            "$dbussystemconfdir"/teamd.conf
+
         # Install a configuration snippet to prevent the automatic creation of
         # "Wired connection #" DHCP connections for Ethernet interfaces
         inst_simple "$moddir"/initrd-no-auto-default.conf /usr/lib/NetworkManager/conf.d/


dustymabe:
I have tested this locally.

modules.d/35network-manager/module-setup.sh
```diff
@@ -38,6 +38,12 @@ install() {
         inst "$dbussystem"/org.freedesktop.NetworkManager.conf
         inst_multiple nmcli nm-online
 
+        # teaming support under systemd+dbus
+        inst_multiple -o                          \
+           "$systemdsystemunitdir"/teamd@.service \
+           "$dbussystem"/teamd.conf               \
+           "$dbussystemconfdir"/teamd.conf
```
dustymabe, line 45: @jpirko - the fedora rpm currently defines the dbus conf in `/etc/dbus-1/system.d/teamd.conf`. Should it be modified to place it in `/usr/share/dbus-1/system.d/teamd.conf` instead?

Here I'm picking it up from either place in case it changes.

dustymabe:
cc @bengal

modules.d/35network-manager/module-setup.sh
```diff
@@ -38,6 +38,12 @@ install() {
         inst "$dbussystem"/org.freedesktop.NetworkManager.conf
         inst_multiple nmcli nm-online
 
+        # teaming support under systemd+dbus
+        inst_multiple -o \
+            "$systemdsystemunitdir"/teamd@.service \
```
bengal, line 43: Currently, NetworkManager spawns teamd directly and then waits that teamd appears on D-Bus to communicate with it.
This line above is not needed because teamd is not started as a systemd service.

The rest LGTM, I also tried the patch and it works well. Thanks!

dustymabe: fixed in latest push. Thanks @bengal.

dustymabe:
@haraldh - when we merge this do you think we could get it backported to F34?

modules.d/35network-manager/module-setup.sh
bengal: LGTM now.

dustymabe:
> LGTM now.

Thanks @bengal - any other reviewers we need to wait on or can this be merged now?

johannbg (approved):
LGTM

dustymabe:
Any ideas how to fix the failing `packit` test?

johannbg:
@dustymabe packit tests passing are not a requirement so you can just ignore it.  Anyways to fix a failing packit test requires us to open up an issue with the packit team, which then will look into it along with infra with high probability that it's not our PR's causing failure, not packit itself but downstream maintainer or infra making some mistake or upgrade ( this has happened before ).

haraldh (approved):

PR #1550, opened 2021-06-28
fix(dmsquash-live): set correct path to checkisomd5

mikhailnov:
isomd5sum (https://github.com/rhinstaller/isomd5sum) installs the binary into /usr/bin.
/bin/checkisomd5 does not exist in the initrd in ROSA rosa2019.1, but /usr/bin/checkisomd5 does exist.
Maybe "ExecStart=/bin/checkisomd5" works ok on Fedora because /bin is a symlink to /usr/bin?
isomd5sum package in Fedora has only the file /usr/bin/checkisomd5 and does not have /bin/*.

Here we could set "ExecStart=checkisomd5" and avoid relying on any absolute path,
but it is supported in systemd >= 239 from 06.2018 (https://github.com/systemd/systemd/blob/5b8fdb1/NEWS#L4420),
probably it is better to avoid breakages on systems with older systemd like RHEL 7 in this case.
There is no reason why anyone would install checkisomd5 into /bin but not /usr/bin.

### DIFF
--- a/modules.d/90dmsquash-live/checkisomd5@.service
+++ b/modules.d/90dmsquash-live/checkisomd5@.service
@@ -6,7 +6,7 @@ Before=shutdown.target
 [Service]
 Type=oneshot
 RemainAfterExit=no
-ExecStart=/bin/checkisomd5 --verbose %f
+ExecStart=/usr/bin/checkisomd5 --verbose %f
 StandardInput=tty-force
 StandardOutput=inherit
 StandardError=inherit


haraldh (approved):

mikhailnov:
It still does not work even with this change :( Adding "rd.live.check" to kernel cmdline causes just a black screen

image

image

it was the same before this fix. But this fix is still valid.

johannbg:
This wont work on any distribution that has not concluded the /usr "merge" and any distribution that has concluded the /usr merge should be symlinking /bin to /usr/bin so this should just work and based on whatever issue you are dealing with in the first place this change did not fix it. 

In this scenario I would say continuing to using /bin as the default path until we decide to "officially" stop supporting distributions that have not yet finished /usr merge so this PR is a nack from me and suggest Rosa carries that patch downstream until we officially stop supporting non usr merged distro's.

mikhailnov:
There is no distro that packages checkisomd5 to /bin, so on _any_ distro dracut will put it into /usr/bin/ inside the initrd, so the proposed variant will work every where, won't it?

> stop supporting distributions that have not yet finished /usr merge
I see that you very much like to stop supporting something :-) Let's better stop supporting systemd < 240 (see description of the PR)

mikhailnov:
As for usr merge... I do not see sense in existense of both `/bin` and `/usr/bin` inside the initrd, I would suggest to rework installation of binaries so that it installs binaries from `/bin` into `/usr/bin` (and symlink `/bin` to `/usr/bin` inside the initrd if it makes sense)

But I am not against usr merge in a soft scenario: keep bin and sbin separate, but move them to usr. This work just requires a lot of time.

johannbg:
> I see that you very much like to stop supporting something :-) Let's better stop supporting systemd < 240 (see description of the PR)

Saw it and was aware of it prior to it anyways it's still recommended to use absolute paths which is what we will continue to do.

mikhailnov:
Ok. Please explain why "This wont work on any distribution that has not concluded the /usr "merge"" ? If I understand correctly, it is not so.

johannbg:
> But I am not against usr merge in a soft scenario: keep bin and sbin separate, but move them to usr. This work just requires a lot of time.

I'm not sure what you mean by soft scenarios but the fact is that many distributions have not done the /usr merge and other distributions have not completed it and decide to paper over that lack of completions via symlinks.

PR #1552, opened 2021-06-29
fix(network-manager): don't pull in systemd-udev-settle

dustymabe:
We get a nice warning about it being deprecated:

```
systemd-udev-settle.service is deprecated. Please fix nm-initrd.service not to pull it in.
```

The service is deprecated because its purpose was to wait for the
discovery of all hardware, but it didn't guarantee that (see the
systemd-udev-settle man page).

NM now runs as an independent service and can deal with devices showing
up at any point, but it does need udev to be started. For now just
Want/After systemd-udev-trigger.

### DIFF
--- a/modules.d/35network-manager/nm-initrd.service
+++ b/modules.d/35network-manager/nm-initrd.service
@@ -1,7 +1,7 @@
 [Unit]
 DefaultDependencies=no
-Wants=systemd-udev-settle.service
-After=systemd-udev-settle.service
+Wants=systemd-udev-trigger.service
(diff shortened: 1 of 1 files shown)


dustymabe:
Considering the case - should we be using `systemd-udev-trigger.service` or `systemd-udevd.service` here?

dustymabe:
cc @bengal

johannbg (approved):

bengal:
LGTM, thanks.

haraldh (approved):

PR #1556, opened 2021-07-08
fix(kernel-modules): add more modules on RISC-V

andreas-schwab:
Include the same extra modules for ARM also on RISC-V and add drivers/mmc/host
and drivers/spi to the list.


johannbg (changes requested):
You need to change your commit msg
from
`fix(kernel-modules): Add more modules on RISC-V`
to
`fix(kernel-modules): add more modules on RISC-V`
to pass the commisery check.

Other than that LGTM

johannbg (approved):

johannbg:
@haraldh ping

haraldh (approved):

PR #1558, opened 2021-07-09
99squash: keep ld cache under initdir

pfliu:
When running kdump on PowerPC, the following bug is hit:
[    0.391629] Freeing unused kernel memory: 5568K
[    0.391634] This architecture does not have kernel memory protection.
/bin/sh: error while loading shared libraries: libc.so.6: cannot open shared object file: No such file or directory
[    0.392214] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00
[    0.392214]
[    0.392223] CPU: 6 PID: 1 Comm: init Not tainted 4.18.0-319.el8.ppc64le #1
[    0.392228] Call Trace:
[    0.392234] [c00000000c703c10] [c000000008ecb94c] dump_stack+0xb0/0xf4 (unreliable)
[    0.392243] [c00000000c703c50] [c000000008167324] panic+0x148/0x3c4
[    0.392249] [c00000000c703cf0] [c000000008170474] do_exit+0xcd4/0xd40
[    0.392255] [c00000000c703dc0] [c0000000081705b0] do_group_exit+0x60/0x110
[    0.392261] [c00000000c703e00] [c000000008170684] sys_exit_group+0x24/0x30
[    0.392268] [c00000000c703e20] [c00000000800b408] system_call+0x5c/0x70

This is due to the non-conventional library path:
ldd /bin/bash
        linux-vdso64.so.1 (0x00007fffbdc90000)
        libtinfo.so.6 => /lib64/libtinfo.so.6 (0x00007fffbda80000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fffbda50000)
        libc.so.6 => /lib64/glibc-hwcaps/power9/libc-2.28.so (0x00007fffbd830000)
                            ^^^
        /lib64/ld64.so.2 (0x00007fffbdcb0000)

During the process of building kdumpimg, all of dynamic library info
have been cached in ld.so.cache. Hence this bug can be simplely resolved
by keeping ld cache under $initdir/etc

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/99squash/module-setup.sh
+++ b/modules.d/99squash/module-setup.sh
@@ -28,6 +28,11 @@ installpost() {
         [[ $squash_dir == "$i"/* ]] || mv "$i" "$squash_dir"/
     done
 
+    # initdir also needs ld.so.* to make ld.so work
+    inst /etc/ld.so.cache
+    inst /etc/ld.so.conf
+    inst_dir /etc/ld.so.conf.d/
+
     # Create mount points for squash loader
     mkdir -p "$initdir"/squash/
     mkdir -p "$squash_dir"/squash/


pfliu:
Note for the testing,  Since the latest dracut can not build squash image on my RHEL8. I use "set -x" to observe the behavior. It works.

pfliu:
Discuss with Kairui offline. There is a big change of squashfs, and after switch-root, all the content in $initdir can not been seen. So cp /etc/ld.so* instead of move them.

This is the change in V2

haraldh:
please run `shfmt` and fix your commit message, as mentioned in https://github.com/dracutdevs/dracut/blob/master/docs/HACKING.md to satisfy the lint tests.

haraldh:
@ryncsn ping

johannbg:
Hmm this sounds like a bug in kdump. Why does kdump not create the correct symlink?

johannbg (changes requested):
Change it 
from
`99squash: keep ld cache under initdir`
to 
`fix(squash): keep ld cache under initdir`

You also need to fix your formatting issues of your changes. 
you can do so by running from the git directory via 
`sudo dnf -y install shfmt && shfmt -s -w modules.d/99squash/module-setup.sh`

pfliu:
> Hmm this sounds like a bug in kdump. Why does kdump not create the correct symlink?
Actually, it is a bug with ./src/install/dracut-install.c. Anyway, /etc/ld.so.* is required to protect against non-standard library search path I.e. if something special in /etc/ld.so.conf. 

As the systemd has all ready installed ld.so.* in squash, so initdir should need the same

pfliu:
> You need to fix your commit msg.
> Change it
> from
> `99squash: keep ld cache under initdir`
> to
> `fix(squash): keep ld cache under initdir`
> 
> You also need to fix your formatting issues of your changes.
> you can do so by running from the git directory via
> `sudo dnf -y install shfmt && shfmt -s -w modules.d/99squash/module-setup.sh`

Thanks for reminding.

I have updated v3 following the guideline.

pfliu:
@johannbg

It turns out that it is hard to create a symlink by dracut-install. This is caused by the ldd.

On my system, I have the following info for the library
[root@ibm-p9z-06-lp3 ~]# ldd /usr/bin/bash
        linux-vdso64.so.1 (0x00007fffb05b0000)
        libtinfo.so.6 => /lib64/libtinfo.so.6 (0x00007fffb03a0000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fffb0370000)
        libc.so.6 => /lib64/glibc-hwcaps/power9/libc-2.28.so (0x00007fffb0150000)  //use cache
        /lib64/ld64.so.2 (0x00007fffb05d0000)

While after removing ld cache, there is a different result
[root@ibm-p9z-06-lp3 ~]# rm -rf /etc/ld.so.cache
[root@ibm-p9z-06-lp3 ~]# ldd  /usr/bin/bash
        linux-vdso64.so.1 (0x00007fff98100000)
        libtinfo.so.6 => /lib64/libtinfo.so.6 (0x00007fff97f00000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fff97ed0000)
        libc.so.6 => /lib64/glibc-hwcaps/power9/libc.so.6 (0x00007fff97cb0000) //no cache
        /lib64/ld64.so.2 (0x00007fff98120000)

Both behavior are OK for ld.so, but raise a critical obstacle for dracut to work in the 1st case.

Since dracut-install.c heavily relies on the result of ldd, and it can not expect definitely result. So copy the ld.so.cache is the best and efficient solution method at present.

Thanks

modules.d/99squash/module-setup.sh
```diff
@@ -25,7 +25,16 @@ installpost() {
...
+        [[ $squash_dir == "$i"/* ]] || [[ "$initdir/etc" == "$i" ]] || mv "$i" "$squash_dir"/
+    done
+    # keep ld cache under $initdir
+    mkdir -p "$squash_dir"/etc
+    for i in "$initdir"/etc/*; do
+        if [[ $i =~ "$initdir"/etc/ld.so* ]]; then
+            cp -r "$i" "$squash_dir"/etc
+        else
+            mv "$i" "$squash_dir"/etc
+        fi
```
ryncsn, line 37: Hi Pingfan,

I think we can just add one line of code here: `inst /etc/ld.so.cache`, `/etc/ld.so.cache` is the only file affects early squash loader. No need to move&copy things again and again.

pfliu:
update to v4 based on ryncsn's comment

johannbg:
@pfliu @ryncs mentions that ld's cache file is the only thing that is needed as in`inst /etc/ld.so.cache` should suffice based on his comment so why do you need all the other files for and the directory? Have you tested this?

ryncsn:
@johannbg Pingfan is right, ld.so.conf is also needed as it may affect LD_LIBRARY_PATH, this is referenced by ld.so. The new commit looks good to me.

ryncsn (approved):

johannbg (approved):

PR #1559, opened 2021-07-09
Print Entering dracut emergency mode instade of Entering emergency mode

wtmpx:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-manager/nm-initrd.service
+++ b/modules.d/35network-manager/nm-initrd.service
@@ -1,7 +1,7 @@
 [Unit]
 DefaultDependencies=no
-Wants=systemd-udev-settle.service
-After=systemd-udev-settle.service
+Wants=systemd-udev-trigger.service
+After=systemd-udev-trigger.service
 After=dracut-cmdline.service
 Wants=network.target
 Before=network.target
--- a/modules.d/35network-manager/nm-initrd.service
+++ b/modules.d/35network-manager/nm-initrd.service
@@ -1,7 +1,7 @@
 [Unit]
 DefaultDependencies=no
-Wants=systemd-udev-trigger.service
-After=systemd-udev-trigger.service
+Wants=systemd-udev-settle.service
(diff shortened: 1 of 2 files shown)


modules.d/35network-manager/nm-initrd.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 DefaultDependencies=no
-Wants=systemd-udev-settle.service
-After=systemd-udev-settle.service
+Wants=systemd-udev-trigger.service
+After=systemd-udev-trigger.service
```
haraldh, line 4: ???

haraldh (changes requested):

modules.d/35network-manager/nm-initrd.service
wtmpx: Sorry, no idea. can you reject above change?

haraldh: > Sorry, no idea. can you reject above change?

no.... you have to get your git commit history right

haraldh:
no we have 3 commits.. maybe squash them :)

johannbg:
I'm nack-ing this PR beside it being to messy to merge as well as if you think that the emergency mode msg is to similar to systemd then make a PR against systemd to change it's message to clarify that you are in it since dracut emergency mode pre-dates systemd. That said it does not matter if the emergency mode msg is the same or similar ( which is being done for consistency ) since the user has been dropped into emergency mode in which it's expected that individual dealing with the emergency knows what he or she is doing thus has no problem figuring it out at which stage of the OS stack, the OS entered it...

wtmpx:
> if you think that the emergency mode msg is to similar to systemd then make a PR against systemd to change it's message to clarify that you are in it since dracut emergency mode pre-dates systemd.

I think for normal user when they need to explain where the system stop booting a clear message would be much easier/clear.

PR #1563, opened 2021-07-16
Fix memory leak

dtardon:
This pull request fixes a memory leak in dracut-install.

## Checklist

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -419,7 +419,7 @@ static char *get_real_file(const char *src, bool fullyresolve)
         char linktarget[PATH_MAX + 1];
         ssize_t linksz;
         _cleanup_free_ char *fullsrcpath;
-        char *abspath = NULL;
+        _cleanup_free_ char *abspath = NULL;
         struct stat sb;
 
         if (sysrootdirlen) {
@@ -482,7 +482,7 @@ static char *get_real_file(const char *src, bool fullyresolve)
         }
 
         log_debug("get_real_file('%s') => '%s'", src, abspath);
-        return abspath;
+        return TAKE_PTR(abspath);
 }
 
 static int resolve_deps(const char *src)
--- a/src/install/macro.h
+++ b/src/install/macro.h
@@ -281,4 +281,14 @@ do {                                                                    \
             sizeof(type) <= 4 ? 10 :                                    \
             sizeof(type) <= 8 ? 20 : sizeof(int[-2*(sizeof(type) > 8)])))
 
+
+/* Takes inspiration from Rusts's Option::take() method: reads and returns a pointer, but at the same t  ime resets it to
+ * NULL. See: https://doc.rust-lang.org/std/option/enum.Option.html#method.take */
+#define TAKE_PTR(ptr)                           \
+        ({                                      \
+                typeof(ptr) _ptr_ = (ptr);      \
+                (ptr) = NULL;                   \
+                _ptr_;                          \
+        })
+
 #include "log.h"


haraldh (approved):

haraldh:
git commit title of `Introduce TAKE_PTR() macro` should be: `feat(install): introduce TAKE_PTR() macro`

johannbg:
@dtardon you are using title case in your description in the commit message. 
You need to change the commit msg
from
`feat(install): Introduce TAKE_PTR() macro`
to 
`feat(install): introduce TAKE_PTR() macro`

Other than that LGTM

johannbg (approved):

src/install/dracut-install.c
```diff
@@ -419,7 +419,7 @@ static char *get_real_file(const char *src, bool fullyresolve)
         char linktarget[PATH_MAX + 1];
         ssize_t linksz;
         _cleanup_free_ char *fullsrcpath;
-        char *abspath = NULL;
+        _cleanup_free_ char *abspath = NULL;
```
sshedi, line 422: I see few problems in this fix.
https://man7.org/linux/man-pages/man3/asprintf.3.html says that
upon failure of asprintf `contents of strp are undefined`

1. Here if asprintf fails, function is returning NULL to caller in few places. but if we we use `_cleanup_free_` in declaration, worst case scenario is passing a garbage address to freep function during free.
And in all the calling functions this pointer is freed.

Instead it's better to use `_cleanup_free_` here https://github.com/dracutdevs/dracut/blob/6c71ba4121ae64ccd13fefba68ca327ac623810f/src/install/dracut-install.c#L1240
 
2. There is recursion in this function, so `abspath` should be freed after every recursive call right?

3. As every caller is freeing this returned pointer, it should have resulted in a double free but it's not because of `TAKE_PTR` macro.

I believe, this is a special case where we can't use `_cleanup_free_`

I have attached a sample program, please check. Correct me if my analysis is wrong, I'm just trying help.
I have attached C file in .txt extension, github is not allowing me to upload C file.
[test.txt](https://github.com/dracutdevs/dracut/files/6832655/test.txt)

sshedi (changes requested):
Good catch, needs further fixing.
It will be good if you run valgrind on this function with some test trigger function with multiple scenarios.

johannbg:
@dtardon you need to fix your commit msg so we can merge your PR.

sshedi:
> @dtardon you need to fix your commit msg so we can merge your PR.
@johannbg - don't you think there will be a problem if asprintf fails and abspath holds an undefined value?
Am I missing something here?

src/install/dracut-install.c
zboszor: I was the one who introduced this `get_real_file()` function and the issue with `asprintf()` passing back garbage in the pointer was taken into consideration back then.

There are two outcomes of `asprintf()`: either there's a failure indicated by the return value, in which case you MUST NOT call `free()` on the passed back pointer since it may (trust me, it will!) contain garbage, or `asprintf()` returns the length of the string printed into the successfully allocated area and the valid pointer which you must `free()` later.

If you look closely, every time `asprintf()` returns a negative number (error) this function does `_exit(EXIT_FAILURE)` (which is a change made by @haraldh ) or returns `NULL` immediately so the caller can fail gracefully. Otherwise the pointer is passed back to the caller of `get_real_file()` which will eventually `free()` it. You can't do any better than that. There's no need for the TAKE_PTR macro here and there's no point of this MR then.

zboszor: Also, dracut-install is a short-lived executable, the amount of leaks during its runtime is not really an issue.

sshedi: @zboszor - totally agree with you on your points.

I did a quick search in `dracut-install.c` there are many places where `_cleanup_free_` is used for first argument of asprintf. Instead of fixing all over the place, simple solution will be writing a wrapper for `asprintf` which takes care of setting pointer to NULL upon failure and it solves this issue once and for all with the overhead of changing asprintf function names with wrapper function name (one time thing).

The changes made so far can be kept as it is (because it solves the problem of freeing memory on recursion) and TAKE_PTR solves double free.

zboszor: But this may introduce new crashes, since code sites after checking the return value of `asprintf()` (and exiting or returning immediately in case of failure) do expect a valid (i.e. non-`NULL`) pointer.

zboszor: BTW, the double free is introduced in this same MR by adding `_cleanup_free_` to `abspath` which is fixed by adding the `TAKE_PTR` macro to the return value. The current code seems to be correct as is. Am I missing something?

zboszor: I found what I was missing. Alternative MR is https://github.com/dracutdevs/dracut/pull/1564

sshedi: Your PR #1564 solves the problem in just one function. But the issue has to be mitigated in all places.

Instead we can do it cleanly by writing a wrapper around asprintf, something like
```
#define _GNU_SOURCE
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <stdarg.h>

int _asprintf(char **restrict strp, const char *restrict fmt, ...)
{
    int ret = -1;
    va_list args;

    if (strp == NULL || fmt == NULL) {
        return ret;
    }

    va_start(args, fmt);
    ret = vasprintf(strp, fmt, args);
    if (ret < 0) {
        *strp = NULL;
    }
    va_end(args);

    return ret;
}

int main(void)
{
    char *s = NULL;

    if (_asprintf(&s, "%s", "hello world") < 0) {
        perror("dracut_asprintf failed\n");
        return -1;
    }

    printf("%s\n", s);
    free(s);

    return 0;
}
```

And put it in util.c and use it in all places. This way we can use _cleanup_free_ with source pointer and a lot of boiler plate code can be avoided.

So, my advice to you is, close #1564 and let @dtardon take care of it here in this PR.
He found this issue, let us give him inputs and let him come up with the solution.

zboszor: I can's still see no point in setting the pointer to NULL because

1. you still have to spell out the error check and exit after your proposed wrapper, and
2. code after calling asprintf() do expect a valid non-NULL pointer, so more code changes would be needed to handle if the pointer is NULL
3. frankly, it doesn't make sense to declare every pointer variable with automatic cleanup, especially for ones that you return back.

If anything, this  kind of wrapper may be preferable:
```
#define dracut_asprintf(strp, fmt, __VA_ARGS__) \
{ \
    if (asprintf(strp, fmt, __VA_ARGS__) < 0) { \
        log_error("Out of memory!"); \
        _exit(EXIT_FAILURE); \
    } \
}
```
This would preserve the existing behaviour and shorten `dracut-install.c` at the same time.

sshedi: Yeah, agreed. I had this idea as well. In fact we can use a macro `#define asprintf(...)` and we don't have to do any other changes. But felt this won't scale well, in case if we want to do some other things if asprintf fails, it will keep getting harder to implement everything in a macro.

However, a function gives more control and makes code more readable. Let the maintainers decide what works best for them :)

PR #1567, opened 2021-07-19
feat(lvm-systemd): introducing the lvm-systemd module

teigland:
In this new module

- A new udev rule is run for each lvm device (PV) that
  appears, and the udev rule runs pvscan directly on each
  device.  pvscan reads only the specified device, which
  avoids excessive device scanning when many PVs appear.

- New pvscan options inform the udev rule when an LV is
  complete and ready to be activated, based on the arrival
  of the given device.

- pvscan is run via a shell wrapper that compares complete
  LV names to those named by rd.lvm.lv, so that only LVs
  named by rd.lvm.lv are activated (this avoids unnecessary
  activations that may delay startup.)

- Based on the points above, the udev rule knows when a
  required LV can be successfully activated, and it will
  run a single lvm command to activate each required LV.
  This avoids running many unnecessary (premature) activation
  commands that can delay startup.

- systemd-run is used to run the activation commands because
  the lvm command may take longer to run than udev permits.

- In case rd.lvm.vg is used (in place of, or in addition to
  rd.lvm.lv), adjust the points agove to refer to VG.

- This method is very similar to the new autoactivation
  method that lvm is adopting when run from root.

This pull request changes...

## Checklist

Fixes #


teigland:
The main question I have at this point is how dracut should choose between 90lvm and 90lvm-systemd modules.  Using 90lvm-systemd requires a version of lvm where pvscan includes the option --checkcomplete, e.g. pvscan --help | grep checkcomplete.

johannbg:
@teigland you need a meta module for that ( for an example the 09dbus is such a module ) .
Basically the existing lvm module needs to be renamed, the new meta module taking it's name ( to not break existing setups ) and then the meta module will pull in either the existing module or the new module depending on the output of the pvscan

teigland:
> @teigland you need a meta module for that ( for an example the 09dbus is such a module ) .
> Basically the existing lvm module needs to be renamed, the new meta module taking it's name ( to not break existing setups ) and then the meta module will pull in either the existing module or the new module depending on the output of the pvscan

OK, thanks for the pointer.  I added a couple commits to that branch: renaming 90lvm to 90lvm-initqueue, and a new 90lvm that copies the dbus example to choose between 90lvm-systemd and 90lvm-initqueue.

coiby:
Hi,

This PR fails `make syncheck` which is required by `make rpm` whose failure in turn causes the failiure of kdump tests for dracut #1163,
```
posix syntax check: modules.d/90lvm-systemd/pvscan-udev-initrd.sh
checking for [[: modules.d/90lvm-systemd/pvscan-udev-initrd.sh
checking for echo -n: modules.d/90lvm-systemd/pvscan-udev-initrd.sh
modules.d/90lvm-systemd/pvscan-udev-initrd.sh contains echo -n
``` 

Looking at `modules.d/90lvm-systemd/pvscan-udev-initrd.sh`, using `echo -n` seems to be justified. I don't know why syncheck forbids `echo -n`.

modules.d/90lvm-systemd/pvscan-udev-initrd.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
```
haraldh, line 1:
```suggestion
#!/bin/bash
```

If you want to use arrays.

haraldh (changes requested):
see also
https://github.com/dracutdevs/dracut/blob/master/docs/HACKING.md
https://github.com/dracutdevs/dracut/blob/master/docs/BASH.md

coiby:
> see also
> https://github.com/dracutdevs/dracut/blob/master/docs/HACKING.md
> https://github.com/dracutdevs/dracut/blob/master/docs/BASH.md

@haraldh I couldn't find out why forbidding 'echo -n' usage by checking the above docs or the original commit message,
```
$ git show 24f32511f
commit 24f32511fc8ae616aa967446ce33ab0a7fc27077 (origin/pr/81)
Author: Mike Gilbert <floppym@gentoo.org>
Date:   Tue Jul 14 11:12:16 2015 -0400

    syncheck: Look for 'echo -n' usage in modules

diff --git a/Makefile b/Makefile
index 98a2fbf6..f3a5c07d 100644
--- a/Makefile
+++ b/Makefile
@@ -212,7 +212,8 @@ syncheck:
                 [ "$${i##*/}" = "module-setup.sh" ] && continue; \
                 read line < "$$i"; [ "$${line#*bash*}" != "$$line" ] && continue; \
                [ $$V ] && echo "posix syntax check: $$i"; bash --posix -n "$$i" ; ret=$$(($$ret+$$?)); \
-               [ $$V ] && echo "checking for [[: $$i"; if grep -Fq '[[ ' "$$i" ; then ret=$$(($$ret+1)); echo "$$i contains [["; fi \
+               [ $$V ] && echo "checking for [[: $$i"; if grep -Fq '[[ ' "$$i" ; then ret=$$(($$ret+1)); echo "$$i contains [["; fi; \
+               [ $$V ] && echo "checking for echo -n: $$i"; if grep -Fq 'echo -n ' "$$i" ; then ret=$$(($$ret+1)); echo "$$i contains echo -n"; fi \
        done;exit $$ret
        @ret=0;for i in *.sh mkinitrd-dracut.sh modules.d/*/*.sh \
                        modules.d/*/module-setup.sh; do \
```

Do you know the rationale?

haraldh:
`echo -n` is not posix compatible. Just use printf.

haraldh:
Your git commit messages also need fixing. See https://github.com/dracutdevs/dracut/blob/master/docs/HACKING.md

coiby:
> `echo -n` is not posix compatible. Just use printf.

It make sense now, thanks!

teigland:
I made the changes above and squashed commits in this branch:
https://github.com/teigland/dracut/tree/lvmsd-4

I don't understand github -- do I create a new pull request from the new branch, or update this pull request somehow?

coiby:
> I made the changes above and squashed commits in this branch:
> https://github.com/teigland/dracut/tree/lvmsd-4
> 
> I don't understand github -- do I create a new pull request from the new branch, or update this pull request somehow?

If you push the changes to teigland/dracut/tree/lvmsd-2, the PR will be updated automatically.

coiby:
This PR still fails packit/rpm-build-fedora-rawhide-x86_64 because RPMs failed to be built. You need to change pkgbuild/dracut.spec as well to include this new module.

johannbg:
@coiby We dont expect contributors to be expert in downstream packaging solution so it's not a requirement that the package builds on Fedora or on other linux distributions so it's more of a nice to have if the contributor updates the spec file but it's entirely optional.

coiby:
> @coiby We dont expect contributors to be expert in downstream packaging solution so it's not a requirement that the package builds on Fedora or on other linux distributions so it's more of a nice to have if the contributor updates the spec file but it's entirely optional.

@johannbg Thanks for the info! In such situation, should I propose a detailed change in the format of patch or will you maintainers take care of this situation when merging the PR?

johannbg:
@coiby We usually follow up with a PR if the contributor does not update the spec file but feel free to do so once this has been merged, it's irrelevant who does it, it currently just needs to be done.

The only reason the spec file exist here upstream was for @haraldh convenience but given there seem to be others poking the components maintainership downstream nilli vanilli, we might just stop shipping it altogether in the near future ( it's existence acts as an hindrance to contribution to the project ) and just let those downstream maintainers poking it there, do their due diligence and maintain the spec file there as well.
 
From a downstream packaging standpoint it should be cleaned up anyway since basically what should be done is package dracut into a core component and sub-package each module with proper dependency on the component it uses and relies upon or the relevant dracut module should be a sub-package of the component it's for ( in this case lvm, but the lvm module should still be maintained here ) and that sub-package ( lvm-dracut ) depend on dracut but it varies how distributions seem to be handling this.

coiby:
> @coiby We usually follow up with a PR if the contributor does not update the spec file but feel free to do so once this has been merged, it's irrelevant who does it, it currently just needs to be done.
> 
> The only reason the spec file exist here upstream was for @haraldh convenience but given there seem to be others poking the components maintainership downstream nilli vanilli, we might just stop shipping it altogether in the near future ( it's existence acts as an hindrance to contribution to the project ) and just let those downstream maintainers poking it there, do their due diligence and maintain the spec file there as well.
> 
> From a downstream packaging standpoint it should be cleaned up anyway since basically what should be done is package dracut into a core component and sub-package each module with proper dependency on the component it uses and relies upon or the relevant dracut module should be a sub-package of the component it's for ( in this case lvm, but the lvm module should still be maintained here ) and that sub-package ( lvm-dracut ) depend on dracut but it varies how distributions seem to be handling this.

@johannbg Thanks for the detailed explanation! Since packit need an RPM spec file and is only for Fedora, will it be dropped in the near future as well?

johannbg:
@coiby well yes since it would serve no purpose at that point other than maybe be of assistance of the kdump upstream + we just added packit for experimental purpose and beside it being limited to a single distribution thus *only* beneficial to the IBM/RH ecosystem ( which immediately causes exclusion not inclusion from the project ),  the main purpose of that experiment was to use it to save time, not to increase it and as it turns out using packit costs more time than it frees up.

If a packit kit test fails then it costs 

a) The contributors time.
b) Our time looking into why the tests failed
c) The packit team time ( if the failure is unrelated to the contributed change )
d) RH infrastructure time ( looking into the backend failure )

And as it turns out the most failure that happen with packit are due to RH/Fedora infra making some changes or somebody screwing up rawhide for one reason or another and when that happens it triggers atleast an half an hour of all the people involved free or paid time so using packit literally is not worth it, even to IBM/RH.

teigland:
updated the branch to address a couple automated check warnings about shell quoting

teigland:
Anyone have any suggestions about what else can be done here?

PR #1569, opened 2021-07-23
fix(base): do not quote $initargs for switch_root

floppym:
We want word splitting to occur so that the arguments are passed
separately, and we don't end up passing an empty string if no arguments
are specified.

Bug: https://bugs.gentoo.org/803548
Fixes: 2fabaaa62dcfd31e593ca45e1374e55adae29d6b

### DIFF
--- a/modules.d/99base/init.sh
+++ b/modules.d/99base/init.sh
@@ -387,7 +387,8 @@ if [ -f /etc/capsdrop ]; then
         }
 else
     unset RD_DEBUG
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

floppym:
The testing failures seem unrelated to this change.

johannbg:
@floppym we have a few required to pass tests, which your PR passed, the rest of the tests are more like nice to have.

haraldh (approved):

PR #1570, opened 2021-07-27
fix(network-manager): ensure safe content of /tmp/dhclient."$ifname".dhcpopts

thom311:
NetworkManager leaves state files behind in `"/run/NetworkManager/devices"`.
These files are in keyfile format (glib's GKeyFile API).

Other users want to parse the dhcpopts file, for example anaconda ([1]).

We need to write the dhcpopts file in a format that is defined and easy
to handle from a shell script. As already previously, this format is
a bash script that sets certain variables. That means, to load the file,
the user could execute it as bash script. But this is dangerous, as the
file contains potentially untrusted data from the network.
Optimally, users still don't trust the .dhcpopts file to be safe for
executing! It would be far better if they too try to parse the file
instead of executing it. That is not trivial however.

Anyway, make sure we properly quote and handle the content so that it also
executing is safe. Note only that, we need to handle the two layers of
backslash escaping.

In this case we parse DHCP options from the state file. They are themselves
backslash escaped UTF-8 strings (C escape sequences), which then are stored
via keyfile API. The properly parse them, we would first load the file with
GKeyFile (which undoes one level of backslash escaping) and then
use g_str_compress() (to undo the second level). Mimic that with bash.

[1] https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html
[2] https://github.com/rhinstaller/anaconda/blob/b3411d6780aa0d76ee1e81a38710ec05a2d1978b/dracut/fetch-kickstart-net.sh#L30

Signed-off-by: Thomas Haller <thaller@redhat.com>

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 type source_hook > /dev/null 2>&1 || . /lib/dracut-lib.sh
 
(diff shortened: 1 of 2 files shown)


thom311:
CC @rvykydal @bengal

johannbg:
@thom311 you will need to rewrite your PR in a non bash dependent manner...

modules.d/35network-manager/nm-run.sh
```diff
@@ -24,11 +24,49 @@ if [ -s /run/NetworkManager/initrd/hostname ]; then
...
+kf_unescape() {
+    # Another layer of unescaping. While values in GKeyFile format
+    # are backslash escaped, the original strings are backslash escaped
+    # too. This will undo the second layer of escaping.
+    printf "%b" "$1"
+}
+
+printf_q() {
+    # There is no `printf %q` in POSIX shell.
+    bash -c 'printf "%q" "$@"' printf "$@"
```
haraldh, line 51: If you need bash, then just change the interpreter of this file to `#!/bin/bash` and let `module-setup.sh` require the `bash` module. Calling `bash` is nonsense :)

haraldh (changes requested):

modules.d/35network-manager/nm-run.sh
thom311: I had that in version1.

docs/HACKING.md file says `Hooks MUST be POSIX compliant`.

haraldh: Well, not if the module requires the `bash` module.

haraldh: We have several exceptions there.

thom311: OK.  The HACKING file is the not clear to me.

Because it makes it sound as if the problem is that this script (`nm-run.sh` is a hook, right?) gets sourced by another script, and it's not clear that networkmaanger module depending on bash will make that other script run under bash.

I pushed a fixup. If this is the right way, I'd readd more bashism (e.g.`local` is really nice).

bengal:
LGTM

johannbg:
> LGTM

The NM module is still missing the requirement on the bash module in other words he has not resolved the requested change. He's just decided to install bash for whatever reason...

haraldh:
'shfmt -s' returned error 1 finding the following formatting issues:

```
----------
--- modules.d/35network-manager/nm-run.sh.orig
+++ modules.d/35network-manager/nm-run.sh
@@ -54,7 +54,7 @@
 }
 
 dhcpopts_create() {
-    kf_parse root-path   new_root_path   < "$1"
+    kf_parse root-path new_root_path < "$1"
     kf_parse next-server new_next_server < "$1"
 }
 
----------
```

You can reformat the above files to meet shfmt's requirements by typing:

  shfmt -s -w filename

haraldh (changes requested):
I know... it doesn't look as good as before... 


'shfmt -s' returned error 1 finding the following formatting issues:

```
----------
--- modules.d/35network-manager/nm-run.sh.orig
+++ modules.d/35network-manager/nm-run.sh
@@ -54,7 +54,7 @@
 }
 
 dhcpopts_create() {
-    kf_parse root-path   new_root_path   < "$1"
+    kf_parse root-path new_root_path < "$1"
     kf_parse next-server new_next_server < "$1"
 }
 
----------
```

You can reformat the above files to meet shfmt's requirements by typing:

  shfmt -s -w filename

thom311:
ah right. Sorry, I wasn't aware of shfmt (but it's documented in the HACKING file).

changed.

haraldh (approved):

johannbg (approved):

PR #1586, opened 2021-08-16
fix(usrmount): do not empty _dev variable

puleglot:
Currently $_dev is always overridden with the value returned by
label_uuid_to_dev(). This results in an empty value if $_dev is a
device path. Fix this by calling label_uuid_to_dev() conditionally.

Bug: https://bugs.gentoo.org/807971

## Checklist

Fixes d3532978de04c78f53664dad7b37705a49a7ee54


johannbg (approved):
LGTM

haraldh (approved):

PR #1594, opened 2021-08-31
fix(install): extend hwcaps library handling to libraries under glibc-hwcaps

dtardon:
## Changes
This ensures that we install base libraries for alternative library implementations under [glibc-hwcaps subdirectory](https://sourceware.org/pipermail/libc-alpha/2020-June/115250.html).

## Checklist


johannbg (approved):
LGTM

lnykryn (approved):
Looks good to me and we did some internal testing and it fixes our issues.

PR #1597, opened 2021-09-08
fix(dracut.sh): remove wrong $ in loop sequence

aafeijoo-suse:
This pull request removes a lonely $ placed in a loop sequence, which does not reference any variable.

## Checklist


johannbg (approved):
LGTM

haraldh (approved):

PR #1598, opened 2021-09-09
Call plymouth from the /usr/bin/ to fit the freedesktop UsrMerge feature

qiangzhao:
This pull request changes...

## Changes
   modules.d/99base/dracut-lib.sh

## Checklist
- [    ] I have reviewed and updated any documentation if relevant
        It doesn't need documentation.
- [ x ] I am providing new code and test(s) for it

Fixes #
Hi @haraldh could you please take a look at my request?
I only modify a little place. Call the plymouth from the /usr/bin/ instead of /bin/. It is because the freedesktop UsrMerge feature, modern linux distribution (eg: fedora, openSUSE...) have all migrate "/bin/" and "/sbin/" to "/usr/bin/" "/usr/sbin/". So it's better to update. Thank you so much for your valuable time in checking.

https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/


haraldh:
Well, I initiated the usr merge initially :) All is working well on Fedora, where `/bin` is a symlink to `usr/bin`. What is your problem here?

qiangzhao:
> Well, I initiated the usr merge initially :) All is working well on Fedora, where `/bin` is a symlink to `usr/bin`. What is your problem here?

The problem is there have some linux didn't follow the change of freedesktop yet, which means their /bin/ is a seperate directory. At the same time, plymouth project default install the "plymouth" program into "/usr/bin/". 
So you call it from the "/bin/" does not work in these scenarios,  but the absolute path is most safety.
Thank you!

qiangzhao:
> Well, I initiated the usr merge initially :) All is working well on Fedora, where `/bin` is a symlink to `usr/bin`. What is your problem here?

The problem is there have some linux didn't follow the change of freedesktop yet, which means their /bin/ is a seperate directory. At the same time, plymouth project default install the "plymouth" program into "/usr/bin/". 
So you call it from the "/bin/" does not work in these scenarios,  but the absolute path is most safety.
Thank you!

qiangzhao:
Thank you Harald for the approval.

johannbg:
The safest path to use is /bin since distributions that either have conducted the usr merge or in midst of doing should have a symlink to /usr/bin and those that dont are use /bin. 
According to plymouths own README, plymouth ships with two binaries which are placed under /sbin/plymouthd and /bin/plymouth and it's own upstream configuration file matches that aswell which means upstream plymouth is not making the asumption that distribution that uses it have completed the /usr move so this PR is a NACK from me since things should just be left as is...

qiangzhao:
> The safest path to use is /bin since distributions that either have conducted the usr merge or in midst of doing should have a symlink to /usr/bin and those that dont are use /bin.
> According to plymouths own README, plymouth ships with two binaries which are placed under /sbin/plymouthd and /bin/plymouth and it's own upstream configuration file matches that aswell which means upstream plymouth is not making the asumption that distribution that uses it have completed the /usr move so this PR is a NACK from me since things should just be left as is...

Hi @johannbg :
Thank you for the detailed analysis. 
So I double checked the plymouth's sources,  I found it use $bindir and $sbindir to indicate the install position. and these variable are defined in the system(eg: rpm env). the README didn't update this info. and mainstream operating Linux release all install plymouth into "/usr/bin", (ubuntu, fedora, opensuse, debian, arch...), this is a trend. 
As you said, "the usr merge or in midst of doing should have a symlink to /usr/bin and those that dont are use /bin", I agree the effect has no big difference normally. But I also notice this is an error handler, and "/bin" is a links to absolute path, which means it also link to system's /usr/bin/, not point to "/oldroot/usr/bin/plymouth", in this way, it is exactly the same as the last operation.
Hope you can take my opinion into consideration, Thank you.

qiangzhao:
@johannbg :

qiangzhao:
Hi @johannbg : I am sorry to bother you, based on the previous comment, Could you approve this request? Thank you!

johannbg:
@qiangzhao As I previously stated this PR is a NACK from me so that would be a no + You will need to fix your commit message before this can be merge regardless whether I ACK or NACK this...

PR #1600, opened 2021-09-10
fix(dracut.sh): require root to run hostonly mode

aafeijoo-suse:
The base module install function (modules.d/99base/module-setup.sh) cannot read /etc/shadow in hostonly mode if the user is not root ([line here](https://github.com/dracutdevs/dracut/blob/master/modules.d/99base/module-setup.sh#L38)), but dracut allows to create the initramfs anyway. This causes that it is not possible to log in to the emergency shell with the newly created initramfs.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1079,6 +1079,10 @@ fi
 
 [[ $hostonly == yes ]] && hostonly="-h"
 [[ $hostonly != "-h" ]] && unset hostonly
+if [[ $hostonly ]] && [[ $EUID != 0 ]]; then
+    printf "%s\n" "dracut: hostonly mode must be run as root! Please use 'sudo'." >&2
+    exit 1
+fi
 
 case $hostonly_mode in
     '')


johannbg:
@aafeijoo-suse in which use cases is the initramfs generated while not having root privileges to begin with?

aafeijoo-suse:
The dracut script is located within a `bin` directory (not `sbin`), with `a+x` permissions, it allows to create the initramfs image in any place (not just /boot), and also the code has several conditions that check if the user is root... I'm starting to use dracut and have a habit of calling a command with sudo only if it complains, which is why I came across this issue last week. But, anyway, it's not a big deal.

ddiss:
My preference be to just get an error when attempting to access unpermitted paths. Although I don't have strong feelings about it.

> in which use cases is the initramfs generated while not having root privileges to begin with?

I use non-root dracut every day for generating quick `hostonly=no` kernel test images for qemu. Please don't neglect this (IMO valid) use case :-)

PR #1605, opened 2021-09-17
allow to disable running nm-initrd-generator when rd.nm-initrd-generator=0

coiby:
With `rd.nm-initrd-generator=0` kexec-tools can copy the network configurations files in
/etc/NetworkManager/system-connections/ or /etc/sysconfig/network-scripts/ifcfg
to the kdump initramfs directly. Thus it would be much easier for kexec-tools
to set up network to dump vmcore to a remote fs.
## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-manager/nm-lib.sh
+++ b/modules.d/35network-manager/nm-lib.sh
@@ -4,14 +4,16 @@ type getcmdline > /dev/null 2>&1 || . /lib/dracut-lib.sh
 
 nm_generate_connections() {
     rm -f /run/NetworkManager/system-connections/*
-    if [ -x /usr/libexec/nm-initrd-generator ]; then
-        # shellcheck disable=SC2046
-        /usr/libexec/nm-initrd-generator -- $(getcmdline)
-    elif [ -x /usr/lib/nm-initrd-generator ]; then
-        # shellcheck disable=SC2046
-        /usr/lib/nm-initrd-generator -- $(getcmdline)
-    else
-        derror "nm-initrd-generator not found"
+    if getargbool 1 rd.nm-initrd-generator; then
+        if [ -x /usr/libexec/nm-initrd-generator ]; then
+            # shellcheck disable=SC2046
+            /usr/libexec/nm-initrd-generator -- $(getcmdline)
+        elif [ -x /usr/lib/nm-initrd-generator ]; then
+            # shellcheck disable=SC2046
+            /usr/lib/nm-initrd-generator -- $(getcmdline)
+        else
+            derror "nm-initrd-generator not found"
+        fi
     fi
 
     if getargbool 0 rd.neednet; then


johannbg:
@coiby has this change been acked by NM upstream?

coiby:
> @coiby has this change been acked by NM upstream?

@johannbg Thanks for raising this question. I didn't know this 35network-manager module is actually contributed by NM developers. I'll ask if NM developers can ack this change first.

johannbg:
@coiby yup close collaboration. @bengal can you have a look at this PR?

bengal:
Can you please explain better what you are trying to achieve? If you don't want the generator to run, can't you just use a command line without `ip=` or any other network-related argument?

Since the new option only affects nm-initrd-generator, maybe it should be parsed directly by the generator itself (meaning: the generator will create no connections when it finds the `rd.nm-initrd-generator=0` argument). The generator is part of the NM repository. 

What do you think?

coiby:
> @coiby yup close collaboration. @bengal can you have a look at this PR?

Thanks for introducing the right contact to me!

coiby:
> Can you please explain better what you are trying to achieve? If you don't want the generator to run, can't you just use a command line without `ip=` or any other network-related argument?

Currently, kexec-tools parses ifcfg-* or .nmconnection to build up dracut cmdline parameter like `ip=` to set up kdump network which is tedious and error-prone. A better way is to directly copy the needed connection profiles to initrd. But in practice, 
 at minimum rd.neednet is still needed to make dracut start NM properly. Unfortunately,  with rd.neednet, nm-initrd-generator would generate a connection profile. So I need explicitly  to tell dracut to stop calling nm-initrd-generator.
 
> 
> Since the new option only affects nm-initrd-generator, maybe it should be parsed directly by the generator itself (meaning: the generator will create no connections when it finds the `rd.nm-initrd-generator=0` argument). The generator is part of the NM repository.
> 
> What do you think?

If it's better to implement it nm-initrd-generator, I would be glad to do it in nm-initrd-generator.

bengal:
> Currently, kexec-tools parses ifcfg-* or .nmconnection to build up dracut cmdline parameter like `ip=` to set up kdump network which is tedious and error-prone. A better way is to directly copy the needed connection profiles to initrd. But in practice,
> at minimum rd.neednet is still needed to make dracut start NM properly. Unfortunately, with rd.neednet, nm-initrd-generator would generate a connection profile. So I need explicitly to tell dracut to stop calling nm-initrd-generator.

Got it now, makes sense.

> If it's better to implement it nm-initrd-generator, I would be glad to do it in nm-initrd-generator.

I think there are two possibilities:

- have a NM-specific argument (as `rd.nm-initrd-generator=0` or `rd.nm.generate=0`  or whatever). In that case it should be implemented in the generator itself;

- have a generic argument that works also with other network modules (legacy, wicked, systemd-resolved) and tells the module to only configure network based on the configuration from the initrd, without parsing the kernel command line.

In general I think it's better to avoid that each network module supports different incompatible arguments. But I would not be opposed to a NM-specific argument. 

What do others think? @johannbg , @haraldh

coiby:
> I think there are two possibilities:
> 
> * have a NM-specific argument (as `rd.nm-initrd-generator=0` or `rd.nm.generate=0`  or whatever). In that case it should be implemented in the generator itself;
> * have a generic argument that works also with other network modules (legacy, wicked, systemd-resolved) and tells the module to only configure network based on the configuration from the initrd, without parsing the kernel command line.

@johannbg also gave the same suggestion for the old #1469.  I am not in favor of this idea because of the following reasons,

- Although for NM and wicked, to implement this switch is very easy, much effort is needed for legacy network which doesn't have a network configruation files generator like nm and wicked. And for systemd-networkd, I failed to configure it to bring up the network - connection and to find the switch to turn on/off network configruation files generation.
- The name "legacy" implies support for legacy network could be dropped.
- Even I implement this feature for other network managers, some effort is needed by an application to copy the relevant network configurations files to initrd and make it sure works as expected. So if no application needs it, it could be a waste of effort. Btw, [kexec-tools](https://src.fedoraproject.org/fork/coiby/rpms/kexec-tools/commits/direct_nm) has almost completed the work to set up kdump network by copying connection profiles to initrd directly [1].

> 
> In general I think it's better to avoid that each network module supports different incompatible arguments. But I would not be opposed to a NM-specific argument.
> 
> What do others think? @johannbg , @haraldh

haraldh:
In this case nm-initrd-generator should understand the flag itself, IMHO. No need for dracut to carry a application specific option and documentation for it.

PR #1606, opened 2021-09-17
fix(multipath): drop ExecStop= setting from service unit

lucab:
This removes the `ExecStop=` field from multipath.service.
Some CI runs do rarely encounter a failure related to this
service in initrd, which seems to be stemming from a socket
I/O race between the client and the server on shutdown.
It looks like the client (`multipathd shutdown`) can lose the race,
hit an I/O error, and cause the whole unit to fail (even if the server
managed to shutdown properly already).

Notably, the upstream unit does not have such stop command
as the daemon can already perform a graceful exit through
its signal handler.

As such, this commit re-aligns the two units by dropping the
custom stop-command, thus avoiding any of the existing
races.

Refs:
 * https://github.com/coreos/fedora-coreos-tracker/issues/803
 * https://github.com/opensvc/multipath-tools/blob/0.8.7/multipathd/multipathd.service

### DIFF
--- a/modules.d/90multipath/multipathd.service
+++ b/modules.d/90multipath/multipathd.service
@@ -14,11 +14,11 @@ ConditionKernelCommandLine=!rd_NO_MULTIPATH
 ConditionKernelCommandLine=!multipath=off
 
 [Service]
-Type=simple
+Type=notify
+NotifyAccess=main
 ExecStartPre=-/sbin/modprobe dm-multipath
 ExecStart=/sbin/multipathd -s -d
 ExecReload=/sbin/multipathd reconfigure
-ExecStop=/sbin/multipathd shutdown
 
 [Install]
 WantedBy=sysinit.target


jlebon:
/cc @bmarzins

Looks like upstream used to have `ExecStop=` as well until https://github.com/opensvc/multipath-tools/commit/521513abdb78dc5721866f600141f94a994ad866, where it moved to `Type=notify`. Seems like we should do that here too then?

On an related note, this might actually allow us to simplify a bunch of gnarly things in FCOS/RHCOS too, assuming multipathd notifies `READY=1` only when paths are up.

benmarz:
/cc @mwilck 

> On an related note, this might actually allow us to simplify a bunch of gnarly things in FCOS/RHCOS too, assuming multipathd notifies `READY=1` only when paths are up.

It depends on what you mean by "when paths are up". multipathd makes no guarantee about the state of the path devices when it issues READY=1 after starting up.  It does only issue READY=1 after discovering all the paths currently existing on the system, and creating or updating the multipath devices based on those paths. It does not wait for uevents for any newly created devices, however. So udev may not have processed the new device when multipathd notifies READY=1.

lucab:
@jlebon done, more alignment with upstream unit is indeed good. Though I believe the startup notification does not work the way you are hoping.

jlebon:
@lucab, just to sanity-check, does the same [parallel test](https://github.com/coreos/fedora-coreos-tracker/issues/803#issuecomment-921678376) which lead you here now pass?

jlebon:
> @lucab, just to sanity-check, does the same [parallel test](https://github.com/coreos/fedora-coreos-tracker/issues/803#issuecomment-921678376) which lead you here now pass?

Discussed this OOB with Luca. The parallel test was done in the real root by artificially calling `multipathd shutdown`, which is what's hitting the errors. With this, we're removing the call to `multipathd shutdown` entirely, and just relying on `multipathd` handling systemd's signal correctly.

mwilck:
Am I understanding correctly that the test that you were making this PR for was a broken artificial test?
If yes, it seems that this PR is trying to fix an invalid error. I fail to see how it can come to pass that `multipathd shutdown` called by systemd as `ExecStop` would fail to connect to the socket. 

One thing that can happen is that `multipathd shutdown` times out because multipathd is too busy to service the client request. In the initrd, this is very unlikely but not totally impossible. Using SIGTERM comes down to the same series of shutdown actions as far as multipathd is concerned and provides better asynchronous behavior, so it's actually preferrable over the shutdown command, even though the latter feels more "graceful" in the first place.

Wrt the "notify" service type, @bmarzins is right that the READY notification has nothing to do with any multipath devices being available. If we did that, it might happen that `multipathd.service` would *never* reach READY state. Actually, it's rather the other way around: multipathd signals READY state rather late, when it has finished probing all devices it finds during startup, which may be a minute later after the daemon is actually up and running. According to **systemd.service(5)**, a service should send this notification "when it has finished starting up", which is rather vague. One could argue that this state is already reached before device scanning. IIRC, a side effect of the current behavior is that systemd will not shutdown a service that hasn't reached READY state, so that multipathd shutdown may be delayed if the probing takes a long time.

lucab:
> Am I understanding correctly that the test that you were making this PR for was a broken artificial test?
> If yes, it seems that this PR is trying to fix an invalid error. I fail to see how it can come to pass that multipathd shutdown called by systemd as ExecStop would fail to connect to the socket.

* Yes, the manual test was an artificial setup just to confirm the real existence of a theoretical race. Confirming it that way was much faster than trying to catch it in initrd, but applies to the same client/server setup.
* No, this is not trying to fix an invalid error. Our CI is hitting this race in real bootups within initrd, see https://github.com/coreos/fedora-coreos-tracker/issues/803#issuecomment-891145457.
* No, I don't think there is any failure when connecting to the socket. The `-EIO` is coming from a `recv(2)` [here](https://github.com/opensvc/multipath-tools/blob/0.8.7/multipathd/uxclnt.c#L115), possibly because the server shutdown was quicker than client logic.

mwilck:
>    * No, I don't think there is any failure when connecting to the socket. It looks like there is an `-EIO` is coming from the `recv(2)` [here](https://github.com/opensvc/multipath-tools/blob/0.8.7/multipathd/uxclnt.c#L115), possibly because the server shutdown was quicker than client logic.

Right, this can happen. Thanks for pointing it out. EIO is returned by the low-level receiving code if it encounters an underrun. @bmarzins: I need to double-check this wrt my latest patch series, which might actually aggravate the issue.

The question is: if the daemon receives a "shutdown" client command, should it 1. exit ASAP, or 2. make sure to send the response before triggering the shutdown? multipathd currently does 1.), and I think that isn't totally wrong. OTOH, sending "ok" shouldn't cause an extensive delay, and this is a socket protocol after all, so the client rightfully expects a response.

lucab:
@mwilck assuming you still want to more or less stick close to systemd `ExecStop` behavior, the daemon has several degrees of freedom. The client instead is more constrained, as it is supposed to synchronously wait till proper daemon shutdown. In the end, daemon design likely ends up driven by client requirements.

I'd suggest to move further protocol rework discussion to its own ticket on `multipath-tools` though. This PR is independent from that, and it is a quick change to wholly avoid ending up in the racing conditions.

dustymabe:
> As for this PR; I'm fine with it.

anyone available to LGTM and merge?

johannbg (changes requested):
Change it 
from
`90multipath: drop ExecStop= setting from service unit`
to
`fix(multipath): drop ExecStop= setting from service unit`

And at somepoint and in a seperated PR the 
`ExecStartPre=-/sbin/modprobe dm-multipath` 
should probably be dropped from that type unit file as well.
Anyway's you need to fix the commit msg before we can merge this...

lucab:
@johannbg amended, PTAL.

johannbg (approved):

PR #1609, opened 2021-09-20
fix(dracut.sh): check if kernel has zstd support

tpgxyz:
This pull request changes... in compression logic

## Checklist

Fixes commit https://github.com/dracutdevs/dracut/commit/591118c56da2bfcea060e3b7671bc87b23c0e44a

In my opinion something went wrong in PR review process:
- why introduce external check for zstd in-kernel support
- why zstd support in kernel must be checked now? Why nobody bother for gz/bzip2/lz4/others by following this logic? Does zstd is some kind of special one ?

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2378,13 +2378,12 @@ case $compress in
         ;;
     zstd)
         compress="$DRACUT_COMPRESS_ZSTD -15 -q -T0"
+        if ! check_kernel_config CONFIG_RD_ZSTD; then
+            dwarn "dracut: kernel has no zstd support compiled in, skipping image compression."
+            compress="cat"
+        fi
         ;;
 esac
-
-if [[ $compress == $DRACUT_COMPRESS_ZSTD* ]] && ! check_kernel_config CONFIG_RD_ZSTD; then
-    dwarn "dracut: kernel has no zstd support compiled in."
-    compress="cat"
-fi
 
 if ! (
     umask 077


aafeijoo-suse:
@tpgxyz I don't agree with this change, because it would only perform the check if `$compress` has no parameters, to set the defaults.

> why zstd support in kernel must be checked now? Why nobody bother for gz/bzip2/lz4/others by following this logic? Does zstd is some kind of special one ?

I think you are right and it would be nice to check if the kernel is configured with the compression option selected... but it's out of the scope of this request.

tpgxyz:
@aafeijoo-suse if nobody set "default" compression then it means:
1. lousy maitenace/packaging
2. done o purpose

ddiss (approved):
This looks fine to me. I agree that it makes sense to cover other compression types, but they can be done separately.

johannbg (changes requested):
A commit msg cannot be both a feature and a fix so can you change it 
from
`feat(dracut.sh): fix for check if kernel has zstd support`
to
`fix(dracut.sh): check if kernel has zstd support`

Otherwise LGTM

tpgxyz:
@johannbg thanks for the review, i applied your requested changes.

johannbg (approved):

haraldh:
Thanks for your contribution.
I think I am in favor of https://github.com/dracutdevs/dracut/pull/1646 ... Please review that one!

tpgxyz:
@haraldh Thanks, i already posted my 3 cents there.

PR #1611, opened 2021-09-24
fix(network-manager): disable tty output if the console is not usable

bengal:
The network-manager module also writes logs to the console, so that it's easier to debug network-related boot issues. If systemd can't open the console, the service fails and network doesn't get configured.

Add a check to disable tty output when the console is not present or not usable.

https://github.com/coreos/fedora-coreos-tracker/issues/943

## Checklist


johannbg:
Hmm this seems somewhat backwards.
I expect/would think that the NM module by default should not be started in debugging mode thus outputting things to the console unless either rd.debug is found on the kernel commandline and or some nm.debug kernel commandline parameter is in use so perhaps a better fix might be to include two unit files, one that starts normall and another that starts if debug options are enabled as in  the "standard" one would have `ConditionKernelCommandLine=!FOO.NM.DEBUG` and starts NM without debug being enabled while the debug one would have `ConditionKernelCommandLine=FOO.NM.DEBUG` in them.

bengal:
> I expect/would think that the NM module by default should not be started in debugging mode thus outputting things to the console unless either rd.debug is found on the kernel commandline and or some nm.debug kernel commandline parameter is in use

Good point, there should be no output on console unless the command line contains `rd.debug`. I've added a commit to change that.

> a better fix might be to include two unit files, one that starts normally and another that starts if debug options are enabled

I'm not sure about this; having two units means that every change must be duplicated in both files and it's easy to get them out of sync by mistake. How about changing only the needed options through a drop-in file instead?

dustymabe:
@johannbg - can we merge this?

johannbg:
@dustymabe I've ack this but it needs second review from @haraldh before it can be merged.

dustymabe:
Thanks for the explanation.

PR #1614, opened 2021-10-03
fix(01fips/module-setup.sh): remove root build requirements.

bsloane1650:
Since /dev/ is mounted as devtmpfs early in the boot process,
the fips modules does not need to statically populate devices in it.
As such, there is no need to run as root when building with the fips modules

This pull request changes...

## Checklist
- [*] I have tested it locally

Fixes #


johannbg:
Closing this PR since it cant be merged due to libcrypts prerequisite on /dev/random /dev/urandom

PR #1625, opened 2021-10-21
fix(install): segfault on popen error

keremet:
Now dracut-install writes message like
`dracut-install: Error 'Cannot allocate memory' initiating pipe stream from 'ldd /bin/sh4 2>&1'`
instead of segmentation fault

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -550,6 +550,10 @@ static int resolve_deps(const char *src)
         ret = 0;
 
         fptr = popen(cmd, "r");
+        if (fptr == NULL) {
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
Change it 
from
`fix(install): fix segfault on popen error`
to
`fix(install): segfault on popen error`

Other than LGTM

keremet:
Done
> You need to fix your commit msg. Change it from `fix(install): fix segfault on popen error` to `fix(install): segfault on popen error`
> 
> Other than LGTM

johannbg (approved):

PR #1626, opened 2021-10-22
fix(dracut-functions.sh): get block device driver if in a virtual subsystem

aafeijoo-suse:
**Preconditions**
- Host-only mode.
- Symlinks of all block devices needed to boot the system pointing to virtual subsystems.

**Issue**
The system does not boot because dracut does not install the kernel module of the block device that contains the root filesystem.

**Issue description**
The `get_dev_module` function uses `udevadm info -a` to get the corresponding kernel modules of a /sys/class/\*/\* or /dev/\* device. This function is called in modules.d/90kernel-modules/module-setup.sh ([line 20](https://github.com/dracutdevs/dracut/blob/master/modules.d/90kernel-modules/module-setup.sh#L20)) to detect if dracut must install block device drivers in host-only mode.
The symlinks in /sys/dev/block/ usually point to "real" devices in /sys/devices/pci*. But, we have come across some NVMe systems where the kernel creates the symlinks in /sys/dev/block/ pointing to "virtual" devices instead.
In this case, udevadm never finds any "driver" attributes following up the chain of parent devices.

System where udevadm can find the driver:

```
# ls -l /dev/block
lrwxrwxrwx 1 root root 10 Oct 18 15:03 259:0 -> ../nvme0n1
lrwxrwxrwx 1 root root 12 Oct 18 15:03 259:1 -> ../nvme0n1p1
# ls -l /sys/dev/block/259*
lrwxrwxrwx 1 root root 0 Oct 18 15:03 /sys/dev/block/259:0 -> ../../devices/pci0000:00/0000:00:0e.0/nvme/nvme0/nvme0n1
lrwxrwxrwx 1 root root 0 Oct 18 15:03 /sys/dev/block/259:1 -> ../../devices/pci0000:00/0000:00:0e.0/nvme/nvme0/nvme0n1/nvme0n1p1
# udevadm info -a /dev/block/259:1 | grep DRIVER
    DRIVER==""
    DRIVERS==""
    DRIVERS==""
    DRIVERS=="nvme"
    DRIVERS==""
```

System where udevadm cannot find the driver:

```
# ls -l /dev/block
total 0
lrwxrwxrwx 1 root root 10 Oct 18 03:23 259:1 -> ../nvme0n1
lrwxrwxrwx 1 root root 12 Oct 18 03:23 259:2 -> ../nvme0n1p1
lrwxrwxrwx 1 root root 12 Oct 18 03:23 259:3 -> ../nvme0n1p2
lrwxrwxrwx 1 root root 12 Oct 18 03:23 259:4 -> ../nvme0n1p3
lrwxrwxrwx 1 root root 12 Oct 18 03:23 259:5 -> ../nvme0n1p4
lrwxrwxrwx 1 root root 12 Oct 18 03:23 259:6 -> ../nvme0n1p5
# ls -l /sys/dev/block/259*
lrwxrwxrwx 1 root root 0 Oct 18 03:23 /sys/dev/block/259:1 -> ../../devices/virtual/nvme-subsystem/nvme-subsys0/nvme0n1
lrwxrwxrwx 1 root root 0 Oct 18 03:23 /sys/dev/block/259:2 -> ../../devices/virtual/nvme-subsystem/nvme-subsys0/nvme0n1/nvme0n1p1
lrwxrwxrwx 1 root root 0 Oct 18 03:23 /sys/dev/block/259:3 -> ../../devices/virtual/nvme-subsystem/nvme-subsys0/nvme0n1/nvme0n1p2
lrwxrwxrwx 1 root root 0 Oct 18 03:23 /sys/dev/block/259:4 -> ../../devices/virtual/nvme-subsystem/nvme-subsys0/nvme0n1/nvme0n1p3
lrwxrwxrwx 1 root root 0 Oct 18 03:23 /sys/dev/block/259:5 -> ../../devices/virtual/nvme-subsystem/nvme-subsys0/nvme0n1/nvme0n1p4
lrwxrwxrwx 1 root root 0 Oct 18 03:23 /sys/dev/block/259:6 -> ../../devices/virtual/nvme-subsystem/nvme-subsys0/nvme0n1/nvme0n1p5
# udevadm info -a /dev/block/259:6 | grep DRIVER
    DRIVER==""
    DRIVERS==""
    DRIVERS==""
```

**Proposed solution**
Follow virtual subsystem symlinks to get the real device, so udevadm can read its driver attribute.

```
# find /sys/devices/virtual/nvme-subsystem/nvme-subsys0 -maxdepth 1 -type l -exec readlink {} \;
../../../../class/nvme-subsystem
../../../pci0000:20/0000:20:01.3/0000:25:00.0/nvme/nvme0
# udevadm info -a /sys/devices/pci0000:20/0000:20:01.3/0000:25:00.0/nvme/nvme0 | grep DRIVER
    DRIVER==""
    DRIVERS=="nvme"
    DRIVERS=="pcieport"
    DRIVERS==""
```

**Comments**
We received this issue report after upgrading from 049 (where `get_dev_module` didn't exist) to 055. For an end user, the system which used to work fine, now does not boot after an upgrade, which is ugly.
We might think that this is something systemd/udev has to solve, but in my opinion it seems that the "attribute walk" option is doing what it's expected, walking along the chain of parent devices, and the fact is the module that is failing and preventing the system from starting is dracut.
This is just a proposed fix, any other solution to this problem will be very welcome!

**Versions**
Kernel: 5.14.0-1-default
Systemd: 249.4-3.1

## Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -945,5 +945,29 @@ block_is_netdevice() {
 
 # get the corresponding kernel modules of a /sys/class/*/* or/dev/* device
 get_dev_module() {
-    udevadm info -a "$1" | sed -n 's/\s*DRIVERS=="\(\S\+\)"/\1/p'
+    local dev_attr_walk
+    local dev_drivers
+    dev_attr_walk=$(udevadm info -a "$1")
+    dev_drivers=$(echo "$dev_attr_walk" | sed -n 's/\s*DRIVERS=="\(\S\+\)"/\1/p')
+    # if no kernel modules found and device is in a virtual subsystem, follow symlinks
+    if [[ -z $dev_drivers && $(udevadm info -q path "$1") == "/devices/virtual"* ]]; then
+        local dev_vkernel
+        local dev_vsubsystem
+        local dev_vpath
+        dev_vkernel=$(echo "$dev_attr_walk" | sed -n 's/\s*KERNELS=="\(\S\+\)"/\1/p' | tail -1)
+        dev_vsubsystem=$(echo "$dev_attr_walk" | sed -n 's/\s*SUBSYSTEMS=="\(\S\+\)"/\1/p' | tail -1)
+        dev_vpath="/sys/devices/virtual/$dev_vsubsystem/$dev_vkernel"
+        if [[ -n $dev_vkernel && -n $dev_vsubsystem && -d $dev_vpath ]]; then
+            local dev_links
+            local dev_link
+            dev_links=$(find "$dev_vpath" -maxdepth 1 -type l ! -name "subsystem" -exec readlink {} \;)
+            for dev_link in $dev_links; do
+                [[ -n $dev_drivers && ${dev_drivers: -1} != $'\n' ]] && dev_drivers+=$'\n'
+                dev_drivers+=$(udevadm info -a "$dev_vpath/$dev_link" \
+                    | sed -n 's/\s*DRIVERS=="\(\S\+\)"/\1/p' \
+                    | grep -v -e pcieport)
+            done
+        fi
+    fi
+    echo "$dev_drivers"
 }


johannbg (approved):
LGTM

aafeijoo-suse:
Added missing newline character between `udevadm info` calls.

haraldh (approved):
This code could make use of bash arrays as noted in https://github.com/dracutdevs/dracut/blob/master/docs/BASH.md to counter filenames with spaces... but I think/hope the kernel will not allow those in sysfs, so we should be safe.

aafeijoo-suse:
@haraldh thanks for your hint.

PR #1630, opened 2021-10-26
feat(mdraid): enable passing environment variables to mdadm

rmetrich:
This feature is useful for testing, e.g. when testing IMSM without having the required hardware.

With the patch, to test IMSM without the hardware, it's enough to add
the following to kernel command line parameters:
~~~
rd.md.env=IMSM_NO_PLATFORM=1
~~~

## Changes

New `rd.md.env` parameter (multi-valued) to pass environment variables to `mdadm` binary.

## Checklist
- [??] I am providing new code and test(s) for it (I don't know how to test this)

### DIFF
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -475,6 +475,10 @@ MD RAID
     only activate the raid sets with the given UUID. This parameter can be
     specified multiple times.
 
+**rd.md.env=**__ENVVAR=VALUE__::
+    set environment variable ENVVAR before calling mdadm. This parameter can be
+    specified multiple times.
+
 DM RAID
 ~~~~~~~
 **rd.dm=0**::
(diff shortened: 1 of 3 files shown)


johannbg:
Not that there is anything wrong with this PR but I'm personally not all to excited on introducing new kernel commandline parameters due to the fact that people's employeers aren't providing the maintainers with the required hardware they are supposed to be working on.

rmetrich:
Well, please @haraldh and @danimo please comment as well.
If you don't want this, then I'm ok to close this, no problem with that.

haraldh:
Phew... difficult one.. I can see that it is useful... and even for disaster recovery.

But you can easily replace the mdadm binary in a test suite with a special test suite dracut module.

Allowing all environment variables to be set for a special binary sets the wrong example for the storage people, because then the whole mess will never change.

haraldh:
I am sorry, but I would reject this.

rmetrich:
Thanks, that's ok, closing!

PR #1636, opened 2021-11-10
fix(resume): check for presence of /sys/power/resume

sharkcz:
On platforms where the kernel is built without suspend/resume support we
see "cat: /sys/power/resume: No such file or directory" message when
creating an initrd image. Check for the presence of /sys/power/resume
first before reading it.

## Checklist
- [N/A] I am providing new code and test(s) for it

### DIFF
--- a/modules.d/95resume/module-setup.sh
+++ b/modules.d/95resume/module-setup.sh
@@ -13,7 +13,7 @@ check() {
     # Only support resume if hibernation is currently on
     # and no swap is mounted on a net device
     [[ $hostonly ]] || [[ $mount_needs ]] && {
-        swap_on_netdevice || [[ "$(cat /sys/power/resume)" == "0:0" ]] && return 255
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
We dont include module numbers in the description field in commit messages since those numbers are subjected to changes as well as your commit message is lacking `:` thus failing the commisery check so change your commit msg  
from
`fix(95resume) - check for presence of /sys/power/resume`
to 
fix(resume):  check for presence of /sys/power/resume

johannbg:
@sharkcz you have an excess whitespace in the description ( after fix(resume): ) in your commit msg. 
You need to fix that so the commisery test passes.

sharkcz:
oh, thanks, hopefully this time it will be OK :-)

johannbg (approved):

haraldh (approved):

PR #1637, opened 2021-11-10
feat(crypt): check if tpm2-tss module is needed in hostonly mode

aafeijoo-suse:
In hostonly mode, include the tpm2-tss module if any encrypted volumes are configured in `/etc/crypttab` to be decrypted using the TPM2 device (`tpm2-device=` option).

More info on the [crypttab man page](https://www.freedesktop.org/software/systemd/man/crypttab.html#tpm2-device=).

## Checklist

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -18,7 +18,14 @@ check() {
 
 # called by dracut
 depends() {
-    echo dm rootfs-block
+    local deps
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

johannbg:
This belongs in the depends() section of the crypt module and probably requires a bit larger scope.

haraldh (approved):

Issue #1635, opened 2021-11-09
label_uuid_to_dev() modifies the UUID causing boot to fail


joshuacov1:
**Describe the bug**
When booting we can specify a UUID or PARTUUID for the root partition to start with. The kernel and udev bring up the device correctly and create the corresponding symlinks using the values provided by the device without changing them. During boot dracut parses the provided UUID to lower case and thus starts an endless loop wating for the divese to appear. The divice is actually mapped correctly by the kernel (which doesn't tweak the UUID) but because we are waiting for a name with lower charachters the expeted device never appers which drops us at the emergency shell leaving the system unbootable. 
This happens especially on nfts/fat filesystems because technically those don't have a UUID but a searial numbers which are used by the linux tools as UUID.

This was introduced by 
 https://github.com/dracutdevs/dracut/commit/d3532978de04c78f53664dad7b37705a49a7ee54

**Distribution used**
Fedora 34/35
This can be verified when writing a fedora livecd to usb key with fat using livecd-iso-to-disk from livecdtools.fc34 and affects all distributions.

I've filed a bug on bugzilla for this but so far nothing has changed:
https://bugzilla.redhat.com/show_bug.cgi?id=2012496

**Dracut version**
055

**To Reproduce**
When we write a fedora live-iso to a USB-key on a fat/fat32/ntfs partition using the livecd-iso-to-disk.sh script, the script uses a UUID if available to identify the root device. On USB-keys with fat there is no real UUID but this is more like a serial number in the form XXXX-XXXX. Thus we end up with the following command line in syslinux.cfg:
 append initrd=initrd.img root=live:UUID=XXXX-XXXX  rd.live.image rw quiet

 The UUID in those cases consists of upper letters only, which is trimmed to lower case by dracut

**Expected behavior**
Don't change the provided UUID

I've provided a simple patch to fix the problem - see attached.
[0001-Dont-change-the-UUID-of-a-device.patch.txt](https://github.com/dracutdevs/dracut/files/7507857/0001-Dont-change-the-UUID-of-a-device.patch.txt)

joshuacov1:
This is now fixed with commit https://github.com/dracutdevs/dracut/commit/4e858741087a5cfea891bd2c1fd51ea9b830aeaf so I'll close it.

PR #1638, opened 2021-11-11
fix(base): do not change the provided UUID

joshuacov1:
This was introduced by
https://github.com/dracutdevs/dracut/commit/d3532978de04c78f53664dad7b37705a49a7ee54
Bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=2012496

This pull request changes...

## Changes
During boot dracut parses the provided UUID to lower case and thus starts an endless loop wating for the devise to appear. The device is actually mapped correctly by the kernel (which doesn't tweak the UUID) but because we are waiting for a name with lower charachters the expeted device never appers which drops us at the emergency shell leaving the system unbootable.
This happens especially on nfts/fat filesystems because technically those don't have a UUID but searial numbers which are used by the linux tools as UUID.

## Checklist

Fixes #
https://github.com/dracutdevs/dracut/issues/1635

### DIFF
--- a/dracut-initramfs-restore.sh
+++ b/dracut-initramfs-restore.sh
@@ -6,11 +6,6 @@ set -e
 [ -e /run/initramfs/bin/sh ] && exit 0
 [ -e /run/initramfs/.need_shutdown ] || exit 0
 
-# SIGTERM signal is received upon forced shutdown: ignore the signal
-# We want to remain alive to be able to trap unpacking errors to avoid
-# switching root to an incompletely unpacked initramfs
-trap 'echo "Received SIGTERM signal, ignoring!" >&2' TERM
-
 KERNEL_VERSION="$(uname -r)"
 
 [[ $dracutbasedir ]] || dracutbasedir=/usr/lib/dracut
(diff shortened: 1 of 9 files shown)


joshuacov1:
Any thoughts on this?
It actually breaks all boot processes when the squash.img (from live iso) is on vfat/ntfs partitions...

haraldh (approved):

johannbg (changes requested):
Can you change your commit message 
from
`fix(dracut-lib.sh): do not change the provided UUID`
to
`fix(base): do not change the provided UUID`

joshuacov1:
Done

> Can you change your commit message from `fix(dracut-lib.sh): do not change the provided UUID` to `fix(base): do not change the provided UUID`

johannbg:
@joshuacov1 your commit msg still has `fix(dracut-lib.sh): do not change the provided UUID`

joshuacov1:
It should be ok now.

I fetched master in the meantime and this updated the branch, but the last commit is the one to merge. 

> @joshuacov1 your commit msg still has `fix(dracut-lib.sh): do not change the provided UUID`

johannbg:
@joshuacov1 ? As you can see the test is still failing and now you need to fix your commit message and squash all the new ones you just introduced with these actions into a single commit ...

johannbg:
@joshuacov1 look at all the files that you are now changing, this cant be merged like this. It's probably best that you file a new PR at this point. One with the correct commit msg and only your changes and we close this PR.

joshuacov1:
I'm sorry for the mess!

This is the new request:
https://github.com/dracutdevs/dracut/pull/1650

> @joshuacov1 look at all the files that you are now changing, this cant be merged like this. It's probably best that you file a new PR at this point. One with the correct commit msg and only your changes and we close this PR.

PR #1641, opened 2021-11-12
feat(fido2): introducing the fido2 module

aafeijoo-suse:
From systemd v248, the systemd-cryptsetup component allows to unlock encrypted storage using FIDO2 security tokens.

This module adds to the initramfs image the dependencies required to unlock an encrypted root filesystem, which are basically the [libfido2](https://github.com/Yubico/libfido2) library and its missing dependencies.
It also needs the `60-fido-id.rules` and the `fido_id` program added by the systemd-udevd module.

More info on the [systemd-cryptenroll](https://www.freedesktop.org/software/systemd/man/systemd-cryptenroll.html#--fido2-device=PATH) and [crypttab](https://www.freedesktop.org/software/systemd/man/crypttab.html#fido2-device=) man pages.

## Checklist

### DIFF
--- a/modules.d/91fido2/module-setup.sh
+++ b/modules.d/91fido2/module-setup.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+}
+
+# Module dependency requirements.
+depends() {
+    # This module has external dependency on other module(s).
+    echo systemd-udevd
+    # Return 0 to include the dependent module(s) in the initramfs.
+    return 0
+}
+
+# Install the required file(s) and directories for the module in the initramfs.
+install() {
+    # Install required libraries.
+    _arch=${DRACUT_ARCH:-$(uname -m)}
+    inst_libdir_file \
+        {"tls/$_arch/",tls/,"$_arch/",}"libfido2.so.*" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libcryptsetup.so.*" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libcbor.so.*" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libhidapi-hidraw.so.*"
+}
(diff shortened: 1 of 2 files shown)


johannbg:
Hmm it seems that you have started to pushing something that should arguably be handled by downstream package managers into the code which makes me having to ask what's your end game with this approach? 
Not sure if I like the trend to "auto include" modules based on entries in files in /etc  parsing files costs resources and will increase the generation time.

aafeijoo-suse:
I'm not sure what the best approach is, I just wanted to know the upstream view on this feature, which is something that some of us need to implement.

The inclusion of this module in hostonly mode based on the content of the crypttab file prevents an end user having to do it manually, otherwise if the root file system is encrypted it won't be decrypted from the initramfs using the FIDO key.

johannbg:
There is nothing wrong with the module except for
`The inclusion of this module in hostonly mode based on the content of the crypttab file prevents an end user having to do it manually` which belongs in it's hostonly section of `install()`.
It's up to downstream to properly package modules and make sure those modules have correct dependency's on other components that the module depends on and install the relevant packages at install time if the end users requires it. ( For example dracut is not being correctly packaged in Fedora from this point of view )
If the installer does not take care of that at install time then it's expected that the user installs the relevant component he needs at a later date and when doing so the downstream packaging system would install the module, so the users should never hit the use case you are trying to solve right. 
Turning dracut into somekind of package manager ( beyond dracut making sure it's module inner dependency get pulled in the initramfs at initrd generation time ) or have it start parsing files in the masses just adds cost and complexity, all for the sake of circumventing which arguably is a downstream packaging problem. 
On top of that in a modernise distribution then /etc can be nonexistant or read only so that has to be taken into account as well.

aafeijoo-suse:
Thank you for your extensive explanation.

I've added this hostonly check just because I thought I had to, after reading the [module-setup.sh: check()](https://mirrors.edge.kernel.org/pub/linux/utils/boot/dracut/dracut.html#_module_setup_sh_check) section of the man page.

> If the $hostonly variable is set, then the module check() function should be in "hostonly" mode, which means, that the check() should only return 0, if the module is really needed to boot this specific host.

And in this specific case there is no prompt to enter a fallback password; if the FIDO key is not present the system does not boot.

Anyway, your reasoning makes sense to me. So, if I remove the hostonly check, would it be ok for you?

johannbg:
@aafeijoo-suse /etc/crypttab belongs to the crypt module so if you want to parse /etc/crypttab to include other module(s) depending on entry's in /etc/crypttab, that logic belongs in that module not in the fido2 ( or the tpm2-tss ) module. 

That said we are missing a systemd-cryptsetup module and in the long run we have to split out logic in existing init dependent modules ( think systemd vs non systemd based systems ) into their own modules and turn the relevant exisiting modules into meta modules.

johannbg:
@aafeijoo-suse can you move your hostonly check to the crypt module in a seperated PR. Then we can merged this module.

johannbg (approved):

haraldh (approved):

PR #1644, opened 2021-11-19
fix: make the base module dependency explicit

LaszloGombos:
When dracut is called with the "--modules" argument, the base
module needs to be specificed by the caller. Dracut already has
a well functioning dependency mamanegemtn system that could be used
to not make base module special.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/99base/init.sh
+++ b/modules.d/99base/init.sh
@@ -297,8 +297,6 @@ done
 udevadm control --exit
 udevadm info --cleanup-db
 
-debug_off # Turn off debugging for this section
-
 # unexport some vars
 export_n root rflags fstype netroot NEWROOT
 unset CMDLINE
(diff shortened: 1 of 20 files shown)


modules.d/99base/init.sh
```diff
@@ -297,8 +297,6 @@ done
 udevadm control --exit
 udevadm info --cleanup-db
 
-debug_off # Turn off debugging for this section
```
haraldh, line 300: this clutters logs with `rd.debug` a lot...

modules.d/99base/init.sh
```diff
@@ -338,7 +336,6 @@ if getarg init= > /dev/null; then
     done
     unset CLINE
 else
-    debug_off # Turn off debugging for this section
```
haraldh, line 341: this clutters logs with rd.debug a lot...

LaszloGombos:
Recreated as PR 1651, sorry for the churn.

PR #1645, opened 2021-11-19
fix(s390_rules): make collect optional

tblume:
/usr/lib/udev/collect has been removed from systemd v246, so make it optional in the udev rules

This pull request changes...

dependency of s390 udev rules on collect

## Checklist

Fixes #

### DIFF
--- a/modules.d/95dasd_rules/module-setup.sh
+++ b/modules.d/95dasd_rules/module-setup.sh
@@ -26,7 +26,6 @@ check() {
     local found=0
     local bdev
     [ "$_arch" = "s390" -o "$_arch" = "s390x" ] || return 1
-    require_binaries /usr/lib/udev/collect || return 1
 
     [[ $hostonly ]] || [[ $mount_needs ]] && {
         for bdev in /sys/block/*; do
@@ -50,7 +49,6 @@ depends() {
 
 # called by dracut
 install() {
-    inst_multiple /usr/lib/udev/collect
     inst_hook cmdline 30 "$moddir/parse-dasd.sh"
     if [[ $hostonly_cmdline == "yes" ]]; then
         local _dasd
--- a/modules.d/95qeth_rules/module-setup.sh
+++ b/modules.d/95qeth_rules/module-setup.sh
@@ -5,7 +5,6 @@ check() {
     local _arch=${DRACUT_ARCH:-$(uname -m)}
     local _online=0
     [ "$_arch" = "s390" -o "$_arch" = "s390x" ] || return 1
-    require_binaries /usr/lib/udev/collect || return 1
     dracut_module_included network || return 1
 
     [[ $hostonly ]] && {
@@ -56,5 +55,4 @@ install() {
         [ -n "$id" ] && inst_rules_qeth "$id"
     done
 
-    inst_simple /usr/lib/udev/collect
 }
--- a/modules.d/95zfcp_rules/module-setup.sh
+++ b/modules.d/95zfcp_rules/module-setup.sh
@@ -45,7 +45,6 @@ check() {
     local _arch=${DRACUT_ARCH:-$(uname -m)}
     local _ccw
     [ "$_arch" = "s390" -o "$_arch" = "s390x" ] || return 1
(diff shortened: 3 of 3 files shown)


johannbg (changes requested):
@tblume can you update your PR to just drop it since no one should be using it anymore.

tblume:
> @tblume can you update your PR to just drop it since no one should be using it anymore.

@johannbg if we just drop collect we will bind dracut to the new chzdev generated udev rules that work without collect.
That means in turn we will make dracut somewhat dependent on newer s390-tools versions.
We might break older systems with rules that still use collect (if they have a systemd version that still ships collect).
Not sure how many systems would be concerned though.
Should we take that risk?

modules.d/95dasd_rules/module-setup.sh
```diff
@@ -50,7 +49,7 @@ depends() {
 
 # called by dracut
 install() {
-    inst_multiple /usr/lib/udev/collect
+    [[ -f /usr/lib/udev/collect ]] && inst_multiple /usr/lib/udev/collect
```
dtardon, line 52: The check isn't needed. Just call `inst_multiple` with `-o`.

modules.d/95qeth_rules/module-setup.sh
```diff
@@ -56,5 +55,5 @@ install() {
         [ -n "$id" ] && inst_rules_qeth "$id"
     done
 
-    inst_simple /usr/lib/udev/collect
+    [[ -f /usr/lib/udev/collect ]] && inst_simple /usr/lib/udev/collect
```
dtardon, line 58: Same here.

modules.d/95zfcp_rules/module-setup.sh
```diff
@@ -66,7 +65,7 @@ depends() {
 
 # called by dracut
 install() {
-    inst_multiple /usr/lib/udev/collect
+    [[ -f /usr/lib/udev/collect ]] && inst_multiple /usr/lib/udev/collect
```
dtardon, line 68: Same here.

johannbg:
Well from my pov distribitions should not be shipping mix and match of versions of components that make up the core/base OS layer so distributions should never hit the scenario you describe.

If distributions choose to play that game, they should also pay the price for that madness by carrying patches for that downstream. 

Upstreams should not be working around that for them at the cost of code bloat and performance.

tblume:
> Well from my pov distribitions should not be shipping mix and match of versions of components that make up the core/base OS layer so distributions should never hit the scenario you describe.
> 
> If distributions choose to play that game, they should also pay the price for that madness by carrying patches for that downstream.
> 
> Upstreams should not be working around that for them at the cost of code bloat and performance.

ok, thanks, just wanted to make sure we do the right ting.

tblume:
We should remove that too:

-->
 grep -r collect modules.d/
modules.d/95dasd_rules/parse-dasd.sh:ACTION=="add", SUBSYSTEM=="ccw", KERNEL=="$ccw", IMPORT{program}="collect $ccw %k ${ccw} $_drv"
modules.d/95dasd_rules/parse-dasd.sh:ACTION=="add", SUBSYSTEM=="drivers", KERNEL=="$_drv", IMPORT{program}="collect $ccw %k ${ccw} $_drv"
modules.d/95zfcp_rules/parse-zfcp.sh:ACTION=="add", SUBSYSTEM=="ccw", KERNEL=="$ccw", IMPORT{program}="collect $ccw %k ${ccw} zfcp"
modules.d/95zfcp_rules/parse-zfcp.sh:ACTION=="add", SUBSYSTEM=="drivers", KERNEL=="zfcp", IMPORT{program}="collect $ccw %k ${ccw} zfcp"
--<

haraldh (approved):

johannbg (approved):
@tblume are you going to do a seperated PR's for the other removals that you pointed out?

tblume:
> @tblume are you going to do a seperated PR's for the other removals that you pointed out?

Yes, will do, just awaiting feedback from the original provider of the rules.

PR #1646, opened 2021-11-19
Improve logic for selecting initramfs compression

mwilck:
## Changes

Handle missing compression tools, or otherwise lack of support
for the configured compression method, more gracefully. It's normally not necessary to bail out if e.g. *xz* is not installed.

## Checklist
- [x ] I have reviewed and updated any documentation if relevant

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1294,23 +1294,6 @@ if [[ $no_kernel != yes ]] && [[ -d $srcmods ]]; then
         else
             dwarn "$srcmods/modules.dep is missing. Did you run depmod?"
         fi
-    elif ! (command -v gzip &> /dev/null && command -v xz &> /dev/null); then
-        read -r _mod < "$srcmods"/modules.dep
-        _mod=${_mod%%:*}
-        if [[ -f $srcmods/"$_mod" ]]; then
-            # Check, if kernel modules are compressed, and if we can uncompress them
-            case "$_mod" in
-                *.ko.gz) kcompress=gzip ;;
-                *.ko.xz) kcompress=xz ;;
-                *.ko.zst) kcompress=zstd ;;
-            esac
-            if [[ $kcompress ]]; then
-                if ! command -v "$kcompress" &> /dev/null; then
-                    dfatal "Kernel modules are compressed with $kcompress, but $kcompress is not available."
-                    exit 1
-                fi
-            fi
-        fi
     fi
 fi
 
--- a/dracut.sh
+++ b/dracut.sh
@@ -2350,6 +2350,13 @@ if [[ $create_early_cpio == yes ]]; then
     fi
 fi
 
+if [[ $compress && $compress != cat ]]; then
+    if ! command -v "${compress%% *}" &> /dev/null; then
+        derror "dracut: cannot execute compression command '$compress', falling back to default"
+        compress=
+    fi
+fi
+
 if ! [[ $compress ]]; then
     # check all known compressors, if none specified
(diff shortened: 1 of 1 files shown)


mwilck:
@aafeijoo-suse, @tblume, please have a look.

johannbg:
@mwilck explain to me how removing the logic that that detects if kernel modules can be uncompressed to begin with as well as If no compressor is specified tries to find a suitable one ( which arguably is required for quite obvious reasons ), results in "improved logic for selecting initramfs compression" ?

mwilck:
> @mwilck explain to me how removing the logic that that detects if kernel modules can be uncompressed to begin with 

Kernel modules need to be uncompressed by the kernel. Dracut just needs to copy them into the initramfs image. As I noted in the commit message, the `$kcompress` command is never used anywhere in the dracut code. I don't know if dracut used to uncompress modules in the past, but the current code does not. *kmod* doesn't need these tools either, it just links to the respective libraries.

> as well as If no compressor is specified tries to find a suitable one ( which arguably is required for quite obvious reasons ), results in "improved logic for selecting initramfs compression" ?

I haven't removed this logic. The only difference is the case when configuration specifies a compression command that can't be executed. This case is treated like an empty `compress` configuration with my patches. dracut will try to find a suitable command in that case, too, and fall back to the best found, or no compression (`cat`) as a last resort.
This can happen if the distibution ships a default configuration file setting a compression method (e.g. zstd) that deviates from dracut's built-in default, but the user chooses not to install the distro's default tool. Currently this would result in a failure to build the initramfs, and an unbootable system in the worst case. With these patches, that can't happen any more (unless `cat` is missing).

dracut.sh
```diff
@@ -2325,7 +2325,9 @@ if ! [[ $compress ]]; then
         break
     done
     if [[ $compress == cat ]]; then
-        printf "%s\n" "dracut: no compression tool available. Initramfs image is going to be big." >&2
+        dwarn "dracut: no compression tool available. Initramfs image is going to be big." >&2
```
haraldh, line 2347: `dwarn` does not have to be redirected with ` >&2`

dracut.sh
```diff
@@ -2327,6 +2338,7 @@ fi
 if ! [[ $compress ]]; then
     # check all known compressors, if none specified
     for i in $DRACUT_COMPRESS_PIGZ $DRACUT_COMPRESS_GZIP $DRACUT_COMPRESS_LZ4 $DRACUT_COMPRESS_LZOP $DRACUT_COMPRESS_ZSTD $DRACUT_COMPRESS_LZMA $DRACUT_COMPRESS_XZ $DRACUT_COMPRESS_LBZIP2 $DRACUT_COMPRESS_BZIP2 $DRACUT_COMPRESS_CAT; do
+        [[ $i != "$DRACUT_COMPRESS_ZSTD" || $DRACUT_KERNEL_RD_ZSTD ]] || continue
```
haraldh, line 2341: or just unset DRACUT_COMPRESS_ZSTD if `check_kernel_config CONFIG_RD_ZSTD` fails

mwilck: > or just unset DRACUT_COMPRESS_ZSTD if `check_kernel_config CONFIG_RD_ZSTD` fails

I can do that if you want, but tbh it feels a little dirty to me.

haraldh: true

haraldh (approved):

haraldh:
Hmm, has to be rebased... sorry.

johannbg (approved):

johannbg:
@mwilck Once you have rebased we can merge the PR

mwilck:
Rebase done.

PR #1649, opened 2021-11-19
fix(dmsquash-live): do not install systemd files when systemd is not enabled

LaszloGombos:
Systemd is an optional module for the dmsquash-live module. This scenario
is properly handled for other modules (for example livenet module) but not for
dmsquash-live module.

## Checklist

Fixes #


haraldh (approved):

johannbg (changes requested):
Can you fix your commit message so the PR passes the commisery check. 
Change the message from
`fix(dmsquash-live): Do not install systemd dependent files when systemd is not enabled`
to
`fix(dmsquash-live): do not install systemd files when systemd is not enabled`

LaszloGombos:
> Can you fix your commit message so the PR passes the commisery check. Change the message from `fix(dmsquash-live): Do not install systemd dependent files when systemd is not enabled` to `fix(dmsquash-live): do not install systemd files when systemd is not enabled`

Thanks, resolved.

johannbg (approved):

PR #1651, opened 2021-11-23
fix: make the base module dependency explicit

LaszloGombos:
When dracut is called with the "--modules" argument, the base
module needs to be specificed by the caller. Dracut already has
a well functioning dependency management system that could be used
to not make base module special.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/95rootfs-block/module-setup.sh
+++ b/modules.d/95rootfs-block/module-setup.sh
@@ -7,7 +7,7 @@ check() {
 
 # called by dracut
 depends() {
-    echo fs-lib
+    echo base fs-lib
 }
 
 cmdline_journal() {


johannbg:
@LaszloGombos if an modules does not require the base module they should not depend on it. That is the correct behaviour of a well functioning dependency management system...

LaszloGombos:
> @LaszloGombos if a module does not require the base module they should not depend on it. That is the correct behaviour of a well functioning dependency management system...

Agreed. 

I made an attempt to only add the base module as a dependency if a module actually dependent on the base module, but I might have missed a few modules. Do you a good example for a module that this PR make it dependent on base module but in fact is not dependent on base ?

johannbg:
@LaszloGombos meta modules ( like network ) are good examples. Those should not have dependencies ( that they currently do just highlights that there is something there that needs to be looked at ), just like if you encounter an if "systemd" statement, that's a clear indicator that we need turn that module into a meta module in the long run and or that module will be obsoleted, which means that the base module becomes either an obsolete or a meta module ( in the long run ).

LaszloGombos:
> @LaszloGombos meta modules ( like network ) are good examples. 

Thanks for sharing some of the concepts for the different groups of modules (meta vs regular module) @johannbg .

"dracut.sh --modules rootfs-block" currently leads to non-working initramfs image that does not have dracut-lib.sh included as at no point the base image is marked as a dependency that would install dracut-lib.sh into initramfs. Solving this particular case has been my main motivation.

Solving this one case might improve the dependency management for most practical scenarios, while the module organization will continue to improve and evolve.

johannbg (approved):

Conan-Kudo (approved):

PR #1652, opened 2021-11-23
Add errors and warnings when network interface does not exist

pvalena:
End with error, or show a warning when nonexistent device is specified for network setup like `ip=10.12.8.12::10.12.255.254:255.255.0.0:xk12:eth0:off`.

## Changes

I've added the error only for `write-ifcfg.sh`, as I think no such setup should be written. For the rest I added only warnings, as I'm not sure `$dev` is not set to something else, or it can be a non-critical error.

## Checklist

Fixes #1712424

## Additional
I'm new to this - this is my first issue, so I'm not entirely sure how to test this. The test suite does not work for me, locally, but I'll try to create a reproducer.

### DIFF
--- a/modules.d/35network-legacy/ifup.sh
+++ b/modules.d/35network-legacy/ifup.sh
@@ -446,7 +446,11 @@ for p in $(getargs ip=); do
 
     # If this option isn't directed at our interface, skip it
     if [ -n "$dev" ]; then
-        [ "$dev" != "$netif" ] && continue
+        if [ "$dev" != "$netif" ]; then
+            [ ! -e "/sys/class/net/$dev" ] \
+                && warn "Network interface '$dev' does not exist!"
+            continue
+        fi
     else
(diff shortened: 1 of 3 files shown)


johannbg (changes requested):
Can you change your commit message
from
`Add errors and warnings when network interface does not exist`
to
`fix(network): add errors and warnings when network interface does not exist`

can you also run `shfmt -s -w modules.d/35network-legacy/ifup.sh` to fix the formatting errors.

pvalena:
> Can you change your commit message from `Add errors and warnings when network interface does not exist` to `fix(network): add errors and warnings when network interface does not exist`
> 
> can you also run `shfmt -s -w modules.d/35network-legacy/ifup.sh` to fix the formatting errors.

Done.

johannbg (approved):

haraldh (approved):

pvalena:
Thanks!

PR #1655, opened 2021-11-25
fix(img-lib): install rmdir

Henrik66:
rmdir is used in some modules but not properly installed (e.g. img-lib module).
It would make it easier to ration about the dependencies and would save
a bit of intramfs space to just use rm instead.

This pull request changes...

## Checklist
- [ X ] I have tested it locally

Fixes #

### DIFF
--- a/modules.d/99img-lib/module-setup.sh
+++ b/modules.d/99img-lib/module-setup.sh
@@ -14,7 +14,7 @@ depends() {
 
 # called by dracut
 install() {
-    inst_multiple tar gzip dd echo tr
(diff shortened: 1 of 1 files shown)


johannbg:
Sorry I dont follow if rmdir is being used in some modules but not properly installed then make a PR that fixes just that. 
Dont make a PR that replaces one command with another one and removes any belt-and-suspenders that the previous command provided in the process...

johannbg:
I need to close this PR and re-opening it since all the tests got stuck "Waiting for status to be reported"

johannbg (approved):
LGTM

PR #1657, opened 2021-11-28
fix(systemd-resolved): add systemd-resolve user

### LINKED ISSUE #1656: systemd-resolve user missing
**Describe the bug**
When adding the `systemd-resolved` module, the service will not start correctly as the `systemd-resolve` user is missing. This user is used by the service since [v235](https://github.com/systemd/systemd/commit/635f3df5dc540892cca0b9d4ba42ab7215277f7c).

**Distribution used**
Arch rolling

**Dracut version**
dracut 055

**Init system**
systemd 249.7-1-arch

**To Reproduce**
Call dracut with `--add systemd-resolved` and reboot

**Expected behavior**
The service should start

**Additional context**

Cornelicorn:
This pull request adds the `systemd-resolve` user to initramfs when using the `systemd-resolved` module.

## Checklist

Fixes #1656

### DIFF
--- a/modules.d/01systemd-resolved/module-setup.sh
+++ b/modules.d/01systemd-resolved/module-setup.sh
@@ -42,6 +42,15 @@ install() {
         "$systemdsystemunitdir/systemd-resolved.service.d/*.conf" \
         resolvectl
 
+    # install systemd-resolve user/group
+    {
+        grep '^systemd-resolve:' "$dracutsysrootdir"/etc/passwd 2> /dev/null
+    } >> "$initdir/etc/passwd"
+
+    {
+        grep '^systemd-resolve:' "$dracutsysrootdir"/etc/group 2> /dev/null
+    } >> "$initdir/etc/group"
+
     # Enable systemd type unit(s)
     $SYSTEMCTL -q --root "$initdir" enable systemd-resolved.service
 


johannbg:
@Cornelicorn hmm we dont parse/grep and or otherwise use users from /etc/passwd anymore anyways something must have changed in systemd upstream ( like systemd finally splitting all their users into seperated sysusers files, which then requires fixes in more places since probably more users are missing ), all the systemd users should currently be provided via systemd-sysusers module, which this module is dependent upon.  

If systemd upstream has done the right thing ( by splitting it's users into seperated sysuser files ) then we need to try to included that sd-resolve sysuser file from the underlying OS probably with a fallback on our own include sd-resolve sysuser file if the sysuser file is not found.

That fix requires adding `"$sysusers"/foo.conf` in the `install()` section of the module and `"$sysusersconfdir"/foo-user.conf` in the hostonly section of the `install()` section and we probably should be adding that sysuser file to the module as well.

Cornelicorn:
Superseeded by #1658

johannbg:
Somewhat permature closure since we still need to include the new systemd-resolve.conf file otherwise this will just break again with the next sd release

Cornelicorn:
> Somewhat permature closure since we still need to include the new systemd-resolve.conf file otherwise this will just break again with the next sd release

Ah right, forgot about the change in systemd sysuser configs. But maybe this should be tackled in a new PR together with all other sysuser configs.

johannbg:
Agreed

PR #1658, opened 2021-11-29
fix(systemd-sysusers): override systemd-sysusers.service

### LINKED ISSUE #1656: systemd-resolve user missing

Cornelicorn:
Fixes a regression with systemd not running units with ConditionNeedsUpdate set in initrds

## Checklist

Fixes #1656

### DIFF
--- a/modules.d/01systemd-sysusers/module-setup.sh
+++ b/modules.d/01systemd-sysusers/module-setup.sh
@@ -24,6 +24,8 @@ depends() {
 # Install the required file(s) and directories for the module in the initramfs.
 install() {
 
+    inst_simple "$moddir/sysusers-dracut.conf" "$systemdsystemunitdir/systemd-sysusers.service.d/sysusers-dracut.conf"
+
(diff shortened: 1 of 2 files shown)


johannbg (changes requested):
In your commit message you need to seperate the subject and body with an empty line as well as fix the spelling error sytemd-sysusers.service should be systemd-sysusers.service

johannbg (approved):

Conan-Kudo (approved):

PR #1659, opened 2021-11-30
fix(systemd-sysusers): use split systemd sysuser configs

Cornelicorn:
Fixes the regression introduced by https://github.com/systemd/systemd/pull/20056
As discussed in #1656 and #1657

## Changes
Include the new sysusers.d files to allow compatability with future systemd releases. Otherwise the users would be missing inside the initrd.

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-coredump/module-setup.sh
+++ b/modules.d/01systemd-coredump/module-setup.sh
@@ -37,6 +37,7 @@ install() {
         "$systemdsystemunitdir"/systemd-coredump.socket \
         "$systemdsystemunitdir"/systemd-coredump@.service \
         "$systemdsystemunitdir"/sockets.target.wants/systemd-coredump.socket \
+        "$sysusers"/systemd-coredump.conf \
         coredumpctl
 
     # Install the hosts local user configurations if enabled.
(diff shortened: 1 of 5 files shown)


Cornelicorn:
Pushed a new version without Title case in the commit subject

johannbg (changes requested):
You are missing the vendor path and have misplaced the admin/user path which belongs in the hostonly section and we need to have both as in the `"$sysusers"/foo.conf` which is the vendor path and belongs in the install() section of the module and the `"$sysusersconfdir"/foo.conf` which is the administrator override which belongs in the hostonly portion of the install() section 

So move 
`"$sysusersconfdir"/systemd-coredump.conf \`
`"$sysusersconfdir"/systemd-journal.conf \`
`"$sysusersconfdir"/systemd-network.conf \`
`"$sysusersconfdir"/systemd-resolve.conf \`
`"$sysusersconfdir"/systemd-timesync.conf \` 
to the `if [[ $hostonly ]];` portion of the modules and add the missing vendor files
`"$sysusers"/systemd-coredump.conf \`
`"$sysusers"/systemd-journal.conf \`
`"$sysusers"/systemd-network.conf \`
`"$sysusers"/systemd-resolve.conf \`
`"$sysusers"/systemd-timesync.conf \` 
to the install() section of the module.

johannbg (approved):
LGTM

Conan-Kudo (approved):

PR #1660, opened 2021-12-02
fix(url-lib): improve ca-bundle detection

aafeijoo-suse:
This is an attempt to unlock PR #909 (created via https://github.com/dracutdevs/dracut/pull/909#issuecomment-984607481). As far as I could test, it fixes a current issue.

**Description**

The current detection routine for openssl-based libcurl assumes that
libcurl has its own hardcoded path to the ca-bundle. Fix the
cases where curl is compiled with:
```
  --with-ca-fallback --without-ca-path --without-ca-bundle
```
In this case, we must also grep in OpenSSLs libcrypto.

Other changes:
  - Filter reported but non-existant paths.
  - Strip nul bytes returned by grep.
  - Consider that ca-bundles might use '.pem' instead of '.crt'.

## Checklist

### DIFF
--- a/modules.d/45url-lib/module-setup.sh
+++ b/modules.d/45url-lib/module-setup.sh
@@ -15,10 +15,10 @@ depends() {
 
 # called by dracut
 install() {
-    local _dir _crt _found _lib _nssckbi _p11roots _p11root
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

Conan-Kudo (approved):

PR #1661, opened 2021-12-03
fix(dmsquash-live): overlayfs on a block device root

LaszloGombos:
An example kernel command line option for this configuration
root=/dev/sda1 ro rd.live.image rd.live.overlay.overlayfs=1

The change checks if $FSIMG is empty and adds a symlink from
/run/rootfsbase to /run/initramfs/live and changes the place
of the creation of the required mount point directories.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -210,8 +210,6 @@ do_live_overlay() {
             fi
         fi
         if [ -n "$overlayfs" ]; then
-            mkdir -m 0755 -p /run/overlayfs
-            mkdir -m 0755 -p /run/ovlwork
             if [ -n "$readonly_overlay" ] && ! [ -h /run/overlayfs-r ]; then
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@LaszloGombos What do you want to do this for? It'd be nice to have some rationale in the commit message for this change.

LaszloGombos:
> @LaszloGombos What do you want to do this for? It'd be nice to have some rationale in the commit message for this change.

Thanks for the follow-up.

After further investigation, I do not believe this is a new feature. I do personally run this configuration and noticed the regression when I disabled dracut-systemd module (which is optional for the dmsquash-live module). 

I believe this option has been supported for quite some time with dracut-systemd module enabled. Re-reading the dracut documentation I did not found any reason to believe this particular use would be an unsupported configuration.

This PR's goal is to bring the functionality on par between when  dracut-systemd is enabled and when it is not enabled.

johannbg (approved):
LGTM

Conan-Kudo (approved):

PR #1663, opened 2021-12-06
fix(cpio): write zeros instead of seek for padding and alignment

ddiss:
This is a workaround for GRUB2's Btrfs implementation, which doesn't
correctly handle gaps between extents.

A fix has already been proposed upstream via
https://lists.gnu.org/archive/html/grub-devel/2021-10/msg00206.html

Given that this bug is severe, it makes sense to include this minimal
workaround.

Signed-off-by: David Disseldorp <ddiss@suse.de>

## Checklist

## Fixes

https://bugzilla.opensuse.org/show_bug.cgi?id=1190982


johannbg (approved):
LGTM

Conan-Kudo (approved):
Gross, but makes sense sadly.

PR #1665, opened 2021-12-08
fix(network-manager): skip non-directories in /sys/class/net

### LINKED ISSUE #1604: nm-run.sh: /sys/class/net/bonding_masters/ifindex: Not a directory
Hi,

When trying to set up a network bond using dracut in RHEL8.4, I see the (harmless) error:

> dracut-initqueue: cat: /sys/class/net/bonding_masters/ifindex: Not a directory

By inspection, it looks like this will still happen with the latest version of dracut.
I think the following change would improve this:

```
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -60,6 +60,7 @@ dhcpopts_create() {
 }
 
 for _i in /sys/class/net/*; do
+    [ -d "$_i" ] || continue
     state="/run/NetworkManager/devices/$(cat "$_i"/ifindex)"
     grep -q '^connection-uuid=' "$state" 2> /dev/null || continue
     ifname="${_i##*/}"
```

Thanks!

glennmorris:
There can be files in this directory, eg "bonding_masters" if a
network bond is in use.

Fixes #1604

### DIFF
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -60,6 +60,7 @@ dhcpopts_create() {
 }
 
 for _i in /sys/class/net/*; do
+    [ -d "$_i" ] || continue
     state="/run/NetworkManager/devices/$(cat "$_i"/ifindex)"
     grep -q '^connection-uuid=' "$state" 2> /dev/null || continue
     ifname="${_i##*/}"


johannbg (approved):
LGTM

johannbg:
@bengal can you review and give an ack/nack from the NM team.

bengal:
LGTM

Conan-Kudo (approved):

PR #1667, opened 2021-12-10
fix(network-wicked): multiple path corrections

aafeijoo-suse:
Since [wicked-0.6.67](https://github.com/openSUSE/wicked/blob/version-0.6.67/wicked.spec.in#L117), its dbus configuration files can be installed in /etc or /usr/share, what was breaking this module in openSUSE Tumbleweed.

Also, check if it's using `libexec` or `lib` instead of displaying always an error.

## Checklist

### DIFF
--- a/modules.d/35network-wicked/module-setup.sh
+++ b/modules.d/35network-wicked/module-setup.sh
@@ -33,15 +33,23 @@ install() {
 
     inst_dir /etc/wicked/extensions
     inst_dir /usr/share/wicked/schema
-    inst_dir /usr/lib/wicked/bin
+    if [ -d /usr/lib/wicked/bin ]; then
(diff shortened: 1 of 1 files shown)


johannbg (approved):
I'm going to ack this and probably will do a follow up changes once this merged which refactors and cleans up this module a bit more and adds a missing hostonly section. It would be nice if you could review and test those changes on Suse when I do.

aafeijoo-suse:
> I'm going to ack this and probably will do a follow up changes once this merged which refactors and cleans up this module a bit more and adds a missing hostonly section. It would be nice if you could review and test those changes on Suse when I do.

Sure, I'll test your changes. We have also found some missing things related to this module.

Conan-Kudo (approved):

PR #1670, opened 2021-12-11
WIP: fix(network-wicked): various fixes

johannbg:
Add missing license header
Add comments to sections
Add missing requirement on binaries
Add missing dependency on the kernel-network-modules
Add missing dbus configuration files
Add missing hostonly section
Remove creation of /etc/wicked/extensions
Refactor install() to use `inst_multiple -o` instead of several instances of `inst_multiple`
Refactor install() to use exisisting variables instead of hard paths

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-wicked/module-setup.sh
+++ b/modules.d/35network-wicked/module-setup.sh
@@ -1,76 +1,155 @@
 #!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
 
-# called by dracut
+# Prerequisite check(s) for module.
 check() {
-    require_binaries wicked || return 1
 
-    # do not add this module by default
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries ip \
+        pppd \
+        wicked \
+        wickedd \
+        wickedd-auto4 \
+        wickedd-dhcp4 \
+        wickedd-dhcp6 \
+        wickedd-nanny \
+        || return 1
+
+    # Do not add this module by default.
     return 255
 }
 
-# called by dracut
+# Module dependency requirements.
 depends() {
-    echo systemd dbus
+
+    # This module has external dependency on other module(s).
+    echo dbus kernel-network-modules systemd
+
+    # Return 0 to include the dependent module(s) in the initramfs.
     return 0
+
 }
 
-# called by dracut
-installkernel() {
-    return 0
-}
-
-# called by dracut
+# Install the required file(s) and directories for the module.
 install() {
-    local -a wicked_units
+
+    # Add systemd overrides for upstream system units, replaces sed
+    # !!! Note to self yet again snippets cant be used, replace with full units !!!
+    # inst_simple "$moddir/wicked.conf" "$systemdsystemunitdir/wicked.service.d/dracut.conf"
+    # inst_simple "$moddir/wickedd-auto4.conf" "$systemdsystemunitdir/wickedd-auto4.service.d/dracut.conf"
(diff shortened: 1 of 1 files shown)


johannbg:
@aafeijoo-suse here's a module based on the content in upstream wicked which adds probably those things that you forgot + bunch of stuff that can be removed. 

So what remains to be done with this afaikt are two things, remove the unessary things I added based on the upstream content and replace the sed section with actual systemd units until upstream comes up with code/units that work both in real root and initrd which we then could just include as is.

So what you can do is point what stuff can be removed and come up with and test the type units for wicked that are tailored to initrd.

aafeijoo-suse:
Thank you @johannbg, this is definitely a good job to start revamping this module. I'll test it and give you some feedback soonish. I'd like to involve here @mtomaschewski, because he can point us out any missing thing.

One type of issues we start receiving come from systems that use wicked to manage the network along with other dracut network modules. E.g.: the ifcfg module generates a wrong ifcfg configuration if the interface is configured to use DHCP via kernel parameter ("ip=eth0:dhcp"), because it checks the existence of a `/tmp/dhclient."$netif".lease` file ([write-ifcfg.sh#L183](https://github.com/dracutdevs/dracut/blob/master/modules.d/45ifcfg/write-ifcfg.sh#L183)), whereas wicked creates a file called `/tmp/leaseinfo."$netif".dhcp.ipv4` instead. In a case like this, I doubt modifying the ifcfg module would be acceptable, because that would lead to update every dracut network module in base of specific wicked settings.

modules.d/35network-wicked/module-setup.sh
```diff
@@ -1,76 +1,155 @@
...
+            "$systemdsystemconfdir/wickedd-dhcp6.service/*.conf" \
+            "$systemdsystemconfdir"/wickedd-nanny.service \
+            "$systemdsystemconfdir/wickedd-nanny.service/*.conf" \
+            "$systemdsystemconfdir/wickedd-pppd@.service" \
+            "$systemdsystemconfdir/wickedd-nanny.service/*.conf"
+    fi
+
+    # TODO replace this section with systemd units
+    #    for unit in "${wicked_units[@]}"; do
+    #        sed -i 's/^After=.*/After=dbus.service/g' "$initdir/$unit"
```
aafeijoo-suse, line 144: AFAIK dependencies cannot be reset in drop-ins, and it seems the `After` must be set to `dbus.service` only (in every wicked service). If you want to avoid the sed, you will need to create custom services and install them, rather than installing those provided by the system.

modules.d/35network-wicked/module-setup.sh
```diff
@@ -1,76 +1,155 @@
...
+        /usr/share/wicked/schema/scripts.xml \
+        /usr/share/wicked/schema/sit.xml \
+        /usr/share/wicked/schema/team.xml \
+        /usr/share/wicked/schema/tuntap.xml \
+        /usr/share/wicked/schema/types.xml \
+        /usr/share/wicked/schema/vlan.xml \
+        /usr/share/wicked/schema/vxlan.xml \
+        /usr/share/wicked/schema/wicked.xml \
+        /usr/share/wicked/schema/wireless.xml \
+        "$dbussystem"/org.opensuse.Network.conf \
```
aafeijoo-suse, line 82: In my previous PR, I didn't use `$dbussystem` because in SUSE it expands to /usr/share/dbus-1/system.d in dracut.sh, but wicked dbus configuration files can be located in /etc/dbus-1/system.d, depending on the SUSE distribution.

PR #1673, opened 2021-12-14
improve lvm module

teigland:
## Changes

This series makes various updates to the lvm module:
- removes support for the unused lvm snapshot management feature
- removes automated editing of deprecated lvm.conf lvmetad settings
- removes an unused udev rule
- replaces the use of --partial with --activationmode degraded
- removes checks for using options that have existed for many years
- uses the filter generated by dracut when lvm.conf has no filter set
- only attempts to activate an LV when the LV is seen on devices

## Checklist

Fixes #

### DIFF
--- a/modules.d/90lvm/64-lvm.rules
+++ b/modules.d/90lvm/64-lvm.rules
@@ -6,6 +6,14 @@
 
 SUBSYSTEM!="block", GOTO="lvm_end"
 ACTION!="add|change", GOTO="lvm_end"
+
+# If the md device is active (indicated by array_state), then set the flag
+# LVM_MD_PV_ACTIVATED=1 indicating that the md device for the PV is ready
+# to be used.  The lvm udev rule running in root will check that this flag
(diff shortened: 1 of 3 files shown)


johannbg:
@teigland your PR is failing the shell linter in quite few places. Have a look at the detail section of the linter check.

teigland:
> @teigland your PR is failing the shell linter in quite few places. Have a look at the detail section of the linter check.

Hi, I'm uncertain about the lint recommendation to use quotes around $scan_args and $LVs here:

LVSLIST=$(lvm lvs $scan_args --noheading -o lv_full_name,segtype $LVS)

When I tried adding quotes around that $scan_args, it produced a command with single quotes around the value, which I couldn't figure out.

So, in my updated branch I'm going to try adding # shellcheck disable=SC2086 prior to that line, since that seems to be how other similar instances are handled.  If there's a better solution please let me know.  Thanks

johannbg (approved):

PR #1674, opened 2021-12-20
fix(kernel-modules): add mailbox drivers for arm

athierry1:
Mailbox drivers might be needed by some of the devices used in the
initrd. For example, on the Raspberry Pi 4, the MMC 'sdhci-iproc' driver
depends on 'bcm2835-mailbox'.

Signed-off-by: Adrien Thierry <athierry@redhat.com>

## Checklist


johannbg (approved):
LGTM

PR #1675, opened 2021-12-23
fix(kernel-modules-extra): Handle zstd module extension

dirkmueller:
The regular expression here is trying to handle various kernel
module compression schemas and was missing the zst extension
which indicates use of zstd.

This pull request changes...

## Checklist

Fixes #


johannbg (changes requested):
You will need to fix your commit msg so it passes the commisery check
change it  from 
`fix(kernel-modules-extra): Handle zstd module extension`
to
`fix(kernel-modules-extra): handle zstd module extension`

Other than that LGTM

johannbg (approved):

PR #1678, opened 2021-12-29
Introducing pkcs11 module

johannbg:
Add a pkcs11 ( library ) module which should ( in theory ) work with systemd + pkcs11. 
If people have the time and hw to test this feel free to do so and provide feedback how that went. 

## Checklist

Fixes #

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -21,11 +21,14 @@ depends() {
     local deps
     deps="dm rootfs-block"
     if [[ $hostonly && -f "$dracutsysrootdir"/etc/crypttab ]]; then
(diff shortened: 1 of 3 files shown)


johannbg:
I could change this to a full blown pcsd module if required/need but something tells me that there might be some race condition involved with doing that...

coiby:
To successfully build the RPM package, this module needs to be added to the `%files` section of pkgbuild/dracut.spec.

johannbg:
@coiby I was aware of that + removing and adding modules to the spec file is optional not a requirement.
 
In otherwords it's more of an nice to have since we ( and other upstreams to ) cant ( and should not ) expect contributors to be experts in downstream packaging solution and their guidelines prior to them contributing to projects and I've have had complaints from people ( in private ) not contributing to dracut due to exactly that. 

Anyways thanks for the pointer I have fixed it now.

PR #1679, opened 2022-01-01
feat(pcsc): introducing the pcsc module

johannbg:
Introducing the pcsc module.

Since I lack the required hw for this it would be nice if someone who has the time and the hw ( like nitrokey ) could try to unlock a LUKS2 partition using X509 certificate to test this and provide feedback how that went.

## Checklist

Fixes #

### DIFF
--- a/modules.d/91pcsc/module-setup.sh
+++ b/modules.d/91pcsc/module-setup.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries pcscd || return 1
+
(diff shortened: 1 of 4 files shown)


johannbg:
@LudovicRousseau if you have the time it would be nice if you could review and provide feedback for the module ( and maybe test it since I lack the required hw to do so )

LudovicRousseau:
The patch looks correct.
I do not use LUKS myself so I do not plan to test it.

I asked for help on the MUSCLE https://muscle.apdu.fr/ mailing list https://lists.infradead.org/pipermail/pcsclite-muscle/2022-January/001227.html

johannbg:
> The patch looks correct. I do not use LUKS myself so I do not plan to test it.

No problem I just noticed that nitrokey has partnered with both Arch and Gentoo's and provided it's developers with nitrokey. There are active contributors from both of these communities here upstream so they might end up helping with testing this. 

Thanks for your reviewing this.

PR #1681, opened 2022-01-05
udev-rules: make sure the hwdb with ifname renaming rules is around

lkundrak:
## Changes

The stable network interface naming relies on hwdb. In one particular
case actually: the Dell "idrac" interface.

The systemd-udevd module actually already includes the hwdb for hostonly
images, but we need it to work for non-hostonly images as well as those
that don't utilize the systemd-udevd module as well.

Let's make udev-rules also pull in the binary hwdb for hostonly builds.
On Fedora 35 this adds 2M compressed, 11M uncompressed. That is quite
some extra fat, but it's consistent with what systemd-udevd does.

For non-hostonly ones, pull in only what we know we need in binary form.
The extra space consumed is negligible (300B uncompressed).

## Checklist
- [ x ] I have reviewed and updated any documentation if relevant
- [ :( ] I am providing new code and test(s) for it

### DIFF
--- a/modules.d/95udev-rules/module-setup.sh
+++ b/modules.d/95udev-rules/module-setup.sh
@@ -97,4 +97,16 @@ install() {
 
     inst_libdir_file "libnss_files*"
 
+    if [[ $hostonly ]]; then
+        inst_multiple -o "${udevdir}"/hwdb.bin
+        inst_multiple -H -o "${udevconfdir}"/hwdb.bin
+    else
+        inst_dir "${udevdir}"/hwdb.d
+        inst_simple "${udevdir}"/hwdb.d/20-net-ifname.hwdb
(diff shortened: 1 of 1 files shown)


johannbg (commented):
Eventually support for systemd in udev-rules will be dropped and it exclusively be used for non systemd based systems while systemd based system will use systemd-udevd instead that said it makes no sense to only include 20-net-ifname.hwdb and include all hwdb's only when the case is hostonly ( we dont add vendor paths to hostonly but I can see why you chose to do that there )  it's all or nothing situation so this PR wont work as is.

Now that does not mean that we cant solve what you are trying to achieve, it just means we need to be smarter how we approach it, which probably will require a bit larger scope to be able to achieve that.

johannbg:
Keeping this one open until it get's cleaned up as a part of a larger scope.

PR #1682, opened 2022-01-06
fix(network-manager): wait for the actual connection

zboszor:
Wait for the actual connection instead of daemon startup.
This fixes dmsquash-live failures.

This pull request changes the ExecStart= line for nm-wait-online-initrd.service, option -s has to be removed from nm-online.
I am using  dmsquash-live combined with network-manager under Yocto 3.4 to remote boot devices.
Since NM pulls up the network interface asynchronously and nm-wait-online-initrd.service (nm-online -s) didn't wait for the interface to actually get an IP address, it made the success or failure for dmsquash-live to pull the live rootfs image timing sensitive.

Opinion needed: can we just remove option "-s" as in the commit or should this module provide two different wait-online services and install one or the other into the initramfs if dmsquash-live is also used?

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-manager/nm-wait-online-initrd.service
+++ b/modules.d/35network-manager/nm-wait-online-initrd.service
@@ -8,7 +8,7 @@ ConditionPathExists=/run/NetworkManager/initrd/neednet
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/nm-online -s -q -t 3600
+ExecStart=/usr/bin/nm-online -q -t 3600
 RemainAfterExit=yes
 
 [Install]


johannbg (dismissed):

johannbg:
I personally would rather avoid us having to fall into the rabbit hole of start shipping files and configuration within modules depending on other modules being installed because it can become quite messy, quite fast.

But maybe trying to avoid that while multiple choices exist in the linux ecosystem is just wishful thinking anyways it's probably best to ship such dependency with the module in question ( dmsquash-live ), then that module would include a type unit if the NM module was being used.

@bengal ping for review and thoughts

zboszor:
> I personally would rather avoid us having to fall into the rabbit hole of start shipping files and configuration within modules depending on other modules being installed because it can become quite messy, quite fast.
> 
> But maybe trying to avoid that while multiple choices exist in the linux ecosystem is just wishful thinking anyways it's probably best to ship such dependency with the module in question ( dmsquash-live ), then that module would include a type unit if the NM module was being used.
> 
> @bengal ping for review and thoughts

Regarding multiple choices: originally there was only dhclient to work with netboot in dracut and it was practically synchronous so there was no timeout failure for downloading the live rootfs and setting up /dev/root.

With NM, getting IP from DHCP became asynchronous but until dhclient was available, I could choose it to generate the netboot initramfs.

Now here's the thing: ISC DHCPd is obsoleted and replaced by ISC Kea in Yocto 3.2, mainly because the last release of the former doesn't build with new BIND releases. For the record, dhclient is part of DHCPd.

Fedora apparently patches either BIND or DHCPd so it can provide both DHCPd and Kea which Yocto doesn't do.

So, I was somewhat forced to use NM in initramfs and this is where I found out the timing sensitivity.

Anyway, I know this change may be detrimental to others, e.g. if there's no network connection, then it prevents the machine from booting. Which is fine if the goal is netboot / PXE boot, but definitely not in other cases.

Probably the solution that would satisfy everyone is that nm-wait-online-initrd.service should use a script in ExecStart= which can deduce from /proc/cmdline whether option -s must be omitted or not.

johannbg:
>So, I was somewhat forced to use NM in initramfs and this is where I found out the timing sensitivity.

I would actually think that embedded community was gravitating to using iwd or networkd so out of curiosity did you try using systemd's own built-in DHCP client in the initramfs? 

Systemd's dhcp client used to be somewhat limited in the past ( it was not up to the dhcp spec in it's early days ) but I think Yu has hashed all those issues out so if you tried using it I'm curios how that went. 

> Probably the solution that would satisfy everyone is that nm-wait-online-initrd.service should use a script in ExecStart= which can deduce from /proc/cmdline whether option -s must be omitted or not.

If the solutions cries for having systemd type units calling a script or escaping to shell for whatever reason, it just indicates that there some engineering failure upstream that needs fixing. 

In otherwords this might be something that needs to be fixed in NM codebase itself ( an upstream issue ) not worked around in dracut or in the type units for NM for it's module.

zboszor:
> > So, I was somewhat forced to use NM in initramfs and this is where I found out the timing sensitivity.
> 
> I would actually think that embedded community was gravitating to using iwd or networkd so out of curiosity did you try using systemd's own built-in DHCP client in the initramfs?

No, because when I had to make the dhclient -> NM switch (more than a year ago), I didn't know that the livenet -> network -> systemd-networkd dependency choice existed. Back then NM worked just fine. Now I have tuned a lot of things around it to suit my PXE booted system so it's quite entrenched. It just started occurring very recently that the  timing sensitivity can be observed. Sometimes the system boots just fine as is but most likely not. Removing the nm-option -s fixed it for good and made the network setup+DHCP synchronous as far as the livenet/dmsquash-live module is concerned.

> 
> Systemd's dhcp client used to be somewhat limited in the past ( it was not up to the dhcp spec in it's early days ) but I think Yu has hashed all those issues out so if you tried using it I'm curios how that went.
> 
> > Probably the solution that would satisfy everyone is that nm-wait-online-initrd.service should use a script in ExecStart= which can deduce from /proc/cmdline whether option -s must be omitted or not.
> 
> If the solutions cries for having systemd type units calling a script or escaping to shell for whatever reason, it just indicates that there some engineering failure upstream that needs fixing.

I think not. The purpose of option -s is clear from "nm-online --help". It waits only for NM startup, i.e. it practically makes sure the local interface exists with a 127.x.x.x IP address but it doesn't wait for any real interface to get its IP address. While it suits many use cases, it doesn't work for the livenet dracut module. The fact that it worked for me for a long time is pure luck.

> 
> In otherwords this might be something that needs to be fixed in NM codebase itself ( an upstream issue ) not worked around in dracut or in the type units for NM for it's module.

johannbg:
@bengal how do you want to proceed with this?

bengal:
Hi, sorry for the delay, I was on vacation. In case I don't reply, you can also ping others from NM team as @lkundrak  or @thaller :)

> The purpose of option -s is clear from "nm-online --help". It waits only for NM startup, i.e. it practically makes sure the local interface exists with a 127.x.x.x IP address but it doesn't wait for any real interface to get its IP address.

That's not correct. `-s` waits that NM has attempted to activate every connection. A connection attempt ends either when the connection is successful (i.e. it gets an address if it needs to) or when it fails (for DHCP, after the timeout specified by `rd.net.timeout.dhcp=`).

If you found that NM didn't wait for an address, that seems a configuration issue or a bug in NM; it would be helpful if you could provide a description of the scenario, what command line is used and possibly attach dracut debug logs.

`nm-online` without `-s` doesn't seem useful in dracut because it only waits for the first connection to be established, ignoring others. If the command line generates multiple connection profiles, they should be all honored.

zboszor:
> > The purpose of option -s is clear from "nm-online --help". It waits only for NM startup, i.e. it practically makes sure the local interface exists with a 127.x.x.x IP address but it doesn't wait for any real interface to get its IP address.
> 
> That's not correct. `-s` waits that NM has attempted to activate every connection.

Not on my side:
```
$ nm-online --help
Usage:
  nm-online [OPTION?]

Waits for NetworkManager to finish activating startup network connections.

Help Options:
  -h, --help                  Show help options

Application Options:
  -q, --quiet                 Don't print anything
  -s, --wait-for-startup      Wait for NetworkManager startup instead of a connection
  -t, --timeout=<timeout>     Time to wait for a connection, in seconds (without the option, default value is 30)
  -x, --exit                  Exit immediately if NetworkManager is not running or connecting
```
So it's the lack of `-s` is what waits for a connection. Also, a comment in `NetworkManager-wait-online.service` under Fedora 35:
```
# `nm-online -s` waits until the point when NetworkManager logs
# "startup complete". That is when startup actions are settled and
# devices and profiles reached a conclusive activated or deactivated
# state. It depends on which profiles are configured to autoconnect and
# also depends on profile settings like ipv4.may-fail/ipv6.may-fail,
# which affect when a profile is considered fully activated.
# Check NetworkManager logs to find out why wait-online takes a certain
# time.
```
So, in order to make `nm-online -s` work as I need, I would also need to put profile configuration into the initramfs. I found the help text earlier and it was easy enough to omit `-s`.

> A connection attempt ends either when the connection is successful (i.e. it gets an address if it needs to) or when it fails (for DHCP, after the timeout specified by `rd.net.timeout.dhcp=`).

... or according to the Fedora comment, marked as **activated or deactivated** when `-s` is present.

> If you found that NM didn't wait for an address, that seems a configuration issue or a bug in NM; it would be helpful if you could provide a description of the scenario, what command line is used and possibly attach dracut debug logs.

Here's the complete pxelinux.cfg/default file: 
```
# PXEBOOT default configuration file
DEFAULT linux
LABEL linux
  SAY Loading system resources...
  LINUX http://192.168.5.80/tftpboot/bzImage-genericx86-64.bin
  INITRD http://192.168.5.80/tftpboot/initrd.bin
  APPEND ro quiet rhgb audit=0 selinux=0 LANG=en.US.UTF-8 rd.shell ip=dhcp root=live:http://192.168.5.80/tftpboot/squashfs.bin
  ONERROR reboot.c3
```
I will get you the rdsos file when I can get around to build a new squashfs live image that adds back `-s`. 

> `nm-online` without `-s` doesn't seem useful in dracut because it only waits for the first connection to be established, ignoring others. If the command line generates multiple connection profiles, they should be all honored.

Which is fine by me, the PXE booted machines only have one Ethernet socket, i.e. only one connection is expected. I am happy to carry this patch in my custom distro layer for Yocto indefinitely if we can't eventually come to a solution that would work for every scenario.

thaller:
Zitat von Zoltán Böszörményi ***@***.***>:

>> > The purpose of option -s is clear from "nm-online --help". It  
>> waits only for NM startup, i.e. it practically makes sure the local  
>> interface exists with a 127.x.x.x IP address but it doesn't wait  
>> for any real interface to get its IP address.
>>
>> That's not correct. `-s` waits that NM has attempted to activate  
>> every connection.
>
> Not on my side:
> ```
> $ nm-online --help
> Usage:
>   nm-online [OPTION?]
>
> Waits for NetworkManager to finish activating startup network connections.
>
> Help Options:
>   -h, --help                  Show help options
>
> Application Options:
>   -q, --quiet                 Don't print anything
>   -s, --wait-for-startup      Wait for NetworkManager startup  
> instead of a connection
>   -t, --timeout=<timeout>     Time to wait for a connection, in  
> seconds (without the option, default value is 30)
>   -x, --exit                  Exit immediately if NetworkManager is  
> not running or connecting
> ```
> So it's the lack of `-s` is what waits for a connection. Also, a  
> comment in `NetworkManager-wait-online.service` under Fedora 35:
> ```
> # `nm-online -s` waits until the point when NetworkManager logs
> # "startup complete". That is when startup actions are settled and
> # devices and profiles reached a conclusive activated or deactivated
> # state. It depends on which profiles are configured to autoconnect and
> # also depends on profile settings like ipv4.may-fail/ipv6.may-fail,
> # which affect when a profile is considered fully activated.
> # Check NetworkManager logs to find out why wait-online takes a certain
> # time.
> ```
> So, in order to make `nm-online -s` work as I need, I would also  
> need to put profile configuration into the initramfs. I found the  
> help text earlier and it was easy enough to omit `-s`.
>
>> A connection attempt ends either when the connection is successful  
>> (i.e. it gets an address if it needs to) or when it fails (for  
>> DHCP, after the timeout specified by `rd.net.timeout.dhcp=`).
>
> ... or according to the Fedora comment, marked as **activated or  
> deactivated** when `-s` is present.
>
>> If you found that NM didn't wait for an address, that seems a  
>> configuration issue or a bug in NM; it would be helpful if you  
>> could provide a description of the scenario, what command line is  
>> used and possibly attach dracut debug logs.
>
> Here's the complete pxelinux.cfg/default file:
> ```
> # PXEBOOT default configuration file
> DEFAULT linux
> LABEL linux
>   SAY Loading system resources...
>   LINUX http://192.168.5.80/tftpboot/bzImage-genericx86-64.bin
>   INITRD http://192.168.5.80/tftpboot/initrd.bin
>   APPEND ro quiet rhgb audit=0 selinux=0 LANG=en.US.UTF-8 rd.shell  
> ip=dhcp root=live:http://192.168.5.80/tftpboot/squashfs.bin
>   ONERROR reboot.c3
> ```
> I will get you the rdsos file when I can get around to build a new  
> squashfs live image that adds back `-s`.
>
>> `nm-online` without `-s` doesn't seem useful in dracut because it  
>> only waits for the first connection to be established, ignoring  
>> others. If the command line generates multiple connection profiles,  
>> they should be all honored.
>
> Which is fine by me, the PXE booted machines only have one Ethernet  
> socket, i.e. only one connection is expected. I am happy to carry  
> this patch in my custom distro layer for Yocto indefinitely if we  
> can't eventually come to a solution that would work for every  
> scenario.
>
> --
> Reply to this email directly or view it on GitHub:
> https://github.com/dracutdevs/dracut/pull/1682#issuecomment-1009617257
> You are receiving this because you were mentioned.
>
> Message ID: ***@***.***>



Excuse it seems like you got the wrong E-mail in the CC. I have  
nothing to do with this I'm just a Student of Chemistry and did some  
lectures on Computer Science.

With best Regards
Tobias Haller

bengal:
> Excuse it seems like you got the wrong E-mail in the CC.

Sorry for the spam, I meant @thom311.

bengal:
> > That's not correct. `-s` waits that NM has attempted to activate every connection.

> Not on my side:
>  ` -s, --wait-for-startup      Wait for NetworkManager startup instead of a connection`

The definition of what "startup" means is given in the man page, and it clearly states that it waits for every connection to finish.

As you said, in your specific scenario you have only one connection profile and thus nm-online without `-s` can possibly work; however it is not suitable in general because there can be multiple profiles.

> Here's the complete pxelinux.cfg/default file:
> `APPEND ... ip=dhcp ...`

If you try to manually inspect how the profile generated for `ip=dhcp` looks like, you should get something like:
```
$ /usr/libexec/nm-initrd-generator --stdout -- ip=dhcp

*** Connection 'default_connection' ***

[connection]
id=Wired Connection
uuid=beb51395-2c54-4ffc-bbc7-b6e2be61698f
type=ethernet
autoconnect-retries=1
multi-connect=3
permissions=

[ethernet]
mac-address-blacklist=

[ipv4]
dhcp-timeout=90
dns-search=
may-fail=false
method=auto

[ipv6]
addr-gen-mode=eui64
dhcp-timeout=90
dns-search=
method=auto
```

Note the `ipv4.may-fail=false` property, which makes NM wait until DHCPv4 completes before considering the connection as activated. When using `-s`, this means that NM will wait until the interface gets an address via DHCPv4 before the "online" target is reached.

If this doesn't happen in your scenario, it would be helpful to have a log at logs with `rd.debug` to understand why.

johannbg:
@zboszor, @bengal, @thom311 since it's quite clear we wont be merging this and it makes little sense ( to me ) to be debuging an issue on a PR, is it not best to move this to upstream NM and create and resolve this as an issue there? ( or we can create an issue for this here and the dialog moved there ) ?

zboszor:
I am sorry for the silence. I haven't been able to re-test my initramfs without this patch + with `rd.debug` but I will find some time to do it and report back. Can we keep this issue open until I do it?

johannbg:
@zboszor sure we will keep it open.

johannbg:
@zboszor it's been three weeks now, how much longer should we keep this open?

zboszor:
> @zboszor it's been three weeks now, how much longer should we keep this open?

You can close this, I was and still am too busy with work.
I will carry this patch in my Yocto layer until I have time to figure out what was going on without it.
Thanks for your patience.

PR #1684, opened 2022-01-09
fix(memstrack): drop bash runtime requirement

ryncsn:
This makes the scripts POSIX compatible and bash is no longer needed.

## Changes

Use nohup instead of disown, and let systemd track the service properly.

## Checklist


johannbg (approved):
LGTM

PR #1685, opened 2022-01-09
feat(dracut.sh): add --aggresive-strip option

ryncsn:
Dracut currently calls `eu-strip` or `strip` with -g, which only strips
out .debug_* sections. symtab and strtab are kept, but are not required
for runtime, and people will rarely need to do binary level debugging
work in initramfs.

So introduce a --aggresive-strip options, try strip out all sections
that are not required for runtime. This can help reduce the binary size
by a lot.

For example, the size of libc.so is reduced by a lot when stripped
with no option than with -g.

    3014184 libc-2.28.orig.so
    2970920 libc-2.28.strip-g.so
    1460904 libc-2.28.strip.so

Signed-off-by: Kairui Song <kasong@tencent.com>

## Checklist


johannbg (approved):
LGTM

PR #1686, opened 2022-01-10
fix(network): wrong test of wicked unit

aafeijoo-suse:
The test for the wicked service is never met because it does not have execute permission.

## Checklist


johannbg (approved):
LGTM

PR #1688, opened 2022-01-12
fix(S390_rules): remove collect based udev rule creators

tblume:
The collect binary has been removed, so the rules are broken

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/95dasd_rules/parse-dasd.sh
+++ b/modules.d/95dasd_rules/parse-dasd.sh
@@ -1,57 +1,22 @@
 #!/bin/bash
 
-create_udev_rule() {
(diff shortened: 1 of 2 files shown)


johannbg (changes requested):
Minor nitpick the scope of the commit message is wrong as in the scope of the commit msg should be the module name (without numbers) so could you split your commit message into two, one that says
`fix(dasd_rules): remove collect based udev rule creators`
and another one
`fix(zfcp_rules): remove collect based udev rule creators`

Other than that LGTM

johannbg (approved):
LGTM

PR #1689, opened 2022-01-13
fix(dracut-shutdown): add cleanup handler on failure

rmetrich:
It may happen that `dracut-shutdown.service` fails, for example on timeout due to very low bandwidth.
In such case, for hardening purposes, a new `dracut-shutdown-onfailure.service` unit doing `dracut-shutdown.service` cleanup needs to execute to make sure switching root to an incomplete initramfs won't occur later.

See also [RHBZ #1924587](https://bugzilla.redhat.com/show_bug.cgi?id=1924587).

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/98dracut-systemd/dracut-shutdown-onfailure.service
+++ b/modules.d/98dracut-systemd/dracut-shutdown-onfailure.service
@@ -1,0 +1,13 @@
+#  This file is part of dracut.
+#
+# See dracut.bootup(7) for details
+
+[Unit]
+Description=Service executing upon dracut-shutdown failure to perform cleanup
+Documentation=man:dracut-shutdown.service(8)
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+ExecStart=-/bin/rm /run/initramfs/shutdown
+StandardError=null
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,7 @@ ifneq ($(enable_documentation),no)
 endif
 	if [ -n "$(systemdsystemunitdir)" ]; then \
 		mkdir -p $(DESTDIR)$(systemdsystemunitdir); \
+		ln -srf $(DESTDIR)$(pkglibdir)/modules.d/98dracut-systemd/dracut-shutdown-onfailure.service $(DESTDIR)$(systemdsystemunitdir)/dracut-shutdown-onfailure.service; \
 		ln -srf $(DESTDIR)$(pkglibdir)/modules.d/98dracut-systemd/dracut-shutdown.service $(DESTDIR)$(systemdsystemunitdir)/dracut-shutdown.service; \
 		mkdir -p $(DESTDIR)$(systemdsystemunitdir)/sysinit.target.wants; \
 		ln -s ../dracut-shutdown.service \
--- a/modules.d/98dracut-systemd/dracut-shutdown.service
+++ b/modules.d/98dracut-systemd/dracut-shutdown.service
@@ -10,6 +10,7 @@ Wants=local-fs.target
 Conflicts=shutdown.target umount.target
 DefaultDependencies=no
 ConditionPathExists=!/run/initramfs/bin/sh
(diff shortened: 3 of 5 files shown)


rmetrich:
See also related plymouth PR https://github.com/freedesktop/plymouth/pull/6
(plymouth must run after `dracut-shutdown-on-failure` when the latter executes)

rmetrich:
Serial console output on success (similar to current code):
~~~
[  OK  ] Stopped Restore /run/initramfs on shutdown.
         Starting Tell Plymouth To Jump To initramfs...
[  OK  ] Stopped target Local File Systems.
         Unmounting /boot...
[  OK  ] Started Tell Plymouth To Jump To initramfs.
[  OK  ] Unmounted /boot.
~~~

Serial console output on failure (no jump to initramfs):
~~~
[  OK  ] Stopped Restore /run/initramfs on shutdown.
         Starting Service executing upon dra…down failure to perform cleanup...
[  OK  ] Stopped target Local File Systems.
         Unmounting /boot...
[  OK  ] Started Service executing upon drac…utdown failure to perform cleanup.
[   50.910373] XFS (vda1): Unmounting Filesystem
[  OK  ] Unmounted /boot.
~~~

johannbg:
@rmetrich looking at that bz.rh report I think you should do two things before proceeding further a) try to find what's the actual root cause for the unpacking failing in the first place and b) try to duplicate it on a distribution which has more modern core/baseOS stack since RHEL releases are made out of outdate core/base OS components.

rmetrich:
I agree RHEL is quite behind but IMHO hardening is still valuable.

I saw this recently on HP hardware which was sending burst of Ctrl-Alt-Del. There was a PR to harden this which is now upstream (commit #b9ba3c8bb8f0f1328cd1ffaa8dbf64585b28c474).
More generally hardware issues may happen that lead to getting to emergency prompt, I see this regularly, e.g. Virtual DVD being unmounted due to iLO timeout while shutdown of installation occurs.

johannbg:
With regards to the dvd/ilo issue is not the issue there that the admin has not increased the idle timeout in the firmware? In properly designed OS it needs to be consistent throughout the OS ( which is not the case in Fedora/rhel ) and the relevant hw firmware as well.

johannbg (approved):

modules.d/98dracut-systemd/dracut-shutdown-onfailure.service
```diff
@@ -0,0 +1,12 @@
...
+# See dracut.bootup(7) for details
+
+[Unit]
+Description=Service executing upon dracut-shutdown failure to perform cleanup
+Documentation=man:dracut-shutdown.service(8)
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+ExecStart=/bin/sh -c '/bin/rm /run/initramfs/shutdown 2>/dev/null || true'
```
marcosfrm, line 12:
```
ExecStart=-/bin/rm /run/initramfs/shutdown
StandardError=null
```
?

johannbg: > ```
> ExecStart=-/bin/rm /run/initramfs/shutdown
> StandardError=null
> ```
> 
> ?

You will need to clarify your ? 

What do you see wrong with the on failiure unit *trying* to execute /bin/rm /run/initramfs/shutdown ( hence the - ) and it dumping it's own output to /dev/null?

How would you solve this differently?

marcosfrm: It is _my suggestion_ to avoid calling `sh` needlessly...

johannbg: @marcosfrm I somehow manage to miss the entire context of you response, your absolutly correct. @rmetrich can you fix your PR as @marcosfrm suggests or is there any special reason you are escaping to shell to call the command?

johannbg (approved):

PR #1691, opened 2022-01-19
Skip creating the dracut initrd in more cases

keszybz:
See https://github.com/systemd/systemd/pull/22175 for the corresponding changes in kernel-install.

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -4,6 +4,7 @@ COMMAND="$1"
 KERNEL_VERSION="$2"
 BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
+INITRD_OPTIONS_SHIFT=4
 
 # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
 # In this case, do not create the initrd.
@@ -27,6 +28,9 @@ fi
 ret=0
 case "$COMMAND" in
     add)
+        # If the initrd was provided on the kernel command line, we shouldn't generate our own.
+        [ "$#" -gt "$INITRD_OPTIONS_SHIFT" ] && exit 0
+
         INITRD_IMAGE_PREGENERATED=${KERNEL_IMAGE%/*}/initrd
         if [[ -f ${INITRD_IMAGE_PREGENERATED} ]]; then
             # we found an initrd at the same place as the kernel
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -6,8 +6,14 @@ BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
 
 # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
-# So, let's skip to create initrd.
+# In this case, do not create the initrd.
 if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
+    exit 0
+fi
+
+# Skip this plugin if we're using a different generator. If nothing is specified,
+# assume we're wanted since we're installed.
+if [ "${KERNEL_INSTALL_INITRD_GENERATOR:-dracut}" != "dracut" ]; then
     exit 0
 fi
 
--- a/install.d/51-dracut-rescue.install
+++ b/install.d/51-dracut-rescue.install
@@ -7,6 +7,11 @@ KERNEL_VERSION="$2"
 BOOT_DIR_ABS="${3%/*}/0-rescue"
 KERNEL_IMAGE="$4"
 
+# Skip this plugin if we're using a different generator. If nothing is specified,
+# assume we're wanted since we're installed.
+if [ "${KERNEL_INSTALL_INITRD_GENERATOR:-dracut}" != "dracut" ]; then
+    exit 0
+fi
 
 dropindirs_sort()
 {


johannbg:
@keszybz as usual fix your commit message so it passes the commisery check, that said we expect dracut being used if installed, if it's not being used it should not be installed on the system in the first place so this should not be a necessary change.

johannbg:
Nack-ing this PR since it literally makes no sense for a initrd generator to disable it sole purpose of being installed in the first place, which is to generate initrd's nor for generators themselves to support multiple initrd generators being install at the same time or for vendor/distributions to support such scenarios.

This PR will not solve a problem, advance or improve anything for genrators in the generators world or on the core/base OS layer or users and is seemingly only being done so systemd/mkosi can workaround deficiency in it's own tool/generator design...

keszybz:
I added those miserable prefixes.

johannbg:
What's the root problem your trying to solve with those prefixes? Is it because mkosi-initrd cant be used as a standalone generator replacement? ( it's existance begs the question why there was a need to invent yet another generator ) Is it because distributions arent shipping mkosi-initrd? Is it because distributions are packaging and shipping dracut ( and in some distributions are shipping multiple generator so fixing it just in dracut wont solve anything ) thus have dependency on it/them. Dracut works fine with systemd as is as do the other generators so whats the root cause for this change?

keszybz:
> Is it because mkosi-initrd cant be used as a standalone generator replacement?

No, quite the opposite. I want to use it exclusively.

>  it's existance begs the question why there was a need to invent yet another generator 

I don't think it's necessary to discuss this here, though I would be happy to go into detail if necessary. Even ignoring mkosi-initrd, there are other generators like mkinitcpio used under arch, and Debian has something custom too… So kernel-install has to be prepared to have multiple initrd generators installed.

> Is it because distributions arent shipping mkosi-initrd?

As far as I know, they aren't, because there isn't anything to ship yet.

> Is it because distributions are packaging and shipping dracut ( and in some distributions are shipping multiple generator so fixing it just in dracut wont solve anything ) thus have dependency on it/them.

Yes, as mentioned in the commit message, this is the pretty much the reason: not that distributions are shipping dracut in general, but that it may be uninstallable because of dependencies. But even if it wasn't uninstallable, I would want to be able to have multiple generators installed in parallel and select which one to use through some configuration. This commit adds support for that configuration switch.

>  Dracut works fine with systemd as is as do the other generators so whats the root cause for this change?

In general you only want to have one generator active. Whether it uses systemd or not is not relevant. This pull request is about not trying to generate the initrd multiple times.

behrmann:
To maybe flesh out the examples a bit: 

Arch packages both mkinitcpio and dracut, but defaults (well as much as one can speak of defaults in Arch) to the former. Both mkinitcpio and dracut ship snippets for kernel-install, but only mkinitcpio ships pacman hooks to generate initrds when certain packages change. Currently, people wanting to give dracut a spin need need to add the hooks themselves, which is certainly not difficult, but having a mechanism to switch the active initrd generator would allow to just ship a single hook for all of them (as a side: currently the hook for mkinitcpio doesn't use kernel-install) and it would be easier than having to uninstall the other generator(s) or to switch them off by symlinking the kernel-install files to `/dev/null`

I don't really know whether kernel-install is actually usable in Debian (they have their own hierarchy of garnly scripts in `/etc/kernel` that are called by postinst scripts in the kernel and other packages), but because of they broad dependencies and suggest of packages it is actually quite easy to end up with both dracut and initramfs-tools installed, actually `dracut` is among the listed reverse dependencies of initramfs-tools (`apt-cache rdepends initramfs-tools`).

I think having a generic way fpr multiple initrd generators to be easily coinstallable is a win.

johannbg:
> > it's existance begs the question why there was a need to invent yet another generator
> 
> I don't think it's necessary to discuss this here, though I would be happy to go into detail if necessary. Even ignoring mkosi-initrd, there are other generators like mkinitcpio used under arch, and Debian has something custom too… So kernel-install has to be prepared to have multiple initrd generators installed.

Given that it already works quite well ( and has done so for years ) for distributions such as Debian, Arch, Gentoo and probably more shipping and using multiple generators solutions is evidence to the contrary that kernel-install has to be prepared to have multiple initrd generators installed. 

> 
> > Is it because distributions are packaging and shipping dracut ( and in some distributions are shipping multiple generator so fixing it just in dracut wont solve anything ) thus have dependency on it/them.
> 
> Yes, as mentioned in the commit message, this is the pretty much the reason: not that distributions are shipping dracut in general, but that it may be uninstallable because of dependencies.

That applies to all generators that distributions are shipping and to me it seems you are just trying to work around having to sort out those dependency with those distributions shipping generator(s) for mkosi. 

> But even if it wasn't uninstallable, I would want to be able to have multiple generators installed in parallel and select which one to use through some configuration. This commit adds support for that configuration switch.

Yes I'm aware of that but here's the thing afaikt there is absolutly no benefit for upstream generators to support thus use that, there is no benefit for distribution to support that either or for users to have multiple generators simultaneously installed. 

Zero benefits so why should upstream generators support that when it adds no value to them? 

There is plethora of reason why distribution should chose to ship multiple generators but no good reason why those distribution should support having more then one of those solution installed at any given time so why should they even support that?

> 
> > Dracut works fine with systemd as is as do the other generators so whats the root cause for this change?
> 
> In general you only want to have one generator active. Whether it uses systemd or not is not relevant. This pull request is about not trying to generate the initrd multiple times.

To me this pull-request is about not having *mkosi* trying to generate the initrd multiple times since so far I have not seen any other project having that requirement.

johannbg:
> To maybe flesh out the examples a bit:
> 
> Arch packages both mkinitcpio and dracut, but defaults (well as much as one can speak of defaults in Arch) to the former. Both mkinitcpio and dracut ship snippets for kernel-install, but only mkinitcpio ships pacman hooks to generate initrds when certain packages change. Currently, people wanting to give dracut a spin need need to add the hooks themselves, which is certainly not difficult, but having a mechanism to switch the active initrd generator would allow to just ship a single hook for all of them (as a side: currently the hook for mkinitcpio doesn't use kernel-install) and it would be easier than having to uninstall the other generator(s) or to switch them off by symlinking the kernel-install files to `/dev/null`

Arch also could just fix that for those users by providing that missing package hook for all the generators it ships ( I think they are actually three not two ) could it not? + it would be a better solution than supporting multiple generators installed at the same time.
 
> I don't really know whether kernel-install is actually usable in Debian (they have their own hierarchy of garnly scripts in `/etc/kernel` that are called by postinst scripts in the kernel and other packages), but because of they broad dependencies and suggest of packages it is actually quite easy to end up with both dracut and initramfs-tools installed, actually `dracut` is among the listed reverse dependencies of initramfs-tools (`apt-cache rdepends initramfs-tools`).
> 
> I think having a generic way fpr multiple initrd generators to be easily coinstallable is a win.

All I see here are downstream packaging/implementation problems which should be worked through in those communities not seeing any solution that generators upstream should support or otherwise implement/fix here.

behrmann:
> Given that it already works quite well ( and has done so for years ) for distributions such as Debian, Arch, Gentoo and probably more shipping and using multiple generators solutions is evidence to the contrary that kernel-install has to be prepared to have multiple initrd generators installed.

I'm not sure this assumption is correct. I'm sure most Arch users use mkinitcpio and by default that - by default - doesn't use kernel-install. Debian, if I remember correctly, (and therefore Ubuntu?) still defaults to initramfs-tools and that doesn't even have snippets for kernel-install (which maybe isn't so bad after all).

Debian (and its derivatives) alone make up a huge chunk of the Linux installation base, so I assume a large part doesn't use kernel-install at all. I think it would be nice for more tooling to standardise on this and since having multiple initrd generators co-installed something that does happen in the wild, I think it makes sense to deal with this in kernel-install and have a nice way of preventing the initrd generators stepping on each others feet.

johannbg:
>  having multiple initrd generators co-installed something that does happen in the wild, 

Can you give me an example of where it already does happen in the wild?

behrmann:
On Debian you can have dracut and initramfs-tools co-installed, but Debian doesn't use kernel-install and the problem of generating multiple initrds upon installation of a new kernel is sidestepped by Debian splitting dracut and initramfs-tools into multiple packages (in the case of dracut quite a few actually), specifically a `dracut` and an `initramfs-tools` package that specify who actually gets to generate the image and respective `*-core` packages, that contain the machinery to do the job. This gets the job done, but I'd argue that it's a lot more heavy-handed than this would be.

On Arch you can have dracut and mkinitcpio co-installed and both install kernel-install hooks. If someone were to use kernel-install, which most users probably don't, since this is not how it's set up by default, they would need to symlink the hook files of the of the initrd generator they don't intend to use. This is what I do on my machines using dracut on Arch. (If people don't use kernel-install they will have to make their own pacman hooks for dracut and disable the hooks for mkinitcpio, the problem is the same as here, just in a slightly different flavour.)

I'm not too familiar with the initrd generator situation in the RH family of distros or with Suse. I assume both default to dracut and usually don't see something else used. I don't know who there is actively using kernel-install. I'm sure or the RH distros @keszybz can say more about this. 

From my vantage point it does seem like for most distros there is at least the unspoken assumption, that it should be possible to have multiple initrd generators installed at the same time. If this is the case only one should generate the initrd. Debian solves this as Debian does by splitting packages. Arch doesn't seem to have addressed this issue and leaves people to hack their own solutions together and the default case works because it is assumed people use mkinitcpio.

keszybz:
Exactly.

> I'm not too familiar with the initrd generator situation in the RH family of distros or with Suse. 

RH-family uses dracut exclusively. Not sure about Suse, but I think they do too.

And in case this wasn't clear: yes, we're working on a new generator, and expect that even if it wasn't very common in the past, it will happen more often.

johannbg:
> And in case this wasn't clear: yes, we're working on a new generator, and expect that even if it wasn't very common in the past, it will happen more often.

I'm not sure what you mean by expect that, expect what exactly?

I'm perfectly aware atleast two new generators got written last year and people are writing new generators all the time ( to make them fast apparently, as soon as the evolve beyond being simple shell script or something written in new language, those generators dont last and I suspect mkosi will go down that path as well because replacing well establish generators is more work than most people realise and getting distro's to default to a new generator is probably a 2- 4 years work of convincing them of doing so with all the "fun" involved in doing that ) but I still dont see any particular reason why those need to be somehow standardized.

behrmann:
Well, I won't argue why standardisation is good, because I think the three of us all agree that standardising Linux userland plumbing is a good thing. 

My point is, that this is better design. Right now all kernel-install hooks that are placed in `kernel/install.d` will be run. This is fine for distros like Fedora that only use dracut; and I think this is what was in mind when this was designed. When there are multiple initrd generators in the game, the picture does get murkier, because we need to make sure that the right initrd generator does the job and that they don't step on each others toes - overwrite each others outputs or do unnecessary work. 

This *is* already possible today switching off hooks in `/etc/kernel/install.d` or by more heavy-handed solutions on the level of package managers. From what I see this will make kernel-install more pluggable and I'd hope that this will standardize the stack further down and allow to get rid of custom solutions in distros.

keszybz:
Systemd 251 has been released with the plumbing for KERNEL_INSTALL_INITRD_GENERATOR. This can be merged now.

keszybz:
→ #1825.

johannbg:
@keszybz this wont be merged.

poettering:
@johannbg any chance this can be revisited? the kernel-install plugin API is defined with `$KERNEL_INSTALL_INITRD_GENERATOR` and `$KERNEL_INSTALL_UKI_GENERATOR`, but the dracut plugin does not implement this. This is disappointing, as this means dracut simply doesn't play nice with alternative initrd implementations, making it very hard for people to switch forth and back.

The interface is documented here:

https://www.freedesktop.org/software/systemd/man/latest/kernel-install.html#Environment%20variables%20exported%20for%20plugins

If this documentation is not enough, then we are happy to extend it, but just ignoring it sucks, because the whole idea of "kernel-install" is that different components can play nicely together, and a certain set of alternatives are supported.

keszybz:
166456331d55cdc23946c11315dc2c88aab15911 does something similar. I'll close my pull request.

poettering:
ah, good

aafeijoo-suse:
> if ( accidentally ) committed in the codebase

It was not an accident, it was reviewed and agreed in #2495 by 3 people.

PR #1701, opened 2022-02-03
fix(tpm2-tss): add tss user and group

### LINKED ISSUE #1700: tpm2-tss does not create tss user and group
**Describe the bug**
On boot before switching to real root, errors (`journalctl -p3`) regarding the tss user normally in `/etc/passwd` are reported:

```
systemd-tmpfiles[421]: /usr/lib/tmpfiles.d/tpm2-tss-fapi.conf:2: Failed to resolve user 'tss': No such process
systemd-tmpfiles[421]: Failed to parse ACL "default:group:tss:rwx": Invalid argument. Ignoring
systemd-tmpfiles[421]: /usr/lib/tmpfiles.d/tpm2-tss-fapi.conf:4: Failed to resolve user 'tss': No such process
systemd-tmpfiles[421]: Failed to parse ACL "default:group:tss:rwx": Invalid argument. Ignoring
systemd-udevd[489]: /lib/udev/rules.d/60-tpm-udev.rules:3 Unknown user 'tss', ignoring
systemd-udevd[489]: /lib/udev/rules.d/60-tpm-udev.rules:4 Unknown user 'tss', ignoring
systemd-udevd[489]: /lib/udev/rules.d/60-tpm-udev.rules:4 Unknown group 'tss', ignoring
```

**Distribution used**
Gentoo

**Dracut version**
055

**Init system**
systemd

**To Reproduce**
1. Install tpm2-tools
2. Enable kernel support for TPM 2
2. `dracut -a 'systemd tpm2-tss' --force`

**Expected behavior**
Should not have this error. The `tss` user maybe should be in the initrd (easiest way to solve this).

**Additional context**
`/etc/passwd` and `/etc/group` in the initrd do not list the `tss` user/group.

```
 $ cat $EXTPATH/etc/passwd
adm:x:3:4:adm:/var/adm:/bin/false
systemd-network:x:111:116:System user; systemd-network:/dev/null:/sbin/nologin
root:x:0:0::/root:/bin/sh
nobody:x:65534:65534:System user; nobody:/var/empty:/sbin/nologin
 $ cat $EXTPATH/etc/group
systemd-journal:x:121:
wheel:x:10:tatsh
adm:x:4:adm,daemon
utmp:x:406:
root:x:0:
systemd-network:x:116:
tty:x:5:tatsh
dialout:x:998:
kmem:x:9:
input:x:249:tatsh
video:x:27:tatsh,sddm
render:x:985:
sgx:x:405:
audio:x:18:tatsh
lp:x:7:lp,tatsh
disk:x:6:adm
cdrom:x:991:tatsh
tape:x:26:
kvm:x:78:
floppy:x:11:
```

aafeijoo-suse:
The `tss` user and group are required by `60-tpm-udev.rules`

```
# tpm devices can only be accessed by the tss user but the tss
# group members can access tpmrm devices
KERNEL=="tpm[0-9]*", TAG+="systemd", MODE="0660", OWNER="tss"
KERNEL=="tpmrm[0-9]*", TAG+="systemd", MODE="0660", OWNER="tss", GROUP="tss"
```

## Checklist

Fixes #1700

### DIFF
--- a/modules.d/91tpm2-tss/module-setup.sh
+++ b/modules.d/91tpm2-tss/module-setup.sh
@@ -55,4 +55,8 @@ install() {
         {"tls/$_arch/",tls/,"$_arch/",}"libcurl.so.*" \
         {"tls/$_arch/",tls/,"$_arch/",}"libjson-c.so.*"
 
+    # Add user and group for tpm2-tss
+    grep '^tss:' "$dracutsysrootdir"/etc/passwd 2> /dev/null >> "$initdir/etc/passwd"
+    grep '^tss:' "$dracutsysrootdir"/etc/group 2> /dev/null >> "$initdir/etc/group"
+
 }


Tatsh:
Tested this patch and it fixes the issue.

johannbg:
@aafeijoo-suse we dont grep users from /etc/passwd anymore since /etc can be nonexistant these days, it's more of an legacy action, Can you update your patch to use systemd sysuser snippet instead which we include in with the module and use instead..

aafeijoo-suse:
Thanks, now I see what is happening here: I could reproduced this bug in each distribution I tested because all of them are missing https://github.com/dracutdevs/dracut/commit/dcbe23c1, so the `systemd-sysusers` service didn't start and didn't do its job in the initrd. After applying this fix, no further action is required.

victor-timofei:
It seems this is not fixed, I am still getting this error on gentoo in version 059

PR #1702, opened 2022-02-04
feat(shutdown): explicitly link reboot, halt and poweroff to systemctl

aafeijoo-suse:
If the running system is using systemd, it provides commands `reboot`, `halt` and `poweroff` via `systemctl` verbs.

Some distributions (SUSE, Debian, Arch,...) ship the legacy SysV `reboot`, `halt` and `poweroff` commands as links to `systemctl` in a separate package, causing the need to add an additional dependency in their dracut installation package, which will be avoided by this PR.

## Checklist

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -137,7 +137,7 @@ install() {
         journalctl systemctl \
         echo swapoff \
         kmod insmod rmmod modprobe modinfo depmod lsmod \
-        mount umount reboot poweroff \
+        mount umount \
         systemd-run systemd-escape \
         systemd-cgls systemd-tmpfiles \
         systemd-ask-password systemd-tty-ask-password-agent \
--- a/modules.d/99shutdown/module-setup.sh
+++ b/modules.d/99shutdown/module-setup.sh
@@ -14,7 +14,16 @@ depends() {
 # called by dracut
 install() {
     local _d
-    inst_multiple umount poweroff reboot halt losetup stat sleep timeout
+    local _systemctl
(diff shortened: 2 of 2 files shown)


johannbg:
@aafeijoo-suse We just spiral down a rabbit hole if we start catering to downstream ( there are ca 500 distro's out there ) and their deviations and different stage of evolution in other words this is something that should be solved/carried downstream until it's fixed *there*

aafeijoo-suse:
Thanks for your feedback. Although I was expecting this answer, I wanted to give it a try, because is a common downstream issue with, IMHO, a fairly reasonable solution and I'd rather cherry pick an upstream commit than having to deviate from it due to some requirement.

johannbg:
Yes I'm pretty set about fixing problems where they belong, at the root of the source of the problem because if you dont then the underlying problem will never get fixed because people are like water they always go the shortest path. 

Your PR here is a good example of that, it's easier to workaround this in dracut than it is to have distribution adapt and or evolve and if merged we would end up with a larger code base as a result of that, which adds to *our* cost of development.

aafeijoo-suse:
I wouldn't say this is an easy workaround only beneficial downstream, because this ensures the use of `systemctl` if it's available. Also, we could simplify this:

https://github.com/dracutdevs/dracut/blob/d754e1c6f081a6501cb7fdcb5caaa6c4977235af/modules.d/99base/dracut-lib.sh#L467-L469

with just `halt -f` (as `halt` is equivalent to `systemctl start halt.target --job-mode=replace-irreversibly --no-block`).

Anyway, I understand your point of view, so thank you again for taking your time to review this proposal.

PR #1703, opened 2022-02-07
S390 rules fix inst rules arguments

tblume:
Arguments to the inst_rules_wildcard function need quoting
rules naming *-s390x-* was obsolete and broken

## Checklist
- [ X] I have tested it locally

Fixes #


johannbg:
I'm going ahead and merging this to make sure that this fix will end up in the release which should hopefully happen in the coming week.

PR #1704, opened 2022-02-08
[RFC] Dracut module args support for dracut cmdline

liutgnu:
This patch set:
1) allows dracut cmdline pass arguments to specific dracut modules as mod.arg="val".
2) decouple compressors for dracut and dracut-squash. With the patch set, the dracut-squash module can receive customized compressor by squash.comp="zstd"
3) It's a RFC patch set, which didn't update the usage docs, in case of further comments and modifications. 

@daveyoung @ryncsn 

## Changes
[PATCH 1/3] Dracut module args support for dracut cmdline
[PATCH 2/3] Allows lsinitrd to show module args
[PATCH 3/3] Decouple the compressor for dracut and dracut-squash

## Checklist
- [N] I have reviewed and updated any documentation if relevant

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -478,7 +478,15 @@ while :; do
 done
 
 # get output file name and kernel version from command line arguments
+dracut_modules_args=()
 while (($# > 0)); do
+    module_args_regex="^\w+\.\w+=(\S|\t|\ )+$"
+    [[ "$1" =~ $module_args_regex ]] && {
+        dracut_modules_args+=("$1")
+        shift
+        continue
+    }
+
     case ${1%%=*} in
         ++include)
             shift 2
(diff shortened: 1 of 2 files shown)


ryncsn:
Hi, Tao,

I think this is not really a generic issue, and a bit overkill to bring a new way to pass arguments just for fixing one specific issue (squash module may need different compressor).

I think just a new `--squash-compressor` argument, or use some environment variable like `DRACUT_SQUASH_COMPRESSOR` is a better idea.

Let's see how other people think about it.

liutgnu:
Hi maintainers,

Do you have any comments and suggestions for this? Thank you very much!
@johannbg @haraldh @danimo 

Thanks,
Tao Liu

johannbg:
I agree with @ryncsn that introducing module specific arguments is a bad idea and a better alternative would be something along the lines of --squash-compressor argument

liutgnu:
> I agree with @ryncsn that introducing module specific arguments is a bad idea and a better alternative would be something along the lines of --squash-compressor argument

Thanks for the comments, @johannbg. I will try to implement one as "--squash-compressor".

ryncsn:
> 



> > I agree with @ryncsn that introducing module specific arguments is a bad idea and a better alternative would be something along the lines of --squash-compressor argument
> 
> Thanks for the comments, @johannbg. I will try to implement one as "--squash-compressor".

Hi Tao,

BTW, maybe you can pass extra arguments with "--squash-compressor" just like "--squash-compressor 'xz -Xdict ....'", just refer to dracut's "--compress", it also allow passing extra arguments with quoted string, splitted by space.

liutgnu:
> > 
> 
> > > I agree with @ryncsn that introducing module specific arguments is a bad idea and a better alternative would be something along the lines of --squash-compressor argument
> > 
> > 
> > Thanks for the comments, @johannbg. I will try to implement one as "--squash-compressor".
> 
> Hi Tao,
> 
> Since you mentioned you wanted to pass in extra ARGS for squash compressor, maybe you can still pass extra arguments with "--squash-compressor" just like "--squash-compressor 'xz -Xdict ....'", just refer to dracut's "--compress", it also allow passing extra arguments with quoted string, splitted by space.

OK, thanks for the info! I'll give it a shot and resubmit a PR.

liutgnu:
The v2 PR has been sent at https://github.com/dracutdevs/dracut/pull/1707

PR #1711, opened 2022-02-17
Add zfs detection

savyajha:
zfs detection is currently done by the zfs dracut module installed by
downstream, resulting in a lot of duplicated code. This commit puts zfs
detection into the core dracut logic, allowing for detection of zfs
partitions to be done at the same time as all others. It also allows for
dracut to correctly create a `root=` cmdline parameter for zfs.

Signed-off-by: Savyasachee Jha <hi@savyasacheejha.com>

This pull request changes...

## Changes

- Adds a `zfs_devs` function analogous to `btrfs_devs` in `dracut-functions.sh`
- Adds detection of zfs devices using function in the above point
- If the rootfs is on a zfs partition, then allows for setting a `root=zfs:<zfs dataset>` cmdline parameter

## Checklist

Fixes #
Nothing

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1446,6 +1446,10 @@ for line in "${fstab_lines[@]}"; do
                 push_host_devs "$i"
             done
         done
+    elif [[ $3 == zfs ]]; then
+        for mp in $(zfs_devs "$1"); do
+            push_host_devs "$mp"
+        done
     fi
     push_host_devs "$dev"
     host_fs_types["$dev"]="$3"
@@ -1498,7 +1502,13 @@ if [[ $hostonly ]] && [[ $hostonly_default_device != "no" ]]; then
                 [[ $mp == "/" ]] && root_devs+=("$i")
                 push_host_devs "$i"
             done
+        elif [[ $(find_mp_fstype "$mp") == zfs ]]; then
+            for i in $(zfs_devs "$(findmnt -n -o SOURCE "$mp")"); do
+                [[ $mp == "/" ]] && root_devs+=("$i")
+                push_host_devs "$i"
+            done
         fi
+
     done
 
     # TODO - with sysroot, /proc/swaps is not relevant
@@ -1549,6 +1559,10 @@ if [[ $hostonly ]] && [[ $hostonly_default_device != "no" ]]; then
             push_host_devs "$_dev"
             if [[ $_t == btrfs ]]; then
                 for i in $(btrfs_devs "$_m"); do
+                    push_host_devs "$i"
+                done
+            elif [[ $_t == zfs ]]; then
+                for i in $(zfs_devs "$_d"); do
                     push_host_devs "$i"
                 done
             fi
@@ -2576,6 +2590,9 @@ freeze_ok_for_fstype() {
     fstype=$(stat -f -c %T -- "$outfile")
     case $fstype in
         msdos)
+            return 1
+            ;;
+        zfs)
             return 1
             ;;
         btrfs)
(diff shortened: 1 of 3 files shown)


savyajha:
Done! Apologies for not having done that earlier.

johannbg (approved):
LGTM

dracut.sh
```diff
@@ -1446,7 +1446,14 @@ for line in "${fstab_lines[@]}"; do
                 push_host_devs "$i"
             done
         done
+    elif [[ $3 == zfs ]]; then
+        for mp in $1; do
+            for i in $(zfs_devs "$mp"); do
+                push_host_devs "$i"
+            done
```
nabijaczleweli, line 1453: this isn't any different from `push_host_devs $(zfs_devs "$mp")`, is it?

modules.d/95rootfs-block/module-setup.sh
```diff
@@ -44,6 +44,9 @@ cmdline_rootfs() {
         printf " root=%s" "$(shorten_persistent_dev "$(get_persistent_dev "$_dev")")"
     fi
     _fstype="$(find_mp_fstype /)"
+    if [[ ${_fstype} == "zfs" ]]; then
+        printf " root=ZFS=%s" "$(findmnt -n -o SOURCE /)"
```
nabijaczleweli, line 48: in upstream we treat the `root=zfs:<dataset>` format as canonical so it'd be nice if this were that

nabijaczleweli: also, if the root dataset has spaces in it (and this is valid and supported), you need to replace them with pluses, i.e.
```
         local _root_ds="$(findmnt -n -o SOURCE /)" 
         printf " root=zfs:%s" "${_root_ds// /+}"
```

dracut.sh
savyajha: No, no it's not. 😅

modules.d/95rootfs-block/module-setup.sh
savyajha: I will do that.

savyajha:
Made changes as suggested by zfs devs.

nabijaczleweli (approved):
If supporting root devices with spaces in is something we want to support, then:
```sh
zpool list -HvP "${_mp%%/*}" | awk -F$'\t' '$2 ~ /^\// {print $2}'
```
in `zfs_devs()` instead; that being said, AFAICT, udev replaces spaces with `\x20` and they'd get broken at tokenisation in the for loop anyway, so

savyajha:
@nabijaczleweli Added support for the case you've mentioned. The device paths are resolved to the real devices. `dracut` will, when required, convert them into persistent device names.

dracut.sh
```diff
@@ -1551,6 +1562,10 @@ if [[ $hostonly ]] && [[ $hostonly_default_device != "no" ]]; then
                 for i in $(btrfs_devs "$_m"); do
                     push_host_devs "$i"
                 done
+            elif [[ $_t == zfs ]]; then
+                for i in $(zfs_devs "$_d"); do
+                    push_host_devs "$i"
+                done
```
nabijaczleweli, line 1568: same applies here, afaict this can just be `push_host_devs $(zfs_devs "$_d")`; either way, nice, short, sweet, and infinitely better than the 02 we ship in openzfs now

savyajha: Made the change, and made a similar change for the other time `zfs_devs` is called.

savyajha: `shfmt` gave errors with this change, so I reverted it back to the way the `btrfs` stuff is implemented.

nabijaczleweli:
I have, yeah :)

behlendorf:
Looks good.

PR #1712, opened 2022-02-18
fix(dmsquash-live): permanent overlay on the same drive as LiveCD .iso

LaszloGombos:
An example kernel command line arguments for this configuration
iso-scan/filename=distro.iso root=live:CDLABEL=ISO
rd.live.image rd.live.overlay=/dev/sda:/overlay.img

iso-scan would mount /dev/sda first and keep it mounted. The change
allows detecting if the permanent overlay drive is already mounted.
It also ensures that the mount is writable, as permanent overlay
requires writable storage.

This pull request changes...

## Checklist

Fixes #


johannbg (approved):
LGTM

PR #1714, opened 2022-02-19
fix(github): remove packit

johannbg:
It's been a year since I initially made the PR to include packit.

Having it has resulted in increasing the project's cost of development
( CoD ) thus defeating the initial purpose of including it, which was an attempt
to do the opposite as in to reduce it, thus removing it.

My take on this after a year in use is that the concept is flawed,
limited to a single distribution and cannot work in practice without
adding to upstreams and it's contributors CoD but other projects milage
may vary.

This pull request changes...

## Checklist

Fixes #


johannbg:
I heard no objections or justifications why we should still be keeping it hence dropping it.

PR #1715, opened 2022-02-19
refactor(base): use blockdev@<devname>.target instead of <devname>.device

savyajha:
The blockdev@<devname>.target is a unit made for ordering block devices
for systemd services. It seems to be the canonical way of doing this as
given in the man page systemd.special. This PR aims to get dracut to
use the same target unit throughout in base, as well as in the
dracut-systemd module.

Signed-off-by: Savyasachee Jha <hi@savyasacheejha.com>

This pull request changes...

## Changes
All instances of `initrd.target.wants/<devname>.device` replaced with `initrd.target.wants/blockdev@<devname>.target`. I have tested this locally and seen no breakage in booting.

## Checklist

Fixes #
Nothing

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -78,6 +78,7 @@ install() {
         "$systemdsystemunitdir"/timers.target \
         "$systemdsystemunitdir"/paths.target \
         "$systemdsystemunitdir"/umount.target \
+        "$systemdsystemunitdir"/blockdev@.target \
         \
         "$systemdsystemunitdir"/sys-kernel-config.mount \
         \
--- a/modules.d/98dracut-systemd/rootfs-generator.sh
+++ b/modules.d/98dracut-systemd/rootfs-generator.sh
@@ -28,9 +28,9 @@ EOF
     fi
 
     _name=$(dev_unit_name "$1")
-    if ! [ -L "$GENERATOR_DIR"/initrd.target.wants/"${_name}".device ]; then
+    if ! [ -L "$GENERATOR_DIR"/initrd.target.wants/blockdev@"${_name}".target ]; then
         [ -d "$GENERATOR_DIR"/initrd.target.wants ] || mkdir -p "$GENERATOR_DIR"/initrd.target.wants
-        ln -s ../"${_name}".device "$GENERATOR_DIR"/initrd.target.wants/"${_name}".device
+        ln -s ../../../../lib/systemd/system/blockdev@.target "$GENERATOR_DIR"/initrd.target.wants/blockdev@"${_name}".target
     fi
 
     if ! [ -f "$GENERATOR_DIR"/"${_name}".device.d/timeout.conf ]; then
--- a/modules.d/99base/dracut-dev-lib.sh
+++ b/modules.d/99base/dracut-dev-lib.sh
@@ -72,10 +72,10 @@ set_systemd_timeout_for_dev() {
     _timeout=${_timeout:-0}
 
     _name=$(dev_unit_name "$1")
-    if ! [ -L "${PREFIX}/etc/systemd/system/initrd.target.wants/${_name}.device" ]; then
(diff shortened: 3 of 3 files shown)


johannbg:
@savyajha on which distro which has which systemd release did you locally test this on?

savyajha:
On Debian Bullseye, @johannbg.

johannbg:
@savyajha yeah suspected that's much an old systemd release, if you can, try testing these ( and any future systemd related changes ) on systemd 250+ release which should be available in either Debian unstable/experiment and or both.

savyajha:
I will, thank you for the advice @johannbg. I've got an old desktop lying around somewhere which has Arch on it, I guess it's time to shake the rust off and see if it still works. If it doesn't I'll spin up VMs and experiment.

Regardless, I think that the `blockdev@.target` bit remains the same from 247 -> 250. At least, there's no change in the manpages.

johannbg:
> I will, thank you for the advice @johannbg. I've got an old desktop lying around somewhere which has Arch on it, I guess it's time to shake the rust off and see if it still works. If it doesn't I'll spin up VMs and experiment.
> 
> Regardless, I think that the `blockdev@.target` bit remains the same from 247 -> 250. At least, there's no change in the manpages.

Your pr is poking a hornest nest of udev rules and systemd dependencies, the scope of this change is not as simple as reading a man page and perform search and replace.

Systemd's man pages should be quite clear about the intent and purposes of type device units and block device targets and there are things in the code that often cant be found in man pages or changes that work in older release but not newer ones etc.

savyajha:
I shall get back after testing it on a newer machine.

savyajha:
@johannbg I haven't been able to check it with systemd 250 yet, but it works perfectly normally on Fedora 35 with a btrfs root.

savyajha:
@johannbg Apologies for the ping, but does this require any more testing/work from my side? It's been tested on:

- Debian Bullseye (systemd 247)
- Fedora 35 (systemd 249)
- Arch Linux (systemd 250)

and found to cause no issues with booting.

johannbg:
@savyajha there is nothing that you further need to do, it is I that need to find the time to properly look into this ( go through udev rules, systemd type unit dependency, upstream systemd code etc ) but as it stands now there is a high propability it will get nack-ed.

savyajha:
A small additional datapoint here: systemd-fstab generator generates `<mountpoint>.mount` files of the following type (example copied from one of my own systems):

```
# Automatically generated by systemd-fstab-generator

[Unit]
Documentation=man:fstab(5) man:systemd-fstab-generator(8)
SourcePath=/etc/fstab
Requires=systemd-fsck@dev-disk-by\x2duuid-a6223870\x2db98c\x2d49bb\x2daa8c\x2dd33466a23ee4.service
After=systemd-fsck@dev-disk-by\x2duuid-a6223870\x2db98c\x2d49bb\x2daa8c\x2dd33466a23ee4.service
After=blockdev@dev-disk-by\x2duuid-a6223870\x2db98c\x2d49bb\x2daa8c\x2dd33466a23ee4.target

[Mount]
Where=<mount point>
What=/dev/disk/by-uuid/a6223870-b98c-49bb-aa8c-d33466a23ee4
Type=xfs
Options=rw,noatime,attr2,inode64,logbufs=8,logbsize=256k,sunit=8192,swidth=16384,noquota,largeio,swalloc,nodev,nosuid
```
It seems systemd naturally does not use `<devname>.device` in generated unit files. It uses `blockdev@<devname>.target`. I do think that is what systemd expects `blockdev@<devname>.target` to be used for.

LaszloGombos:
@savyajha It seems this PR does not break any of the existing tests, but it is hard for me to conclude if the code in this PR at least executed by some of the tests. One way to confirm this is if you intentionally make a typo/bug in your code and confirm that some tests are failing.

In general good to have a test case with PRs.

Also, can you think of an actual benefit for this change (other than source code maintenance and readability). Are they any other executables/tools that could be interfering with <devname>.device or take advantage of blockdev@<devname>.target

PR #1717, opened 2022-02-21
fix(install): restore musl support

LaszloGombos:
__GLIBC_PREREQ is only defined in glibc. This change reverts
the recent change from ef0f848a67fdd0a0dab135acbd1cd7fa0179a95c.
The portability of the code outweights the benefits of the
earlier change.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/src/install/util.c
+++ b/src/install/util.c
@@ -24,17 +24,15 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <sys/types.h>
+#include <sys/syscall.h>
 
 #include "util.h"
 
-#if __GLIBC_PREREQ(2, 30) == 0
-#include <sys/syscall.h>
 #ifndef SYS_gettid
 #error "SYS_gettid unavailable on this system"
 #endif
 
 #define gettid()    ((pid_t) syscall(SYS_gettid))
-#endif /*__GLIBC_PREREQ */
 
 size_t page_size(void)
 {


LaszloGombos:
This change will restore the capability to be able to build dracut on alpine using the musl c library.

johannbg:
Portability of code in itself is not an argument nor somekind of code quality measure point thus it bares no more or less weight or benefits in an argument then non portable code as in it's entirely up to the designers whether multiple platform support is an objective, sometimes it is and sometimes it isn't and there is nothing wrong, better or one approach outweighting the other with that.

Not saying your PR is not valid just pointing out your justification for it is mood ( you could have simply said it was an regression ) anyways @sshedi you might want to chime in on this since you made the change and thus you might be affected if and then when we revert it.

sshedi:
`#define _GNU_SOURCE` is needed and it is present in musl library as well.

musl introduced `gettid()` in 2020
link: https://git.musl-libc.org/cgit/musl/commit/?id=d49cf07541bb54a5ac7aec1feec8514db33db8ea

So, you can't use it with older version of musl anyway. So, musl should add some condition check macros to handle this scenario. This api is present only in musl v1.2.2 and above.

And not all glibc systems support `SYS_gettid` calls.

To keep things simple, add a check in https://github.com/dracutdevs/dracut/blob/master/configure and if gettid not found, abort the build. I think this is the easy way out of this issue.

LaszloGombos:
Thanks @johannbg and @sshedi for the quick feedback.

@sshedi checking for SYS_gettid build time works well both for glibc and musl, so the GLIBC_PREREQ test is not required.

sshedi:
> Thanks @johannbg and @sshedi for the quick feedback.
> 
> @sshedi checking for SYS_gettid build time works well both for glibc and musl, so the GLIBC_PREREQ test is not required.

That's good. It would be good to have this check in configure script and check it there. 
And we can move `#define gettid()    ((pid_t) syscall(SYS_gettid))` to `util.h`, I think this will be lot cleaner than the existing approach.

LaszloGombos:
> It would be good to have this check in configure script and check it there.

Thanks @sshedi for the review and feedback. 
I'd like to keep the scope of this PR to resolve the regression and restore musl support.
There is room for more refactoring, perhaps this can be discussed as a separate PR with the reviewers.

johannbg (approved):
LGTM

PR #1718, opened 2022-02-21
Fix shell options leaking from the 10i18n module.

MichalHe:
The findkeymap function in the `10i18n/module-setup.sh` manipulates the shell  options and relies on restoring them using the trap. However, as the function might be called recursively, each recursive invocation changes the signal handler to its
own. As the recursion is entered with shell options already modified, the changed trap handler is replaced with restoration to the modified shell options, not the original ones. This patch wraps the findkeymap function so that the shellopts are manipulated and restored outside the recursion.

## Changes
- wrap the recursive `findkeymap` function with a wrapper in which the shell options are manipulated and restored correctly

## Checklist

### DIFF
--- a/modules.d/10i18n/module-setup.sh
+++ b/modules.d/10i18n/module-setup.sh
@@ -30,11 +30,7 @@ install() {
     I18N_CONF="/etc/locale.conf"
     VCONFIG_CONF="/etc/vconsole.conf"
 
-    findkeymap() {
-        # shellcheck disable=SC2064
-        trap "$(shopt -p nullglob globstar)" RETURN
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
Sorry for the late review we are all quite busy. 

minor nitpick you need to fix your commit message 
change it from
`fix: stop leaking shell options from the 10i18n module`
to
`fix(10i18n): stop leaking shell options`

other than that LGTM

MichalHe:
@johannbg I've changed the commit message according to your comment and rebased the PR against the current master.

johannbg (approved):

PR #1727, opened 2022-02-24
fix(iscsi): Remove unneeded iscsi NOP-disable code

gonzoleeman:
Commit 7374943ae3d06 ("fix(iscsi): add support for the new iscsiadm
"no-wait" (-W) command") added some code to try to disable iscsi's
NOPs, since they are considered "bad" for an iscsi root disc,
but that code doesn't work because the session is already established
at this point. Open-iscsi will change so that it automatically
disables NOPs for the root disc, so this code can just go away.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/95iscsi/iscsiroot.sh
+++ b/modules.d/95iscsi/iscsiroot.sh
@@ -82,15 +82,6 @@ handle_firmware() {
         if [ $_res -ne 0 ]; then
             warn "iscsiadm: Log-in to iscsi target failed"
         else
-            # get a list of connected targets
-            tgts=$(cat /sys/firmware/ibft/target*/target-name | sort -u)
-            # disable NOPs for each FW target
-            for tgt in ${tgts}; do
-                iscsiadm -m node -T "${tgt}" \
-                    --op update \
-                    --name 'node.conn[0].timeo.noop_out_interval' --value 0 \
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
You need to change your commit msg since you have a title case in your description.
Change it
from
`fix(iscsi): Remove unneeded iscsi NOP-disable code`
to 
`fix(iscsi): remove unneeded iscsi NOP-disable code`

Other than that LGTM, 

@tblume might want to chime in on this and or ack/nack it.

gonzoleeman:
> You need to change your commit msg since you have a title case in your description. Change it from `fix(iscsi): Remove unneeded iscsi NOP-disable code` to `fix(iscsi): remove unneeded iscsi NOP-disable code`
> 
> Other than that LGTM,
> 
> @tblume might want to chime in on this and or ack/nack it.

Thank you, I didn't know about the non-upper-case convention.

Assuming all else is fine, should I modify my branch and force push it, or should I cancel this request and push a new branch?

johannbg:
Modify your branch and force push

tblume:
> @tblume might want to chime in on this and or ack/nack it.

ack

gonzoleeman:
I updated the commit message to remove upper case from the subject line, and I force-pushed the update.

johannbg (approved):

PR #1728, opened 2022-02-25
fix(dracut.sh): apply microcode updates properly in debian

savyajha:
Instead of `<microcode>.early`, Debian has some microcode files with names
of the sort `<microcode>.initramfs`.

This pull request changes...

## Changes

Allows dracut to pull in microcode data of the form <microcode>.initramfs, which seems to be the convention in debian.

## Checklist

Fixes #


johannbg:
@savyajha it's unrealistic to expect upstreams to carry downstream deviation for all the ca 500 linux distribution out there without said distribution having a sound technical reason for doing what they are doing causing the deviation from other distributions. 

To me this looks like a downstream packaging bug which should be resolved within the Debian community with the maintainers of the affected components.

savyajha:
I will take it up with them. Thanks!

PR #1734, opened 2022-03-03
fix(dracut.sh): drop restorecon call

dtardon:
This pull request drops `restorecon` call from dracut.sh, as AFAICT there is no need for it. `cp` already applies appropriate label on the target file (unless `--preserve=context` is used).

## Changes
- drop `restorecon` call

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2558,8 +2558,6 @@ else
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

PR #1741, opened 2022-03-11
[RFC/WIP] dmsquash-live: add support for iso-scan.ram kernel parameter

champtar:
iso-scan.ram copy the ISO to RAM allowing in place reinstalls
without pulling install source from the network

An exemple in place kickstart reinstall using kexec:
```
echo ks.cfg | cpio -H newc -o >> initrd.img
kexec -l vmlinuz --initrd=initrd.img --command-line="ks=file:/ks.cfg \
iso-scan/filename=/AlmaLinux-8.5-x86_64-dvd.iso \
iso-scan.ram=yes \
inst.repo=hd:LABEL=AlmaLinux-8-5-x86_64-dvd"
kexec -e
```

This pull request changes...

## Changes
Allow to load ISO in RAM

## Checklist

Fixes #

### DIFF
--- a/modules.d/90dmsquash-live/iso-scan.sh
+++ b/modules.d/90dmsquash-live/iso-scan.sh
@@ -8,6 +8,8 @@ isofile=$1
 
 [ -z "$isofile" ] && exit 1
 
+copytoram=$2
+
 ismounted "/run/initramfs/isoscan" && exit 0
 
 mkdir -p "/run/initramfs/isoscan"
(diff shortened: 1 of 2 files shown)


LaszloGombos:
Could we re-use the existing rd.live.ram=1 parameter instead of introducing the new iso-scan.ram=yes parameter ?

champtar:
`rd.live.ram` is specific to live images, `iso-scan.ram` (maybe `iso-scan/ram` would be a better name) is not specific to live images at all.
It might make sense to load the full ISO in RAM even for live images in some weird cases, reusing `rd.live.ram` give us less control on what we want to do, and would requires more code changes.

PR #1743, opened 2022-03-12
fix(dmsquash-live): mount live device with the correct type

LaszloGombos:
fstype is not the correct type for the live device.

This pull request changes...

## Checklist

Fixes #


LaszloGombos:
This bug has been undetected for a period of time as $fstype is usually set to auto. If $fstype is not set to auto mount likely fails.

johannbg (approved):

PR #1746, opened 2022-03-13
feat(kernel-modules-base): split up kernel-modules dracut module

LaszloGombos:
Basic kernel module installation to initramfs stays in kernel-modules.
rd.driver command line options are now handled in kernel-modules-base.
kernel-modules now depends on kernel-modules-base to maintain
current functionality.

The main motivation for this change is to enable rd.driver
command line options (via the kernel-modules-base module)
when kernel-modules dracut module is not installed
(e.g. when --add-drivers is used to manage modules manually).

This pull request changes...

## Checklist

Fixes #


johannbg:
Hmm why would you not want to install the kernel dracut module as is?

LaszloGombos:
> why would you not want to install the kernel dracut module as is?

As an example (this happens to be one of the ways I use dracut)

dracut --add-drivers 'autofs4 overlay isofs ahci' --modules "dmsquash-live kernel-modules"
gives me drivers under /lib/modules that is about 70M in size (uncompressed).

dracut --add-drivers 'autofs4 overlay isofs ahci' --modules "dmsquash-live kernel-modules-base" on the other hand is only about 2MB in size for drivers and can boot just as well in my environment. 

This might not be a popular use case, but beyond the significant difference in size in this particular example, I think there is an argument for better modularity and let each module be good at one thing and one thing only.

PR #1748, opened 2022-03-14
fix(98dracut-systemd): fix service dependencies

dtardon:
This pull request fixes dependencies of `dracut-systemd.service`. `dracut-systemd.service` is not an early boot service, therefore it should not use `DefaultDependencies=no`. This automatically fixes the service's ordering dependencies, as in its current state it is missing `Before=shutdown.target umount.target`.

## Checklist


johannbg (changes requested):
We dont include module numbers in commit messages since those are subjected to change so change your commit message
from
`fix(98dracut-systemd): fix service dependencies`
to
`fix(dracut-systemd): fix service dependencies`

Other than that LGTM

johannbg (approved):

PR #1749, opened 2022-03-14
fix(udev-rules): add cdrom udev rules by default

LaszloGombos:
Include cdrom udev rules by default, even if
systemd-udevd or dmsquash-live dracut modules
are not installed.

This pull request changes...

## Checklist

Fixes #


johannbg (approved):
LGTM

PR #1751, opened 2022-03-16
fix(bluetooth): make $dbussystem/bluetooth.conf optional

aafeijoo-suse:
Do not display an error message if the `$dbussystem/bluetooth.conf` file don't exist.
This line was added in commit https://github.com/dracutdevs/dracut/commit/34b1dd2e26c343e9000094db01a7985b6851adf1 to include the missing `/usr/share/dbus-1/system.d/bluetooth.conf` in Arch Linux, but not all Linux distributions ship this file (e.g.: Fedora, SUSE...).

## Checklist

### DIFF
--- a/modules.d/62bluetooth/module-setup.sh
+++ b/modules.d/62bluetooth/module-setup.sh
@@ -55,7 +55,7 @@ install() {
     shopt -q -s nullglob globstar
     local -a var_lib_files
 
-    inst_multiple \
+    inst_multiple -o \
         "$dbussystem"/bluetooth.conf \
         "${systemdsystemunitdir}/bluetooth.target" \
         "${systemdsystemunitdir}/bluetooth.service" \


johannbg:
The file exist on Fedora ( and probably Suse too ) it's just being placed in the wrong location ( /etc ) in otherwords it's a downstream packaging bug that needs fixing.

johannbg:
This also could be an upstream ( default configuration ) issue in which upstream bluez incorrectly places the file in /etc by default and that error get's mirrored downstream but I assume that @hadess and the rest of the bluez maintainers are well aware that they ( like every other upstreams ) are in the role of the vendor thus should be defaulting to use /usr not /etc.

The alternative is for us to assume the role of the vendor and ship that file ourselves with the bluetooth module ( which we may have to do anyway ) 

Either way we should not be papering over the problem when the file is not found and regardless of what we do upstream bluez should fix this if it's a configuration bug ( which defaults to use /etc as opposed to /usr ) with them or downstream packages issue which incorrectly places the file in /etc.

aafeijoo-suse:
> The alternative is for us to assume the role of the vendor and ship that file ourselves with the bluetooth module ( which we may have to do anyway )

I don't think this is the best solution... it's a file that belongs to bluez.

> Either way we should not be papering over the problem when the file is not found where it's expected to be and regardless of what we do upstream bluez should fix this if it's a configuration bug ( which defaults to use /etc as opposed to /usr ) with them or downstream with it's packaging issue which incorrectly places the file in /etc.

I think you're right, it's not a dracut issue that bluez or downstream packagers put the configuration file in the wrong place. But, on the other hand is common to find the -o option in dracut modules when installing files that must be on the system... that's why I created this PR.

johannbg:
> > This also could be an upstream ( default configuration ) issue in which upstream bluez incorrectly places the file in /etc by default and that error get's mirrored downstream but I assume that @hadess and the rest of the bluez maintainers are well aware that they ( like every other upstreams ) are in the role of the vendor thus should be defaulting to use /usr not /etc.
> 
> You might want to discuss this upstream on the linux-bluetooth mailing-list, because the topic of configuration files was recently there.
> 
> As a sidenote, I don't maintain the bluez package in Fedora, and I'd appreciate if you could ask questions when tagging me (and likely other folks), rather than talk about me in the third person.

Never said you did maintain the package in Fedora I was mentioning you as an upstream. The rest of your reply I take into consideration and have a look at the upstream mailinglist discussion.

johannbg:
> > The alternative is for us to assume the role of the vendor and ship that file ourselves with the bluetooth module ( which we may have to do anyway )
> 
> I don't think this is the best solution... it's a file that belongs to bluez.

That a file belongs to upstream does not mean that it's tailored to be run in an initrd Image and what we need to look at ( just like any other upstream ) is how much does it cost the projects in terms of development, maintainance, documentation etc. 

Us shipping our own files as a part of a module ( that file could even be sourced from upstreams directly ) costs less ( in terms of maintainance and code size ) then us including files from downstream because then we dont have to deal with the needless fragmentation on the core/baseOS layer.

> I think you're right, it's not a dracut issue that bluez or downstream packagers put the configuration file in the wrong place. But, on the other hand is common to find the -o option in dracut modules when installing files that must be on the system... that's why I created this PR.

Yeah maybe we should deprecate the -o option if it's taking away people incentives to properly fix things at the root cause of the problem. 

In anycase that's a part of a larger dialog.

I will merge this but we really need to push people in the right direction to fix problems where they belong not give them option to workaround them here upstream since if we do then those issues will never get properly fixed.

johannbg (approved):

PR #1759, opened 2022-03-24
Fixes two bugs which cause minutes long boot times caused by unneeded…

tuppero:
… arping to the default gateway

This pull request changes dracut-network.
Fixes two bugs which cause minutes long boot times caused by unneeded "arping -q -f -w 60" to the default gateway
(when unresponsive) even if the NFS server is on the local network.
This was evident in logs like "dracut-initqueue[x]: Resolving x.x.x.34 via ARP on br0 failed".
## Changes
modules.d/40network/net-lib.sh
## Checklist
- [ x] I am providing new code and test(s) for it

Fixes #
- get_netroot_ip() had first two arguments interchanged in 'splitsep ":" "$1" prefix server rest'
- ip_is_local() was returning negated values from 'strstr "$(ip route get "$@" 2> /dev/null)" " via "'


modules.d/40network/net-lib.sh
```diff
@@ -101,7 +101,8 @@ get_netroot_ip() {
 }
 
 ip_is_local() {
-    strstr "$(ip route get "$@" 2> /dev/null)" " via "
+    strstr "$(ip route get "$@" 2> /dev/null)" " via " || return 0
+    return 1
```
nabijaczleweli, line 105: isn't this just `! strstr ...`?

tuppero: Yes, it's the same negation. (I was just following the style of the functions before it).
Thanks

tuppero:
[modules.d/40network/net-lib.sh] still has two obvious bugs.

pvalena:
Please fix the CI issue: `Error: commit_message:1:1: error: use of type tag that's neither 'feat', 'fix' nor whitelisted (build, chore, ci, docs, perf, refactor, revert, style, test, improvement)`

Otherwise LGTM.

LaszloGombos:
@tuppero Are you in a position to edit the commit message title (not the github title).

The following first line (or something similar) would make the commit message checker happy

> fix(network): avoid arping the gateway for a faster boot

LaszloGombos:
not stale

johannbg (dismissed):
Let's start by having the comment message fixed so it passess the commisery check  as @LaszloGombos has already requested.

LaszloGombos (approved):
LGTM

We could make further changes on the syntax on follow-up PRs, but this resolves important bugs that should be included in the next release.

modules.d/40network/net-lib.sh
aafeijoo-suse: The old test is wrong, but now with this one if the `ip` command fails for any reason, `ip_is_local` returns true. Maybe it's good to save the command output in a variable and check that it's not null before calling `strstr`. E.g., see `wait_for_if_link`, `wait_for_if_up`, `wait_for_route_ok`,...

LaszloGombos: good catch, followed up with a commit

aafeijoo-suse (approved):

PR #1760, opened 2022-03-24
feat(connman): introduce connman support module

zboszor:

## Changes

Add a new 35connman module.

This module supersedes issue #1682 as almost all of my Yocto images use connman and the remote booted one was the odd one out. This module allows using connman as the network tool both in initramfs and the squashfs root that the initramfs is generated from.

The network-connman module depends on systemd, since connman, unlike NetworkManager, doesn't have a configure-and-exit mode so a scripted network initialization is not possible. Also, connman deletes the connections when exiting.

## Checklist
- [x] I am providing new code

Fixes #

### DIFF
--- a/modules.d/35connman/cm-config.sh
+++ b/modules.d/35connman/cm-config.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+type cm_generate_connections > /dev/null 2>&1 || . /lib/cm-lib.sh
+
+if [ -n "$netroot" ] || [ -e /tmp/net.ifaces ]; then
+    echo rd.neednet >> /etc/cmdline.d/connman.conf
+fi
+
+if getargbool 0 rd.debug -d -y rdinitdebug -d -y rdnetdebug; then
+    if [ -n "$DRACUT_SYSTEMD" ]; then
+        # Enable tty output if a usable console is found
+        # shellcheck disable=SC2217
+        if [ -w /dev/console ] && (echo < /dev/console) > /dev/null 2> /dev/null; then
+            mkdir -p /run/systemd/system/cm-initrd.service.d
+            cat << EOF > /run/systemd/system/cm-initrd.service.d/tty-output.conf
+[Service]
+StandardOutput=tty
+EOF
+            systemctl --no-block daemon-reload
+        fi
+    fi
+fi
+
+cm_generate_connections
--- a/modules.d/35connman/cm-initrd.service
+++ b/modules.d/35connman/cm-initrd.service
@@ -1,0 +1,24 @@
+[Unit]
+DefaultDependencies=no
+Wants=systemd-udev-trigger.service
+After=systemd-udev-trigger.service
+After=dracut-cmdline.service
+Wants=network.target
+Before=network.target
+RequiresMountsFor=/var/lib/connman
+After=dbus.service
+ConditionPathExists=/run/connman/initrd/neednet
+
+[Service]
+Type=dbus
+BusName=net.connman
+Restart=on-failure
+ExecStart=/usr/sbin/connmand -n
+StandardOutput=null
+CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_BIND_SERVICE CAP_NET_RAW CAP_SYS_TIME CAP_SYS_MODULE CAP_SYS_ADMIN
+ProtectHome=true
+ProtectSystem=full
+
+[Install]
+WantedBy=initrd.target
+Also=cm-wait-online-initrd.service
--- a/modules.d/35connman/cm-lib.sh
+++ b/modules.d/35connman/cm-lib.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+type getcmdline > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+cm_generate_connections() {
+    if getargbool 0 rd.neednet; then
+        mkdir -p "$hookdir"/initqueue/finished
+        echo '[ -f /tmp/cm.done ]' > "$hookdir"/initqueue/finished/cm.sh
+        mkdir -p /run/connman/initrd
+        : > /run/connman/initrd/neednet # activate ConnMan services
+    fi
+}
--- a/modules.d/35connman/cm-run.sh
+++ b/modules.d/35connman/cm-run.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+type source_hook > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+if [ -e /tmp/cm.done ]; then
+    return
+fi
+
+while read -r _serv; do
+    ifname=$(connmanctl services "$_serv" | grep Interface= | sed 's/^.*Interface=\([^,]*\).*$/\1/')
+    source_hook initqueue/online "$ifname"
+    /sbin/netroot "$ifname"
+done < <(connmanctl services | grep -oE '[^ ]+$')
+
+: > /tmp/cm.done
--- a/modules.d/35connman/cm-wait-online-initrd.service
+++ b/modules.d/35connman/cm-wait-online-initrd.service
@@ -1,0 +1,16 @@
+[Unit]
+DefaultDependencies=no
+Requires=cm-initrd.service
+After=cm-initrd.service
+Before=network-online.target
+Before=dracut-initqueue.service
+ConditionPathExists=/run/connman/initrd/neednet
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/connmand-wait-online
+RemainAfterExit=yes
+
+[Install]
+WantedBy=initrd.target
+WantedBy=network-online.target
--- a/modules.d/35connman/module-setup.sh
+++ b/modules.d/35connman/module-setup.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# called by dracut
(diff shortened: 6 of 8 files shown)


modules.d/35network-connman/module-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+    return 0
+}
+
+# called by dracut
+installkernel() {
+    return 0
+}
+
+# called by dracut
+install() {
```
johannbg, line 23: Here you are missing optional install of /usr/lib/connman/wired-setup in conjuction with the ExecStartPre=-/usr/lib/connman/wired-setup in cm-initrd.service. 

If it's not needed remove that ExecStartPre line from the type unit.

johannbg (changes requested):
Remove the module number from the commit message we dont include those since those numbers are subjected to change.

modules.d/35network-connman/cm-config.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+type cm_generate_connections > /dev/null 2>&1 || . /lib/cm-lib.sh
+
+if [ -n "$netroot" ] || [ -e /tmp/net.ifaces ]; then
+    echo rd.neednet >> /etc/cmdline.d/35-neednet.conf
```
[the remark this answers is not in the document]
zboszor: It's a modified copy of the network-manager module. I thought the number has meaning there.

modules.d/35network-connman/cm-initrd.service
```diff
@@ -0,0 +1,26 @@
+[Unit]
+DefaultDependencies=no
+Wants=systemd-udev-trigger.service
+After=systemd-udev-trigger.service
+After=dracut-cmdline.service
+Wants=network.target
+Before=network.target
+RequiresMountsFor=/var/lib/connman
+After=dbus.service
+Conflicts=systemd-resolved.service
```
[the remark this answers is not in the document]
zboszor: That's what Yocto adds to the service.

modules.d/35network-connman/module-setup.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    require_binaries sed grep || return 1
```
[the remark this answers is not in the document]
zboszor: Then the network-manager module also misses them and fails during the install phase.

modules.d/35network-connman/module-setup.sh
```diff
@@ -0,0 +1,45 @@
...
+install() {
+    # We don't need `ip` but having it is *really* useful for people debugging
+    # in an emergency shell.
+    inst_multiple ip sed grep
+
+    inst connmand
+    inst connmanctl
+    inst connmand-wait-online
+    inst "$dbussystem"/connman.conf
+    inst_multiple -o /etc/connman/main.conf
```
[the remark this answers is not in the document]
zboszor: Please explain.

modules.d/35network-connman/module-setup.sh
zboszor: > If it's not needed remove that ExecStartPre line from the type unit.

Removed the ExecStartPre line.

modules.d/35network-connman/module-setup.sh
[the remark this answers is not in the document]
zboszor: There's no difference between network-manager and network-connman regarding netroot:
```
$ grep netroot 35network-{manager,connman}/*
35network-manager/nm-config.sh:if [ -n "$netroot" ] || [ -e /tmp/net.ifaces ]; then
35network-manager/nm-run.sh:    /sbin/netroot "$ifname"
35network-connman/cm-config.sh:if [ -n "$netroot" ] || [ -e /tmp/net.ifaces ]; then
35network-connman/cm-run.sh:    /sbin/netroot "$ifname"
```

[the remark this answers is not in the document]
zboszor: It's the network module that installs netroot so it's not actually needed to be a required binary:
```
install() {
    inst_script "$moddir/netroot.sh" "/sbin/netroot"
    ...
```

modules.d/35network-connman/cm-initrd.service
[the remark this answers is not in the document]
zboszor: Removed the Conflicts line.

modules.d/35network-connman/cm-config.sh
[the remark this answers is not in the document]
zboszor: I renamed the file.

zboszor:
> Remove the module number from the commit message we dont include those since those numbers are subjected to change.

Done.

modules.d/35network-connman/module-setup.sh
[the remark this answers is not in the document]
zboszor: Nevermind, I think I found what I needed to do.

modules.d/35network-connman/module-setup.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    require_binaries sed grep connmand connmanctl connmand-wait-online || return 1
```
johannbg, line 5: You are still missing a requirement on netroot which is called in cm-run.sh

johannbg (changes requested):
Still a missing requirement on netroot other than that looks good to me, I'll probably merge this once that is fixed

modules.d/35network-connman/module-setup.sh
zboszor: See my previous comments. There is no "netroot" binary anywhere which is installed by the "40network" module.
Neither of the `35network-*` modules need it as a required binary.

zboszor:
```
$ grep netroot modules.d/??network*/module-setup.sh
modules.d/40network/module-setup.sh:    inst_script "$moddir/netroot.sh" "/sbin/netroot"
$
```

zboszor: @johannbg I just added a new commit to add netroot as a required binary for every `network-*` module. If it's needed for network-connman, then it won't break CI for network-legacy and network-manager.

zboszor: @johannbg As I suspected, it did break CI for every test using network-legacy or network-manager. I will drop the second commit.

johannbg: Yeah you should not have added netroot as a requirement for every network-* module since each module has it's own requirements and meta modules should not be used for somekind of "shared" dependencies since meta module and modules listed in meta modules should work independendly of the meta module itself and contain everything necessary to work properly by themselves.

btw somethin I missed in the review can you drop `network-` from the module name and simply call the module connman so it's in line with network-manager and systemd-networkd.

zboszor: I renamed the module to "connman".

johannbg: Yup it's necessary and dont forget to update the commit message in the process ( feat(connman): ... )

zboszor: Then I will copy the netroot.sh script from the network module and install that if that's what you want. Keep in mind that connman will be the sole module doing this. None of systemd-networkd, network-legacy, network-wicked or network-manager needs anything, they rely on the upper level network module. Why do you want connman to be the odd one out?

zboszor: I copied netroot.sh from 40network and install it from the connman module now.

johannbg (approved):
LGTM

zboszor:
> LGTM

@johannbg Wouldn't it look even better if the netroot script wasn't copied (allowing deviation in the long run) but install it from the network module directory like:
```
inst_script "$moddir/../40network/netroot.sh" "/sbin/netroot"
```
or symlink it instead of keeping a separate copy?

johannbg:
@zboszor we should not be putting something in the meta module ( like the entire existing install() section in the network module ). 

There is no guarantee that the modules defined in the meta module use what is provided there thus you end up installing something on the host that is not used ( increased image size for no gain ) and if it's used, it forces vendors/users to also include the meta module for the actual module that is being used, to work.

Meta modules are also hard to express from a downstream packaging point of view since they dont have any clear dependency's and on top of that, with them, we are adding to the cost of development since we now have two modules to think of, the meta module + the actual module being used.

So in the long run we remove the added historic crap that currently exist in the meta module and just have them strictly include other modules depending on some conditions what's install on the host system ( we will use meta modules for example when we split modules based on the init system in use as in systemd vs non systemd so we dont end up breaking vendors/users system ) and have the modules themselves include everything that they require to work.

Conan-Kudo:
Could we please prefix this with `network-` like the others have?

johannbg:
> Could we please prefix this with `network-` like the others have?

No since adding a modules role in front of a module name is nonsense.  
And dont you just mean like wicked has which was a mistake on Daniels behalf calling it that.

network <-- no prefix and adding would just be silly network-network 
network-legacy <-- needed to be called something different since network was turned into a meta module 
network-manager <-- no prefix
systemd-networkd <-- not prefix
connman <-- no prefix
iwd <-- will have no prefix
which leaves just network-wicked which should have just been called wicked as upstream named the project.

Conan-Kudo (approved):
fine, lgtm

PR #1762, opened 2022-03-25
Avoid using nullglob option and fix dracut-install

pvalena:
Glob is in most cases not expected to evaluate to null. To avoid unexpected failures, don't use the `nullglob` option.

I've found no code (WRT the modifications applied) where nullglob is needed, as the variable is either checked explicitly by `[[ -f $file ]]`, or passed to `inst_multiple -o` where the file string is actually expected instead. I think more code is actually expecting an invalid argument instead of missing one (f.e. `cp`), so this increases resiliency.

## Changes
 - removed `nullglob` option use
 - alternative implementation for findkeymap (10i18n module) with `mapfile` (recursion overwrites `trap` setting)
 - fixed `dracut-install` not to fail when `-o` is specified, but no source is specified

## Checklist
_ _ _ _

## Additional

E.g., a PR which I don't think should be needed: https://github.com/msekletar/prefixdevname/pull/1

## Tests
keymapfile: https://gist.github.com/f6048df4d7f2a2d609974283a674b9fd
dracut-install: https://gist.github.com/07171bb36863394728852810b21c026f

### DIFF
--- a/docs/BASH.md
+++ b/docs/BASH.md
@@ -16,8 +16,8 @@ Don't use `dirname`, use:
 If you set `shopt` in a function, reset to its default state with `trap`:
 ```shell
 func() {
-  trap "$(shopt -p nullglob globstar)" RETURN
-  shopt -q -s nullglob globstar
+  trap "$(shopt -p globstar)" RETURN
+  shopt -q -s globstar
 }
 ```
 
@@ -51,8 +51,8 @@ func() {
 Or collect the filenames in an array, if you need them more than once:
 ```shell
 func() {
-    trap "$(shopt -p nullglob globstar)" RETURN
-    shopt -q -s nullglob globstar
+    trap "$(shopt -p globstar)" RETURN
+    shopt -q -s globstar
 
     filenames=( /usr/lib*/**/lib*.a )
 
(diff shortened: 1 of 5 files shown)


pvalena:
`dracut-install` needs some more work:

```
/usr/lib/dracut/dracut-install -D $(pwd) -H -o
[ 1034.107802] dracut-install[1354]: segfault at 0 ip 00007f64e0f80644 sp 00007fff441cdc68 error 4 in libc.so.6[7f64e0eca000+176000]                                                         "ibm-p8-kvm-03-guest-0" 08:28 28-Mar-22
[ 1034.111051] Code: 84 00 00 00 00 00 f3 0f 1e fa c5 f9 6e c6 89 f8 25 ff 0f 00 00 c4 e2 7d 78 c0 c4 41 31 ef c9 zsh: segmentation fault (core dumped)

```

pvalena:
I've run some tests (automated & manual) and fixed the remaining issues.

## Tests
keymapfile: https://gist.github.com/f6048df4d7f2a2d609974283a674b9fd
dracut-install: https://gist.github.com/07171bb36863394728852810b21c026f

johannbg (approved):
LGTM

pvalena:
Thanks!

lnykryn (approved):
LGTM

PR #1764, opened 2022-03-29
improvement: Add support for single password boot on non-systemd setups.

jovaska1337:
`ask_for_password()` in the crypt module (`modules.d/90crypt/crypt-lib.sh`) can now store passwords in the kernel keyring via `keyctl`, mimicing the behavior of `systemd-ask-password`. This allows `pam_gdm.so` (a standalone PAM module from GDM) to work on non-systemd setups and unlock a GNOME keyring when autologin is enabled without asking for a password. (given that the keyring password is the same as the disk encryption password)

## Changes
- `keyctl` (and `tee`) added as an optional includes in the crypt module. (detected with `find_binary()` in `install()`)
- `ask_for_password()` will pipe passwords to `keyctl` (via a helper script) if it's available and the `--key-name` option is used.
- The helper script (`modules.d/90crypt/password-helper.sh`) is required due to the way plymouth handles the `--command` parameter.
- `modules.d/90crypt/cryptroot-ask.sh` will now try to store the result of `ask_for_password()` in the kernel keyring with the name "cryptsetup", allowing `pam_gdm.so` to work.

## Checklist

### DIFF
--- a/modules.d/90crypt/crypt-lib.sh
+++ b/modules.d/90crypt/crypt-lib.sh
@@ -48,6 +48,10 @@ crypttab_contains() {
 # --tty-echo-off
 #   Turn off input echo before tty command is executed and turn on after.
 #   It's useful when password is read from stdin.
+# --key-name name
+#   Insert the password into the user keyring for later usage.
+#   (This can be used by pam_gdm.so to unlock a GNOME keyring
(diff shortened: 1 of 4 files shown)


jovaska1337:
This should now pass all the required checks.

PR #1765, opened 2022-03-29
Wrap __GLIBC_PREREQ usage with __GLIBC__

ismaell:
Fixes build against musl 1.2.2+; earlier versions don't implement it but
there's no way to know it's presence other than an autoconf-style test.


ismaell:
Fixed in ce55a85ed5d902c19d75895508856f96ec2ceb1a.

PR #1769, opened 2022-04-01
fix(resume): correct call to block_is_netdevice function

aafeijoo-suse:
The `block_is_netdevice` function requires the device in <maj:min> format, but the `swap_devs` array is populated with the content of /proc/swaps, which prints the devices using their paths from /dev.

This causes the check method to never detect if swap is mounted on a network device.

## Checklist


johannbg (approved):
LGTM

PR #1775, opened 2022-04-06
Avoid calling unavailable dracut-logger functions from boot scripts

aafeijoo-suse:
The dracut-logger functions are only available during the initrd generation.
For internal initrd scripts that run at boot, dracut-lib provides `warn()` and `info()`.

This PR should cover all the wrong calls:
```
# grep -r -n -w -e dfatal -e derror -e dwarn -e dwarning -e dinfo -e ddebug -e dtrace modules.d | grep -v module-setup
modules.d/35network-manager/nm-lib.sh:14:        derror "nm-initrd-generator not found"
modules.d/45ifcfg/write-ifcfg.sh:107:        derror "Cannot find network interface '$_netif'!"
```

## Checklist


johannbg (approved):
LGTM

PR #1777, opened 2022-04-07
fix(nvmf): set executable bit on `nvmf-autoconnect.sh`

bgilbert:
## Changes

`nvmf-autoconnect.sh` is installed in `/sbin`, and run via `initqueue` in the special case of FC transport without a `discovery.conf` and without kernel arguments.  If the executable bit is not set, `initqueue` fails with `Invalid command`.

Caught by CI checking for non-executable files in binary directories: https://github.com/coreos/fedora-coreos-tracker/issues/1155

## Checklist
- [ ] I have tested it locally - no access to FC-transport NVMe


johannbg (approved):
LGTM

PR #1779, opened 2022-04-09
UEFI: Use kernel version if no VERSION value is given in os-release

saint-erk:
This pull request changes the uefi section of dracut.sh to create a local temporary copy of os-release if the system-provided os-release file does not specify a value for VERSION. (Systemd-boot / gummiboot expects that VERSION is present to automatically files saved to /EFI/Linux/*.efi, and some distributions like Void Linux don't set a VERSION value.)

## Checklist
- [N/A] I have reviewed and updated any documentation if relevant
- [Code only] I am providing new code and test(s) for it

Fixes #N/A


johannbg:
That's a nack we will not workaround operating system vendor who does not populating values of os-release fields that other software depends upon nor will we workaround software that depends on values in the os-release file and does not work when it does not find one. 

You either need work with the void linux community to populate that field with whatever the void community agrees upon or you need to file a bug against sd-boot for being dependent on that field in the first place.

PR #1782, opened 2022-04-11
fix(lvm): add missing grep requirement

aafeijoo-suse:
Since commit https://github.com/dracutdevs/dracut/commit/7ffc5e38 `lvm_scan.sh` requires `grep`.

https://github.com/dracutdevs/dracut/blob/a7a4b76c4ad5794f5f8a24ecd5b495f1512d05f7/modules.d/90lvm/lvm_scan.sh#L55-L65

## Checklist


johannbg (approved):
LGTM

PR #1784, opened 2022-04-20
feat(virtiofs): virtiofs root filesystem support

germag:
Currently, dracut can't mount virtiofs as root filesystem.
Make possible to boot virtual machines off virtiofs,
passing "root=virtiofs:\<mtag\>", using the mount tag <mtag>
as root filesystem. Alternatively this module also supports
"rootfstype=virtiofs root=\<mtag\>"

## Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -228,6 +228,9 @@ udev-rules:
 virtfs:
   - modules.d/95virtfs/*
 
+virtiofs:
+  - modules.d/95virtiofs/*
+
 zfcp:
   - modules.d/95zfcp/*
 
--- a/modules.d/95virtiofs/mount-virtiofs.sh
+++ b/modules.d/95virtiofs/mount-virtiofs.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/sh
+
+if [ "${fstype}" = "virtiofs" -o "${root%%:*}" = "virtiofs" ]; then
+    if ! { modprobe virtiofs || strstr "$(cat /proc/filesystems)" virtiofs; }; then
+        die "virtiofs is required but not available."
+    fi
+
+    mount -t virtiofs -o "$rflags" "${root#virtiofs:}" "$NEWROOT" 2>&1 | vinfo
+    if ! ismounted "$NEWROOT"; then
+        die "virtiofs: failed to mount root fs"
+        exit 1
+    fi
+
+    info "virtiofs: root fs mounted (options: '${rflags}')"
+
+    [ -f "$NEWROOT"/forcefsck ] && rm -f -- "$NEWROOT"/forcefsck 2> /dev/null
+    [ -f "$NEWROOT"/.autofsck ] && rm -f -- "$NEWROOT"/.autofsck 2> /dev/null
+fi
+:
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -928,6 +928,25 @@ NOTE:
     If "exportname" instead of "port" is given the standard port is used.
     Newer versions of nbd are only supported with "exportname".
 
+VIRTIOFS
+~~~~~~~~
+**root=**virtiofs:__<mount-tag>__::
+    mount virtiofs share using the tag <mount-tag>.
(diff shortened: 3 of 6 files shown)


modules.d/95virtiofs/mount-virtiofs.sh
```diff
@@ -0,0 +1,40 @@
...
+virtiofs_fstype_opt() {
+	[ -n "$fstype" -a "$fstype" = "virtiofs" ]
+}
+
+virtiofs_root_opt() {
+	[ -n "$root" -a -z "${root%%virtiofs:*}" ]
+}
+
+virtiofs_root() {
+	[ virtiofs_fstype_opt -o virtiofs_root_opt ]
```
dtardon, line 12: This line doesn't do what you think it does. It's the equivalent of 
```sh
[ -n 'virtiofs_fstype_opt' -o -n 'virtiofs_root_opt' ]
```
You probably meant to write
```sh
virtiofs_fstype_opt || virtiofs_root_opt
```

modules.d/95virtiofs/mount-virtiofs.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+load_module() {
+	if ! (modprobe virtiofs || strstr "$(cat /proc/filesystems)" virtiofs); then
```
dtardon, line 16: Please use
```sh
if ! { modprobe virtiofs || strstr "$(cat /proc/filesystems)" virtiofs; }; then
```
instead. There's no need to run a subshell here.

modules.d/95virtiofs/mount-virtiofs.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/sh
+
+virtiofs_fstype_opt() {
+	[ -n "$fstype" -a "$fstype" = "virtiofs" ]
```
dtardon, line 4: It's not necessary to test `$fstype` for non-emptiness if you then compare it with a concrete string, is it?

modules.d/95virtiofs/mount-virtiofs.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+virtiofs_root_opt() {
+	[ -n "$root" -a -z "${root%%virtiofs:*}" ]
```
dtardon, line 8: I think that
```sh
[ ${root%%:*} = "virtiofs" ]
```
is a bit easier to understand. But YMMV.

modules.d/95virtiofs/mount-virtiofs.sh
```diff
@@ -0,0 +1,40 @@
...
+	fi
+}
+
+mount_root() {
+	local fstag=${root#virtiofs:}
+	mount -t virtiofs -o "$rflags" "$fstag" "$NEWROOT" 2>&1 | vinfo
+}
+
+if virtiofs_root; then
+	load_module
```
dtardon, line 27: Btw, are all these small functions, each only called from a single place, really necessary? The code would be half of the current length without them and IMHO still perfectly understandable...

modules.d/95virtiofs/mount-virtiofs.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+load_module() {
+	if ! (modprobe virtiofs || strstr "$(cat /proc/filesystems)" virtiofs); then
+		die "virtiofs is required but not available."
+	fi
+}
+
+mount_root() {
+	local fstag=${root#virtiofs:}
```
dtardon, line 22: IMHO there's no need for a variable here. Just use the expression directly.

.github/labeler.yml
```diff
@@ -228,6 +228,9 @@ udev-rules:
 virtfs:
   - modules.d/95virtfs/*
 
+virtiofs:
```
dtardon, line 231: Nitpick: I'd split this into a separate commit, as it's an infrastructure thing, not a part of the feature.

germag:
v3:
- Fix shellcheck error
When I ran 'shfmt -s' locally (in v2), it found no issues

johannbg (approved):
LGTM

Nice touch with the label, Appreciated thanks.

PR #1634, opened 2021-11-07
fix(dracut-initramfs-restore.sh): add missing default paths

johannbg:
Adding missing default paths.
Fixes #1628

This pull request changes...

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

Fixes #

Issue #1628, opened 2021-10-25
dracut-initramfs-restore fails if kernel+initrd are located in /boot/Default instead of /boot/${MACHINE_ID}


steffenweber:
**Describe the bug**
During shutdown, the following error is logged: `dracut-initramfs-restore[1234]: No initramfs image found to restore!`

**Distribution used**
Gentoo Linux

**Dracut version**
dracut 055

**Init system**
systemd 249

**To Reproduce**
The systemd script [kernel-install](https://github.com/systemd/systemd/blob/v249/src/kernel-install/kernel-install) prefers the directory `/boot/Default` (if it exists) over `/boot/${MACHINE_ID}`. The script [dracut-initramfs-restore.sh](https://github.com/dracutdevs/dracut/blob/master/dracut-initramfs-restore.sh) doesn't seem to know about this location and therefore fails to find the initrd file during shutdown.

**Expected behavior**
The script should find the initrd location and the error message should go away.

**Additional context**
See https://github.com/systemd/systemd/pull/16777 for details on the `/boot/Default` directory.

johannbg:
@steffenweber could you confirm or deny that PR #1634 fixes things for you?

steffenweber:
I can confirm that the patch fixes the issue. Thank you!

PR #1786, opened 2022-04-22
Add missing dracut default output files

aafeijoo-suse:
Consistent with issue #1628 and PR #1634, add missing dracut default output files.

When using systemd-boot and kernel-install, dracut runs a hook script (install.d/50-dracut.install) to generate the initrd at the correct place in the ESP. However, when calling dracut manually the initrd is saved in the traditional place in /boot. The same happens with lsinitrd, it does not detect the default initrd.

Fix the auto detection that doesn't notice the bootloader spec installation.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1028,7 +1028,15 @@ stdloglvl=$((stdloglvl + verbosity_mod_l))
 
 if ! [[ $outfile ]]; then
     if [[ $machine_id != "no" ]]; then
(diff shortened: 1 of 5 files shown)


johannbg (commented):
"Default" is not a valid machine-id, where does this come from?

johannbg:
> "Default" is not a valid machine-id, where does this come from?

Oh never mind I remember now.

johannbg (approved):

PR #1790, opened 2022-04-25
fix(fedora.conf): vi binary is missing

mmatsuya:
This pull request changes...

Changes
/usr/bin/vi is just a wrapper shell script to run /usr/libexec/vi which is the vi binary.
/usr/bin/vi is available in initramfs.img, but /usr/libexec/vi is missing.
As as result, the vi command never works.
This PR is to add /usr/libexec/vi.

## Checklist

### DIFF
--- a/dracut.conf.d/fedora.conf.example
+++ b/dracut.conf.d/fedora.conf.example
@@ -8,7 +8,7 @@ i18n_install_all="yes"
 
 stdloglvl=3
 sysloglvl=5
-install_optional_items+=" vi /etc/virc ps grep cat rm "
+install_optional_items+=" vi /usr/libexec/vi /etc/virc ps grep cat rm "
 prefix="/"
 environment=/usr/lib/environment.d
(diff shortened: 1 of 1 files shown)


johannbg:
The reason why the other one was autoclosed is simply because I have very limited time to work on dracut ( and other things for that matter ) for the past weeks which is why I was always removing the stale bot label on PR's which you yourself ( and anyone else ) can do by simply commenting on the PR's 

Filing a new PR when the previous PR did not get any review will get you nowhere. 
 
Any PR with reference to RHEL will be closed since those issues should be dealt with by RH customer support, it's team and it's processes. 

Fix your comment body to match the one you had previously on your PR and fix your commit msg title to `fix(fedora.conf): vi binary is missing` and then I will merge this.

mmatsuya:
@johannbg, Thanks! I understood the situation. 
I changed the msg title and commit message.

Let me know if there is anything else I can do.

johannbg (approved):

PR #1791, opened 2022-04-25
Use Documentation= to point to man page

dtardon:
This pull request improves dracut's service files a tiny bit.

## Changes
- Remove reference to dracut.bootup(7) from two services that are not part of boot process.
- Use Documentation= directive instead of a comment to refer to a man page.

## Checklist

### DIFF
--- a/modules.d/98dracut-systemd/dracut-cmdline-ask.service
+++ b/modules.d/98dracut-systemd/dracut-cmdline-ask.service
@@ -1,9 +1,8 @@
 #  This file is part of dracut.
-#
-# See dracut.bootup(7) for details
 
 [Unit]
 Description=dracut ask for additional cmdline parameters
+Documentation=man:dracut.bootup(7)
 DefaultDependencies=no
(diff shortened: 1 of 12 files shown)


johannbg (changes requested):
Drop the puncation in your commit msg 
`fix(dracut-systemd): drop misleading man page ref.`
Multiple man page reference should be kept in a single Documentation= line so fix where applicable.

dtardon:
> Drop the puncation in your commit msg `fix(dracut-systemd): drop misleading man page ref.`

Sigh. That dot ends an abbreviation, not a sentence... Is the use of that commisery tool really worth it?

> Multiple man page reference should be kept in a single Documentation= line so fix where applicable.

As you wish.

johannbg:
> > Drop the puncation in your commit msg `fix(dracut-systemd): drop misleading man page ref.`
> 
> Sigh. That dot ends an abbreviation, not a sentence... Is the use of that commisery tool really worth it?

Yes since I plan on automate the release process.

PR #1794, opened 2022-04-26
Fix firmware loading paths, supprt .zst-suffixed firmware

nabijaczleweli:
## Changes
The firmware loading path is
1. contents of /sys/module/firmware_class/parameters/path (`fw_path_para`), if any
2. /lib/firmware/updates/$(uname -r)
3. /lib/firmware/updates
4. /lib/firmware/$(uname -r)
5. /lib/firmware

Always.

Repeated single-character strstr()s -> strpbrk() and stat()s with unused buffers -> access()es.

Also check for `.zst`-suffixed firmware, coming in 5.19: https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/driver-core.git/commit/?h=driver-core-next&id=23cfbc6ec44e5e80d5522976ff45ffcdcddfb230

62bluetooth `inst_multiple`s its firmware (for some reason?); glob at the end of the unglobbed paths to catch compressed fw.

### DIFF
--- a/modules.d/62bluetooth/module-setup.sh
+++ b/modules.d/62bluetooth/module-setup.sh
@@ -28,24 +28,23 @@ depends() {
 installkernel() {
     instmods bluetooth btrtl btintel btbcm bnep ath3k btusb rfcomm hidp
     inst_multiple -o \
-        /usr/lib/firmware/ar3k/AthrBT* \
-        /usr/lib/firmware/ar3k/ramps* \
-        /usr/lib/firmware/ath3k-1.fw \
-        /usr/lib/firmware/BCM2033-MD.hex \
-        /usr/lib/firmware/bfubase.frm \
-        /usr/lib/firmware/BT3CPCC.bin \
-        /usr/lib/firmware/brcm/*.hcd \
-        /usr/lib/firmware/mediatek/mt7622pr2h.bin \
-        /usr/lib/firmware/qca/nvm* \
-        /usr/lib/firmware/qca/crnv* \
-        /usr/lib/firmware/qca/rampatch* \
-        /usr/lib/firmware/qca/crbtfw* \
-        /usr/lib/firmware/rtl_bt/* \
-        /usr/lib/firmware/intel/ibt* \
-        /usr/lib/firmware/ti-connectivity/TIInit_* \
-        /usr/lib/firmware/nokia/bcmfw.bin \
-        /usr/lib/firmware/nokia/ti1273.bin
-
+        /lib/firmware/ar3k/AthrBT* \
+        /lib/firmware/ar3k/ramps* \
+        /lib/firmware/ath3k-1.fw* \
+        /lib/firmware/BCM2033-MD.hex* \
+        /lib/firmware/bfubase.frm* \
+        /lib/firmware/BT3CPCC.bin* \
+        /lib/firmware/brcm/*.hcd* \
+        /lib/firmware/mediatek/mt7622pr2h.bin* \
+        /lib/firmware/qca/nvm* \
+        /lib/firmware/qca/crnv* \
+        /lib/firmware/qca/rampatch* \
+        /lib/firmware/qca/crbtfw* \
+        /lib/firmware/rtl_bt/* \
+        /lib/firmware/intel/ibt* \
+        /lib/firmware/ti-connectivity/TIInit_* \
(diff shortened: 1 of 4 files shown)


nabijaczleweli:
Would appreciate guidance w.r.t. what "tags"(??) I'm supposed to put in the commit messages

johannbg:
> Would appreciate guidance w.r.t. what "tags"(??) I'm supposed to put in the commit messages

See HACKING.md in the docs directory

johannbg:
From an OS design perspective vendor-supplied operating system resources should resided within /usr and  /bin, /sbin, /lib  and be symlinks to the /usr file system

What your PR is based on seems to be the kernels firmware`s api search paths thou not technically wrong it's not what vendor ( distributions ) are/should be using so arguably it's a bug with the kernel firmware`s api search paths that it does not include searching under /usr/lib/firmware/ these days

nabijaczleweli:
Thankfully, the system has already been designed by the kernel (firmware lives in `/lib/firmware`), and distributor + local administrator (`/lib` is `/usr/lib` sometimes). On our target system (the initrd), `/lib` is `/usr/lib` always, but that doesn't matter, since that's a superset of a split-/usr system.
This is strictly a conformance issue (we *must* match the search paths and methodology *exactly* to the kernel or we may not pick up the same firmware as it has in the real root).
I don't really see your point.

nabijaczleweli:
Bumping this after a month

johannbg:
Btw this is practically an ack already, minor issues with the commit messages ( I was planning on updating the docs once I notice what you did after I responded with regards to the commit msg but have not had the time to do that nor something else ) I'm going to see if we cant sort that commit message out and any other issue that I might find and have this merged by the end of the week.

modules.d/62bluetooth/module-setup.sh
```diff
@@ -28,24 +28,23 @@ depends() {
...
-        /usr/lib/firmware/qca/crbtfw* \
-        /usr/lib/firmware/rtl_bt/* \
-        /usr/lib/firmware/intel/ibt* \
-        /usr/lib/firmware/ti-connectivity/TIInit_* \
-        /usr/lib/firmware/nokia/bcmfw.bin \
-        /usr/lib/firmware/nokia/ti1273.bin
-
+        /lib/firmware/ar3k/AthrBT* \
+        /lib/firmware/ar3k/ramps* \
+        /lib/firmware/ath3k-1.fw \
```
aafeijoo-suse, line 33: Missing * here, it should be `/lib/firmware/ath3k-1.fw*`

modules.d/62bluetooth/module-setup.sh
```diff
@@ -28,24 +28,23 @@ depends() {
...
+        /lib/firmware/BCM2033-MD.hex* \
+        /lib/firmware/bfubase.frm* \
+        /lib/firmware/BT3CPCC.bin* \
+        /lib/firmware/brcm/*.hcd \
```
aafeijoo-suse, line 37: Missing * here, it should be `/lib/firmware/brcm/*.hcd*`

aafeijoo-suse (changes requested):
I think you should change your commit messages by adding the affected module/part in parentheses, always avoiding using the module number, because it may change in the future. E.g. with 5ab7a78634859141103f92cd87a002f6f0ee8f20: `fix(bluetooth): accept compressed firmwares in inst_multiple`

nabijaczleweli:
Messages updated, `*`es added

johannbg:
Not quite there yet... 
Squash  `fix(cli): default to correct firmware search paths` and `fix(cli): default to correct firmware search paths in dracut-install` together and change the commit message to 
`fix(dracut): default to correct firmware search paths`
also change the commit message from
`feat(rt): support ZSTD-compressed firmware with .zst suffix `
to
`feat(dracut-install): support ZSTD-compressed firmware with .zst suffix `

johannbg:
As I mentioned previously the docs need updating, just see previous commit messages touching these files.

johannbg (changes requested):
Commit msg needs sqashing and rewording as previously mention in a comment.

nabijaczleweli:
squashed, reworded

johannbg (approved):

johannbg:
Thanks for your contribution sorry for the commit nitpicks and the time it took to review and merge this.

PR #1796, opened 2022-04-27
dracut-install cleanup

nabijaczleweli:
## Changes
See individual commit messages; draft because on top of #1794


src/install/dracut-install.c
```diff
@@ -183,9 +183,7 @@ static char *convert_abs_rel(const char *from, const char *target)
         target_dir_p = strndup(target, dirlen);
         if (!target_dir_p)
                 return strdup(from + strlen(destrootdir));
-
-        realpath_p = realpath(target_dir_p, NULL);
-        if (realpath_p == NULL) {
+        if (realpath(target_dir_p, relative_from) == NULL) {
```
marcosfrm, line 186: `relative_from` as a temporary buffer to build `realfrom` and `realtarget` is fine, but its name is confusing. Maybe rename it to `buf` or whatever? (comments below will need update in this case)

nabijaczleweli:
bad bot

nabijaczleweli:
rebased

src/install/dracut-install.c
```diff
@@ -539,26 +538,26 @@ static int resolve_deps(const char *src)
...
+                log_error("Error '%s' initiating pipe stream for %s", strerror(errno), ldd);
+                exit(EXIT_FAILURE);
+        }
 
-        log_debug("%s", cmd);
+        log_debug("%s %s", ldd, fullsrcpath);
+        if (vfork() == 0) {
+                dup2(fds[1], 1);
+                dup2(fds[1], 2);
+                execlp(ldd, ldd, fullsrcpath, (char *)NULL);
```
marcosfrm, line 547: This will break at least the "command not found" case, that will be considered successful (`ret = 0`). However current `popen()` code is also broken for non-english locales: `sh` and `ldd` output are internationalized and error messages do not match.

nabijaczleweli: Wait, this actually depends on the "command not found" bash output? `sh -c` obviously outputs `sh: 1: ldd: not found`, like it has for the past 30+ years. Wild. Added a "command not found" write to compensate.

The latter fixed by setting `LC_ALL=C`.

marcosfrm: I wonder if checking *wstatus* could simplify things, like:

```c
        pid_t pid = fork();
        if (pid == 0) {
                dup2(fds[1], 1);
                dup2(fds[1], 2);
                putenv("LC_ALL=C");
                execlp(ldd, ldd, fullsrcpath, (char *)NULL);
                _exit(EXIT_FAILURE);
        }
        close(fds[1]);

        while (waitpid(pid, &ret, 0) < 0) {
                if (errno != EINTR) {
                        ret = 1;
                        break;
                }
        }

        if (ret != 0) {
                log_error("ERROR: failed to run '%s %s'", ldd, fullsrcpath);
                return 1;
        }

        while (getline(&buf, &linesize, fptr) >= 0) {

        ...
```

(dropped `err`)

Then clean up `strstr()` calls below, removing the ones which correspond to unclean exit codes.

nabijaczleweli: well, there's three strings
> "you do not have execution permission"
> "unable to find sysroot"
> "command not found"

The last one is obviously from bash, but (my) ldd (or ld.so) doesn't contain the other two. Are they all foldable together to $? != 0?

marcosfrm: I am on Fedora 36 here btw. "you do not have execution permission" was added in

https://github.com/dracutdevs/dracut/commit/b4dc22cab7b8b071c8a150b9c61edaa247bacb6a

hence it must stay too.

marcosfrm: **Or**, if the `noexec` case makes `ldd` exit code != 0, then drop it. Needs testing.

nabijaczleweli: Turns out this is a Debian patch, `all/local-ldd.diff`: 
```
-    test -x "$file" || echo 'ldd:' $"\
-warning: you do not have execution permission for" "\`$file'" >&2
```

Applied != 0 => ret += 1; retained all except for "command not found"

src/install/dracut-install.c
```diff
@@ -618,7 +619,15 @@ static int resolve_deps(const char *src)
         }
 
         fclose(fptr);
-        wait(&err);
+        while (waitpid(ldd_pid, &err, 0) == -1) {
+            if (errno != EINTR) {
+                if (err != 0) {
```
marcosfrm, line 624: `waitpid()` returns the terminated pid (> 0) when sucessfull, but its exit status still can be abnormal: non-zero exit code, killed by signal... hence `err != 0` check and error message print need to be outside the loop[*]. If `waitpid() < 0` and `errno != EINTR` things are broken and we must assume everything failed: `err` in this case *cannot* be trusted (I think most implementations do not change it, not sure). Set it to anything non-zero.

`write(2, "command not found", ...)` is not needed. Also, to print meaningful exit codes/signals in the error message, it would need WIFEXITED, WEXITSTATUS, WIFSIGNALED, WTERMSIG, macros. Overkill in my opinion.

[*] `cp()` function (`normal_copy` label) is buggy in this regard too, because those `log_error()` will not run in case of normal errors, when `waitpid()` succeeds (missing `cp` binary, etc).

nabijaczleweli: I was clearly in no state to be writing this last night, because this is an obvious error; I've fixed this and cp.

src/install/dracut-install.c
```diff
@@ -618,7 +618,13 @@ static int resolve_deps(const char *src)
         }
 
         fclose(fptr);
-        wait(&err);
+        int err = -1;
```
marcosfrm, line 613: Shell exit code convention fits. I like it.

Small corner case: if `waitpid()` fails, the fail condition (`err = -1`) will not reflect in `ret`. Not a big matter though.

807c3c3180fe3417295466e0bec99a26a387b048
3f819029ddde17544ebfdeaaa3f8f7e0664cc423
b2d3f8610b3d5a3d385164f5459d0c3938831d11
bb96d3dfe4c74a71b7f5a0070efeb05d6b6eea1b

are ok for me.

nabijaczleweli: Hm, I'd expected it to be some form of either WEXITSTATUS() == 0x7f or WTERMSIG() == 0x7f or something but it seems that, indeed, it's neither WIFSIGNALED nor WIFEXITED; a minor rearrangement (and simplification of the return handling) makes this rare (AFAICT corresponding to fork()==-1) case `255`.

nabijaczleweli: Moved these to #1845.

marcosfrm:
Please move the last three commits to a different PR to facilitate review.

fix: return valid path on abs_rel error
feat: canonicalise parent of abs_rel() from, too
perf: don't allocate abs_rel() target parent realpath

nabijaczleweli:
Rebased with #1845 and #1851 trimmed off

src/install/dracut-install.c
```diff
@@ -356,23 +356,21 @@ static int library_install(const char *src, const char *lib)
...
         else
-                log_debug("Lib install: '%s'", p);
+                log_debug("Lib install: '%s'", lib);
         ret += r;
 
         /* also install lib.so for lib.so.* files */
-        q = strstr(p, ".so.");
+        q = strstr(lib, ".so.");
         if (q) {
                 q[3] = '\0';
```
marcosfrm, line 369: `lib` will truncated after this point and it is used again below. Current code *precisely* avoids that.

src/install/dracut-install.c
```diff
@@ -1259,7 +1247,7 @@ static char **find_binary(const char *src)
                         continue;
                 }
 
-                if (lstat(fullsrcpath, &sb) != 0) {
+                if (faccessat(AT_FDCWD, fullsrcpath, F_OK, AT_SYMLINK_NOFOLLOW) != 0) {
                         log_debug("stat(%s) != 0", fullsrcpath);
```
marcosfrm, line 1251: While at it... `lstat` here.

src/install/dracut-install.c
```diff
@@ -2117,7 +2097,6 @@ int main(int argc, char **argv)
         hashmap_free(items_failed);
         hashmap_free(modules_loaded);
 
-        free(destrootdir);
```
marcosfrm, line 2120: Still need `free()` if `realpath(i, NULL)` succeeds.

```diff
diff --git a/src/install/dracut-install.c b/src/install/dracut-install.c
index 2f1e00dc..9a9733d9 100644
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -2028,7 +2028,7 @@ int main(int argc, char **argv)
         if (!(destrootdir = realpath(i, NULL))) {
                 log_error("Environment DESTROOTDIR or argument -D is set to '%s': %m", i);
                 r = EXIT_FAILURE;
-                goto finish;
+                goto finish2;
         }

         items = hashmap_new(string_hash_func, string_compare_func);
@@ -2037,7 +2037,7 @@ int main(int argc, char **argv)
         if (!items || !items_failed || !modules_loaded) {
                 log_error("Out of memory");
                 r = EXIT_FAILURE;
-                goto finish;
+                goto finish1;
         }

         if (logdir) {
@@ -2047,7 +2047,7 @@ int main(int argc, char **argv)
                 if (logfile_f == NULL) {
                         log_error("Could not open %s for logging: %m", logfile);
                         r = EXIT_FAILURE;
-                        goto finish;
+                        goto finish1;
                 }
         }

@@ -2080,7 +2080,9 @@ int main(int argc, char **argv)
         if (arg_optional)
                 r = EXIT_SUCCESS;

-finish:
+finish1:
+        free(destrootdir);
+finish2:
         if (logfile_f)
                 fclose(logfile_f);

```

Too much? :)

marcosfrm (commented):
`lstat()` on line 890 can be replaced by `faccessat()` too.

nabijaczleweli:
all applied

marcosfrm:
Take a look at

https://github.com/marcosfrm/dracut/commit/67e26afda46c69c834185943b0fac1a5b980219b

it allows use `lib` until `library_install()` end. Feel free to add it to this PR if I not missed anything.

marcosfrm:
Or I can open a new PR for the `dirname()` stuff.

LGTM

marcosfrm:
Ping?

> Take a look at
> 
> [marcosfrm@67e26af](https://github.com/marcosfrm/dracut/commit/67e26afda46c69c834185943b0fac1a5b980219b)
> 
> it allows use `lib` until `library_install()` end. Feel free to add it to this PR if I not missed anything.

Actually the last `strdup()` is required because POSIX `basename()`, as it expects `char *path` and switching to the GNU version possibly can break other libcs -- musl only implements the POSIX version if i remember correctly.

nabijaczleweli:
Month bump

marcosfrm:
Commit messages still need changes.

nabijaczleweli:
Appended `(dracut-install)` to all that touch the `.c`

marcosfrm:
For a clean git history, needs rebase after 7f2b09339c063279f736f6d4f079f6dd7ed10577

marcosfrm:
LGTM

marcosfrm:
Can this go in to get more testing?

nabijaczleweli:
bad bot

marcosfrm:
ping

nabijaczleweli:
pong. looks fine on my end?

nabijaczleweli:
so true, bestie

johannbg:
unresolved conversation and un-requested re-reviews ( @marcosfrm acked in comment section not in the review process )  has the tendency to get pr ignored from those of us that are quite busy anyways let's not merge this until in 059 milestone

johannbg (approved):

aafeijoo-suse (approved):
Everything I was able to check and test went well, so let's merge this.

Issue #1798, opened 2022-04-28
iSCSI: iscsiroot: do not exit if discovery one netroot failed


wenchao-hao:
**Describe the bug**
iSCSI: iscsiroot: do not exit if discovery one netroot failed

**Distribution used**
Fedora 35

**Dracut version**
dracut-55

**Init system**
systemd

**To Reproduce**
1. specify two netroot iSCSI devices in cmdline, the first one is not acceasible, and second is accessible
2. during initrd, it found first failed, and would not login second

**Expected behavior**
If one netroot failed, do not exit discovery flow, and continue to handle other netroots

**Additional context**
Add any other context you like about the problem here.

PR #1799, opened 2022-04-29
fix(iscsi): do not exit in handle_netroot() if discovery failed

wenchao-hao:
This pull request changes...

## Changes

If user specify multiple netroot in cmdline, and one of it is not accessible, previous implement would exit discovery flow.
After this chang, it would continue trying other netroot.

Actually, the previous implement is not reasonable in another way, if we specify 3 netroot, and the second netroot would fail, the others are ok. Then the first netroot would connect success, and the third would not be connected.

## Checklist
- [✓] I have tested it locally
- [✓] I have reviewed and updated any documentation if relevant
- [✓] I am providing new code and test(s) for it

Fixes https://github.com/dracutdevs/dracut/issues/1798

### DIFF
--- a/modules.d/95iscsi/iscsiroot.sh
+++ b/modules.d/95iscsi/iscsiroot.sh
@@ -229,7 +229,7 @@ handle_netroot() {
             echo "$target"
         done
     })
-    [ -z "$targets" ] && echo "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed with status $?" && exit 1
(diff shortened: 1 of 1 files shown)


johannbg (changes requested):
Change your commit msg 
from
`iscsiroot: do not exit in handle_netroot() if discovery failed`
to 
`fix(iscsi): do not exit in handle_netroot() if discovery failed`

johannbg (approved):

PR #1800, opened 2022-05-01
ci(integration): replace Fedora 33 with Fedora latest

johannbg:
Fedora 33 reached end-of-life on 2021-11-30, replacing it with
Fedora latest

Alphabetically ordering the distributions in the container section

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -14,11 +14,11 @@ jobs:
         strategy:
             matrix:
                 container: [
-                        "fedora:33",
-                        "fedora:rawhide",
-                        "opensuse:latest",
                         "arch:latest",
                         "debian:latest",
+                        "fedora:latest",
+                        "fedora:rawhide",
+                        "opensuse:latest",
                 ]
                 test: [
                         "04",


LaszloGombos:
Non-reviewer looks good to me.

johannbg:
@LaszloGombos someone that reviews automatically becomes a reviewer, the more the merrier, the thing is with this change is that it might be somewhat redundant since we already are running all the test on fedora latest.

johannbg:
+ RH is not doing anything special for the project worthy of giving Fedora somekind of more test coverage over other distro's so we might just as well just reduce this to Fedora rawhide or just have it fedora-latest.

LaszloGombos:
>so we might just as well just reduce this to Fedora rawhide or just have it fedora-latest.

Does it still make sense to keep the test/container/Dockerfile-Fedora-3* maintained in the tree ? Happy to submit a PR to remove them. Perhaps remove Dockerfile-Fedora-rawhide as well as stick to Fedora-latest like for other distros ?

johannbg:
> > so we might just as well just reduce this to Fedora rawhide or just have it fedora-latest.
> 
> Does it still make sense to keep the test/container/Dockerfile-Fedora-3* maintained in the tree ? Happy to submit a PR to remove them. Perhaps remove Dockerfile-Fedora-rawhide as well as stick to Fedora-latest like for other distros ?

Sticking to Fedora-latest is only fair to other distribution and saves us from having to support other multiple downstream releases ( like all the Debian releases, it would be hard to justifie a nack to those requests ) and each docker file existance ( like every other file ) only adds to the CoD of the project. Looking downstream the 56 release was not built for F36<= and rawhide triggered most of the packit faliures, so there is that, anyways @pvalena is looking into the Fedora situation so he might want to chime in here.

johannbg:
@LaszloGombos I've been thinking about this and if you have time, make the PR to remove all the Fedora Dockerfiles except Fedora latest and if you have even more time have a look at making a PR for an Gentoo image, pick an openrc one if openrc is still the default init system in Gentoo.

johannbg:
This was superseeded in #1810

PR #1803, opened 2022-05-03
fix(dracut.sh): always check that MACHINE_ID is not empty 

aafeijoo-suse:
On new installations, /etc/machine-id may exist and be empty, and also /efi or /boot/efi may be a mount point, leading to an invalid initramfs output file.

Excerpt from a failed openQA test:
```
2022-05-01 09:03:19 <3> install(3949) [Ruby] lib/cheetah.rb(log_stream_line):211 Error output: dracut: Executing: /usr/bin/dracut --kver=5.17.4-1-default --force
2022-05-01 09:03:19 <3> install(3949) [Ruby] lib/cheetah.rb(log_stream_line):211 Error output: dracut: Can't write to /boot/efi//5.17.4-1-default: Directory /boot/efi//5.17.4-1-default does not exist or is not accessible.
2022-05-01 09:03:19 <3> install(3949) [Ruby] lib/cheetah.rb(record_status):183 Status: 1
2022-05-01 09:03:19 <3> install(3949) [Ruby] yast2/execute.rb(rescue in popup_error):232 Execution of command "[["/usr/bin/dracut", "--force", "--regenerate-all"]]" failed.
```

Checking also that `MACHINE_ID` is not empty in `lsinitrd.sh`, though it is not critical, just following what is being done previously in its code.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1089,9 +1089,13 @@ if ! [[ $outfile ]]; then
             outfile="$dracutsysrootdir/lib/modules/${kernel}/initrd"
         elif [[ -e $dracutsysrootdir/boot/vmlinuz-${kernel} ]]; then
             outfile="$dracutsysrootdir/boot/initramfs-${kernel}.img"
-        elif [[ -z $dracutsysrootdir ]] && mountpoint -q /efi; then
+        elif [[ -z $dracutsysrootdir ]] \
+            && [[ $MACHINE_ID ]] \
+            && mountpoint -q /efi; then
             outfile="/efi/${MACHINE_ID}/${kernel}/initrd"
-        elif [[ -z $dracutsysrootdir ]] && mountpoint -q /boot/efi; then
+        elif [[ -z $dracutsysrootdir ]] \
+            && [[ $MACHINE_ID ]] \
+            && mountpoint -q /boot/efi; then
             outfile="/boot/efi/${MACHINE_ID}/${kernel}/initrd"
         else
             outfile="$dracutsysrootdir/boot/initramfs-${kernel}.img"
--- a/lsinitrd.sh
+++ b/lsinitrd.sh
@@ -133,9 +133,11 @@ else
         image="/lib/modules/${KERNEL_VERSION}/initrd"
     elif [[ -f /boot/initramfs-${KERNEL_VERSION}.img ]]; then
         image="/boot/initramfs-${KERNEL_VERSION}.img"
-    elif mountpoint -q /efi; then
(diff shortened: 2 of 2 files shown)


johannbg:
We break any over-mounting usecases ( such as immutable image as in /etc being read-only ) if we dont allow existing but empty /etc/machine-id.

And if it exist but is empty and is not read only then it will always be populated by systemd. 

Anyway what am I missing here, why should we care if /etc/machine-id is not empty? 

That said we start out being ro instead of rw ( propably due to some historic unix reason which has no base for existance on the 21 century, on linux ) which probably is breaking usecases related to machine-id so we might want to change the default to be rw.

johannbg (approved):

aafeijoo-suse:
> And if it exist but is empty and is not read only then it will always be populated by systemd so there should never be a situation where it is empty.
> 
> Anyway what am I missing here, why should we care if /etc/machine-id is not empty ( not against the check but something is broken if you have missing machine-id path )?

Yes, this was a surprise to me too... I didn't expect it either until I was notified about reports of broken installations...

johannbg:
@aafeijoo-suse if you can duplicate those issues then try adding rootflags=rw and see if that helps. 
See for example [1] https://github.com/coreos/ignition/blob/main/dracut/30ignition/ignition-remount-sysroot.service that's working around the ro issue. 

The thing is we remount rw for the fsck check so why not just switch the default from ro to rw since being ro is kinda pointless in that case.

aafeijoo-suse:
> @aafeijoo-suse if you can duplicate those issues then try adding rootflags=rw and see if that helps. See for example [1] https://github.com/coreos/ignition/blob/main/dracut/30ignition/ignition-remount-sysroot.service that's working around the ro issue.
> 
> The thing is we remount rw for the fsck check so why not just switch the default from ro to rw since being ro is kinda pointless in that case.

Yes, it's easy to reproduce this issue, just try to perform a new installation on a UEFI system using a [Tumbleweed ISO](https://openqa.opensuse.org/tests/2323205/asset/iso/openSUSE-Tumbleweed-DVD-x86_64-Snapshot20220430-Media.iso) containing the dracut code before applying this fix. I'm not sure what's the process that YaST follows, but I'll check it out. Thanks for your hint!

Screenshot_Tumbleweed-UEFI-kiwi-error_2022-05-04_12:18:58

johannbg:
@aafeijoo-suse you might want to read https://systemd.io/BUILDING_IMAGES/ and pass that to the yast team

PR #1804, opened 2022-05-03
ci: on debian, default shell is not bash

LaszloGombos:
## Checklist

Fixes #

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -1,7 +1,7 @@
 .PHONY: all check clean $(wildcard TEST-??-*)
 
 $(wildcard TEST-??-*):
-	@[ "$$EUID" == "0" ] || { echo "'check' must be run as root! Please use 'sudo'."; exit 1; }
+	@[ "$(shell id -u)" = 0 ] || { echo "'check' must be run as root! Please use 'sudo'."; exit 1; }
 	@{ \
 		[ -d $@ ] || exit 0; \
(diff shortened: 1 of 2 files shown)


johannbg:
@LaszloGombos we are not going to be doing somekind of downstream spesific distribution shenanigans ( it does not scale and it adds to our cost of development ) , not even in our test cases, for example Debian should have completed it's /usr move by now so there should exist a symbolic link to /usr/bin/bash from /bin/bash

LaszloGombos:
@johannbg Thanks for the review. Always appreciate it and never take it for granted. 

> we are not going to be doing somekind of downstream spesific distribution shenanigans ( it does not scale and it adds to our cost of development ).
Acknowledged. I am aware of this project goal and I also do agree.

> for example Debian should have completed it's /usr move by now so there should exist a symbolic link to /usr/bin/bash from /bin/bash
I agree with the sentiment that for the /usr move, Debian seems to be behind.

I do however think that "/bin/bash" is not "distribution shenanigans". Dracut source tree seems to have about 180+ references to "/bin/bash" and less than 10 references to "/usr/bin/bash".

Since you raised this specifically, I will revert the "/usr/bin/bash" --> "/bin/bash" part of the change regardless, it does not impact test execution.

johannbg (approved):
LGTM

Issue #1805, opened 2022-05-03
CI on debian is not yet passing


LaszloGombos:
Running GitHub Action tests on Debian does not yet succeed.  

This is likely an issue with the CI setup and not with the dracut application itself. The intent of this bug report is to acknowledge the CI failure and collaborate on the solution (and not to suggest that dracut itself has some issues).

The error message:

> [  OK  ] Reached target Switch Root.
>          Starting Switch Root...
> [FAILED] Failed to start Switch Root.

Link to a recent run: https://github.com/dracutdevs/dracut/runs/6281914684

johannbg:
"Can't install systemd base" comes from [TEST-04-FULL-SYSTEMD/test.sh](https://github.com/dracutdevs/dracut/blob/79170aafba9ba1a852ffb2fdcc8277d216056671/test/TEST-04-FULL-SYSTEMD/test.sh) we probably needs some elif statements involving dpkg there ( L83 ) we have one for rpm and packman but not dpkg?

johannbg:
There can be dragons in other test cases as well which we will need to adjust for Debian.

LaszloGombos:
Thanks @johannbg for the great hint. PR https://github.com/dracutdevs/dracut/pull/1806 .

johannbg:
Merged hence closing

PR #1806, opened 2022-05-04
fix(test): add support for dpkg to pass the test on debian

LaszloGombos:
## Checklist

Fixes Issue #1805


johannbg (approved):
Yay we did it Debian is working in the test suite, Great work thanks! 
Now Gentoo is the only one remaining :)

PR #1807, opened 2022-05-04
fix(test): call into livecd-tools package to generate iso

LaszloGombos:
## Checklist


johannbg:
@LaszloGombos are livecd-tools available outside Fedora ( in Arch,Debian,Gentoo,OpenSuse)?

johannbg:
Also you are missing entire reasoning on why you want to make this change

LaszloGombos:
> are livecd-tools available outside Fedora ( in Arch,Debian,Gentoo,OpenSuse)?

I do not think so as livecd-tools has a dependency on DNF/rpm. I just assumed that the current create.py script as well has a dependency on Fedora/rpm, but I might be very well wrong on that.

> Also you are missing entire reasoning on why you want to make this change

I should have made this more clear. The goal is to fix this test and enable it at least on Fedora, as it looks to me that this is not yet running on CI and I assume because it would fail. 

It would be indeed even better to generate an iso for the test in a more distro-independent manner, but this seems more work than just make it first run on at least on Fedora first.

As it stands there are no tests running on CI for the dmsquash-live module.

johannbg:
> > are livecd-tools available outside Fedora ( in Arch,Debian,Gentoo,OpenSuse)?
> 
> I do not think so as livecd-tools has a dependency on DNF/rpm. I just assumed that the current create.py script as well has a dependency on Fedora/rpm, but I might be very well wrong on that.

We cannot have our test suite being bound to a single distribution.
 
> > Also you are missing entire reasoning on why you want to make this change
> 
> I should have made this more clear. The goal is to fix this test and enable it at least on Fedora, as it looks to me that this is not yet running on CI and I assume because it would fail.

Not all tests need to be run on every PR's but exist as a part of the test suite meant to run locally or within a company as part of larger test suite. 

When we reach the state of our test suite that only runs test based what has been changed, we can enable all our test cases.

> It would be indeed even better to generate an iso for the test in a more distro-independent manner, but this seems more work than just make it first run on at least on Fedora first.

Yes making real changes that works for everyone is hard and it being Fedora could be changed in the future. 
Gentoo might be better distro for test coverage or Arch which follows spesifications and upstream more closely than Fedora does.

> As it stands there are no tests running on CI for the dmsquash-live module.

There are resons for that, I mean some jobs are too expensive to do on every PR,  
 
That said our test infrastructure needs a complete overhaul, most of it is a relic from the past or originally designed as being run in an environment internal to RH thus spesific to Fedora.

Ps. Changes made to test files should be prefixed with test(testcase): .... not with fix, if unsure use ci:

LaszloGombos:
Opened issue https://github.com/dracutdevs/dracut/issues/1809 and closing this PR to take it out from the review queue. Needs a new and better approach.

PR #1810, opened 2022-05-06
ci(integration): maintain only Fedora latest

LaszloGombos:
## Checklist


johannbg (approved):

.github/workflows/integration.yml
```diff
@@ -14,11 +14,10 @@ jobs:
         strategy:
             matrix:
                 container: [
-                        "fedora:33",
-                        "fedora:rawhide",
-                        "opensuse:latest",
                         "arch:latest",
                         "debian:latest",
+                        "fedora:latest",
```
pvalena, line 19: Fedora latest is latest stable (released), and I presume that it will not receive dracut updates.
Is that intention to drop Rawhide, which the updates will actually go to?

johannbg: Each primary ( Arch, Debian,Fedora, Gentoo, OpenSuse which btw are the only distribution we will ever test for otherwise we end up having to add gazillion distro ) has one slot in our CI so we have been adding tests for missing primary's and removing the additional slots Fedora had and the test suit is still run on Fedora ( for now anyway arguably Arch ( which more closesly follows systemd or Gentoo ( which has both systemd and an alternative init ) would be a better choice to test on.

When we maintained dracut in fedora, the dracut releases in fedora was the same on *all* fedora releases in otherwords it followed the kernel's maintainership model which arguably all core/baseOS components should be doing anyways any fedora favorism that once existed is being obsoleted and fedora is just being put on par with the other primaries. 

At one point we might even drop the spec file since we cant expect contributors to be experts in downstream packaging solutions which is why it's optional now to update the spec file ( it used to be mandatory ).

PR #1811, opened 2022-05-06
fix(iscsi): iscsiroot: remove redundant ifaces config in initrd

### LINKED ISSUE #1797: iSCSI module-setup: should not package all files in /etc/iscsi/ to initrd
**iSCSI module-setup: should not package all files in /etc/iscsi/ to initrd**
iscsiadm -m discovery -t st -p xxx would using userdefined iface to perform discovery if any userdefined iface exists. While if we added these userdefined iface in initrd, while this iface is not accessible to IP specified in netroot, it results in discovery failed. We might want using default iface of iSCSI to perform discovery.

**Distribution used**
Fedora35

**Dracut version**
dracut-055

**Init system**
systemd

**To Reproduce**
1. specify an iSCSI iface which bound to a specify netif, this netif can not connect to remote iSCSI IP during inited
2. specify netroot of iSCSI devices
3. failed to login iSCSI devices during initrd

**Expected behavior**
Do not include userdefined ifaces in initrd

**Additional context**
Add any other context you like about the problem here.

wenchao-hao:
As issue in https://github.com/dracutdevs/dracut/issues/1797, there are two way to address it,
First is not add these files in initrd by modify module-setup, another one is to remove the redundant ifaces before discoverying.

Here we remove redundant ifaces configure before discoverying as a workaround to address this issue.

## Changes
iscsiroot: remove redundant ifaces configure before discoverying

## Influence
The removal is before iscsiroot create new ifaces according to cmdline, so it would not affect it.

## Checklist
- [✓] I have tested it locally
- [✓] I have reviewed and updated any documentation if relevant
- [✓] I am providing new code and test(s) for it

Fixes #1797

### DIFF
--- a/modules.d/95iscsi/module-setup.sh
+++ b/modules.d/95iscsi/module-setup.sh
@@ -198,14 +198,9 @@ install() {
         "$systemdsystemunitdir"/sockets.target.wants/iscsid.socket \
         "$systemdsystemunitdir"/sockets.target.wants/iscsiuio.socket
 
+    inst_simple /etc/iscsi/iscsid.conf
     if [[ $hostonly ]]; then
-        local -a _filenames
-
-        inst_dir /etc/iscsi
-        mapfile -t -d '' _filenames < <(find /etc/iscsi -type f -print0)
-        inst_multiple "${_filenames[@]}"
-    else
-        inst_simple /etc/iscsi/iscsid.conf
+        inst_simple /etc/iscsi/initiatorname.iscsi
     fi
 
     # Detect iBFT and perform mandatory steps


johannbg:
The first approach is better fix

wenchao-hao:
> The first approach is better fix

Updated.

johannbg:
Not an expert in iscsi but your fix is now excludes every file in the /etc/iscsi directory with the exception of /etc/iscsi/iscsid.conf in otherwords your fix is equallent of just building the image with hostonly="no".   
On top of that in modern OS design the assumption that /etc exists or is writeable cant be made.
Now would it be simpler fix to just exclude the offending file(s) in the find command using one of find's switches?

wenchao-hao:
> Not an expert in iscsi but your fix is now excludes every file in the /etc/iscsi directory with the exception of /etc/iscsi/iscsid.conf in otherwords your fix is equallent of just building the image with hostonly="no".

There are two types of file in /etc/iscsi

The first type is configure file, /etc/iscsi/iscsid.conf and /etc/iscsi/initiatorname.iscsi, which are necessary.
The second is DB files which generated by iscsiadm command during user commands. These DB files are ususally not related with initrd which should not be included in initrd(these DB files would be created in initrd by iscsiroor according to netroor:iscsi:xxx parameters).  Actually, these files are in /var/lib/iscsi rather than /etc/iscsi in some distribution such as centos.

/etc/iscsi/initiatorname.iscsi is generated by rd.iscsi.initiator from cmdline in iscsiroot, so I did not include this file in my change too. Maybe we should add /etc/iscsi/initiatorname.iscsi to initrd if not building the image with hostonly="no".  What do you think?

wenchao-hao:
@johannbg

johannbg:
Anything included from /etc should belong in a host only section these days.

PR #1812, opened 2022-05-06
ci: add Gentoo Linux container

LaszloGombos:
## Checklist

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -32,6 +32,7 @@ jobs:
                     - { dockerfile: 'Dockerfile-OpenSuse-latest',   tag: 'opensuse:latest' }
                     - { dockerfile: 'Dockerfile-Arch',              tag: 'arch:latest' }
                     - { dockerfile: 'Dockerfile-Debian',            tag: 'debian:latest' }
+                    - { dockerfile: 'Dockerfile-Gentoo',            tag: 'gentoo:latest' }
         steps:
             -   name: Check out the repo
                 uses: actions/checkout@v2
--- a/.github/workflows/manualtest.yml
+++ b/.github/workflows/manualtest.yml
@@ -16,6 +16,7 @@ on:
                     - "arch"
                     - "debian"
                     - "opensuse"
+                    - "gentoo"
 
 jobs:
     test:
--- a/test/container/Dockerfile-Gentoo
+++ b/test/container/Dockerfile-Gentoo
@@ -1,0 +1,31 @@
+FROM docker.io/gentoo/portage:latest as portage
+
+# kernel and its dependencies in a separate builder
+FROM docker.io/gentoo/stage3:musl as kernel
+COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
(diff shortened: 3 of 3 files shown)


johannbg:
@LaszloGombos what was the current status on this one?

johannbg:
@LaszloGombos you need to fix your manifest as can be seen in the failing test :smile:  
Try include the full path in the FROM line in Dockerfile-Gentoo, for example instead of using `FROM gentoo/portage:latest` use `FROM docker.io/gentoo/portage:latest`
Btw is it better to use portage instead of the stage3 images for Gentoo?

LaszloGombos:
> Try include the full path in the FROM line in Dockerfile-Gentoo,

Ok, updated the PR. Thanks for the hint !

LaszloGombos:
Log of building the container to help with the reviews: https://github.com/dracutdevs/dracut/runs/7707115197
Also attached, as not sure about the retention policy
[log.txt](https://github.com/dracutdevs/dracut/files/9275943/log.txt)
.

LaszloGombos:
CC @aafeijoo-suse to help with CI and/or move this PR forward somehow.

aafeijoo-suse:
#1816 would be useful right now... :)

aafeijoo-suse:
I found this useful tool which allows to locally test these kind of github actions: https://github.com/nektos/act

Although the docker tags look good, it throws the same error:
```
# act -j nosystemd
[Integration Test/nosystemd] Cannot parse container options: --privileged -v /dev:/dev
[Integration Test/nosystemd] 🚀  Start image=ghcr.io/dracutdevs/gentoo:latest
[Integration Test/nosystemd]   🐳  docker pull image=ghcr.io/dracutdevs/gentoo:latest platform= username= forcePull=false
[Integration Test/nosystemd] 🏁  Job succeeded
Error: Error response from daemon: manifest unknown
```

LaszloGombos:
@aafeijoo-suse Thanks for helping out.

> Error: Error response from daemon: manifest unknown

I might not fully understand the full CI setup of this project, but I think the current CI is not prepared for PRs that changes container code - like this PR. 

I think the reason for the failure is because the new containers generation (container.yml) and running the test on the containers (integration.yml) runs paralel and integration will run earlier before the container is pushed to the registry. 

This is not a new problem and I have observed this in other PRs as well (e.g. https://github.com/dracutdevs/dracut/pull/1827). We had this problem for quite some time, and resolving this problem is outside of the scope of "introducing Gentoo container" PR.

Most importantly, after the PR lands and the container is pushed, subsequent CI runs and PRs would not have the test failing. 
 
TBH, this kind of PRs are so rare, that I am not sure if we want to address resolving this race in the CI system - lot of complexity for little benefit. 

One trivial work around is land the container code first and do the test integration PR later when the container is available - which is also the proposed approach on https://github.com/dracutdevs/dracut/pull/1827 

I will update this PR and just introduce the Gentoo container. Hope this will also make the review easier as the PR will be smaller.

LaszloGombos:
not stale

johannbg (approved):

Issue #1809, opened 2022-05-05
16-DMSQUASH test is failing


LaszloGombos:
Goal is to fix the 16-DMSQUASH test - ideally so that the test is not distro and not package manager dependent. 

Previous discussion on this - https://github.com/dracutdevs/dracut/pull/1807 that can be used as a guideline on considerations for a good solution.

LaszloGombos:
@johannbg Would you find it an acceptable solution to download a Fedora iso for the purpose of running this test ?

(e.g. https://download.fedoraproject.org/pub/fedora/linux/releases/35/Server/x86_64/iso/Fedora-Server-netinst-x86_64-35-1.2.iso is "only" 0.5G).

Perhaps this is a better alternative than computing the iso as part of running the test from a .ks file ?

PR #1813, opened 2022-05-08
fix(test): dmsquash-live test without an iso

LaszloGombos:
Remove fedore specific parts and use squash filesystem as live root.

This pull request changes...

## Checklist

Fixes [1809](https://github.com/dracutdevs/dracut/issues/1809)


johannbg (approved):
LGTM

PR #1814, opened 2022-05-09
feat(gensplash): remove module

LaszloGombos:
splashutils Gentoo package (a dependency for the gensplash dracut module)
is non existent in the official Gentoo repositories.

## Checklist


johannbg:
So you want to remove a module just because the component for it does not exist in Gentoo? Have you checked if the module is not being used by other distro's or if upstream is inactive?

johannbg (approved):
Ok so I've looked into it and afaikt it looks like a dead project which no one else is using. 
Let's allow this to bake here for a couple of days to see if anyone objects the modules removal.

LaszloGombos:
> Ok so I've looked into it and afaikt it looks like a dead project which no one else is using.
Thanks.

> Let's allow this to bake here for a couple of days to see if anyone objects the modules removal.
Make sense.

For the benefit of other readers - https://wiki.gentoo.org/wiki/Fbsplash

johannbg:
Note that splash is still supported in gentoo's gen kernel  [gen_configkernel](https://gitweb.gentoo.org/proj/genkernel.git/tree/gen_configkernel.sh?id=6421096929ebe8819fa0edeaa80d1496f802ff6f#n196), [gen_initramfs](https://gitweb.gentoo.org/proj/genkernel.git/tree/gen_initramfs.sh?id=6421096929ebe8819fa0edeaa80d1496f802ff6f#n566), [gen_determineargs](https://gitweb.gentoo.org/proj/genkernel.git/tree/gen_determineargs.sh?id=6421096929ebe8819fa0edeaa80d1496f802ff6f#n97)
which is a bit weird since it's been deprecated in gentoo and propably trigger some errors if SPLASH="no" is not set in genkernel.conf so for whatever reason gentoo did not remove the support for it in gen kernel when the component got removed from the distro.

johannbg:
@floppym note we are about to drop this module not sure if it will affect gentoo since weirdly enough genkernel still seemingly supports splash.

johannbg:
A week has passed and no one screamed hence merging...

PR #1816, opened 2022-05-11
ci(github) add manual trigger to all workflows

LaszloGombos:
This pull enables manual triggers to all workflows. 
Manual triggers allow to not wait for containers to build on schedule once a day or re-run a flaky test on a PR.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,5 +1,6 @@
 name: Container
 on:
+    workflow_dispatch:
     schedule:
         -  cron: '30 11 * * *'   # every day at 4:40
     push:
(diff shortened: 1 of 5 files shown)


johannbg:
You need to fix your commit msg but yeah I'm not so sure if this is something that we want.

johannbg:
To explain a bit better our test cases should not be flaky ( they used to be flaky but we got passed that ), if they are that's something we need to address ( + we only have few required test to pass which are those that truly matter ) and secondly building containers once a day should be sufficient.

LaszloGombos:
closing as project has no interest in this github tool.

PR #1825, opened 2022-05-26
Skip creating the dracut initrd when initrd is provided or a different initrd generator is configured

keszybz:
Updated version of https://github.com/dracutdevs/dracut/pull/1691.
(Updated == "with adjusted commit messages". For some reason github was ignoring my push on the closed PR.)

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -4,6 +4,7 @@ COMMAND="$1"
 KERNEL_VERSION="$2"
 BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
+INITRD_OPTIONS_SHIFT=4
 
 # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
 # In this case, do not create the initrd.
@@ -27,6 +28,9 @@ fi
 ret=0
 case "$COMMAND" in
     add)
+        # If the initrd was provided on the kernel command line, we shouldn't generate our own.
+        [ "$#" -gt "$INITRD_OPTIONS_SHIFT" ] && exit 0
+
         INITRD_IMAGE_PREGENERATED=${KERNEL_IMAGE%/*}/initrd
         if [[ -f ${INITRD_IMAGE_PREGENERATED} ]]; then
             # we found an initrd at the same place as the kernel
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -6,8 +6,14 @@ BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
 
 # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
-# So, let's skip to create initrd.
+# In this case, do not create the initrd.
 if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
+    exit 0
+fi
+
+# Skip this plugin if we're using a different generator. If nothing is specified,
+# assume we're wanted since we're installed.
+if [ "${KERNEL_INSTALL_INITRD_GENERATOR:-dracut}" != "dracut" ]; then
     exit 0
 fi
 
--- a/install.d/51-dracut-rescue.install
+++ b/install.d/51-dracut-rescue.install
@@ -7,6 +7,11 @@ KERNEL_VERSION="$2"
 BOOT_DIR_ABS="${3%/*}/0-rescue"
 KERNEL_IMAGE="$4"
 
+# Skip this plugin if we're using a different generator. If nothing is specified,
+# assume we're wanted since we're installed.
+if [ "${KERNEL_INSTALL_INITRD_GENERATOR:-dracut}" != "dracut" ]; then
+    exit 0
+fi
 
 dropindirs_sort()
 {


johannbg:
As previously nacked.

keszybz:
Did you see the comments from various people trying to explain to you why this is wanted?

keszybz:
Jörg and I have been trying to explain that this is just a small change to make dracut play better with other systems. So maybe try to re-read the discussion.

johannbg:
Maybe you should re-read my response it's quite apparent that the "mkosi" team and I might add the same mkosi team that has referred to dracut being a monopoly which is outright absurd since the only "dracut monopoly" that exist is the one that the developers of mkosi they themselves created and has nothing to do with dracut, in otherwords an monopoly of their own creation which they are planning to replacing with another monopoly of their own creation which probably seen by upstream mkosi as being "progress" o_O as well as now trying to make a fool out of myself anyways so I continue is forcing initramfs projects into somekind of self made standard to circumvent the fact that downstreams are shipping and using multiple initramfs generating tools because having to support those tools in mkosi is proving to be quite difficult task to achive ( it's easier to force those tools into somekind of "standard" ) 

If there were any technical argument that literally benefitted any distribution or initramfs generation project outside mkosi upstream this would have been merged regardless of any remarks like dracut being somekind of monopoly and what not but there is not and I did not read any comment that justified that it was on the previous PR ( if it had been I would have re-opened it and merged it ).

johannbg:
And I might add it's quite alarming seeing that systemd upstream has no problem twisting and turning and forcing projects into systemd's team and surrounding projects, own selfmade standards that only benefits other projects the same systemd team or members of that team is working on and just expect that every people, project and distributions just swallow that act of corruption down.

I suggest that next time mkosi/systemd team decides to create a "standard" it collaborates with other upstreams and downstream ( if relevant ) prior to deciding on their own what that "standard" might be as opposed to plan on forcing everyone to abide whatever it comes up with through PR's to systemd.

behrmann:
I just stumbled in here from #1691 and I had to think a bit about chiming in here...

Maybe as a disclaimer first: I'm a happy dracut user across all systems I run (Debian at `$DAYJOB` and Arch in private). I do like @keszybz's idea for mkosi-initrd (but I don't see that changing my usage of dracut on the near future). I don't really talk to anybody outside of GH issues; I just like flicking through PRs and if I have anything to I throw in I do that. I don't have an agenda (at least I think I don't), I just like things working similar across systems and I want people to get along. Also, I'd love for everybody to just forget about the whole monopoly thing. That was poorly worded, but I think that was made clear, everybody apologized and I hope we can move on from that.

That all being said: I genuinely think this is a useful change for dracut. Sure, it can just not be merged and as far as I see behaviour should not change to how it is now, but adding this should make juggling different initrd generators easier for people using kernel-install on systems where dracut is not the default - like Debian and Arch.

I feel right now there's seems to be a questioning of motivations here, that I don't think is helpful, even though I understand where it is coming from. Maybe some sort of call could be arranged to talk this through - as I think talking might be more amenable to resolving any misunderstandings than long form letter writing or using GH issues as a chat. :)

johannbg:
There is no incentive nor benefits for initramfs generation projects to support being simultaneously used or installed alongside other initramfs generation projects regardless if downstream or other projects require such support since the initramfs generation projects would effectively be shooting themselves in the foot if they decide to support that since such support increases the CoD of the projects since from that point on the relevant project that did decide to support it would not only having to deal with issues related with themselves but also any other initramfs generation project being installed along side it. 

If we take Arch for example it ships 4 initramfs generation tools, Archboot, Booster, Dracut, Mkinitcpio,  [1] . 
Let's say you have an user that decide to install all of them because for whatever reason he thinks that's a wise thing to do and there is an issue with one of them, he in his infinit wisdom has no clue which one it is so he just randomly files an issue with one of the 4 upstreams the initramfs generation tools that happened to be installed on his system. 
The unfortunate upstream that has been chosen then has to spend the time and energy ( which now has been increased by factor of 3 ) to figure out if it's relevant to them.

On top of that you will have to adjust every component in the distribution to support multiple initramfs generation tools being installed at the same time as well as the initramfs generation projects themselves as well as also make sure they aren't stepping on each other toes in otherwords they will be forced to align with eachother, which kills all creativity and benefits each of those project might have over eachother. 

Sorry to say I'm not seeing the justification for nor the technical benefit for upstream initramfs generation tools support being installed alongside one another nor for distributions to do that. 

All I see is a waste of resources from upstream/downstream having to support that ( vs not doing so ), with all the benefit of such support going to the tool(s) that require such support, tools such as mkosi.

To get this merged, it will have to be showned how this is benefitial to the project ( not downstream nor the tool(s) in question ) to support this. 

Sorry, if that's disappointing to hear but it is what it is. 

1. https://wiki.archlinux.org/title/Category:Initramfs

keszybz:
> To get this merged, it will have to be showned how this is benefitial to the project ( not downstream nor the tool(s) in question ) to support this.

Sure. The file that is being patched here is invoked by `kernel-install`, and provides a plugin for `kernel-install`. For users it is beneficial for this plugin to behave correctly, i.e. follow the interface defined by `kernel-install`. (I assume that you accept the fact that something that is beneficial for users of dracut is beneficial for dracut.)

`kernel-install` has a few different ways in which it can be invoked. In one of the ways an initrd is passed in. (This might be useful for example when creating a new boot entry with new options for a kernel+initrd combo that was already installed earlier, or when using an initrd with some custom modifications, etc). Right now this dracut plugin will unconditionally generate an initrd, which is takes quite a bit of time and is unneeded and frustrating to the user.

The second patch is for the case where the user wants to use a different initrd generater. If they decided to do that, and the dracut plugin misbehaves and tries to generate an initrd anyway, this is annoying for the user. They will get it to work anyway, e.g. by uninstalling dracut or masking the plugin, but this extra work shouldn't be necessary. Again, I assume that you assume that something that is good for the users of the project is good in general.

johannbg:
I'll ask again 

How is this benefitial for *upstream* dracut and other initramfs generation projects where the maintainance burden resides.

keszybz:
Are you working on this project for the benefit of the users, or for some other abstract purpose?

johannbg:
It should be quite apparent that upstream initramfs generation projects having to effectively support one another or coexist nicely if you will, means increased load on those projects vs the corner case user ( yes regular users aren´t swapping the tool that is used to generate initramfs on their system on daily bases like their underwear ) who's experience enough to simply uninstall the initramfs generation tool that's not the distribution default and install and use one of the other one initramfs generation projects the distribution provides.

behrmann:
> Are you working on this project for the benefit of the users, or for some other abstract purpose?

Let's not go down that path, even if we're frustrated. We all want to improve UX of all users here, be they end users, sysadmins like myself and upstreams. I'm still sure we can get on the same page here. :)

> There is no incentive nor benefits for initramfs generation projects to support being simultaneously used or installed alongside other initramfs generation projects regardless if downstream or other projects require such support since the initramfs generation projects would effectively be shooting themselves in the foot if they decide to support that since such support increases the CoD of the projects since from that point on the relevant project that did decide to support it would not only having to deal with issues related with themselves but also any other initramfs generation project being installed along side it.
>
> If we take Arch for example it ships 4 initramfs generation tools, Archboot, Booster, Dracut, Mkinitcpio, [1] . Let's say you have an user that decide to install all of them because for whatever reason he thinks that's a wise thing to do and there is an issue with one of them, he in his infinit wisdom has no clue which one it is so he just randomly files an issue with one of the 4 upstreams the initramfs generation tools that happened to be installed on his system. The unfortunate upstream that has been chosen then has to spend the time and energy ( which now has been increased by factor of 3 ) to figure out if it's relevant to them.
>
> On top of that you will have to adjust every component in the distribution to support multiple initramfs generation tools being installed at the same time as well as the initramfs generation projects themselves as well as also make sure they aren't stepping on each other toes in otherwords they will be forced to align with eachother, which kills all creativity and benefits each of those project might have over eachother.
>

First of all, I fully concede the point, that there is little benefit in having multiple initrd generators installed. On a system that is meant to be used by users instead of tinkerers there should be one and only one initrd generator doing the job of generating the initrd. It is evident, that this can be solved by having only a single initrd generator installed and in a properly administered system this will in almost all likelihood be the case.

That being said there is a significant amount of edge cases where systems are not properly configured - be it through tinkering from the side of the user or some surprising, unexpected behaviour through overly-long, maybe optional dependency chains (something I had happen to me on Debian more than once).

All of these edge cases can (and will) arise today - by getting initramfs-tools through some weird dependency chain when one only wanted dracut on Debian or your example of some zealous user installing all the initrd generators on Arch.

The changes here are not a panacea against any of these situations, since they obviously require buy in from users - users here being initrd generators, they have to support `KERNEL_INSTALL_INITRD_GENERATOR`. 

I *do* think that this feature is of explicit benefit to initrd generators in general and dracut in particular, since it allows to properly define the state a dracut-using system should be in (set `KERNEL_INSTALL_INITRD_GENERATOR=dracut` in `/etc/kernel/install.conf`). This is something that downstream distros that insist on having multiple initrd generators coinstallable can put into their configuration hooks and that upstreams like dracut can put into their issue templates (e.g. please check "I have checked that my system is configured like so").

Of course you can also define a supported system for dracut as one that has dracut and only dracut installed as initrd generator, but on distros where multiple initrd generators are coinstallable this might situation might happen by accident. I argue it is far easier to ensure that a certain config file contains a certain value.

I do not think that the code in this particular PR is a long-term maintenance burden. I disagree that not merging it prevents the stated problems from having setups with multiple initrd generators, since they may arise today already just the way you described, but I do think this can decrease maintenance burden in the future, as the proper system state can be described more precisely making it easier to avoid the broken parts of configuration space.

lnykryn:
Btw, this will kill two birds with one stone. There is also an issue, that Fedora and probably other distributions.
Right now when we are installing the initial package set, we of course install kernel and this trips the kernel-install, which creates initrd. But this one is useless, since you might still miss some other packages required for your setup. That's why anaconda deletes this one and at the end creates a new one manually. This could give anaconda a knob they could use to disable initrd generation and hence really improving the installation speed,

lnykryn (approved):

johannbg:
@lnykryn given that you approved those changes that can directly affect the load on the resources that the project has, have you suddenly now decided after years of inactivity to start contributing to the project full time?

lnykryn:
I am fine with the maintenance of the kernel-install script. I've already done some research in this area and it should be fine.

johannbg:
Sorry I dont follow. Opening this pandora's box will require additional resources to the project as whole in the long run.

lnykryn:
Btw, to address the worry that people will fill bugs of other initrd generators to Dracut can be addressed pretty easily. We will just ask users to add their configuration to the report.

On the downstream side, this actually makes things easier. Currently, I have the issue that I want to have Dracut installed but also I am playing with using initrds pre-generated on a different machine (by dracut as well), so I just want to have the possibility to run dracut manually if things go south.

Also, what is the recommended solution for my issue? E.g. I want to run kernel install and not generated any initrd.

johannbg:
Interesting how many issues have you gone through in the past year, heck let's say last two years to be able to make that assessment?

Winterhuman:
Just to add to the discussion, I think I'm in support of this as well, not for supporting multiple generators, but for checking what generator is being used without needing to check for files or packages.

If I had a bash script which only supported specific generators, I could check `$KERNEL_INSTALL_INITRD_GENERATOR` for them and return an error if they aren't the supported values. The alternative is needing to hook into the package manager to check for the generator's package (which means checking what package manager is being used) or checking for certain files or directories (which could be moved, renamed, or be left-overs from a previous install).

lnykryn:
We agreed with @johannbg that we should introduce this in Fedora first as a downstream patch and if it will get its use and it will not bring any problems, we will reopen this PR.

PR #1827, opened 2022-05-27
feat(dracut-install): Add an optional file with hashes of included files

jamacku:
## Changes

This patch is adding hashing functionality into dracut-install.
From every source is generated a hash which is stored into a file
together with the destination path. Function uses MD5 algorithm from
openssl library.

eg.: f222f1bfcee4a4d02abfcaf24de19ab0 /root/tmp/dest/usr/bin/test

This can be used to track changes in source.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ mandir ?= ${prefix}/share/man
 CFLAGS ?= -O2 -g -Wall -std=gnu99 -D_FILE_OFFSET_BITS=64 -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2
 bashcompletiondir ?= ${datadir}/bash-completion/completions
 pkgconfigdatadir ?= $(datadir)/pkgconfig
+OPENSSL_LIBS ?= -DOPENSSL -lssl -lcrypto
 
 man1pages = man/lsinitrd.1
 
@@ -73,7 +74,7 @@ src/install/util.o: src/install/util.c src/install/util.h src/install/macro.h sr
 src/install/strv.o: src/install/strv.c src/install/strv.h src/install/util.h src/install/macro.h src/install/log.h
 
 src/install/dracut-install: $(DRACUT_INSTALL_OBJECTS)
-	$(CC) $(LDFLAGS) -o $@ $(DRACUT_INSTALL_OBJECTS) $(LDLIBS) $(FTS_LIBS) $(KMOD_LIBS)
+	$(CC) $(LDFLAGS) -o $@ $(DRACUT_INSTALL_OBJECTS) $(LDLIBS) $(FTS_LIBS) $(KMOD_LIBS) $(OPENSSL_LIBS)
 
 logtee: src/logtee/logtee.c
 	$(CC) $(LDFLAGS) -o $@ $<
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -19,6 +19,9 @@
 
 #define PROGRAM_VERSION_STRING "2"
 
+#define HASH_BUFF_SIZE 1024
+#define HASH_FILE "source_hash.txt"
+
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
@@ -43,6 +46,7 @@
 #include <fts.h>
 #include <regex.h>
 #include <sys/utsname.h>
+#include <openssl/evp.h>
 
 #include "log.h"
 #include "hashmap.h"
@@ -68,6 +72,7 @@ static bool arg_modalias = false;
 static bool arg_resolvelazy = false;
 static bool arg_resolvedeps = false;
 static bool arg_hostonly = false;
+static bool arg_hash_source = false;
 static bool no_xattr = false;
 static char *destrootdir = NULL;
 static char *sysrootdir = NULL;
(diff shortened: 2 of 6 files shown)


johannbg:
@jamacku can you update your PR to include in the docker images the missing libraries and header files that are needed to be linked to openssl as in openssl-devel for Fedora based platform, libssl-dev for Debian based platforms etc.

jamacku:
@johannbg I updated Dockerfiles.

johannbg:
@jamacku the test are still spectacularly failing so you need to have a better look at that

jamacku:
> @jamacku the test are still spectacularly failing so you need to have a better look at that

Changes of Dockerfiles in this PR has no impact on integration tests, since containers are updated only on `push` events.

https://github.com/dracutdevs/dracut/blob/master/.github/workflows/container.yml#L53

LaszloGombos:
> Changes of Dockerfiles in this PR has no impact on integration tests, since containers are updated only on push events.

I think this is one of the cases where having an option for manual trigger could be useful for the project  - as proposed earlier in https://github.com/dracutdevs/dracut/pull/1816

src/install/dracut-install.c
```diff
@@ -2059,6 +2118,41 @@ int main(int argc, char **argv)
...
+
+                r = lstat(libdir, &sb);
+                if (r != 0) {
+                        if (errno != ENOENT) {
+                                log_error("ERROR: stat '%s': %m", libdir);
+                                r = EXIT_FAILURE;
+                                goto finish;
+                        } else {
+                                log_info("mkdir '%s'", libdir);
+                                r = dracut_mkdir(libdir);
```
marcosfrm, line 2137: Does not check for success...

src/install/dracut-install.c
```diff
@@ -740,6 +747,37 @@ static int dracut_mkdir(const char *src)
         return 0;
 }
 
+static int hash_file_md5(char *file_path, char *dst_path)
```
marcosfrm, line 750: These can be `const char *`.

jamacku:
Thanks for the review, @marcosfrm. Could you please have another look?

src/install/dracut-install.c
```diff
@@ -928,6 +967,19 @@ static int dracut_install(const char *orig_src, const char *orig_dst, bool isdir
                 } else {
                         log_info("cp '%s' '%s'", fullsrcpath, fulldstpath);
                         ret += cp(fullsrcpath, fulldstpath);
+
+                        if (arg_hash_source) {
+                                /* create hash of source and save it into the file */
+                                if (dst[0] != '/') {
+                                        if (_asprintf(&dstpath, "/%s", dst) < 0) {
```
marcosfrm, line 974: `_asprintf()` takes care of logging and exiting with failure in case of error. Drop the `if` block.

src/install/dracut-install.c
```diff
@@ -995,6 +1047,7 @@ static void usage(int status)
                "  -s --mod-filter-symbol    Filter kernel modules by symbol regexp\n"
                "  -S --mod-filter-nosymbol  Exclude kernel modules by symbol regexp\n"
                "  -N --mod-filter-noname    Exclude kernel modules by name regexp\n"
+               "  -g --source-hash          Install file (/usr/lib/dracut/source_hash.txt) with list of all files and coresponding MD5 hashes"
```
marcosfrm, line 1050: "Install /usr/lib/dracut/source_hash.txt with ..." is a bit shorter.

src/install/dracut-install.c
```diff
@@ -2079,6 +2138,47 @@ int main(int argc, char **argv)
...
+                                goto finish;
+                        } else {
+                                log_info("mkdir '%s'", libdir);
+                                r = dracut_mkdir(libdir);
+                                if (r != 0) {
+                                        r = EXIT_FAILURE;
+                                        goto finish;
+
+                                hashmap_put(items, i, i);
+                        }
```
marcosfrm, line 2163: This `}` should be before `hashmap_put()`.

src/install/dracut-install.c
```diff
@@ -2079,6 +2138,47 @@ int main(int argc, char **argv)
...
+                                r = EXIT_FAILURE;
+                                goto finish;
+                        } else {
+                                log_info("mkdir '%s'", libdir);
+                                r = dracut_mkdir(libdir);
+                                if (r != 0) {
+                                        r = EXIT_FAILURE;
+                                        goto finish;
+
+                                hashmap_put(items, i, i);
```
marcosfrm, line 2162: `i`? Do you mean `libdir`?

Makefile
```diff
@@ -27,6 +27,7 @@ mandir ?= ${prefix}/share/man
 CFLAGS ?= -O2 -g -Wall -std=gnu99 -D_FILE_OFFSET_BITS=64 -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2
 bashcompletiondir ?= ${datadir}/bash-completion/completions
 pkgconfigdatadir ?= $(datadir)/pkgconfig
+OPENSSL_LIBS ?= -DOPENSSL -lssl -lcrypto
```
marcosfrm, line 30: `configure` needs to error out if openssl is not present, like kmod:

https://github.com/dracutdevs/dracut/blob/master/configure#L58

It is guaranteed `OPENSSL_LIBS` will be set then.

marcosfrm (commented):
I wonder if the "minimalist" distros will like openssl as a new mandatory dependency.

src/install/dracut-install.c
marcosfrm: Actually it should not contain `destrootdir`. Must add the *file* path to the hashmap too, below, after successful `fopen()`.

Makefile
LaszloGombos: Perhaps only set the flag in configure if the library is installed to deal with "minimalist" distro's ?

marcosfrm: Doable with #ifdefery in dracut-install.c.

PR #1832, opened 2022-06-02
fix(nvmf): deprecate old nvmf cmdline options

### LINKED ISSUE #1831: Man page quotes non-functional nvmf parameter names
In `man 7 dracut.cmdline` the `nvmf` parameters are prefixed with `rd.`:

 * `rd.nvmf.hostnqn`
 * `rd.nvmf.hostid`
 * `rd.nvmf.discover`
 
The acutal [implementation](https://github.com/dracutdevs/dracut/blob/f63f411d52df613936082d646ab072447b8b9d7f/modules.d/95nvmf/parse-nvmf-boot-connections.sh#L117) however only looks for `nvmf.hostnqn`, `nvmf.hostid`, and `nvmf.discover` with no `rd.` prefix.

aafeijoo-suse:
- Deprecate old nvmf cmdline options without the `rd.` prefix.
- Add missing `rd.nonvmf` cmdline option to man page.

## Checklist

Fixes #1831

### DIFF
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -895,6 +895,9 @@ NOTE: letters in the MAC-address must be lowercase!
 
 NVMf
 ~~~~
+**rd.nonvmf=0**::
+    Disable NVMf
+
 **rd.nvmf.hostnqn=**__<hostNQN>__::
     NVMe host NQN to use
 
--- a/modules.d/95nvmf/module-setup.sh
+++ b/modules.d/95nvmf/module-setup.sh
@@ -78,17 +78,17 @@ cmdline() {
         [ -z "$trtype" ] && return 0
         nvme list-subsys "${PWD##*/}" | while read -r _ _ trtype traddr host_traddr _; do
             [ "$trtype" != "${trtype#NQN}" ] && continue
-            echo -n " nvmf.discover=$trtype,${traddr#traddr=},${host_traddr#host_traddr=}"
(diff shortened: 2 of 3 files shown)


johannbg:
@aafeijoo-suse hmm dont we need to provide a transition period from "old" to "new" encase there are setups in the wild using the non rd prefixed parameters?

johannbg:
@aafeijoo-suse dont get me wrong I'm not saying we should since I'm perfectly aware that people wont adapt until they have to so I personally would want to rip the bandage off so to speak immediately instead of allowing "transition" period of some sort so I'm fine with merging this either way.

johannbg (approved):
LGTM

aafeijoo-suse:
You made me hesitate so I had to double-check it, but adding `-d` with the old option name to `getarg` just prints a deprecation warning and this old option name doesn't stop working.

johannbg:
@aafeijoo-suse yeah it worked just fine, I'm just conflicted if we should or should not be providing backwards compatibilty with the old prefixes thus I needed a second opinion on the matter.

aafeijoo-suse:
> @aafeijoo-suse hmm **_dont_** we need to provide a transition period from "old" to "new" encase there are setups in the wild using the non rd prefixed parameters?

Oh, the "don't" instead of "do" made me misunderstand your question :)

Yes, we know that allowing deprecated options without announcing an expiration date or something keeps people using it, but in this case would be too abrupt in my opinion.

Issue #1778, opened 2022-04-08
[crypt] Detached LUKS header causes erroneous `/etc/block_uuid.map`


atweiden:
**Describe the bug**

Any LUKS encrypted device with detached header causes `blkid $_dev -s UUID -o value` to return nil [here in `modules.d/90crypt/module-setup.sh`](https://github.com/dracutdevs/dracut/commit/bf5c53a33f36d15a87297d4492624d137c3cd9fa#r70780210).

This results in an erroneous `/etc/block_uuid.map` being written to the initramfs, which greatly inhibits successful bootup of systems having detached headers on the LUKS root volume. A Dracut rescue shell is spawned.

**Distribution used**

Void Linux

**Dracut version**

053

**Init system**

runit

**To Reproduce**

In Void Linux [live environment](https://github.com/atweiden/voidiso/releases/download/2022.04.19/voidlinux-2022.04.19-x86_64.iso):

```bash
git clone https://github.com/atweiden/voidvault
cd voidvault
export RAKULIB="$(realpath lib)"
export RAKUDO_HOME=/usr/lib/raku
export PATH="$(realpath bin):$PATH"
# bootstrap void
voidvault new 1fa
```

Reboot after running script to completion.

**Expected behavior**

This issue calls into question the validity of `/etc/block_uuid.map` as an approach. While `/etc/block_uuid.map` could be made to include _something_ here, rather than useless trailing whitespace, what would work? No UUID is possible to obtain.

**Additional context**

Unrelated to https://github.com/dracutdevs/dracut/issues/1566 — omitting `rd.luks.name` doesn’t rectify bootup errors.

reagentoo:
Is there any solution how to use detached headers for root using dracut?

atweiden:
> Your problem is using UUID instead of PARTUUID. With a detached header, you should be using the partition's UUID since there can be no filesystem UUID

The headerless volume won’t have a PARTUUID if it exists on an unpartitioned device.

Upon reviewing #1838, it doesn’t seem to hinge on having a `PARTUUID`.

#1838 may provide a workaround, though. If it works for headerless LUKS volumes on partitioned devices, I don’t see why it wouldn’t also work for the same on unpartitioned devices.

> When they accept my pull request, this issue will be resolved.

Considering #1838 doesn’t fix the problem with `/etc/block_uuid`, no, it won’t.

LaszloGombos:
@atweiden 

[...]

PR #1838, opened 2022-06-07
Added cmdline options to enable use of detached headers with rootfs.

SentToDevNull:
Detached headers may be mounted from external media or placed in the initramfs under /headerdiskmount.

Tested on all my own systems (Gentoo).

### Example

I store my detached header (e.g. "definitely_not_a_luks_header.img") on a separate disk (e.g. /dev/vda1) and use the following dracut cmdline options in my GRUB config:
```
menuentry "Gentoo" {
  linux /vmlinuz-superkernel \
    rd.luks.partuuid=3c9d04af-02 \
    rd.luks.header=definitely_not_a_luks_header.img \
    rd.luks.header.disk=/dev/vda1 \
    root=/dev/mapper/luks-3c9d04af-02 \
    rw \
    quiet
  initrd /initramfs-superkernel.img
}
```

### DIFF
--- a/modules.d/90crypt/cryptroot-ask.sh
+++ b/modules.d/90crypt/cryptroot-ask.sh
@@ -134,6 +134,16 @@ fi
 
 unset allowdiscards
 
+# parse for header file
+luks_header_filename=$(getargs "rd.luks.header")
+luks_header_disk=$(getargs "rd.luks.header.disk")
+if [ -n "$luks_header_filename" ]; then
+    cryptsetupopts="$cryptsetupopts --header /headerdiskmount/$luks_header_filename"
+fi
+
+unset luks_header_filename
+unset luks_header_disk
+
 # fallback to passphrase
 ask_passphrase=1
 
--- a/modules.d/90crypt/parse-crypt.sh
+++ b/modules.d/90crypt/parse-crypt.sh
@@ -34,7 +34,16 @@ else
     PARTUUID=$(getargs rd.luks.partuuid -d rd_LUKS_PARTUUID)
     SERIAL=$(getargs rd.luks.serial -d rd_LUKS_SERIAL)
     LUKS=$(getargs rd.luks.uuid -d rd_LUKS_UUID)
+    HEADER=$(getargs rd.luks.header -d rd_LUKS_HEADER)
+    HEADER_DISK=$(getargs rd.luks.header.disk -d rd_LUKS_HEADER_DISK)
     tout=$(getarg rd.luks.key.tout)
+
+
+    if [ -n "$HEADER_DISK" ]; then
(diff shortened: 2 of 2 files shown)


reagentoo:
Thanks for this work. I use the next cmd list with genkernel-9999:
```
ro root=UUID=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
apparmor=1 security=apparmor
dolvm root_trim=yes scandelay
udev.children-max=1
acpi_enforce_resources=lax
crypt_root=PARTUUID=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
root_key=/keyfiles/gentoo.key
root_keydev=PARTUUID=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
root_header=/keyfiles/gentoo.hdr
root_headerdev=PARTUUID=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
```
In case of dracut - can `rd.luks.header.disk` accept PARTUUID?

atweiden:
> Fixes the issue mentioned by @reagentoo here: #1778

Since this PR doesn’t address [the issue with](https://github.com/dracutdevs/dracut/commit/bf5c53a33f36d15a87297d4492624d137c3cd9fa#r70780210) `/etc/block_uuid.map` at all, by definition it can’t solve #1778.

It could be a workaround, however.

atweiden:
Would [this](https://github.com/dracutdevs/dracut/pull/1838/files#diff-4c818a1f98f56ac87a7515fcf1476a6d2d6cc7944ffcdaa24efacdddabe07f42R141) leave the device mounted at the hard coded path `/headerdiskmount` after the root volume is unlocked?

What was the rationale for adding a second instance of `cryptsetupopts="$cryptsetupopts --header ..."`:

```bash
if [ -n "$luks_header_filename" ]; then
  cryptsetupopts="$cryptsetupopts --header /headerdiskmount/$luks_header_filename"
fi
```

... below the [existing code](https://github.com/dracutdevs/dracut/blob/29e9718f11c5a9346e75f025cea060cf9b456c01/modules.d/90crypt/cryptroot-ask.sh#L112) which looks to do much of the same:

```bash
header=*)
  cryptsetupopts="${cryptsetupopts} --${1}"
```

I can’t tell if this PR handles the scenario where users install the detached LUKS header onto a LUKS encrypted volume, e.g. if `/boot` is encrypted and installed to a USB drive, and the detached header is stored within `/boot`. At first glance, the approach appears to make the assumption the header is stored unencrypted on a device. Perhaps this might be worked around by including the detached header in the initramfs, but I don’t immediately see how this PR facilitates that, either.

PR #1839, opened 2022-06-08
fix(bluetooth): install firmware files from /lib

aafeijoo-suse:
The generic path to use should be `/lib` since distributions that follow usr merge
have a symlink to `/usr/lib`, and those that don't, use `/lib`.

## Checklist


marcosfrm:
See https://github.com/dracutdevs/dracut/pull/1794 (particularly 5ab7a78634859141103f92cd87a002f6f0ee8f20)

aafeijoo-suse:
> See #1794 (particularly [5ab7a78](https://github.com/dracutdevs/dracut/commit/5ab7a78634859141103f92cd87a002f6f0ee8f20))

Thanks for letting me know, I missed it.

PR #1841, opened 2022-06-09
feat(dracut): support parallel execution with --parallel

mwilck:
Add an option `--parallel` which can currently be used with
`--regenerate-all`. With `--regenerate-all --parallel`, dracut will be
run in parallel for all kernels found.

Also introduce a config file equivalent setting: `parallel="yes"`.

In my testing I've found a linear speed up in execution time.

## Changes

 * add option `--regenerate-parallel`

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -273,6 +273,9 @@ Creates initial ramdisk images for preloading modules
   --kernel-image [FILE] Location of the kernel image.
   --regenerate-all      Regenerate all initramfs images at the default location
                          for the kernel versions found on the system.
+  -p, --parallel        Use parallel processing if possible (currently only
+                        supported --regenerate-all)
+                        images simultaneously.
   --version             Display version.
 
 If [LIST] has multiple arguments, then you have to put these in quotes.
@@ -368,7 +371,7 @@ rearrange_params() {
     TEMP=$(
         unset POSIXLY_CORRECT
         getopt \
-            -o "a:m:o:d:I:k:c:r:L:fvqlHhMN" \
+            -o "a:m:o:d:I:k:c:r:L:fvqlHhMNp" \
             --long kver: \
             --long add: \
             --long force-add: \
@@ -447,6 +450,7 @@ rearrange_params() {
             --long keep \
             --long printsize \
             --long regenerate-all \
+            --long parallel \
             --long noimageifnotneeded \
             --long early-microcode \
             --long no-early-microcode \
@@ -813,7 +817,8 @@ while :; do
             ;;
         --keep) keep="yes" ;;
         --printsize) printsize="yes" ;;
-        --regenerate-all) regenerate_all="yes" ;;
+        --regenerate-all) regenerate_all_l="yes" ;;
+        -p | --parallel) parallel_l="yes" ;;
         --noimageifnotneeded) noimageifnotneeded="yes" ;;
         --reproducible) reproducible_l="yes" ;;
         --no-reproducible) reproducible_l="no" ;;
@@ -869,37 +874,6 @@ while (($# > 0)); do
 done
 
 [[ $sysroot_l ]] && dracutsysrootdir="$sysroot_l"
-
-if [[ $regenerate_all == "yes" ]]; then
-    ret=0
-    if [[ $kernel ]]; then
-        printf -- "--regenerate-all cannot be called with a kernel version\n" >&2
-        exit 1
-    fi
-
-    if [[ $outfile ]]; then
-        printf -- "--regenerate-all cannot be called with a image file\n" >&2
-        exit 1
-    fi
-
-    ((len = ${#dracut_args[@]}))
-    for ((i = 0; i < len; i++)); do
-        [[ ${dracut_args[$i]} == "--regenerate-all" ]] \
-            && unset dracut_args["$i"]
-    done
-
-    cd "$dracutsysrootdir"/lib/modules || exit 1
-    for i in *; do
-        [[ -f $i/modules.dep ]] || [[ -f $i/modules.dep.bin ]] || continue
-        "$dracut_cmd" --kver="$i" "${dracut_args[@]}"
-        ((ret += $?))
-    done
-    exit "$ret"
-fi
-
-if ! [[ $kernel ]]; then
-    kernel=$(uname -r)
-fi
 
 export LC_ALL=C
 export LANG=C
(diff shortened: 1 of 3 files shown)


aafeijoo-suse:
I like the idea, we gain speed on systems with multiple kernels, although the log gets messy.
By the way, you need to fix some formatting issues, `shfmt -s -w dracut.sh` will do the job (see https://github.com/dracutdevs/dracut/runs/6814635981?check_suite_focus=true)

mwilck:
> I like the idea, we gain speed on systems with multiple kernels, although the log gets messy. By the way, you need to fix some formatting issues, `shfmt -s -w dracut.sh` will do the job (see https://github.com/dracutdevs/dracut/runs/6814635981?check_suite_focus=true)

Fixed.

johannbg:
I'm not particular happy about --regenerate-all[=yes] and --regenerate-all=parallel since it will just end up confusing users, that said is there any argument against simply running things in parallel besides logs getting messy when it's run? ( which arguably is to be expected from parallel executions meaning it's a none issue from my pov )

mwilck:
If you prefer doing this unconditionally, fine with me. I opted for `--regenerate-all=parallel` because it doesn't change the default behavior (`--regenerate-all` still works as usual), and I wanted to avoid introducing a new, separate option for this.

If you fear that people get confused (why would they?), perhaps I should just leave the feature undocumented (i.e. undo the man page change)?

johannbg:
> If you prefer doing this unconditionally, fine with me. I opted for `--regenerate-all=parallel` because it doesn't change the default behavior (`--regenerate-all` still works as usual), and I wanted to avoid introducing a new, separate option for this.
> 
> If you fear that people get confused (why would they?), 

Because regenerate-all already implies you are going to regenerate all initramfs images and now you have added an "=yes" switch into the mix which obviously implies that there should be a "=no" to accommodate it, which in turn should do something different so users might for example expect that it regenerates not all but "some" initramfs images or regenerate just one initramfs image etc.

The same applies to  --regenerate-all=parallel then users expect being able to configure an swith opposide to that etc.

Having a "run-parallel" option ( and parallel=on/off in conf file ) is a better solution compared to having --regenerate-all=parallel to trigger this or --regenerate-all=parallel-off to disable it if set to be the default from my pov but ymmv.

aafeijoo-suse:
> ... besides logs getting messy when it's run? ( which arguably is to be expected from parallel executions meaning it's a none issue from my pov )

I agree, I don't see it as a problem, plus we can document that this will happen during parallel build.

> Having a "run-parallel" option ( and parallel=on/off in conf file ) is a better solution compared to having --regenerate-all=parallel to trigger this or --regenerate-all=parallel-off to disable it if set to be the default from my pov but ymmv.

I'd also vote to have a separate option for this, and keep the meaning of `--regenerate-all` as it is now, but that's also just a personal preference.

mwilck:
I've changed the option syntax.

sinyb:
May I suggest that instead of adding sub-options to "--regenerate-all", we introduce new option (-p is currently unused, or --parallel), that would have no effect for single kernel/initramfs?
That way it would be trivial to add "parallel=yes" to default config file and have wanted behavior with little effort. 
Programs that run dracut "behind the scene" (like zypper) would not have to know anything about the change...

sinyb:
> I've changed the option syntax.

```
/etc/dracut.conf
regenerate_parallel="yes"
```
has no effect with "dracut --regenerate-all  --force", while
`regenerate-parallel="yes"`
causes error "/etc/dracut.conf: line 5: regenerate-parallel=yes: command not found"

mwilck:
> ```
> /etc/dracut.conf
> regenerate_parallel="yes"
> ```
> 
> has no effect with "dracut --regenerate-all --force", while `regenerate-parallel="yes"` causes error "/etc/dracut.conf: line 5: regenerate-parallel=yes: command not found"

Hm, this is tricky. `--regenerate-all` (and thus, `regenerate_parallel`) is evaluated before reading the config file. Therefore with this patch, setting `regenerate_parallel` can't be set in the config file.

mwilck:
I pushed another commit that allows using `regenerate_all="yes"` in the conf file.

johannbg:
I think it's better to just call this parallel which allows for it to be potentially extended in the future as opposed to call it --regenerate-parallel and just document it can only be used along with  --regenerate-all at the moment and add an short switch of -p to it as well also remember to update the shell completion.

mwilck:
> I think it's better to just call this parallel which allows for it to be potentially extended in the future as opposed to call it --regenerate-parallel and just document it can only be used along with --regenerate-all at the moment and add an short switch of -p to it as well also remember to update the shell completion.

You're not concerned that `--parallel` might confuse users by raising unrealistic expectations?

sinyb:
> > I think it's better to just call this parallel which allows for it to be potentially extended in the future as opposed to call it --regenerate-parallel and just document it can only be used along with --regenerate-all at the moment and add an short switch of -p to it as well also remember to update the shell completion.
> 
> You're not concerned that `--parallel` might confuse users by raising unrealistic expectations?

You mean like using parallel versions of zstd/xz/bzip2/gz/whatever in addition to creating multiple inirds in parallel?

I wouldn't be too concerned. Whoever comes that far to use --parallel has probably read man page or at least --help, where its behavior is properly documented...

Speaking of which, why not use parallel compression too? :)
Is that anything else that can be parallelized?

mwilck:
> Speaking of which, why not use parallel compression too? :) 

Yes, it's used with `compress=zstd` (`zstd -15 -q -T0`). But only zstd yields reproducible results with multithreading, therefore it's not enabled for other algorithms.

sinyb:
> Yes, it's used with `compress=zstd` (`zstd -15 -q -T0`). But only zstd yields reproducible results with multithreading, therefore it's not enabled for other algorithms.

Nice to know, thanks! 
It seems that it is already default on my Opensuse Tumbleweed, I get the same results with or without " --compress=`zstd -15 -q -T0` ":
time dracut --regenerate-all -f
real    0m6.960s
user    0m38.493s
sys     0m12.846s
(for 6 kernels, I have added regenerate_parallel="yes" to .conf)

mwilck:
OK, renamed the option to `-p | --parallel` as requested.

dracut.sh
```diff
@@ -961,6 +935,62 @@ for f in $(dropindirs_sort ".conf" "$confdir" "$dracutbasedir/dracut.conf.d"); d
...
+
+    if [[ $outfile ]]; then
+        printf -- "--regenerate-all cannot be called with a image file\n" >&2
+        exit 1
+    fi
+
+    ((len = ${#dracut_args[@]}))
+    for ((i = 0; i < len; i++)); do
+        case ${dracut_args[$i]} in
+            --regenerate-all | --regenerate-parallel)
```
johannbg, line 959: You forgot to change `--regenerate-parallel`  to `-p`, `--parallel` here

johannbg (changes requested):
Change your commit message
from `feat: support parallel execution with --parallel`
to 
`feat(dracut): support parallel execution with --parallel`

Install and run shfmt -s -w dracut.sh to fix the formatting issues with your changes
and update `--regenerate-parallel` to `-p` `--parallel` where you forgot it.

dracut.sh
mwilck: right, thanks. But as `--parallel` is supposed to be independent of `--regenerate-all`, maybe I should pass the flag down to the child processes instead? (currently it won't have any effect as `--regenerate-all` is removed).

mwilck: re-pushed; still removing `--parallel`. Tell me what you prefer.

johannbg: Let's keep this as is

johannbg (approved):
LGTM once the test have run I'll merge this.

mwilck:
@aafeijoo-suse, any chance to get this in TW soon?

aafeijoo-suse:
> @aafeijoo-suse, any chance to get this in TW soon?

Yes, the update to 057 is on the devel project and will be submitted to factory soon.

PR #1842, opened 2022-06-10
feat(lvm): add new module lvmthinpool-monitor

liutgnu:
Previously dracut didn't support the feature of lvm thinpool autoextend.

The feature is useful to cases such as kdump, when vmcore to be saved to a
lvm thin volume. The thinpool should be able to autoextend, otherwise an
IO error will be caused and leaves an incomplete vmcore.

There is lvm2-monitor.service and dmeventd avaliable, however
considering [1], it is not suitable for kdump and memory limited cases.

This patch achieves the same by parallel looping a shell function in the
background, which calls lvextend periodically. If thredshold reaches,
autoextend it, if not then nothing happens.

[1]: https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/message/YF254ZO3PJ3U56P4OKHV3JNYP2PJUMYX/

Signed-off-by: Tao Liu <ltao@redhat.com>

## Checklist

Fixes #

### DIFF
--- a/modules.d/80lvmthinpool-monitor/module-setup.sh
+++ b/modules.d/80lvmthinpool-monitor/module-setup.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    # No point trying to support lvm if the binaries are missing
+    require_binaries lvm sort tr awk || return 1
+
+    return 255
+}
+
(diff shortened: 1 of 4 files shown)


johannbg (approved):
LGTM

PR #1845, opened 2022-06-17
Fix dracut-install subprocesses

nabijaczleweli:
## Changes
The four commits @marcosfrm mentioned in https://github.com/dracutdevs/dracut/pull/1796/files#r899623860:
  * popen() is awful and, helpfully, splits on the unescaped filename
  * ldd messages may be internationalised
  * fix waiting (and string parsing instead of waiting) for cp and ldd

## Checklist

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -500,12 +500,10 @@ static char *get_real_file(const char *src, bool fullyresolve)
 
 static int resolve_deps(const char *src)
 {
-        int ret = 0;
+        int ret = 0, err;
 
         _cleanup_free_ char *buf = NULL;
-        size_t linesize = LINE_MAX;
-        _cleanup_pclose_ FILE *fptr = NULL;
-        _cleanup_free_ char *cmd = NULL;
+        size_t linesize = LINE_MAX + 1;
         _cleanup_free_ char *fullsrcpath = NULL;
 
         fullsrcpath = get_real_file(src, true);
@@ -513,7 +511,7 @@ static int resolve_deps(const char *src)
         if (!fullsrcpath)
                 return 0;
 
-        buf = malloc0(LINE_MAX);
+        buf = malloc(linesize);
         if (buf == NULL)
                 return -errno;
 
@@ -523,11 +521,11 @@ static int resolve_deps(const char *src)
                 if (fd < 0)
                         return -errno;
 
-                ret = read(fd, buf, LINE_MAX);
+                ret = read(fd, buf, linesize - 1);
                 if (ret == -1)
                         return -errno;
 
-                buf[LINE_MAX - 1] = '\0';
+                buf[ret] = '\0';
                 if (buf[0] == '#' && buf[1] == '!') {
                         /* we have a shebang */
                         char *p, *q;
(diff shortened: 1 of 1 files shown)


src/install/dracut-install.c
```diff
@@ -625,7 +625,12 @@ static int resolve_deps(const char *src)
         }
 
         fclose(fptr);
-        wait(&err);
+        int err = -1;
+        while (waitpid(ldd_pid, &err, 0) == -1 && errno == EINTR)
+                ;
+        ret = WIFSIGNALED(err) ? 128 + WTERMSIG(err) : WEXITSTATUS(err);
```
marcosfrm, line 622: Different behaviour from b2d3f8610b3d5a3d385164f5459d0c3938831d11. It should not always overwrite possible non-zero `ret` from the `strstr()` matches above.

nabijaczleweli: Fixed

src/install/dracut-install.c
```diff
@@ -625,9 +625,14 @@ static int resolve_deps(const char *src)
...
-
-        return ret;
+        int err = 126;
+        while (waitpid(ldd_pid, &err, 0) == -1 && errno == EINTR)
+                ;
+        err = WIFSIGNALED(err) ? 128 + WTERMSIG(err) : WEXITSTATUS(err);
+        if (err != 0)
+                log_error("ERROR: %s %s failed with %d", ldd, fullsrcpath, err);
+
+        return ret ?: err;
```
marcosfrm, line 626: Let's prioritize the exit code, in other words `err ?: ret;`. This was b2d3f8610b3d5a3d385164f5459d0c3938831d11 behaviour.

If you do not mind a couple more lines of code, I would avoid those useless `err = 126;`/`ret = 126;` in the last two commits and simpy return early:

```c
        int err;
        while (waitpid(ldd_pid, &err, 0) == -1) {
                if (errno != EINTR) {
                        log_error("ERROR: waitpid() failed: %m");
                        return 1;
                }
        }
```

nabijaczleweli: Applied

marcosfrm:
Ok from me. Now let's wait for the actual reviewers... :)

johannbg:
> Ok from me. Now let's wait for the actual reviewers... :)

Very little to add from the "actual" reviewers since you have taken care of all that ( great thanks for that! )  + as most upstream (f)oss project we are driven by do-ocracy which means those that have and take the time and do documentation,PR's, reviews etc effectively become "upstream" and thus "the actual reviewers" so feel free to do as many reviews as you have time to do ;)

marcosfrm:
Thanks @johannbg, I was afraid of using "Start a review" button. :)

With this patchset, `resolve_deps()` fails if `ldd` returns non-zero. Unluckily it is a problem, because, for example, "not a dynamic executable" is for `dracut-install` a non-error condition and still `ldd` returns `1`.

@nabijaczleweli what do you think about this? It is simple and does fix the issue.

```c
        err = WIFSIGNALED(err) ? 128 + WTERMSIG(err) : WEXITSTATUS(err);

        /* it is expected ldd failing with 1 ("not a dynamic executable" for example),
           hence only consider error condition non-existing binary, killed by signal, etc */
        if (err >= 126)
                log_error("ERROR: %s %s failed with %d", ldd, fullsrcpath, err);

        return (err >= 126) ?: ret;
```

Sorry for not catching this before.

nabijaczleweli:
applied

marcosfrm:
Thanks!

marcosfrm:
With 057 released, this can go in to get more testing.

johannbg (changes requested):
As pointed out by Marcos you need fix your commit message so fix the commit msg to reflect the change being made, squash the older/outdated commit msg and add the file name of the file you are touching to the commit message as in `fix(dracut.install.c): ...` or `refactor(dracut.install.c): ...`

marcosfrm:
While at it, maybe make both error messages consistent:

```c
                log_error("ERROR: '%s %s' failed with %d", ldd, fullsrcpath, err);

                log_error("ERROR: 'cp --reflink=auto --sparse=auto %s -fL %s %s' failed with %d", preservation, src, dst, ret);
```

nabijaczleweli:
Applied both

marcosfrm:
The first three commits can be merged into one, with a commit message like:

```
refactor(dracut-install): fork() instead of popen(), sanitise line reading

Also set LC_ALL=C to fix ldd output pattern matching in non-english locales.
```

johannbg (approved):
LGTM

PR #1848, opened 2022-06-20
Fix ima evm

### LINKED ISSUE #1847: 98integrity: EVM is always enabled, even when there are no EVM keys
**Describe the bug**
The code in `98integrity/evm-enable.sh` is always enabling EVM, even when the /etc/keys/evm directory is missing or present but empty.

This is because in both cases (missing or empty), the `for` loop in `load_evm_x509` will assign for `PUBKEY` a value (usually `/sysroot/etc/keys/evm/*`).  In POSIX shell if the directory is missing or empty, the expansion will no happen and the iteration variable will have the full path (instead of empty, as one should expect from other languages).

Eventually this will make the function to return 0, and consider that some key has been loaded, and finally write 2 in `/sys/kernel/security/evm`.

**Distribution used**
openSUSE Tumbleweed

**Dracut version**
HEAD

**To Reproduce**
Install dracut and enable the "integrity" module. Make sure that /etc/keys/evm is missing (default) or empty. Create the initrd and check the content of "/sys/kernel/security/evm", if is different from 0 the bug is present.

**Expected behavior**
EVM should not be enabled if there are not keys available in /etc/keys/evm.

aplanas:
This pull request changes...

## Changes

* Drop unused variable
* Do not enable EVM if there are no key in the system

## Checklist

Fixes #1847

### DIFF
--- a/modules.d/98integrity/evm-enable.sh
+++ b/modules.d/98integrity/evm-enable.sh
@@ -99,6 +99,7 @@ load_evm_x509() {
     fi
     # load the default EVM public key onto the EVM keyring along
     # with all the other ones in $EVMKEYSDIR
+    local key_imported=1
     for PUBKEY in ${EVMX509PATH} "${NEWROOT}${EVMKEYSDIR}"/*; do
         if [ ! -f "${PUBKEY}" ]; then
             if [ "${RD_DEBUG}" = "yes" ]; then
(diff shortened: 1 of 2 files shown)


aafeijoo-suse (changes requested):
@aplanas, you need to change the commit messages to make the checker happy:

- `fix(integrity): do not enable EVM if there is no key`
- `fix(integrity): remove unused variable`

aafeijoo-suse:
CC @stefanberger

aplanas:
> @aplanas, you need to change the commit messages

Done, thanks!

stefanberger:
LGTM.

johannbg (approved):
LGTM

aafeijoo-suse (approved):
LGTM

PR #1863, opened 2022-07-02
fix(dracut-systemd): set Environment=DRACUT_SYSTEMD=1 environment variable in dracut-shutdown.service

### LINKED ISSUE #1862: dracut-shutdown.service lacks DRACUT_SYSTEMD=1 environment variable
**Describe the bug**

dracut-shutdown.service lacks DRACUT_SYSTEMD=1 environment variable

**Distribution used**

Debian but easily seen in dracut upstream source code:
https://github.com/dracutdevs/dracut/blob/master/modules.d/98dracut-systemd/dracut-shutdown.service#L11

**Dracut version**
latest git

**Init system**
systemd

**To Reproduce**
Write a shutdown hook and check for DRACUT_SYSTEMD variable existence.

**Expected behavior**
DRACUT_SYSTEMD=1 environment variable should be set during 

**Additional context**

dracut-lib.sh looks at the DRACUT_SYSTEMD environment variable. For example, the `info` and `warn` functions of dracut-lib.sh require a properly set DRACUT_SYSTEMD.

All other dracut systemd units in https://github.com/dracutdevs/dracut/blob/master/modules.d/98dracut-systemd/ are setting the DRACUT_SYSTEMD environment variable too.

adrelanos:
## Changes

* set Environment=DRACUT_SYSTEMD=1 environment variable in dracut-shutdown.service

## Checklist
- [x] I have reviewed and updated any documentation if relevant: none required
- [x] I am providing new code and test(s) for it: none required

Fixes #1862

### DIFF
--- a/modules.d/98dracut-systemd/dracut-shutdown.service
+++ b/modules.d/98dracut-systemd/dracut-shutdown.service
@@ -9,6 +9,7 @@ ConditionPathExists=!/run/initramfs/bin/sh
 OnFailure=dracut-shutdown-onfailure.service
 
 [Service]
+Environment=DRACUT_SYSTEMD=1
 RemainAfterExit=yes
 Type=oneshot
 ExecStart=/bin/true


dtardon (changes requested):
This doesn't fix anything, as `dracut-shutdown.service` just unpacks the "exitrd" to `/run/initramfs`. It's `systemd-shutdown` that runs `/run/initramfs/shutdown`.

lnykryn:
I agree with David, this change does not make any difference.

PR #1866, opened 2022-07-04
fix(virtiofs): make shebangs work on split-usr systems

floppym:
Fixes: c29325a982013a315418cdb2318e90cf77e387b5


LaszloGombos (commented):
LGTM

johannbg (approved):
LGTM however a bit of side note here, at some point in time we will stop supporting split usr system...

PR #1868, opened 2022-07-05
fix(dracut.sh): correct wrong systemd variable paths

aafeijoo-suse:
`systemdsystemconfdir` usually expands to /etc/systemd/system, but the local
configuration of systemd-networkd, systemd-timesyncd, portablectl and systemd
users is saved into directories whose parent is /etc/systemd
(`systemdutilconfdir`).

This error went unnoticed because these four variables are not being used anywhere in the code.

## Checklist


aafeijoo-suse:
.

johannbg (approved):
At somepoint we need to come up with a better naming scheme for variable paths

Issue #1756, opened 2022-03-21
network module has transitive dependency on systemd


HackingM:
When using the `network` module, or any module which depends on it, on an `OpenRC` system the `network` module cannot be installed.

```
dracut: dracut module 'dbus-daemon' depends on 'systemd', which can't be installed
dracut: dracut module 'dbus' depends on 'dbus-daemon', which can't be installed
dracut: dracut module 'network-manager' depends on 'dbus', which can't be installed
dracut: dracut module 'network' depends on 'network-manager', which can't be installed

```
This applies to all Dracut versions >= 54

Presumably there should be some way of using the `network` module without `systemd`.

dtardon:
> When using the `network` module, or any module which depends on it, on an `OpenRC` system the `network` module cannot be installed.
> 
> ```
> dracut: dracut module 'dbus-daemon' depends on 'systemd', which can't be installed
> dracut: dracut module 'dbus' depends on 'dbus-daemon', which can't be installed
> dracut: dracut module 'network-manager' depends on 'dbus', which can't be installed
> dracut: dracut module 'network' depends on 'network-manager', which can't be installed
> ```

I'm afraid have bad news for you. The `network` module is built around `NetworkManager`, so if you don't have that, the fact that the module needs `systemd` is the least of your problems... You might want `network-legacy` instead.

> This applies to all Dracut versions >= 54
> 
> Presumably there should be some way of using the `network` module without `systemd`.

No, there shouldn't. The change to start `NetworkManager` via a service has been intentional. Let's not go back.

johannbg:
The network module is just a meta module that pulls in dependency on what's installed so I'm not following what you are referring to. ( @dtardon the network module is not spesifically built around NM, that's just nonsense ) 

You use a meta module if an distribution provides multiple components that provide the same solutions like for example Fedora who ships both systemd's networkd and NetworkManager ( as well as iwd ) or if you want to have the same dracut configuration between multiple distribution 

Currently there is no other option for none dbus/systemd based system to use anything other than the network legacy module ( that might change if I ever find the time to implement the IWD module but most likely network-legacy would get dropped at that point ) and each module should be package downstream independently in a separate sub-component either of dracut or relevant component it represents with correct dependency on that component but distribution package things as they see fit, not what is technically correct thing to do.

Implementing these network solutions in the wild can be somewhat complex since

iwd can be used standalone, with or without dbus, with NM and with systemd's networkd and with NM+networkd. 
[...]

PR #1869, opened 2022-07-11
fix(network-legacy): check if dhclient has --timeout option

aafeijoo-suse:
It seems dhclient's `--timeout` option is not upstream [1], apparently it's RH-only [2]..., so a pre-check is needed before using it.

[1] https://github.com/isc-projects/dhcp
[2] https://github.com/landgraf/dhcp-fedora/commit/d171ac96

Some tests already take this into account:

https://github.com/dracutdevs/dracut/blob/fe8fa2b0cadbb33e27c8dd8b5851548dcd65835c/test/TEST-20-NFS/test.sh#L73-L75

https://github.com/dracutdevs/dracut/blob/fe8fa2b0cadbb33e27c8dd8b5851548dcd65835c/test/TEST-50-MULTINIC/test.sh#L73-L75

## Checklist

### DIFF
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -740,7 +740,7 @@ NFS
     Default is 1.
 
 **rd.net.timeout.dhcp=**__<arg>__::
-    If this option is set, dhclient is called with "-timeout <arg>".
+    If this option is set, dhclient is called with "--timeout <arg>".
 
 **rd.net.timeout.iflink=**__<seconds>__::
     Wait <seconds> until link shows up. Default is 60 seconds.
--- a/modules.d/35network-legacy/dhcp-multi.sh
+++ b/modules.d/35network-legacy/dhcp-multi.sh
@@ -21,6 +21,13 @@ do_dhclient() {
     _timeout=$(getarg rd.net.timeout.dhcp=)
     _DHCPRETRY=$(getargnum 1 1 1000000000 rd.net.dhcp.retry=)
 
+    if [ -n "$_timeout" ]; then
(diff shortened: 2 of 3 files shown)


johannbg:
This is a nack it's better that we simply stop supporting legacy networking altogether, it's on it's way out in Fedora ( which probably means it's already not supported in Arch ) and if the existing behaviour was somehow reliant on a patch from RH existing in downstream distro's it propably means it got rejected upstream and it makes no sense to add checks to modules to see if an option added by some downstream distro patch exist or not.

johannbg:
> That's ok, but before dropping the network-legacy module, I'd first make sure that any of the other network modules support all of the command line parameters that the legacy module supports (which is something I'm currently reviewing).

@aafeijoo-suse what was the outcome of that reviewing?

aafeijoo-suse:
The initial review is here: #1876 
On paper the features that NM lacks are minor. We cannot test every possible setup, but I think @tblume is having some issues with the iSCSI tests.
On the other hand, I know you're aware that the network-legacy module is the only option for non-systemd distros right now (#1756). SUSE also needs it (and we'll have to maintain it for years). That's all I can say, I see your point and it's not my decision to drop it upstream :)

johannbg (approved):
Ack for the change and keeping the legacy bit's for now however we cannot be carrying the legacy bit's along forever. At some point those none systemd distro's will simply have to adapt more modernized solutions like iwd for their distribution.

johannbg:
@LaszloGombos ping for review.

LaszloGombos (approved):
LGTM

PR #1873, opened 2022-07-15
fix(systemd): set right permissions for the machine-id file

### LINKED ISSUE #1864: `/etc/machine-id` included with wrong permissions, leading to unconfigurable networking
**Describe the bug**
Configuring networking via systemd-networkd and a trivial dhcp config fails with the following log messages:
```
systemd-networkd: DHCPv4 client: Failed to set DUID-EN/Vendor: Permission denied
systemd-networkd: DHCPv4 CLIENT: Failed to set IAID+DUID: Permission denied
```

(These messages only appear when setting SYSTEMD_LOGLEVEL=debug for systemd-networkd). After some time I figured out that the message was in fact not related to configuring the interface but because `/etc/machine-id` was included into the image as 600 `-rw-------` instead of 444 `-r--r--r--`. Fixing the permissions on that file resolves the networking problems.

I did some further testing and determined that dracut assumes the user has an umask of 022 - but I have 077. This caused `/etc/machine-id` to be included as mode 600. I think dracut should set it's own umask before doing any file creation to ensure its assumptions hold up.

**Distribution used**
Gentoo.

**Dracut version**
dracut-056

**Init system**
systemd

**To Reproduce**
Include systemd-networkd and use a config like:
```
[Match]
Name=e*
[Network]
DHCP=yes
```

**Expected behavior**
Dracut should include files with the correct permissions regardless of the user's umask value.

aafeijoo-suse:
Set the right permissions for the `/etc/machine-id` file.

## Checklist

Fixes #1864

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -192,6 +192,7 @@ install() {
 
     if ! [[ -e "$initdir/etc/machine-id" ]]; then
         : > "$initdir/etc/machine-id"
+        chmod 444 "$initdir/etc/machine-id"
     fi
 
     # install adm user/group for journald


dtardon:
> Writing `uninitialized\n` to it, according to https://systemd.io/BUILDING_IMAGES/

I think the first boot semantics makes no sense in initrd, hence `etc/machine-id` should be _empty_ (https://www.freedesktop.org/software/systemd/man/machine-id.html#First%20Boot%20Semantics).

aafeijoo-suse:
> > Writing `uninitialized\n` to it, according to https://systemd.io/BUILDING_IMAGES/
> 
> I think the first boot semantics makes no sense in initrd, hence `etc/machine-id` should be _empty_ (https://www.freedesktop.org/software/systemd/man/machine-id.html#First%20Boot%20Semantics).

Right, thanks for pointing me to the right place in the doc. The other place in the code (which is much more recent than this) that writes the machine-id if it's empty confused me.

https://github.com/dracutdevs/dracut/blob/fe8fa2b0cadbb33e27c8dd8b5851548dcd65835c/modules.d/35network-manager/module-setup.sh#L67-L73

lnykryn (approved):
LGTM

LaszloGombos:
LGTM

johannbg (approved):
LGTM

PR #1874, opened 2022-07-16
New Module: add the 99overlayroot module

TylerHelt0:
This pull request changes...

## Changes
This module will move `/sysroot` to `/run/overlayroot/rootfs` and use the new mountpoint as the lowerdir for an overlayfs mounted onto `/sysroot` . The upper dir and workdir are located at `/run/overlayroot/u` and `/run/overlayroot/w`. This functionality is only enabled if `overlayroot=1` is passed on the cmdline. 

This module will be useful for developers of kiosk experiences as well as those who want to create or use dm-verity root filesystems or other read-only roots. 

It could potentially be expanded to be able to mount persistent devices as the upper filesystem of the overlay. 

## Checklist

Fixes #


lnykryn:
Just a note here, I don't think that GPL v2, which we used for dracut is compatible with AGPL used here.

LaszloGombos:
Are you aware of the rd.live.overlay.overlayfs=1 option ?
Would you be able to compare this new module to what rd.live.overlay.overlayfs=1 provides ?

TylerHelt0:
> Are you aware of the rd.live.overlay.overlayfs=1 option ? Would you be able to compare this new module to what rd.live.overlay.overlayfs=1 provides ?

Yes. Some users may not want to use systemd or otherwise that functionality may not work for their use case. Including this module will provide this functionality on distros that use OpenRC, Runit, etc...

TylerHelt0:
> Just a note here, I don't think that GPL v2, which we used for dracut is compatible with AGPL used here.

I can chnage or remove the license

LaszloGombos:
Similar earlier PR/discussion - https://github.com/dracutdevs/dracut/pull/896

> Some users may not want to use systemd

rd.live.overlay.overlayfs=1 and dracut itself does not require or depend on systemd.

The feature that you seem to describe is tested by https://github.com/dracutdevs/dracut/blob/master/test/TEST-16-DMSQUASH/test.sh#L22 and this test is passing on a Gentoo container (without systemd installed) - https://github.com/dracutdevs/dracut/pull/1812

johannbg:
@TylerHelt0 as @LaszloGombos already points out this is already supported and tested for hence closing...

PR #1877, opened 2022-07-21
fix(dracut-init.sh): add missing hostonly code in the inst_multiple function

aafeijoo-suse:
The `_hostonly_install` variable is declared local in some `inst` functions that
check for the existance of the source files, but not in the `inst_multiple`
function because it is not needed.

## Checklist

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -271,8 +271,12 @@ inst_symlink() {
 }
 
 inst_multiple() {
-    local _ret
-    if "$DRACUT_INSTALL" ${dracutsysrootdir:+-r "$dracutsysrootdir"} ${initdir:+-D "$initdir"} -a ${loginstall:+-L "$loginstall"} ${DRACUT_RESOLVE_DEPS:+-l} ${DRACUT_FIPS_MODE:+-f} "$@"; then
+    local _ret _hostonly_install
+    if [[ $1 == "-H" ]]; then
+        _hostonly_install="-H"
+        shift
+    fi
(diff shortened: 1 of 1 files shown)


lnykryn:
Hmm, do we know why inst_multiple does not process -H? THis seems wrong to me.

aafeijoo-suse:
> Hmm, do we know why `inst_multiple` does not process -H? THis seems wrong to me.

I'm not saying `inst_multiple` does not process -H, but this method does not declare any `local _hostonly_install` variable, so the log entry `${_hostonly_install:+-H}` is wrong. See:

https://github.com/dracutdevs/dracut/blob/fe8fa2b0cadbb33e27c8dd8b5851548dcd65835c/dracut-init.sh#L257-L266

For example, the `inst` method declares it:

https://github.com/dracutdevs/dracut/blob/fe8fa2b0cadbb33e27c8dd8b5851548dcd65835c/dracut-init.sh#L218-L229

lnykryn:
Yeah, I agree, but my question is why the code is different from the other functions.

aafeijoo-suse:
> Yeah, I agree, but my question is why the code is different from the other functions.

The `inst_multiple` function does not perform any checks on input parameters, so the `_hostonly_install` variable is not needed here

See this commit: https://github.com/dracutdevs/dracut/commit/4ef45f13

lnykryn:
I don't know; it still looks like Harald forgot to add the same code in inst_multiple. But probably I am missing something.

johannbg:
I agree with @lnykryn I'm pretty sure Harald simply forgot to add the same code in inst_multiple...

aafeijoo-suse:
ok, 2 against 1, I'm not going to fight here :)

aafeijoo-suse:
.

johannbg (approved):

Issue #1835, opened 2022-06-02
Since last update using `--conf /dev/null` fails. How do I ignore preexisting config files?


oddlama:
**Describe the bug**
Since the last major update, invoking dracut with `--conf /dev/null` fails. If I see correctly, this has been introduced as a response to #1136, where it was considered a bug not to reject `/dev/null`. I've been using `--conf /dev/null --confdir /dev/null` to ignore any existing configuration files, but the linked issue suggests this never worked (?). What is the correct way to make dracut ignore any existing configuration files, now that `/dev/null` is invalid?

**Distribution used**
Latest arch LIVE-USB environment

**Dracut version**
dracut-056

aafeijoo-suse:
You can pass an empty file to `--conf` and a directory with no `.conf` files to `--confdir`.

oddlama:
I was hoping for a solution that doesn't require creating a new directory. Espevially for automated invocation, I need a stateless approach and cannot rely on the state of the user's filesystem.

Is there any rationale behind not allowing /dev/null to disable configuration? It seems like an unnecessary complication from my point of view.

aafeijoo-suse:
I don't see why calling `mktemp` would be an issue in this specific use case, as dracut calls it internally during the generation process.
```
# dracut -f --conf $(mktemp) --confdir $(mktemp -d)
```

oddlama:
> I don't see why calling `mktemp` would be an issue in this specific use case, as dracut calls it internally during the generation process.
> ```
> # dracut -f --conf $(mktemp) --confdir $(mktemp -d)
> ```

Not strictly a technical issue of course, but arguably a usability issue. All classical command-line tools can be used with special files from `/dev` or with command substitution by design (see for example `dd if=/dev/zero ...` or `diff <(echo test) ...`). With dracut I cannot use either of those, as those are not files but character devices or fifos.

I hope I could clarify why it seems strange to me, when a tool actively prohibits this type of usage. This is also why I initially expected there to be a technical reason for dracut to disallow it. I'm still not sure if that's the case, but if it isn't, I think it feels like an artificial restriction with no apparent benefit.

Of course this is a very minor thing overall, but I'd like to say that I would be happy if dracut could be used as described. :)

oddlama:
Excuse my impatience, but I wonder if there has been an update to this issue in the meantime?

pvalena:
I think this is a behaviour change, as `/dev/null` is invalid config file and therefore expected to fail.

PR #1878, opened 2022-07-21
fix(dracut.sh): do not fail on irregular files

pvalena:
If file is not a regular file (test -f), dracut.sh fails,
which is unexpected change of behaviour.
The workaround would be to create an empty file.

## Changes

The following example currently fails:
```
dracut --conf /dev/null
```

After the change 

## Checklist

Fixes: #1835


lnykryn (approved):
LGTM

LaszloGombos:
LGTM

johannbg (approved):
LGTM

PR #1882, opened 2022-07-25
dmsquash-live-root: Run checkisomd5 on correct device

bcl:
When the new grub2 iso is written to a usb drive the disk label points
to a partition that does not include the full iso image. This causes
checkisomd5 to run with the wrong data and it fails.

This patch adds a check that will test to see if there is a parent
device that is a disk, and to run checkisomd5 on it instead of on the
partition pointed to by the label.

When running from an iso this will return the original
/dev/disk/by-label/ path, and when running from a usb drive it will
return the parent device (eg. /dev/sda).

Resolves: rhbz#2107858

## Checklist

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -33,8 +33,35 @@ overlay_size=$(getarg rd.live.overlay.size=)
 getargbool 0 rd.live.overlay.thin && thin_snapshot="yes"
 getargbool 0 rd.live.overlay.overlayfs && overlayfs="yes"
 
+# Take a path to a disk label and return the parent disk if it is a partition
+# Otherwise returns the original path
+function get_check_dev() {
+    local _udevinfo
+    dev_path="$(udevadm info -q path --name "$1")"
+    _udevinfo="$(udevadm info -q property --path "${dev_path}")"
+    strstr "$_udevinfo" "DEVTYPE=partition" || {
+        echo "$1"
+        return
+    }
+    parent="${dev_path%/*}"
+    _udevinfo="$(udevadm info -q property --path "${parent}")"
+    strstr "$_udevinfo" "DEVTYPE=disk" || {
+        echo "$1"
+        return
+    }
+    strstr "$_udevinfo" "ID_FS_TYPE=iso9660" || {
+        echo "$1"
+        return
+    }
+
+    # Return the name of the parent disk device
+    echo "$_udevinfo" | grep "DEVNAME=" | sed 's/DEVNAME=//'
+}
+
+# Find the right device to run check on
+check_dev=$(get_check_dev "$livedev")
 # CD/DVD media check
-[ -b "$livedev" ] && fs=$(blkid -s TYPE -o value "$livedev")
+[ -b "$check_dev" ] && fs=$(blkid -s TYPE -o value "$check_dev")
 if [ "$fs" = "iso9660" -o "$fs" = "udf" ]; then
     check="yes"
 fi
@@ -42,10 +69,10 @@ getarg rd.live.check -d check || check=""
 if [ -n "$check" ]; then
     type plymouth > /dev/null 2>&1 && plymouth --hide-splash
     if [ -n "$DRACUT_SYSTEMD" ]; then
-        p=$(dev_unit_name "$livedev")
+        p=$(dev_unit_name "$check_dev")
         systemctl start checkisomd5@"${p}".service
     else
-        checkisomd5 --verbose "$livedev"
+        checkisomd5 --verbose "$check_dev"
     fi
     if [ $? -eq 1 ]; then
         die "CD check failed!"
--- a/modules.d/90dmsquash-live/module-setup.sh
+++ b/modules.d/90dmsquash-live/module-setup.sh
@@ -22,7 +22,7 @@ installkernel() {
 
 # called by dracut
 install() {
-    inst_multiple umount dmsetup blkid dd losetup blockdev find rmdir
+    inst_multiple umount dmsetup blkid dd losetup blockdev find rmdir grep
     inst_multiple -o checkisomd5
     inst_hook cmdline 30 "$moddir/parse-dmsquash-live.sh"
     inst_hook cmdline 31 "$moddir/parse-iso-scan.sh"


modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -33,19 +33,34 @@ overlay_size=$(getarg rd.live.overlay.size=)
 getargbool 0 rd.live.overlay.thin && thin_snapshot="yes"
 getargbool 0 rd.live.overlay.overlayfs && overlayfs="yes"
 
+# Take a path to a disk label and return the patent disk if it is a partition
```
rvykydal, line 36: a typo, patent -> parent

rvykydal (approved):
Looks good to me, thank you.
But I don't feel strong enough in dracut land so I'd prefer also second eye on the patch.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -33,19 +33,34 @@ overlay_size=$(getarg rd.live.overlay.size=)
...
+# Otherwise returns the original path
+function get_check_dev() {
+    dev_path="$(udevadm info -q path --name "$1")"
+    udevadm info -q property --path "${dev_path}" | grep -q 'DEVTYPE=partition' || { echo "$1"; return; }
+    parent="${dev_path%/*}"
+    udevadm info -q property --path "${parent}" | grep -q 'DEVTYPE=disk' || { echo "$1"; return; }
+    udevadm info -q property --path "${parent}" | grep -q 'ID_FS_TYPE=iso9660' || { echo "$1"; return; }
+
+    # Return the name of the parent disk device
+    udevadm info -q property --path "${parent}" | grep 'DEVNAME=' | sed 's/DEVNAME=//'
```
aafeijoo-suse, line 46: Now the module should install `grep` and `sed` here:

https://github.com/dracutdevs/dracut/blob/fe8fa2b0cadbb33e27c8dd8b5851548dcd65835c/modules.d/90dmsquash-live/module-setup.sh#L25

aafeijoo-suse (commented):
Also, you need to change your commit message from:

`dmsquash-live-root: Run checkisomd5 on correct device`

to something like:

`fix(dmsquash-live): run checkisomd5 on correct device`

See https://github.com/dracutdevs/dracut/blob/master/docs/HACKING.md#commit-messages

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -33,19 +33,34 @@ overlay_size=$(getarg rd.live.overlay.size=)
 getargbool 0 rd.live.overlay.thin && thin_snapshot="yes"
 getargbool 0 rd.live.overlay.overlayfs && overlayfs="yes"
 
+# Take a path to a disk label and return the patent disk if it is a partition
+# Otherwise returns the original path
+function get_check_dev() {
+    dev_path="$(udevadm info -q path --name "$1")"
+    udevadm info -q property --path "${dev_path}" | grep -q 'DEVTYPE=partition' || { echo "$1"; return; }
```
pvalena, line 40: I think it's better to use "strstr" instead.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -33,19 +33,34 @@ overlay_size=$(getarg rd.live.overlay.size=)
...
+    parent="${dev_path%/*}"
+    udevadm info -q property --path "${parent}" | grep -q 'DEVTYPE=disk' || { echo "$1"; return; }
```
pvalena, line 42: Probably best to save this to a variable and use strstr as well.

Also, best to check with ^ and $ on strings with grep (not sure about strstr).

pvalena (changes requested):
TBH I'd rather rewrite this and parse the entries line-by-line using `for` and `case`.

modules.d/90dmsquash-live/dmsquash-live-root.sh
bcl: You can't use ^$, the output ends up as one big string so you are actually testing to see if it is present. Good idea saving it to a variable though.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -33,19 +33,46 @@ overlay_size=$(getarg rd.live.overlay.size=)
...
+        echo "$1"
+        return
+    }
+
+    # Return the name of the parent disk device
+    echo "$_udevinfo" | grep "DEVNAME=" | sed 's/DEVNAME=//'
+}
+
+# Find the right device to run check on
+check_dev=$(get_check_dev "$livedev")
```
FGrose, line 62: How about 
check_dev=$(lsblk -npo PKNAME "$livedev")

bcl: Hm, yeah, that looks like it works for my test cases, but it doesn't have any kind of sanity checking for iso9660 on the parent device. I'm trying to limit the impact of the change as much as possible :)

FGrose: @bcl Don't lines 37-40 already test for iso9660 or udf filesystem types?

bcl: Yes, but they don't fall back to livedev if they fail. In practice this likely doesn't matter, but I wanted the failure case to act as much like it does currently as possible.

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -33,19 +33,46 @@ overlay_size=$(getarg rd.live.overlay.size=)
 getargbool 0 rd.live.overlay.thin && thin_snapshot="yes"
 getargbool 0 rd.live.overlay.overlayfs && overlayfs="yes"
 
+# Take a path to a disk label and return the parent disk if it is a partition
```
LaszloGombos, line 36: Would it make sense to add this function to dracut-lib.sh instead to make it accessible in other modules as well as needed ?

modules.d/90dmsquash-live/module-setup.sh
```diff
@@ -22,7 +22,7 @@ installkernel() {
 
 # called by dracut
 install() {
-    inst_multiple umount dmsetup blkid dd losetup blockdev find rmdir
+    inst_multiple umount dmsetup blkid dd losetup blockdev find rmdir grep sed
```
LaszloGombos, line 25: sed is already installed by the base module, which is a dependency to this dmsquash-live module, so no use to repeat it here.

modules.d/90dmsquash-live/dmsquash-live-root.sh
bcl: I'm not sure. I think it's probably of limited usefulness outside of dmsquash-live.

modules.d/90dmsquash-live/module-setup.sh
bcl: Ah, ok, I'll drop that.

bcl:
@LaszloGombos looks like one of the tests timed out. I don't think I have the power to retry, can you?

johannbg (approved):
LGTM

PR #1885, opened 2022-08-02
fix(dracut-init.sh): propagate the result code returned by dracut-install

aafeijoo-suse:
Currently only the `inst_multiple` function propagates the result code returned by `dracut-install`.
This was missing since 043 (e.g.: https://github.com/dracutdevs/dracut/blob/043/dracut-functions.sh#L751)

## Checklist


lnykryn (approved):
LGTM. Some parts of the code assume that inst and friends return a proper return code.

johannbg (approved):
LGTM

PR #1886, opened 2022-08-02
fix(dmsquash-live-root): check kernel for built-in `overlay` drivers

FGrose:
Do not assume that `overlay` is always a module.  Check first for
`overlay` in `/proc/filesystems`.

Signed-off-by: Federico Vaga <federico.vaga@cern.ch>

See this discussion thread:
https://www.spinics.net/lists/linux-initramfs/msg04900.html

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -177,7 +177,7 @@ do_live_overlay() {
         fi
     fi
     if [ -n "$overlayfs" ]; then
-        if ! modprobe overlay; then
+        if ! { strstr "$(< /proc/filesystems)" overlay || modprobe overlay; }; then
             if [ "$overlayfs" = required ]; then
                 die "OverlayFS is required but not available."
                 exit 1


marcosfrm:
`modprobe foo` should return 0 for builtin modules, no?

```
# modinfo ext4
name:           ext4
filename:       (builtin)
softdep:        pre: crc32c
license:        GPL
file:           fs/ext4/ext4
description:    Fourth Extended Filesystem
author:         Remy Card, Stephen Tweedie, Andrew Morton, Andreas Dilger, Theodore Ts'o and others
alias:          fs-ext4
alias:          ext3
alias:          fs-ext3
alias:          ext2
alias:          fs-ext2
# modprobe ext4
# echo $?
0
```

FedericoVaga:
Your example is correct, and you are right, `modprobe` behaves as you described. However, it does behave like that only if you have this file `/lib/modules/<kver>/modules.builtin` in your filesystem. If you do not have it, you will get the following message:

```
modprobe overlay                                                                                               
```

If you run `dracut` with `--no-kernel` you do not get any ` /lib/modules/<kver>`

What I implemented in this patch is also what has been done already in the NFS module `modules.d/95nfs/nfs-start-rpc.sh ` (see a67a6f957dfd8ca51dea1ac5f8ca51a7646cca0f).

marcosfrm:
Always copy `/lib/modules/$(uname -r)/modules.builtin*` to the initramfs and avoid these hacks?

marcosfrm:
I must say I do not fully know `--no-kernel` use case, hence your fix may be better after all.

FedericoVaga:
> Always copy `/lib/modules/$(uname -r)/modules.builtin*` to the initramfs and avoid these hacks?

If this become a requirement, then it should be done by `dracut` automatically, not by users.

> I must say I do not fully know `--no-kernel` use case, hence your fix may be better after all.

In my case, I do not want any kernel stuff to be in the initrd because I have all I need in the kernel image. I need an initrd stage to perform a couple of actions, and I want this to be decoupled from the kernel. Meaning that whenever I update the kernel I can forget about rebuilding the initrd image, or to update final rootfs.

marcosfrm:
> Meaning that whenever I update the kernel I can forget about rebuilding the initrd image, or to update final rootfs.

Right. Your patch makes sense then. Sorry for the noise. On a side note, kernel should put this information somewhere in `/sys` to allow libkmod detect it without having to parse any index file...

FedericoVaga:
> Right. Your patch makes sense then. Sorry for the noise. On a side note, kernel should put this information somewhere in `/sys` to allow libkmod detect it without having to parse any index file...

It would be a nice option. Something along the line of `/proc/config.gz` and `/sys/kernel/kheaders.tar.xz`

lnykryn (approved):

lnykryn:
LGTM

LaszloGombos:
> Do not assume that overlay is always a module.
Is this only a concern for the the overlay module ? As an example same can be said for the squashfs module.

Perhaps this is an opportunity to introduce this piece of code as a reusable function in dracut-functions.sh, so that other modules can also start using it as needed.

`{ strstr "$(< /proc/filesystems)" $1 || modprobe $1; };`

FedericoVaga:
> Is this only a concern for the the overlay module ? As an example same can be said for the squashfs module.

You are right. As I mentioned, I copied the proposed code from the NFS module, and probably the problem is present in other modules. I'm not a dracut developer, but what you suggest sounds to me the right approach: for sure there are 2 modules needing it :)

LaszloGombos:
One could argue to fix the bug and refactor code later with a follow-up PR - especially if @FedericoVaga is not in a position to do so in this PR.

LGTM.

johannbg (approved):
LGTM

PR #1894, opened 2022-08-08
refactor(url-lib): write curl output directly to file

dtardon:
This pull request makes curl write directly to a file instead of using redirection via stdout. The redirection was originally added in 5c274467d6a4959b7f451f9a211ffbb38d69b641 to workaround https://bugzilla.redhat.com/show_bug.cgi?id=989133 (curl would create no file if the source was empty), but I don't see that problem with current curl.

A nice side-effect of this change is that curl shows download progress.

## Checklist


johannbg (approved):
LGTM

PR #1896, opened 2022-08-13
fix(dmsquash-live): check kernel for built-in squashfs drivers

LaszloGombos:
Check first for squashfs in `/proc/filesystems`.
This is needed in the --no-kernel use case to avoid the error:

Introduce load_fstype function to make it easier to check `/proc/filesystems`
before calling modprobe.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -89,7 +89,7 @@ det_img_fs() {
     blkid -s TYPE -u noraid -o value "$1"
 }
 
-modprobe squashfs
+load_fstype squashfs
 CMDLINE=$(getcmdline)
 for arg in $CMDLINE; do
     case $arg in
@@ -204,7 +204,7 @@ do_live_overlay() {
         fi
     fi
     if [ -n "$overlayfs" ]; then
-        if ! { strstr "$(< /proc/filesystems)" overlay || modprobe overlay; }; then
+        if ! load_fstype overlay; then
             if [ "$overlayfs" = required ]; then
                 die "OverlayFS is required but not available."
                 exit 1
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -1152,3 +1152,10 @@ remove_hostonly_files() {
         done < /lib/dracut/hostonly-files
     fi
 }
+
+# parameter: kernel_module filesystem_name
+# returns OK if kernel_module is loaded
+# modprobe fails if /lib/modules is not available (--no-kernel use case)
+load_fstype() {
+    strstr "$(cat /proc/filesystems)" "$1" || modprobe "$1"
+}


LaszloGombos:
Follow-up to https://github.com/dracutdevs/dracut/pull/1886#issuecomment-1207294432

CC @FGrose

FedericoVaga:
Thanks for fixing it. If I may suggest, this fix should be also applied for the NFS module

    modprobe: FATAL: Module sunrpc not found in directory /lib/modules/<kver>

If it could help, here it is the full list of error I get:

    modprobe: FATAL: Module sunrpc not found in directory /lib/modules/<kver>
    modprobe: FATAL: Module squashfs not found in directory /lib/modules/<kver>                                                                                          

Probably there are more, but I'm not exercising all dracut modules

LaszloGombos:
> If I may suggest, this fix should be also applied for the NFS module

Thanks @FedericoVaga . Indeed. I prefer to fix them one module at a time to keep the code reviews better manageable. virtiofs module is also impacted by a similar concern in addition to the NFS module.

johannbg (approved):
LGTM

LaszloGombos:
> If it could help, here it is the full list of error I get

@FedericoVaga if you're in a position, you could further help by testing https://github.com/dracutdevs/dracut/pull/1919, which should resolve the issue with the NFS module as well.

FedericoVaga:
I still see the error concerning the `squashfs` module. Could it be due to the fact that the output of `/proc/filesystems` is not regular for all filesystems. For example, on my system I have this:

```
~ $ cat /proc/filesystems | grep -E "over|squ"
        squashfs
nodev   overlay
```
Not sure how `strstr()` works, but perhaps is not parsing correctly the output for the squashfs line

LaszloGombos:
@FedericoVaga Thanks for checking.

```
strstr() {                                                                                                                                                                                                   
    [ "${1##*"$2"*}" != "$1" ]                                                                                                                                                                               
}  

```

It does not look seem like `strstr()` is the problem. 

Perhaps there is another module with unchecked `modprobe squashfs`. Do you happened to have the `squash` dracut module in your configuration ?

FedericoVaga:
No, I'm not directly including it, nor it seems included as a dependency

LaszloGombos:
> If it could help, here it is the full list of error I get:

> ```
> modprobe: FATAL: Module sunrpc not found in directory /lib/modules/<kver>
> modprobe: FATAL: Module squashfs not found in directory /lib/modules/<kver>                                                                                          
> modprobe: FATAL: Module overlay not found in directory /lib/modules/<kver>

>```

@FedericoVaga This PR has now also landed, so you should be able to just test from master branch.

Would you be able to pass "rd.shell rd.debug loglevel=7" arguments to initramfs (you might be able to do this from grub shell if you use grub) and try to find the location where there is an attempt to "modprobe squashfs". 

The error might not be from dracut itself directly but from some other tool that dracut calls (indirectly).

FedericoVaga:
@LaszloGombos Sorry for the late answer, I had my holidays. I do not know when I will be able to test it, but as soon as I'm back on this topic I will let you know, or propose a patch if I find the problem and the solution myself

PR #1898, opened 2022-08-13
fix(virtiofs): ismounted has a dependency on the base module

LaszloGombos:
When dracut.sh is called with "--modules virtiofs", make sure
dracut-lib.sh is installed by making the base module a dependency.

This pull request changes...

## Checklist

Fixes #


LaszloGombos:
CC @germag

johannbg (approved):
LGTM

Issue #1897, opened 2022-08-13
Remove bashism from 90dmsquash-live module


LaszloGombos:
Most modules run using /bin/sh. 90dmsquash-live could also benefit from not requiring bash as a shell dependency.

Porting this here for some new contributors to get into contributing.

LaszloGombos:
Ideally the PR for this issue should also change TEST-16-DMSQUASH to exclude bash from initramfs.testing and invoke dracut.sh with "dash dmsquash-live qemu" and make the changes necessary to pass the modified test.

marcosfrm:
@johannbg 
In the light of https://github.com/dracutdevs/dracut/pull/1911#issuecomment-1215166546 , when the "flexibility" justifies removing bash dependency or not? What standard exactly? Can we measure it? My take on this is that shell syntax is already bad and limiting it to dash make things horrible.

johannbg:
It should be quite apparent that this is a nack just like nabijaczleweli PR's were and yes limit ourselves to dash makes things worse not better.

PR #1899, opened 2022-08-13
Debashify modules.d where possible

nabijaczleweli:
Within the modules.d domain:
  * everywhere that required/used bash needlessly is fixed. This also fixes a few implicit ones.
  * everything that uses a bash extension *and only if it uses a bash extension* has a bash shebang. Otherwise it uses a /bin/sh one.
  * there's also like one or two obvious bugs that I fixed
  * everything (that isn't explicit bash code) passes bullseye shellcheck -s {da,k,}sh

This brings the hard bash dependencies to:
  * 02caps (capsh runs bash always(??))
  * 35network-manager for nm-run.sh for printf %q
  * 80cms – i think you *could* make it bash-free but I gave up
  * 90dmsquash-live for exec -a @ntfs-3g
  * 95dasd_rules to parse hexadecimal numbers
  * 99uefi-lib to parse binary data; i think you reasonably *could* make it work with head -c &a. on *some* implementations (or on all with clever rearranging) but meh

Nevertheless, this is -121 net somehow, almost -1 line/file and -10 lines/hour on average!

### DIFF
--- a/modules.d/45url-lib/module-setup.sh
+++ b/modules.d/45url-lib/module-setup.sh
@@ -1,5 +1,4 @@
-#!/bin/bash
-# module-setup for url-lib
+#!/bin/sh
 
 # called by dracut
 check() {
@@ -15,11 +14,11 @@ depends() {
 
 # called by dracut
 install() {
-    local _dir _crt _crts _found _lib _nssckbi _p11roots _p11root
+    local _dir _crt _crts _found _lib _nssckbi
     inst_simple "$moddir/url-lib.sh" "/lib/url-lib.sh"
     inst_multiple -o ctorrent
     inst_multiple curl sed
-    if curl --version | grep -qi '\bNSS\b'; then
+    if curl --version | grep -qwiF 'NSS'; then
         # also install libs for curl https
         inst_libdir_file "libnsspem.so*"
         inst_libdir_file "libnsssysinit.so*"
@@ -28,27 +27,21 @@ install() {
     fi
 
     for _dir in $libdirs; do
-        [[ -d $dracutsysrootdir$_dir ]] || continue
+        [ -d "$dracutsysrootdir$_dir" ] || continue
         for _lib in "$dracutsysrootdir$_dir"/libcurl.so.* "$dracutsysrootdir$_dir"/libcrypto.so.*; do
-            [[ -e $_lib ]] || continue
-            if ! [[ $_nssckbi ]]; then
-                read -r -d '' _nssckbi < <(grep -F --binary-files=text -z libnssckbi "$_lib")
-            fi
-            read -r -d '' _crt < <(grep -E --binary-files=text -z "\.(pem|crt)" "$_lib" | sed 's/\x0//g')
-            [[ $_crt ]] || continue
-            [[ $_crt == /*/* ]] || continue
-            if [[ -e $_crt ]]; then
+            [ -e "$_lib" ] || continue
+            [ -z "$_nssckbi" ] && grep -Fqaz libnssckbi "$_lib" && _nssckbi=1
+            _crt="$(grep -Eazm1 "\.(pem|crt)" "$_lib" | tr -d '\0')"
+            [ "${_crt#/*/}" != "$_crt" ] || continue
+            if [ -e "$_crt" ]; then
                 _crts="$_crts $_crt"
                 _found=1
             fi
         done
     done
-    if [[ $_found ]] && [[ -n $_crts ]]; then
+    if [ -n "$_found" ]; then
         for _crt in $_crts; do
-            if ! inst "${_crt#$dracutsysrootdir}"; then
-                dwarn "Couldn't install '$_crt' SSL CA cert bundle; HTTPS might not work."
-                continue
-            fi
+            inst "${_crt#"$dracutsysrootdir"}" || dwarn "Couldn't install '$_crt' SSL CA cert bundle; HTTPS might not work."
         done
     fi
     # If we found no cert bundle files referenced in libcurl but we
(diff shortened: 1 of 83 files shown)


modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -124,7 +124,7 @@ else
         # at the first glance, but ends with lots and lots of squashfs
         # errors, because systemd attempts to kill the ntfs-3g process?!
         if [ -x "/usr/bin/ntfs-3g" ]; then
-            (exec -a @ntfs-3g ntfs-3g -o "${liverw:-ro}" "$livedev" /run/initramfs/live) | vwarn
+            ntfs-3g -o "${liverw:-ro}" "$livedev" /run/initramfs/live | vwarn
```
marcosfrm, line 127: Bash's `exec -a @ntfs-3g ...` is intentional:

https://systemd.io/ROOT_STORAGE_DAEMONS/

LaszloGombos:
Most tests seem to be failing with the following error:
> dracut: systemd does not work with a prefix starting with "/run/"!

modules.d/00systemd/module-setup.sh
```diff
@@ -186,33 +172,27 @@ install() {
...
         chmod 444 "$initdir/etc/machine-id"
     fi
 
     # install adm user/group for journald
+    # we don't use systemd-networkd, but the user is in systemd.conf tmpfiles snippet
     inst_multiple nologin
-    {
-        grep '^systemd-journal:' "$dracutsysrootdir"/etc/passwd 2> /dev/null
-        grep '^adm:' "$dracutsysrootdir"/etc/passwd 2> /dev/null
-        # we don't use systemd-networkd, but the user is in systemd.conf tmpfiles snippet
```
LaszloGombos, line 203: I think it is useful to keep the comment.

modules.d/00systemd/module-setup.sh
```diff
@@ -186,33 +172,27 @@ install() {
...
+         -e '^adm:' \
+         -e '^systemd-network:' "$dracutsysrootdir"/etc/passwd >> "$initdir/etc/passwd" 2> /dev/null
 
-    {
-        grep '^systemd-journal:' "$dracutsysrootdir"/etc/group 2> /dev/null
-        grep '^wheel:' "$dracutsysrootdir"/etc/group 2> /dev/null
-        grep '^adm:' "$dracutsysrootdir"/etc/group 2> /dev/null
-        grep '^utmp:' "$dracutsysrootdir"/etc/group 2> /dev/null
-        grep '^root:' "$dracutsysrootdir"/etc/group 2> /dev/null
-        # we don't use systemd-networkd, but the user is in systemd.conf tmpfiles snippet
```
LaszloGombos, line 213: I think it is useful to keep the comment.

modules.d/01fips/module-setup.sh
```diff
@@ -13,30 +13,28 @@ depends() {
...
+    if [ -f "${srcmods}/modules.fips" ]; then
         _fipsmodules="$(cat "${srcmods}/modules.fips")"
     else
-        _fipsmodules=""
-
         # Hashes:
-        _fipsmodules+="sha1 sha224 sha256 sha384 sha512 "
-        _fipsmodules+="sha3-224 sha3-256 sha3-384 sha3-512 "
-        _fipsmodules+="crc32c crct10dif ghash "
+        _fipsmodules="sha1 sha224 sha256 sha384 sha512 "
```
LaszloGombos, line 20: This should be _fipsmodules="$_fipsmodules sha1 sha224 sha256 sha384 sha512 "

modules.d/01fips/module-setup.sh
```diff
@@ -13,30 +13,28 @@ depends() {
...
+        _fipsmodules="$fipsmodules sha3-224 sha3-256 sha3-384 sha3-512 "
```
LaszloGombos, line 21: Perhaps you need to use $_fipsmodules instead of $fipsmodules

modules.d/00systemd/module-setup.sh
```diff
@@ -23,10 +23,8 @@ installkernel() {
 
 # called by dracut
 install() {
-    local _mods
-
-    if [[ $prefix == /run/* ]]; then
```
LaszloGombos, line 28: This is the change that seem to break tests

LaszloGombos (commented):
I would suggest to break this PR up into several PRs to not only make it easier for you but also reviewers.

modules.d/90dmsquash-live/dmsquash-live-root.sh
nabijaczleweli: I somehow completely missed the `@`, so it looked superfluous. Considering #1897, would you prefer this to get rolled back to the bash spelling or something like
```sh
(
  ln -s /usr/bin/ntfs-3g /run/@ntfs-3g
  (sleep 1 && rm /run/@ntfs-3g) &
  PATH=/run
  exec @ntfs-3g -o ...
) | vwarn
```

modules.d/00systemd/module-setup.sh
nabijaczleweli: The comment moved above the greps:
```sh
    # install adm user/group for journald
    # we don't use systemd-networkd, but the user is in systemd.conf tmpfiles snippet
    inst_multiple nologin
    grep -e '^systemd-journal:' \
         -e '^adm:' \
         -e '^systemd-network:' "$dracutsysrootdir"/etc/passwd >> "$initdir/etc/passwd" 2> /dev/null
    grep -e '^systemd-journal:' \
         -e '^wheel:' \
         -e '^adm:' \
         -e '^utmp:' \
         -e '^root:' \
         -e '^systemd-network:' "$dracutsysrootdir"/etc/group >> "$initdir/etc/group" 2> /dev/null
```

modules.d/00systemd/module-setup.sh
nabijaczleweli: Yeah; fixed

modules.d/01fips/module-setup.sh
nabijaczleweli: Yeah; fixed.

LaszloGombos:
@nabijaczleweli Can you please also check the linting failures 
https://github.com/dracutdevs/dracut/runs/7827534646?check_suite_focus=true

modules.d/90dmsquash-live/dmsquash-live-root.sh
marcosfrm: \*sigh\* simple feature bash/ksh/zsh's `exec` have. I would avoid that, it is awful.

nabijaczleweli:
I've cleft this (and rebased) into:
  * `[01]*`: #1911
  * `[3-8]*`: #1912
  * `9[01]*`: #1913
  * `95*`: #1914
  * `9[6-9]*`: #1915

Across roughly similar commit counts (17+15+19+23+16).

LaszloGombos:
Thanks @nabijaczleweli  for splitting up this PR into several more manageable PRs. Should we close this PR than ?

PR #1908, opened 2022-08-14
test(16-DMSQUASH): make it pass on arch

LaszloGombos:
Include additional kernel drivers and install squashfs-tools
into the arch container.


johannbg (approved):
LGTM

PR #1911, opened 2022-08-15
Debashify modules.d where possible: [01]*

nabijaczleweli:
#1899 but just modules.d/[01]*


johannbg:
FYI We wont be debashifying any modules. In the future you might want to create an issue prior to such a large scope so you wont be spending time on something that will get nacked.

PR #1915, opened 2022-08-15
Debashify modules.d where possible: 9[6-9]* 

nabijaczleweli:
https://github.com/dracutdevs/dracut/pull/1899 but just modules.d/9[6-9]*

### DIFF
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -1157,5 +1157,5 @@ remove_hostonly_files() {
 # returns OK if kernel_module is loaded
 # modprobe fails if /lib/modules is not available (--no-kernel use case)
 load_fstype() {
-    strstr "$(cat /proc/filesystems)" "$1" || modprobe "$1"
+    grep -qF "$1" /proc/filesystems || modprobe "$1"
 }
--- a/modules.d/96securityfs/module-setup.sh
+++ b/modules.d/96securityfs/module-setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # called by dracut
 check() {
--- a/modules.d/97biosdevname/module-setup.sh
+++ b/modules.d/97biosdevname/module-setup.sh
@@ -1,8 +1,8 @@
-#!/bin/bash
+#!/bin/sh
 
 # called by dracut
 check() {
-    [[ "$mount_needs" ]] && return 1
+    [ -n "$mount_needs" ] && return 1
     require_binaries biosdevname || return 1
     return 0
 }
--- a/modules.d/97masterkey/module-setup.sh
+++ b/modules.d/97masterkey/module-setup.sh
@@ -1,8 +1,8 @@
-#!/bin/bash
+#!/bin/sh
 
 # called by dracut
 check() {
-    [[ $hostonly ]] && {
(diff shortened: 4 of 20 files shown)


modules.d/99base/dracut-lib.sh
```diff
@@ -1157,5 +1157,5 @@ remove_hostonly_files() {
 # returns OK if kernel_module is loaded
 # modprobe fails if /lib/modules is not available (--no-kernel use case)
 load_fstype() {
-    strstr "$(cat /proc/filesystems)" "$1" || modprobe "$1"
```
LaszloGombos, line 1160: Can you please elaborate on this change ? Why are you switching to grep as part of this PR ?

nabijaczleweli: Because I did so in the original in both places I noticed it and trunk moved one of them?

LaszloGombos: I do not understand what "both places" and "trunk moved one of them" means. How is this particular change related to debashify ?

nabijaczleweli: "Both" means two of two, and "trunk moved one of them" means exactly the same: https://github.com/dracutdevs/dracut/commit/922c9e28ed87815cf6ae0b5ee911ff0ef616d1b0. In that case that was spelled with a ksh extension; consider this (in particular this hunk which is marked `refactor: strstr cat -> grep`) a free optimisation pass.

LaszloGombos: > "Both" means two of two, and "trunk moved one of them" means exactly the same:
I genuinely did not understood which "two" you're referring to - you could have just listed a reference/link to the two places.  

Here is the link to this project code of conduct - https://github.com/dracutdevs/dracut/blob/master/docs/CODE_OF_CONDUCT.md

LaszloGombos:
I understand you put in a lot of effort into these changes.  I am concern about the review process for a large PRs.

To further help with the reviews:

1./ I understand these PRs can be reviewed independently, but first the project as a whole needs to discuss and agree a desire to reduce the dependency on bash. It is best to have this discussion in one PR only, so I suggest to keep only one PR (perhaps this PR open) and move other PRs into draft state for now. Once at least one of the PRs are landed and the project reached consensus on this, you might be able to resume reviews for multiple PRs in parallel. 
 
2./ Splitting up PRs based on module numbers are arbitrary. I suggest you consider the semantics and the dependency of the modules for organizing this work - especially for the first PR.

3./ It is generally a good advise not to combine unrelated changes - but especially important when there is a concern raised about the size of the PR. If the title and the commit message indicates Debashify as the main goal, than do not include "free optimisation" in the same PR. Feel free to open separate PRs for the unrelated changes. 

4./ module-setup.sh script is executed during dracut.sh run and dracut.sh still has a hard dependency on bash. I suggest to remove the changes from module-setup.sh from this PR.

5./ It would be good to have a test case that, so that this is not just a one time drive-by effort but also considers the maintenance cost for the project. You can either create a new test case or modify existing test cases (e.g. either TEST-01-BASIC or TEST-16-DMSQUASH) to actually confirm that the test passes without bash being in the generated initramfs.

For your first PR, you should only include the minimal number of changes that make the test pass (but without the PR the test should ideally fail).

nabijaczleweli:
Decided I don't care

PR #1916, opened 2022-08-15
fix(kernel-modules): always include nvmem driver on nvmem_on_arm

lnykryn:
These drivers are need to boot on some SoCs like NXP i.MX
We should include them so installation images will work.

This pull request changes...

## Checklist

Fixes #


LaszloGombos:
LGTM

johannbg (approved):
LGTM

PR #1917, opened 2022-08-15
fix(drm): add video drivers needed on hyper-v and similar

lnykryn:
Due to non-availability of Hyper-V video driver hyperv_drm in kdump
initramfs, the console seems to be in hang state with no text over it.

We should also go through the /sys/bus/vmbus/devices and include drivers
referenced there.

For reference: https://bugzilla.redhat.com/show_bug.cgi?id=2099502


johannbg (approved):
LGTM

PR #1919, opened 2022-08-15
fix(nfs,virtiofs): check kernel for builtin fs drivers

LaszloGombos:
Check first for fs support in `/proc/filesystems` before attempting
to load the kernel module.

This is needed in the --no-kernel use case to avoid the error:


LaszloGombos:
CC @FedericoVaga

FedericoVaga:
From the log I can see that I'm correctly run "dracut-057-40-ge4ad4a97" (this pull request), and I can confirm that the error concerning `sunrpc` is gone.

johannbg (approved):
LGTM

PR #1921, opened 2022-08-16
fix(skipcpio): ignore broken pipe

dtardon:
If lsinitrd is called from a context in which SIGPIPE is ignored (e.g.,
from a systemd unit with default setting of IgnoreSIGPIPE=), the
following line will result in an error being issued:

bin="$($SKIP "$image" | { read -r -N 6 bin && echo "$bin"; })"

An example error from `kdumpctl start` (which internally just calls
`systemctl start kdump.service`):

kdumpctl[1287]: ERROR: src/skipcpio/skipcpio.c:191:main(): fwrite

A minimal reproducer:

systemd-run -t sh -c '/path/to/skipcpio /path/to/any/file | false'

## Checklist

### DIFF
--- a/src/skipcpio/skipcpio.c
+++ b/src/skipcpio/skipcpio.c
@@ -23,6 +23,7 @@
 #define _GNU_SOURCE
 #endif
 
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

PR #1923, opened 2022-08-17
test: do not bring up the network interface if it is not needed

LaszloGombos:
Notice that TEST-04 does not bring up networking interface as
it is not needed. Apply the same principle for all local tests
that do not need networking.

This should make the tests run faster and slightly more reliable.

This change also brings us to a better place where network-legacy
files are only gets included via proper dracut module inclusion.

### DIFF
--- a/test/TEST-01-BASIC/test.sh
+++ b/test/TEST-01-BASIC/test.sh
@@ -38,15 +38,13 @@ test_setup() {
             mkdir -p -- dev sys proc etc var/run tmp
             mkdir -p root usr/bin usr/lib usr/lib64 usr/sbin
         )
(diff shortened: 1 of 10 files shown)


johannbg:
If we dont need networking we should not need any network related tools ( such as ping ip etc, ) either so if we can we should probably drop those as well.

LaszloGombos:
> if we can we should probably drop network related tools ( such as ping ip etc, ) as well.

Thanks for the review. It looks like we can indeed drop more tools. Updated the PR.

LaszloGombos:
not stale

johannbg (approved):

PR #1928, opened 2022-08-18
test(SYSTEMD): make the man command succeed

LaszloGombos:
Document the reason of the existing workaround.

Move the existing logic to the initramfs.testing generation phase
and make the steps more generic.


LaszloGombos:
not stale

johannbg (approved):

PR #1930, opened 2022-08-18
ci: add workflow for manual test run

LaszloGombos:
This workflow allows it to run any of the 29 tests
on any of the 4 containers.

On commit, currently the CI only runs 24 tests out of the 116
possible combinations of test runs.

As the project grows, the usefulness of this tool will also grow.

### DIFF
--- a/.github/workflows/manualtest.yml
+++ b/.github/workflows/manualtest.yml
@@ -1,0 +1,34 @@
+name: Manual test
+
+on:
+    workflow_dispatch:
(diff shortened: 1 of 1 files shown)


LaszloGombos:
Workflow will be under https://github.com/dracutdevs/dracut/actions/workflows/manualtest.yml

workflow


CC @aafeijoo-suse

johannbg (approved):

PR #1934, opened 2022-08-20
feat(overlayfs): add new module overlayfs

### LINKED ISSUE #1904: overlayfs with NFS directory fails
It does not seem possible to use overlayfs (dmsquash-live dracut module) with nfs (nfs dracut module). 
I could not find anything in the documentation that would prevent this configuration.

The following seem to fail:
`"root=nfs:192.168.50.1:/nfs/client rd.live.image rd.live.overlay.overlayfs=1" `

For reference the following works:
`"root=/dev/sda1 rd.live.image rd.live.overlay.overlayfs=1"`

LaszloGombos:
Earlier interests:
 - https://github.com/dracutdevs/dracut/pull/896 (approved by @johannbg stale bot closed it)
 - https://github.com/dracutdevs/dracut/pull/1874 (not integrated with existing modules, extra argument, no tests)

No new command line argument introduced. This new module can be supported with existing arguments and documentation.

## Checklist

Fixes #1904

### DIFF
--- a/modules.d/90dmsquash-live/mount-overlayfs.sh
+++ b/modules.d/90dmsquash-live/mount-overlayfs.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+PATH=/usr/sbin:/usr/bin:/sbin:/bin
+
+getargbool 0 rd.live.overlay.overlayfs && overlayfs="yes"
+getargbool 0 rd.live.overlay.reset -d -y reset_overlay && reset_overlay="yes"
+getargbool 0 rd.live.overlay.readonly -d -y readonly_overlay && readonly_overlay="--readonly" || readonly_overlay=""
+
+ROOTFLAGS="$(getarg rootflags)"
+
+if [ -n "$overlayfs" ]; then
+    if ! [ -e /run/rootfsbase ]; then
+        mkdir -m 0755 -p /run/rootfsbase
+        mount --bind "$NEWROOT" /run/rootfsbase
+    fi
+
+    mkdir -m 0755 -p /run/overlayfs
+    mkdir -m 0755 -p /run/ovlwork
+    if [ -n "$reset_overlay" ] && [ -h /run/overlayfs ]; then
+        ovlfs=$(readlink /run/overlayfs)
+        info "Resetting the OverlayFS overlay directory."
+        rm -r -- "${ovlfs:?}"/* "${ovlfs:?}"/.* > /dev/null 2>&1
+    fi
+    if [ -n "$readonly_overlay" ] && [ -h /run/overlayfs-r ]; then
+        ovlfs=lowerdir=/run/overlayfs-r:/run/rootfsbase
+    else
+        ovlfs=lowerdir=/run/rootfsbase
+    fi
+
+    if ! strstr "$(cat /proc/mounts)" LiveOS_rootfs; then
+        mount -t overlay LiveOS_rootfs -o "$ROOTFLAGS,$ovlfs",upperdir=/run/overlayfs,workdir=/run/ovlwork "$NEWROOT"
+    fi
+fi
--- a/modules.d/90overlayfs/module-setup.sh
+++ b/modules.d/90overlayfs/module-setup.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+check() {
+    [[ $hostonly ]] && return 1
+    return 255
+}
+
+depends() {
+    echo base
+}
+
+installkernel() {
+    instmods overlay
+}
+
+install() {
+    inst_hook mount 01 "$moddir/mount-overlayfs.sh"
+}
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -141,6 +141,9 @@ multipath:
 nvdimm:
   - modules.d/90nvdimm/*
 
+overlayfs:
+  - modules.d/90overlayfs/*
+
 ppcmac:
   - modules.d/90ppcmac/*
 
--- a/modules.d/90dmsquash-live/dmsquash-live-genrules.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-genrules.sh
@@ -15,7 +15,4 @@ case "$root" in
             /sbin/initqueue --settled --onetime --unique /sbin/dmsquash-live-root "${root#live:}"
         fi
         ;;
-    nfs*)
-        cp /sbin/mount-overlayfs "$hookdir/mount/99-mount-overlayfs.sh"
-        ;;
 esac
--- a/modules.d/90dmsquash-live/dmsquash-live-genrules.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-genrules.sh
@@ -15,4 +15,7 @@ case "$root" in
             /sbin/initqueue --settled --onetime --unique /sbin/dmsquash-live-root "${root#live:}"
         fi
         ;;
+    nfs*)
+        cp /sbin/mount-overlayfs "$hookdir/mount/99-mount-overlayfs.sh"
+        ;;
 esac
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -425,22 +425,8 @@ if [ -n "$overlayfs" ]; then
     else
         ln -sf /run/initramfs/live /run/rootfsbase
     fi
-    mkdir -m 0755 -p /run/overlayfs
-    mkdir -m 0755 -p /run/ovlwork
-    if [ -n "$reset_overlay" ] && [ -h /run/overlayfs ]; then
-        ovlfs=$(readlink /run/overlayfs)
-        info "Resetting the OverlayFS overlay directory."
-        rm -r -- "${ovlfs:?}"/* "${ovlfs:?}"/.* > /dev/null 2>&1
-    fi
-    if [ -n "$readonly_overlay" ] && [ -h /run/overlayfs-r ]; then
-        ovlfs=lowerdir=/run/overlayfs-r:/run/rootfsbase
-    else
-        ovlfs=lowerdir=/run/rootfsbase
-    fi
     if [ -z "$DRACUT_SYSTEMD" ]; then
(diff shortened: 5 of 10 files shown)


LaszloGombos:
Moved out overlayfs mount into its own dedicated file so that we can call into it directly for NFS support.

Added a new test case for overlayfs on top of NFS (which is now passing :-))

LaszloGombos:
To facilitate PR reviews (and help to prioritize), I wanted to express that from my contributor perspective, this one PR has by far the biggest impact on the dracut project.

This PR adds a functionality that many would use.

I understand review resources are very limited, so just wanted to communicate that getting feedback on this PR is more important (and likely more impactful) on the project than my other PR.

LaszloGombos:
definitely not stale

LaszloGombos:
CC  @Vogtinator @Mrfai @TylerHelt0

modules.d/90dmsquash-live/mount-overlayfs.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+PATH=/usr/sbin:/usr/bin:/sbin:/bin
```
Vogtinator: Why is that necessary?

modules.d/90dmsquash-live/mount-overlayfs.sh
```diff
@@ -0,0 +1,35 @@
...
+if [ -n "$overlayfs" ]; then
+    if ! [ -e /run/rootfsbase ]; then
+        mkdir -m 0755 -p /run/rootfsbase
+        mount --bind "$NEWROOT" /run/rootfsbase
+    fi
+
+    mkdir -m 0755 -p /run/overlayfs
+    mkdir -m 0755 -p /run/ovlwork
+    if [ -n "$reset_overlay" ] && [ -h /run/overlayfs ]; then
+        ovlfs=$(readlink /run/overlayfs)
```
Vogtinator: This variable has two different meanings, please split them

LaszloGombos: Thanks. Changed the name of the variable here to ovlfsdir .

Mrfai:
Thanks for working on the overlayfs module and adding support for using it with NFS. In my old Debian implementation, I've mounted a tmpfs on top of the rad-only NFS mount to make this writeable. How can I do this with the new implementation? I can't see anything similar in the code.

LaszloGombos:
@Mrfai 

>  In my old Debian implementation, I've mounted a tmpfs on top of the rad-only NFS mount to make this writeable. How can I do this with the new implementation? I can't see anything similar in the code.

`mount -t overlay ... upperdir=/run/overlayfs ..."`

/run is a tmpfs that is already created earlier in the dracut boot process.

Conan-Kudo (approved):
This looks great, thanks for this excellent refactor!

TylerHelt0:
These also don't support my original use case of booting a read-only ext4/btrfs/erofs filesystem protected by dm-verity as the lowerdir of a forgetful overlayfs root for kiosks/stability and to appease users who want security and a 'normal arch experience'. The rd.live options seem to only work on squashfs and ext4 **filesystem images** not **partitions**

LaszloGombos:
Thanks @TylerHelt0 . 

Lot to unbundle.  Appreciate that you're sharing the use-cases with details. I think lack of understanding of all the use-cases is an area to improve. Control over the size of overlay is certainly a good point.

> The rd.live options seem to only work on squashfs and ext4 filesystem images not partitions. 

I would start with this first. We have tests for partitions that seem to be passing -https://github.com/dracutdevs/dracut/commit/027eee902d0d7af4db45fcb7835d49990c9fafa1 .

Are you using upstream latest dracut ? Which distro ?

TylerHelt0:
I'm on Dracut 56-1 from the arch repos. 
I haven't actually tried using these options with an ext4 partition so they may work with the latest upstream. 
Also to note, both of the SFS images I used booted fine with default directory hierarchies and filenames expected by dracut and using only root=live:LABEL= syntax. The overlay image is easily resized by remounting /run. The dm-snapshot image is locked to   the size of the ext4 filesystem (I have not tried modifying the underlying dm devices)

LaszloGombos:
Kept the priority of the mount to 01 for overlayfs, as changing the mount priority is out of scope for this PR.

LaszloGombos:
@TylerHelt0 

> I'm on Dracut 56-1 from the arch repos.

Lot has changed in this area upstream since v56. You should try to build from this upstream repo and try again.  If building from upstream sounds too scary, you will likely able to just grab the modules.d directory from upstream instead of the a full dracut build.

TylerHelt0:
Arch has dracut-git that is built from github. I installed this and it solved an issue with lvm messages printing to console during boot but I haven't had a chance to test the rd.live options. 
I'm mainly interested in those for creating Live USBs. I wonder how fedora handles Copy to Ram. I bet they just use rootfs.img inside squashfs method. Could be wrong. 
I'll report back if I have time to test them using latest dracut-git.

johannbg (approved):

johannbg:
@aafeijoo-suse ping for review

LaszloGombos:
Rebased without the additional merge commit.

aafeijoo-suse (changes requested):

modules.d/90overlayfs/module-setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
aafeijoo-suse: Sorry to be picky, but I wouldn't add a new commit fixing a module that does not exist on the repo yet... Could you please update 137a4273c65f728444196ec418bc93c4e808cbcc with the new content and force push the branch? Thanks!

aafeijoo-suse (commented):
Just a couple of comments:

- In accordance with my previous comment https://github.com/dracutdevs/dracut/pull/1934#discussion_r1007873197, I think it will be better to merge the "refactor(overlayfs)" commit with "feat(overlayfs)", as you are refactoring a module that does not exist on the repo yet.

- One question regarding the new `mount-overlayfs.sh` script, just to verify that everyone is aware:

The current HEAD code in `dmsquash-live-root.sh` only adds a script to the mount hook dir if systemd is not around:

https://github.com/dracutdevs/dracut/blob/cf31fcf804be4dc0fa31885f5185a59b6012cdf4/modules.d/90dmsquash-live/dmsquash-live-root.sh#L440-L444

In your first commit (0b3030cf2a2e2f305d8da777a32c9f1da1d1011a), `mount -t overlay LiveOS_rootfs ...` keeps the same philosophy, but in your second commit (38eab1aab9cc1a2d235ef3ab1e658f7fcae9e2a9) the mount is tried to be done even with systemd present, just with a previous check that LiveOS_rootfs is not mounted yet, which can happens because the dmsquash-generator [is still there](https://github.com/dracutdevs/dracut/blob/master/modules.d/90dmsquash-live/module-setup.sh#L34-L35). Is this done intentionally?

LaszloGombos:
> in your second commit (https://github.com/dracutdevs/dracut/commit/38eab1aab9cc1a2d235ef3ab1e658f7fcae9e2a9) the mount is tried even with systemd present, just with a previous check

Thanks for pointing this out. Yes, this was done intentionally.

Since there is a longer term desire to separate systemd dracut modules into its own modules, we should keep systemd dependency out of overlayfs module. Perhaps at a later time we can introduce a systemd-overlayfs module, but that is not in scope for this PR.

To address the specific concern, I added an extra check to only pull in overlayfs as a dependency for dmsquash-live only if systemd is not installed.

aafeijoo-suse:
> To address the specific concern, I added an extra check to only pull in overlayfs as a dependency for dmsquash-live only if systemd is not installed.

Thanks for your clarification. If this was done intentionally and it's already approved by @Conan-Kudo and @johannbg, maybe you don't need to add this extra check. I just wanted to be sure that everyone is aware.

LaszloGombos:
@aafeijoo-suse Is there anything else you'd like to change on this PR ? Thanks !

aafeijoo-suse (approved):
It's ok, let's add this, we'll see if we need to add the systemd check later. Thanks.

daiaji:
https://github.com/Antynea/grub-btrfs/issues/214
```
cat /etc/dracut.conf.d/myflags.conf 
compress="lz4"
add_dracutmodules+="dmsquash-live"
```
```
cat /boot/grub/grub-btrfs.cfg
    menuentry '  vmlinuz-6.0-x86_64 & initramfs-6.0-x86_64.img & amd-ucode.img' --class snapshots --class gnu-linux --class gnu --class os $menuentry_id_option 'gnulinux-snapshots-ca0a7c48-b433-4667-a759-42b305229791' {
        if [ x$feature_all_video_module = xy ]; then
        insmod all_video
        fi
        set gfxpayload=keep
        insmod btrfs
        if [ x$feature_platform_search_hint = xy ]; then
            search --no-floppy --fs-uuid  --set=root  ca0a7c48-b433-4667-a759-42b305229791
        else
            search --no-floppy --fs-uuid  --set=root ca0a7c48-b433-4667-a759-42b305229791
        fi
        echo 'Loading Snapshot: 2022-12-10 19:12:05 snapshots_root/@.20221210T1912'
        echo 'Loading Kernel: vmlinuz-6.0-x86_64 ...'
        linux "/snapshots_root/@.20221210T1912/boot/vmlinuz-6.0-x86_64" root=UUID=ca0a7c48-b433-4667-a759-42b305229791  quiet udev.log_priority=3 rootflags=subvol="snapshots_root/@.20221210T1912" rd.live.overlay.readonly=1
        echo 'Loading Microcode & Initramfs: amd-ucode.img initramfs-6.0-x86_64.img ...'
        initrd "/snapshots_root/@.20221210T1912/boot/amd-ucode.img" "/snapshots_root/@.20221210T1912/boot/initramfs-6.0-x86_64.img"
    }
```
I tried using dracut to enable overlayfs for root booted from a read-only btrfs snapshot, but added the dmsquash-live module, and added the kernel command line parameter `rd.live.overlay.readonly=1` or `rd.live.overlay.overlayfs=1`, but nothing seems to work.

Are dmsquash-live and overlayfs modules not suitable for this use case?

AdamWill:
This broke Fedora/RHEL installer images, because it does not handle three paths on which overlayfs setup previously was reached in the script. See https://github.com/dracutdevs/dracut/issues/2232 .

PR #1937, opened 2022-08-22
WIP refactor: free up some namespace for out-of-tree mount hooks

LaszloGombos:
This change moves mount hooks into the 10-89 priority space.

iscsi: 01 --> 11
dmsquash-live: 01 --> 22
nbd: 01 --> 12
rootfs-block: 99 --> 70
virtfs: 99 --> 71
watchdog: 00,50,99 --> 10,50,89

## Checklist

Fixes #


LaszloGombos:
not stale

LaszloGombos:
This is part of a larger effort to try to leave some namespace empty (e.g. 0-9, 90-99), so that users of dracut (other projects, distro's, end-users) can fully take advantage of dracut's modular infrastructure (e.g. with out of tree modules or dracut command line arguments).

johannbg:
You need to be a bit more spesific about your larger effort since 90-99 range is intended for the stuff that dracut relies on.

LaszloGombos:
> 90-99 range is intended for the stuff that dracut relies on

I am advocating to explore if we should change this rule to 80-89 (instead of 90-99) - also filed an issue for this larger discussion - https://github.com/dracutdevs/dracut/issues/2015 .

Regardless, this PR is not to make this larger change (which would need more discussion and perhaps also documentation changes).

This PR is scoped to discuss if rootfs-block really need to be 99 priority.

johannbg:
Well you have not explained why priority 99 should be reserved for hooks that should be run last and I would say that rootfs-block really needs to be 99 priority or do you think it's not something that dracut needs to rely on anymore?

LaszloGombos:
Closing to take it out from active PR list, but plan to resume work at later time. Not targeting the upcoming release with these changes.

PR #1943, opened 2022-08-23
refactor(dmsquash-live): move ntfs-3g code to dmsquash-live-ntfs

LaszloGombos:
Since the installation of ntfs-3g binary is in the
dmsquash-live-ntfs module, calling ntfs-3g from dmsquash-live fails
when dmsquash-live-ntfs module is not installed anyways.

## Checklist

### DIFF
--- a/modules.d/90dmsquash-live-ntfs/mount-ntfs-3g.sh
+++ b/modules.d/90dmsquash-live-ntfs/mount-ntfs-3g.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+type vwarn > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+# Symlinking /usr/bin/ntfs-3g as /sbin/mount.ntfs seems to boot
+# at the first glance, but ends with lots and lots of squashfs
+# errors, because systemd attempts to kill the ntfs-3g process?!
+# See https://systemd.io/ROOT_STORAGE_DAEMONS/
+if [ -x "/usr/bin/ntfs-3g" ]; then
+    (
+        ln -s /usr/bin/ntfs-3g /run/@ntfs-3g
+        (sleep 1 && rm /run/@ntfs-3g) &
+        # shellcheck disable=SC2123
+        PATH=/run
+        exec @ntfs-3g "$@"
+    ) | vwarn
+else
+    die "Failed to mount block device of live image: Missing NTFS support"
+    exit 1
+fi
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -120,22 +120,7 @@ else
             exit 1
         fi
     else
-        # Symlinking /usr/bin/ntfs-3g as /sbin/mount.ntfs seems to boot
-        # at the first glance, but ends with lots and lots of squashfs
-        # errors, because systemd attempts to kill the ntfs-3g process?!
-        # See https://systemd.io/ROOT_STORAGE_DAEMONS/
-        if [ -x "/usr/bin/ntfs-3g" ]; then
-            (
-                ln -s /usr/bin/ntfs-3g /run/@ntfs-3g
-                (sleep 1 && rm /run/@ntfs-3g) &
-                # shellcheck disable=SC2123
-                PATH=/run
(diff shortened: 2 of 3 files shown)


LaszloGombos:
not stale

Conan-Kudo (approved):

iammattcoleman (approved):

LaszloGombos:
Rebased !

Conan-Kudo (approved):
Still looks good to me.

modules.d/90dmsquash-live-ntfs/mount-ntfs-3g.sh
```diff
@@ -0,0 +1,20 @@
...
+# at the first glance, but ends with lots and lots of squashfs
+# errors, because systemd attempts to kill the ntfs-3g process?!
+# See https://systemd.io/ROOT_STORAGE_DAEMONS/
+if [ -x "/usr/bin/ntfs-3g" ]; then
+    (
+        ln -s /usr/bin/ntfs-3g /run/@ntfs-3g
+        (sleep 1 && rm /run/@ntfs-3g) &
+        # shellcheck disable=SC2123
+        PATH=/run
+        exec @ntfs-3g -o "${liverw:-ro}" "$livedev" /run/initramfs/live
```
aafeijoo-suse, line 15: This is script is not parsing the arguments passed in `dmsquash-live-root.sh`. This should be :
```
exec @ntfs-3g "$@"
```

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -120,22 +120,7 @@ else
...
-                (sleep 1 && rm /run/@ntfs-3g) &
-                # shellcheck disable=SC2123
-                PATH=/run
-                exec @ntfs-3g -o "${liverw:-ro}" "$livedev" /run/initramfs/live
-            ) | vwarn
-        else
-            die "Failed to mount block device of live image: Missing NTFS support"
-            exit 1
-        fi
+        [ -x "/sbin/mount-ntfs-3g" ] && /sbin/mount-ntfs-3g "${liverw:-ro}" "$livedev" /run/initramfs/live
```
aafeijoo-suse, line 123: And here `/sbin/mount-ntfs-3g` would need the `-o` option:
```
/sbin/mount-ntfs-3g -o "${liverw:-ro}" "$livedev" /run/initramfs/live
```

aafeijoo-suse (changes requested):

modules.d/90dmsquash-live-ntfs/mount-ntfs-3g.sh
iammattcoleman: Good catch! I should've seen this one...

iammattcoleman (changes requested):
@aafeijoo-suse pointed out a couple definite issues.

aafeijoo-suse (approved):

marcosfrm:
I wonder if we can first try `mount -t ntfs3 ...` and only call `ntfs-3g` as a fallback.

```sh
        # prefer ntfs3 kernel driver, fallback to ntfs-3g
        if ! mount -n -t ntfs3 -o "${liverw:-ro}" "$livedev" /run/initramfs/live >/dev/null 2>&1; then
            [ -x "/sbin/mount-ntfs-3g" ] && /sbin/mount-ntfs-3g -o "${liverw:-ro}" "$livedev" /run/initramfs/live
        fi
```

Then add `ntfs3` dependency in `module-setup.sh`.

One day `ntfs3` will register itself as an alias to `ntfs` and all these hacks will be gone! \o/

Conan-Kudo:
Is `ntfs3` even maintained? I remember folks complaining that it seems to be pretty dead post-merge. I don't know if we want to make it prefer something that's kinda dead and probably going to be removed eventually.

marcosfrm:
It is now (5.19+ started getting fixes), in a slow pace.

LaszloGombos:
> I wonder if we can first try mount -t ntfs3 

This is already the case and ntfs3 is already handled by dmsquash-live dracut module.

dmsquash-live-ntfs dracut module is only needed for "old" ntfs - which is one of the reason I wanted to factor this code out from dmsquash-live as I expect less and less usage for dmsquash-live-ntfs over time and no reason to carry this code in dmsquash-live for ntfs3.

johannbg (approved):

PR #1945, opened 2022-08-24
test: list only the required modules explicitly

LaszloGombos:
rootfs-block module would always include the following
modules: udev-rules base fs-lib

This is a follow-up to a1d4041cee5f7b2d8da9192338d5e81315e758ef


johannbg (approved):
LGTM

PR #1951, opened 2022-08-25
test(FULL SYSTEMD): make test 04 use the test dracut modules

LaszloGombos:
Same as https://github.com/dracutdevs/dracut/pull/2070 but for test 04.


johannbg:
Why did you change the package description?

LaszloGombos:
> Why did you change the package description?

To match the name with downstream (https://github.com/archlinux/svntogit-packages/blob/packages/dracut/trunk/PKGBUILD) and with https://github.com/dracutdevs/dracut/blob/master/README.md

LaszloGombos:
not stale

LaszloGombos:
Not a priority for me any more..

test/TEST-04-FULL-SYSTEMD/test.sh
```diff
@@ -149,110 +99,21 @@ Type=oneshot
...
-        ln -fs ../testsuite.service "$initdir"/etc/systemd/system/testsuite.target.wants/testsuite.service
-
-        # make the testsuite the default target
-        systemctl --root="$initdir" set-default testsuite.target
 
-        # install basic tools needed
-        inst_multiple sh bash setsid loadkeys setfont \
-            login sulogin gzip sleep echo mount umount
-        inst_multiple modprobe
+    mkdir -p "$TESTDIR"/overlay/source && cp -a "$TESTDIR"/dracut.*/initramfs/* "$TESTDIR"/overlay/source && rm -rf "$TESTDIR"/dracut.*
```
aafeijoo-suse, line 103: See Line 66 to fix this one (same `mkdir` and `cp`).

test/TEST-04-FULL-SYSTEMD/test.sh
```diff
@@ -54,79 +52,33 @@ test_setup() {
...
+        -i ./fstab /etc/fstab \
+        -i "${basedir}/modules.d/99base/dracut-lib.sh" "/lib/dracut-lib.sh" \
+        -i "${basedir}/modules.d/99base/dracut-dev-lib.sh" "/lib/dracut-dev-lib.sh" \
+        --no-hostonly --no-hostonly-cmdline --nomdadmconf --nohardlink \
+        -f "$TESTDIR"/initramfs.root "$KVERSION" || return 1
+
+    mkdir -p "$TESTDIR"/overlay/source && cp -a "$TESTDIR"/dracut.*/initramfs/* "$TESTDIR"/overlay/source && rm -rf "$TESTDIR"/dracut.* && export initdir=$TESTDIR/overlay/source
+
+    if type -P rpm &> /dev/null; then
+        rpm -ql systemd | xargs -r "$basedir"/dracut-install ${initdir:+-D "$initdir"} -o -a -l
```
aafeijoo-suse, line 69: I'm not sure if we should continue testing systemd like this, this is installing and enabling all units, even some of them are not included by any dracut systemd module (that's why some tweaks are being made in the old test from line 205 to 235, missing with new changes). Maybe we can change the approach of this test, and do not install only the systemd module, but try to install all the dracut systemd modules (line 149).

aafeijoo-suse (commented):
Also, check if this is important (it didn't happen before this PR):

https://github.com/dracutdevs/dracut/actions/runs/3558519205/jobs/5977242900#step:4:236

test/TEST-04-FULL-SYSTEMD/test.sh
```diff
@@ -149,110 +99,21 @@ Type=oneshot
...
-            | while read -r i || [ -n "$i" ]; do
-                i=${i##Exec*=}
-                i=${i##-}
-                inst_multiple -o "$i"
-            done
-
-        # some helper tools for debugging
-        [[ $DEBUGTOOLS ]] && inst_multiple "$DEBUGTOOLS"
-
-        # install ld.so.conf* and run ldconfig
```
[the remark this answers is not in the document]
LaszloGombos: Yes, intentionally omitted. I do not mind adding them back, but I think we should be able to debug tests without changing the test files all the time.

test/TEST-04-FULL-SYSTEMD/test.sh
```diff
@@ -149,110 +99,21 @@ Type=oneshot
...
-
-        # install dbus and pam
-        inst_multiple -o \
-            /etc/dbus-1/** \
-            /etc/pam.d/** \
-            /etc/security/** \
-            /lib64/security/** \
-            /lib/security/**
-
-        # install dbus socket and service file
```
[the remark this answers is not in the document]
LaszloGombos: > the dbus module could be added explicitly
I added dbus dracut modules to initramfs.root. Adding dbus also pulled in additional dracut modules - e.g. "systemd-sysusers".  

This however did not resolve the "Failed to connect to bus: No such file or directory" and introduced a new issue (systemd-sysusers.service fails to start on fedora).

IMHO, perhaps it is good to have this issue as I think https://github.com/dracutdevs/dracut/pull/2067 will resolve this issue, which means that https://github.com/dracutdevs/dracut/pull/2067 is now gaining a test case.

test/TEST-04-FULL-SYSTEMD/test.sh
LaszloGombos: My goal was to keep the test simple and maintainable. The test as is already passing on all target containers. Would adding more code make the test more maintainable ? Or would it increase test coverage for _dracut_ by adding more code ? 

If I understand your suggestion correctly, you proposed to add more code to make initramfs.root more complete, but would that actually test dracut itself, or we would just end up testing systemd itself on the target container ? As I understand the main goal of these tests is to test initramfs.testing and not initramfs.root.
That said, I do not have a strong opinion on this and happy to add back the code from old test from line 205 to 235.

> try to install all the dracut systemd modules (line 149).
I think this is a good idea, but this is a suggestion for another PR. This is a suggestion to change the scope of this test case, and this PR is already complex enough just trying to do the refactoring.

LaszloGombos: > add back the code from old test from line 205 to 235.

I convinced myself to make this change now as the goal of this PR not to optimize, but just refactor.
It seems this additional piece of code only picks up the following 2 additional files:
- /sbin/agetty
- /usr/lib/systemd/systemd-random-seed

test/TEST-04-FULL-SYSTEMD/test.sh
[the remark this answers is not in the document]
aafeijoo-suse: I meant the "_# install ld.so.conf* and run ldconfig_" part.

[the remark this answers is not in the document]
LaszloGombos: That line also is not necessary - see the original related commit - https://github.com/dracutdevs/dracut/commit/7eb0ac977e2d4708f09dfca3ce9050fcfa476292

dracut.sh will build the ld cache now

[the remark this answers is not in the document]
aafeijoo-suse: Ok, I didn't remember that. Thanks for the clarification.

LaszloGombos:
@aafeijoo-suse Is there anything else that i can improve on this PR ?

johannbg (approved):

johannbg:
@LaszloGombos If all outlined issue that @aafeijoo-suse pointed out has been addressed a request from a re-review from him is missing...

test/TEST-04-FULL-SYSTEMD/test.sh
```diff
@@ -149,115 +107,24 @@ Type=oneshot
 StandardInput=tty
 StandardOutput=tty
 EOF
-        mkdir -p "$initdir"/etc/systemd/system/testsuite.target.wants
-        ln -fs ../testsuite.service "$initdir"/etc/systemd/system/testsuite.target.wants/testsuite.service
-
-        # make the testsuite the default target
-        systemctl --root="$initdir" set-default testsuite.target
```
aafeijoo-suse, line 156: I miss this, isn't it necessary?

test/TEST-04-FULL-SYSTEMD/test.sh
```diff
@@ -149,115 +107,24 @@ Type=oneshot
...
-        done
-
-        # some basic terminfo files
-        for _terminfodir in /lib/terminfo /etc/terminfo /usr/share/terminfo; do
-            [ -f ${_terminfodir}/l/linux ] && break
-        done
-        inst_multiple -o ${_terminfodir}/l/linux
-
-        # softlink mtab
-        ln -fs /proc/self/mounts "$initdir"/etc/mtab
```
aafeijoo-suse, line 203: Maybe this would avoid https://github.com/dracutdevs/dracut/actions/runs/3566589541/jobs/5993228395#step:4:447 , not happening before this PR.

test/TEST-04-FULL-SYSTEMD/test.sh
aafeijoo-suse: Ah, sorry, I forgot.

aafeijoo-suse (approved):

PR #1956, opened 2022-08-26
feat(fs-lib): make gentoo container pass test-01

### LINKED ISSUE #1952: consistent mapping of LABEL/UUID to devices
We have two functions mapping LABEL/UUID to devices. One of them is exact match (label_uuid_to_dev) and it is used in most modules.

There is however an other function that does fuzzy matching and returns with a _list_ of devices (devnames). This is to allow to be sloppy with LABEL/UUID and allow for e.g. "UUID=123" device specification.

fstab_mount might end up running  fsck_single on a different device than the actual device that is being mounted.

I suggest to only use devnames in crypt-lib.sh and stick to label_uuid_to_dev on all other cases.

devnames does NOT support PARTLABEL nor PARTUUID, which is supported by label_uuid_to_dev.

LaszloGombos:
## Checklist

Fixes #1952 #2091

### DIFF
--- a/modules.d/99fs-lib/fs-lib.sh
+++ b/modules.d/99fs-lib/fs-lib.sh
@@ -146,11 +146,7 @@ fsck_single() {
     local _drv
 
     [ $# -lt 2 ] && return 255
-    # if UUID= marks more than one device, take only the first one
-    [ -e "$_dev" ] || _dev=$(devnames "$_dev" | while read -r line || [ -n "$line" ]; do if [ -n "$line" ]; then
-        echo "$line"
-        break
-    fi; done)
+    _dev=$(readlink -f "$(label_uuid_to_dev "$_dev")")
     [ -e "$_dev" ] || return 255
     _fs=$(det_fs "$_dev" "$_fs")
     fsck_able "$_fs" || return 255


LaszloGombos:
not stale

aafeijoo-suse (approved):

LaszloGombos:
This PR changes a function that is only called if systemd is not in initramfs.

This issue seem to impact Gentoo container passing test 01 (as it does not have systemd installed).

Full test run - https://github.com/LaszloGombos/dracut/actions/runs/3451869106/jobs/5761388877

johannbg:
Did you forget to update your commit message by any chance?

LaszloGombos:
Gentoo passing (with this PR) - https://github.com/LaszloGombos/dracut/actions/runs/3453311170/jobs/5763808493
Gentoo failing (without this PR) - https://github.com/dracutdevs/dracut/actions/runs/3453323734/jobs/5763829434

LaszloGombos:
@johannbg I will stop assigning if it is not useful. This is to help issues that are already reviewed by one person and need the second review. You can use https://github.com/dracutdevs/dracut/issues?q=is%3Aopen+assignee%3A%40me to quickly find them in the review queue. But if not useful, I will stop.

johannbg:
You never assign issues to individuals without discussing it with the relevant individuals before, be it your dayjob or F/OSS projects. We use what was design for this, which is labels and the label should be 'needs reviewing'

johannbg (approved):

PR #1957, opened 2022-08-26
chore(fs-lib.sh): remove unused fsck_batch

LaszloGombos:
## Checklist

### DIFF
--- a/modules.d/99fs-lib/fs-lib.sh
+++ b/modules.d/99fs-lib/fs-lib.sh
@@ -161,31 +161,6 @@ fsck_single() {
     return $?
 }
 
-# takes list of filesystems to check in parallel; we don't rely on automatic
-# checking based on fstab, so empty one is passed
-fsck_batch() {
-    local FSTAB_FILE=/etc/fstab.empty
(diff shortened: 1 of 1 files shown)


LaszloGombos:
fsck_batch was added over 10 years ago - https://github.com/dracutdevs/dracut/commit/fefab84fffc4bfbe8ee1ed9727e619c5b2f74601
It seems that in the last 10 years it was never used (or tested), better to remove it.

aafeijoo-suse (commented):
I'm not sure it's a good idea to remove a function from a "lib" module. Maybe it is being used by any custom user module.
See:
- https://github.com/void-ppc/foxlet-void-packages/blob/master/srcpkgs/dracut/patches/fix-bash-5.patch
- https://github.com/a00403a/MagicSpecA-E/blob/master/dracut/0057-99fs-lib-export-FSTAB_FILE-before-fsck-call.patch

LaszloGombos:
This is a small amount of code (requiring small maintenance) , so if this PR not straightforward than I prefer to take it out from the PR queue.

PR #1963, opened 2022-08-28
chore(udev-rules): remove obsolete code

LaszloGombos:
Arch linux no longer needs a special load-modules.sh file.

Seems this is one of those temporary workarounds from 2010 that for a long time no longer needed - https://github.com/dracutdevs/dracut/commit/5a4bbf1bda2a93452bba254f5adbb4c1c5276230


johannbg (approved):
LGTM

PR #1964, opened 2022-08-28
chore(base): do not output root entry while installing base

LaszloGombos:
Silence the printing out root entry of passwd file
while including base module.

dracut: *** Including module: base ***
root:x:0:0:root:/root:/bin/bash
dracut: *** Including module: fs-lib ***

## Checklist


johannbg (approved):
LGTM

PR #1965, opened 2022-08-28
chore(udev-rules): organize rules

LaszloGombos:
Move network rules to network meta module.

Mark distro specific rules as such and create a section
for legacy rules files that are not distro specific.

Remove fixme that's not actionable.

Add 95-udev-late.rules, it is in debian
(https://packages.debian.org/sid/arm64/udev/filelist).

This pull request changes...

## Checklist

### DIFF
--- a/modules.d/40network/module-setup.sh
+++ b/modules.d/40network/module-setup.sh
@@ -46,5 +46,15 @@ install() {
     inst_hook cmdline 91 "$moddir/dhcp-root.sh"
     inst_multiple ip sed awk grep pgrep tr
     inst_multiple -o arping arping2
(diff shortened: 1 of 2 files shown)


johannbg:
@LaszloGombos note that there are hidden dragons in poking udev rules so we need to be extra careful when poking those.
That said those udev rules should never be part of a meta module ( meta modules are not somekind of dumping ground for workarounds or generics ) and we should not be carrying/including any downstream distribution specific rules ( we are trying to get rid of any downstream distribution hacks, not add to them ) and or rules related to any forks of udev ( such as eudev )

PR #1966, opened 2022-08-29
feat(dracut.sh): populate uefi_cmdline if no other cmdline is given

pvalena:
feat(dracut.sh): populate uefi_cmdline if no other cmdline is given

    When creating uefi image in hostonly mode (with using --hostonly-cmdline),
    it makes sense to copy current kernel commandline from /proc/cmdline,
    in case there are no other options specified.

    Usually, the cmdline.d/*.conf file is generated by module rootfs-block
    (or other modules), but it might, not handle all cases correctly,
    and specifying --kernel-cmdline every time is not much user-friendly.

## Changes

`/proc/cmdline` will be used as a last option for for uefi `.cmdline` entry.

## Checklist
- [ ] I have tested it locally:
__TBD__

**TODO: document this**

Shouldn't be probably needed/desired to test this, as the host `/proc/cmdline` would need to be remounted.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2598,14 +2598,18 @@ umask 077
 if [[ $uefi == yes ]]; then
     if [[ $kernel_cmdline ]]; then
         echo -n "$kernel_cmdline" > "$uefi_outdir/cmdline.txt"
-    elif [[ $hostonly_cmdline == yes ]] && [ -d "$initdir/etc/cmdline.d" ]; then
-        for conf in "$initdir"/etc/cmdline.d/*.conf; do
-            [ -e "$conf" ] || continue
-            printf "%s " "$(< "$conf")" >> "$uefi_outdir/cmdline.txt"
-        done
+    elif [[ $hostonly_cmdline == yes ]]; then
+        if [ -d "$initdir/etc/cmdline.d" ]; then
+            for conf in "$initdir"/etc/cmdline.d/*.conf; do
+                [ -e "$conf" ] || continue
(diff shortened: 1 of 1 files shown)


dracut.sh
```diff
@@ -2598,14 +2598,18 @@ umask 077
...
-            [ -e "$conf" ] || continue
-            printf "%s " "$(< "$conf")" >> "$uefi_outdir/cmdline.txt"
-        done
+    elif [[ $hostonly_cmdline == yes ]]; then
+        if [ -d "$initdir/etc/cmdline.d" ]; then
+            for conf in "$initdir"/etc/cmdline.d/*.conf; do
+                [ -e "$conf" ] || continue
+                printf "%s " "$(< "$conf")" >> "$uefi_outdir/cmdline.txt"
+            done
+        elif [ -e "$initdir/proc/cmdline" ]; then
```
aafeijoo-suse, line 2607: Is the running system `/proc` mounted in the temporary directory where the initrd is being created?

aafeijoo-suse: The `initdir` is a temporary directory where dracut adds stuff before packing its content into an initramfs image.

https://github.com/dracutdevs/dracut/blob/4012314f204806a67d038f378f525646814d12ea/dracut.sh#L1258-L1259

So, `/var/tmp/dracut.??????/initramfs/proc/cmdline` should never exist while dracut is running.

johannbg (approved):
LGTM

PR #1969, opened 2022-09-01
feat(systemd): install systemd-sysroot-fstab-check

aafeijoo-suse:
`systemd-sysroot-fstab-check` is a symlink to `systemd-fstab-generator` added in
systemd commit https://github.com/systemd/systemd/commit/cd7ad0cb

systemd has a temporary patch until dracut installs it (https://github.com/systemd/systemd/commit/db527621)


johannbg (approved):
LGTM

Issue #1754, opened 2022-03-18
systemd-cryptsetup-generator does not work correctly with /etc/cmdline.d/01-default.conf (dracut-initqueue[479]: failed to start systemd-cryptsetup@luks-a4465013-c35b-4378-ae5f-c2bb7712b89e.service: Unit systemd-cryptsetup@luks-a4465013-c35b-4378-ae5f-c2bb7712b89e.service not found.)


T35R6braPwgDJKq:
**Describe the bug**
On a bios system (VM): i get
```
dracut-initqueue[479]: failed to start systemd-cryptsetup@luks-a4465013-c35b-4378-ae5f-c2bb7712b89e.service: Unit systemd-cryptsetup@luks-a4465013-c35b-4378-ae5f-c2bb7712b89e.service not found.
```
In the maintenance shell I can just call
```
/lib/systemd/systemd-decrypt attach a /dev/disk/by-uuid/4e56edf6-3e8a-4ef3-a084-09b40468bd89
```
and exit the shell - the boot continues

**Distribution used**
Archlinux

**Dracut version**
```
dracut --version
dracut 056
```

**Init system**
Which init system is being used?

**To Reproduce**
The dracut confs commandline
```
lsinitrd /boot/initramfs-linux.img -f etc/cmdline.d/01-default.conf
rw rd.info rd.shell rd.luks.timeout=30 rd.luks.crypttab=no rd.luks.uuid=4e56edf6-3e8a-4ef3-a084-09b40468bd89 root=UUID=a4465013-c35b-4378-ae5f-c2bb7712b89e rootflags=subvol=/@/0/snapshot zswap.enabled=1 zswap.compressor=zstd zswap.zpool=z3fold rd.md=0 rd.lvm=0 rd.multipath=0 rd.fcoe=0 rd.neednet=1 ip=dhcp resume=UUID=a4465013-c35b-4378-ae5f-c2bb7712b89e resume_offset=1268624
```

**Expected behavior**
When I switch to UEFI (`uefi=yes` and `uefi_stub=...`) it works - the password is asked, no timeout happens

**Additional context**
Add any other context you like about the problem here.

T35R6braPwgDJKq:
does anybody know how to tackle this? Booting takes 5 minutes that way

tacerus:
The `cryptsetup@` systemd unit should be started in the format `cryptsetup@<mapper name>` - or is `luks-xxxxxxx` what you use as a LUKS volume mapper name?

[...]

PR #1975, opened 2022-09-07
fix(dracut.sh): Fix setting built in variables

lnussel:
The previous method would only set built-in values if the directory in question did _not_ exist. However, when $dracutsysrootdir was set, the test would check for existence of $dracutsysrootdir and then leave the actual variables empty.

Replace with just setting a default value when empty.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1375,18 +1375,14 @@ esac
 
 abs_outfile=$(readlink -f "$outfile") && outfile="$abs_outfile"
 
-[[ -d $dracutsysrootdir$systemdutildir ]] \
-    || systemdutildir=$(pkg-config systemd --variable=systemdutildir 2> /dev/null)
-
-if ! [[ -d $dracutsysrootdir$systemdutildir ]]; then
-    [[ -e $dracutsysrootdir/lib/systemd/systemd-udevd ]] && systemdutildir=/lib/systemd
-    [[ -e $dracutsysrootdir/usr/lib/systemd/systemd-udevd ]] && systemdutildir=/usr/lib/systemd
+: "${systemdutildir:=$(pkg-config systemd --variable=systemdutildir 2> /dev/null)}"
+if [ -z "$systemdutildir" ]; then
+    systemdutildir=/lib/systemd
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (commented):
Good catch!

Some comments:
- Although the commisery check hasn't been performed yet, I think it will fail until you replace the uppercase F from "fix" on your commit message.
- You will need to rebase this PR when #1868 is merged, or you can add the changes from commit https://github.com/dracutdevs/dracut/pull/1868/commits/ae4c8744 here.
- I don't know if it's a good solution to avoid all the checks against `$dracutsysrootdir`, maybe an alternative could be something like `[[ -z "$VAR" || ! -d "$dracutsysrootdir$VAR" ]] && VAR=...` (this is just my opinion, let's wait for more feedback). For example:
```
# Current code
[[ -d $dracutsysrootdir$systemdutilconfdir ]] \
    || systemdutilconfdir=$(pkg-config systemd --variable=systemdutilconfdir 2> /dev/null)
[[ -d $dracutsysrootdir$systemdutilconfdir ]] \
    || systemdutilconfdir=/etc/systemd

# PR proposal
: "${systemdutilconfdir:=$(pkg-config systemd --variable=systemdutilconfdir 2> /dev/null)}"
: "${systemdutilconfdir:=/etc/systemd}"

# Alternative
[[ -z "$systemdutilconfdir" || ! -d "$dracutsysrootdir$systemdutilconfdir" ]] \
    && systemdutilconfdir=$(pkg-config systemd --variable=systemdutilconfdir 2> /dev/null)
[[ -z "$systemdutilconfdir" || ! -d "$dracutsysrootdir$systemdutilconfdir" ]] \
    && systemdutilconfdir=/etc/systemd
```

lnussel:
rebased with changes from #1754 

Without knowing the history of the code I'd just remove all runtime path autodetection. That's stuff for build time if at all. Not sure how likely it is that systemd was built to find files somewhere else. Distros that really dare might just as well patch dracut or install a config file that overrides the paths (like `dracut.conf.d/fedora.conf.example`).

johannbg (changes requested):
Let's start by having you fix your commit message so it passes the commisery check as @aafeijoo-suse already has pointed out basically remove the latter "fix" word and your PR should pass the commisery check

PR #1976, opened 2022-09-07
fix(dracut): allow to set persistent policy based on /dev/mapper device names

aafeijoo-suse:
Currently, dracut only allows to configure the persistent policy to address
disks and partitions using any directory name found in `/dev/disk`, but there is
no defined way to set the default policy based on `/dev/mapper` device names. So,
for example, if the persistent policy is set by the distribution in any
configuration file, the user cannot set the default if he wants to.

The current workaround is to set `persistent_policy=""` in a conf file or pass
`--persistent-policy " "` (white space) as an argument. This patch aims to
establish a specific value for this case.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -198,8 +198,9 @@ Creates initial ramdisk images for preloading modules
                          necessary network drivers.
   --persistent-policy [POLICY]
                         Use [POLICY] to address disks and partitions.
-                         POLICY can be any directory name found in /dev/disk.
-                         E.g. "by-uuid", "by-label"
+                         POLICY can be any directory name found in /dev/disk
+                         (e.g. "by-uuid", "by-label"), or "mapper" to use
+                         /dev/mapper device names (default).
   --fstab               Use /etc/fstab to determine the root device.
   --add-fstab [FILE]    Add file to the initramfs fstab.
   --mount "[DEV] [MP] [FSTYPE] [FSOPTS]"
@@ -1293,6 +1294,13 @@ else
     printf "%s\n" "dracut: Are you running from a git checkout?" >&2
     printf "%s\n" "dracut: Try passing -l as an argument to $dracut_cmd" >&2
     exit 1
+fi
+
+if [[ $persistent_policy == "mapper" ]]; then
+    unset persistent_policy
+elif [[ -n $persistent_policy && ! -d "/dev/disk/${persistent_policy}" ]]; then
+    dwarn "Invalid persistent policy, your system does not have a /dev/disk/${persistent_policy} directory."
+    unset persistent_policy
 fi
 
 if [[ $enhanced_cpio == "yes" ]]; then
--- a/man/dracut.8.asc
+++ b/man/dracut.8.asc
@@ -391,8 +391,8 @@ Default:
 
 **--persistent-policy** _<policy>_::
     Use _<policy>_ to address disks and partitions.
-    _<policy>_ can be any directory name found in /dev/disk.
-    E.g. "by-uuid", "by-label"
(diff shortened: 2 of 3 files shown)


Conan-Kudo (approved):

dracut.sh
```diff
@@ -1295,6 +1296,13 @@ else
     exit 1
 fi
 
+if [[ $persistent_policy == "mapper" ]]; then
+    unset persistent_policy
```
LaszloGombos, line 1300: Can you help to understand this section ? Why do you unset persistent_policy ?

man/dracut.8.asc
```diff
@@ -391,8 +391,8 @@ Default:
 
 **--persistent-policy** _<policy>_::
     Use _<policy>_ to address disks and partitions.
-    _<policy>_ can be any directory name found in /dev/disk.
-    E.g. "by-uuid", "by-label"
+    _<policy>_ can be any directory name found in /dev/disk (e.g. "by-uuid",
+    "by-label"), or "mapper" to use /dev/mapper device names (default).
```
LaszloGombos, line 395: This man page is for all dracut deployment. Is "mapper" the default for all dracut installations ?

man/dracut.conf.5.asc
```diff
@@ -142,8 +142,8 @@ Configuration files must have the extension .conf; other extensions are ignored.
 
 *persistent_policy=*"__<policy>__"::
     Use _<policy>_ to address disks and partitions.
-    _<policy>_ can be any directory name found in /dev/disk.
-    E.g. "by-uuid", "by-label"
+    _<policy>_ can be any directory name found in /dev/disk (e.g. "by-uuid",
+    "by-label"), or "mapper" to use /dev/mapper device names (default=mapper).
```
LaszloGombos, line 146: This man page is for all dracut deployment. Is "mapper" the default for all dracut installations ?

LaszloGombos (commented):
Would it be possible to add "--persistent-policy mapper" to one of the test cases to have this tested ?

dracut.sh
aafeijoo-suse: See where the `persistent_policy` variable is used:

https://github.com/dracutdevs/dracut/blob/cf31fcf804be4dc0fa31885f5185a59b6012cdf4/dracut-functions.sh#L270-L290

By default (unset), the first check is `/dev/mapper/*`. With this patch I just want to give a name ("mapper") to the default option, because if any distro/package sets the `persistent_policy` in a conf file, and the user wants to change it to `/dev/mapper/*`, there is not a common way to do it (read PR description).

LaszloGombos: Thanks !

dracut.sh
```diff
@@ -1295,6 +1296,13 @@ else
...
+elif [[ -n $persistent_policy && ! -d "/dev/disk/${persistent_policy}" ]]; then
```
LaszloGombos, line 1301: Regarding the use of "-d" in the test - are these always directories or can they be e.g. symlinks to a directory ?

LaszloGombos (approved):
LGTM

dracut.sh
LaszloGombos, line 1300: Perhaps you can also consider not to unset the variable. I would find that easier to read.

aafeijoo-suse: IMHO, this is the least intrusive change, since the effect of the new "mapper" option is to not set any value for the `persistent_policy` variable.

dracut.sh
aafeijoo-suse: AFAIK directories.

johannbg (approved):

johannbg:
Everyone remember to resolve conversation when they have been resolved...

PR #1978, opened 2022-09-10
fix(load_fstype): use $1 if $2 is missing

FGrose:
Use parameter expansion default value if the second argument is missing. The current code mistakenly reuses the null second argument.

## Checklist

### DIFF
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -1157,8 +1157,5 @@ remove_hostonly_files() {
 # returns OK if kernel_module is loaded
 # modprobe fails if /lib/modules is not available (--no-kernel use case)
 load_fstype() {
-    if [ -z "$2" ]; then
-        set -- "$1" "$2"
-    fi
-    strstr "$(cat /proc/filesystems)" "$2" || modprobe "$1"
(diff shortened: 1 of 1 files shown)


modules.d/99base/dracut-lib.sh
```diff
@@ -1157,8 +1157,5 @@ remove_hostonly_files() {
 # returns OK if kernel_module is loaded
 # modprobe fails if /lib/modules is not available (--no-kernel use case)
 load_fstype() {
-    if [ -z "$2" ]; then
-        set -- "$1" "$2"
```
aafeijoo-suse, line 1161: I think what @LaszloGombos wanted to do here was `set -- "$1" "$1"`

FGrose: Yes, I'm proposing `"${2:-"$1"}"` as more concise.

LaszloGombos:
LGTM. Thanks @FGrose  and @aafeijoo-suse

LaszloGombos:
Not sure why commit message checker test is failing.

Changing the PR commit message on the GH interface does not actually change the commit message on the commit itself. You might need to amend your commit for that.

LaszloGombos:
LGTM

johannbg (approved):
LGTM

PR #1983, opened 2022-09-19
fix(dracut-initramfs-restore.sh): fix initramfs path detection

ryncsn:
The path detection is not working on latest Fedora and some other distros, and fails to extract the initramfs. It seems the if statement is broken by a previous commit, so let's fix it.

Fixes: 3d8e1ad ('fix(dracut-initramfs-restore.sh): add missing default paths')
Signed-off-by: Kairui Song <kasong@tencent.com>

## Checklist

Fixes commit 3d8e1ad

### DIFF
--- a/dracut-initramfs-restore.sh
+++ b/dracut-initramfs-restore.sh
@@ -27,14 +27,14 @@ fi
 
 mount -o ro /boot &> /dev/null || true
 
-if [[ -d /efi/loader/entries ]] || [[ -L /efi/loader/entries ]] \
-    || [[ -d /efi/$MACHINE_ID ]] || [[ -L /efi/$MACHINE_ID ]]; then
+if [[ -d /efi/loader/entries || -L /efi/loader/entries ]] \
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (changes requested):
Hmm.. I think this condition is wrong, with `[[ A ]] || [[ B ]] && [[ C ]] || [[ D ]]` if only D is true the condition is met, whereas we should use here `[[ A || B ]] && [[ C || D ]]`. See a similar check (there is no need to check if MACHINE_ID is empty here):

https://github.com/dracutdevs/dracut/blob/4012314f204806a67d038f378f525646814d12ea/dracut.sh#L1128-L1139

ryncsn:
Ah, yes, you are right, the conditions are still wrong, I've fixed it again, now it should be OK.

aafeijoo-suse (approved):
The failure of TEST-10 should be unrelated.

johannbg (approved):
LGTM

PR #1984, opened 2022-09-21
fix(systemd): add missing modprobe@.service

aafeijoo-suse:
`sys-kernel-config.mount` needs `modprobe@configfs.service` since systemd v246.7 (https://github.com/systemd/systemd/commit/42cc2855), so the kernel configfs fails to mount in the initrd.

## Checklist


marcosfrm:
LGTM

johannbg (approved):
LGTM

PR #1991, opened 2022-09-27
Add new dmsquash-live-autooverlay module

iammattcoleman:
This pull request introduces a new module that creates a partition for overlayfs usage in the free space on the root filesystem's parent block device.

Ubuntu 20.04 (the `main` branch) and Fedora 36 (the `fedora-36` branch) Kiwi image descriptions that produce ISO images that demonstrate this feature are available in this repository:
https://github.com/iammattcoleman/overlay-test

## Checklist

### DIFF
--- a/modules.d/90dmsquash-live-autooverlay/create-overlay.sh
+++ b/modules.d/90dmsquash-live-autooverlay/create-overlay.sh
@@ -1,0 +1,119 @@
+#!/bin/sh
+
+type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+if getargbool 0 rd.live.debug -n -y rdlivedebug; then
+    exec > /tmp/create-overlay.$$.out
+    exec 2>> /tmp/create-overlay.$$.out
+    set -x
+fi
+
+gatherData() {
+    overlay=$(getarg rd.live.overlay)
+    if [ -z "$overlay" ]; then
+        info "Skipping overlay creation: kernel command line parameter 'rd.live.overlay' is not set"
+        exit 0
+    fi
+    # shellcheck disable=SC2086
+    if ! str_starts ${overlay} LABEL=; then
+        die "Overlay creation failed: the partition must be set by LABEL in the 'rd.live.overlay' kernel parameter"
+    fi
+
+    overlayLabel=${overlay#LABEL=}
+    # shellcheck disable=SC2086
+    if [ -b /dev/disk/by-label/${overlayLabel} ]; then
+        info "Skipping overlay creation: overlay already exists"
+        exit 0
+    fi
+
+    filesystem=$(getarg rd.live.overlay.cowfs)
+    [ -z "$filesystem" ] && filesystem="ext4"
+    if [ "$filesystem" != "ext4" ] && [ "$filesystem" != "xfs" ] && [ "$filesystem" != "btrfs" ]; then
+        die "Overlay creation failed: only ext4, xfs, and btrfs are supported in the 'rd.live.overlay.cowfs' kernel parameter"
+    fi
+
+    live_dir=$(getarg rd.live.dir)
+    [ -z "$live_dir" ] && live_dir="LiveOS"
+
+    [ -z "$1" ] && exit 1
+    rootDevice=$1
+
+    # The kernel command line's 'root=' parameter was parsed into the $root variable by the dmsquash-live module.
+    # $root contains the path to a symlink within /dev/disk/by-label, which points to a partition.
+    # This script needs that partition's parent block device.
+    # shellcheck disable=SC2046
+    # shellcheck disable=SC2086
+    rootDeviceAbsolutePath=$(readlink -f ${rootDevice})
+    rootDeviceSysfsPath=/sys/class/block/${rootDeviceAbsolutePath##*/}
+    if [ -f "${rootDeviceSysfsPath}/partition" ]; then
+        # shellcheck disable=SC2086
+        partition=$(cat ${rootDeviceSysfsPath}/partition)
+    else
+        partition=0
+    fi
+    # shellcheck disable=SC2086
+    readonly=$(cat ${rootDeviceSysfsPath}/ro)
+    # shellcheck disable=SC2086
+    if [ "$partition" != "1" ] || [ "$readonly" != "0" ]; then
+        info "Skipping overlay creation: unpartitioned or read-only media detected"
+        exit 0
+    fi
+    # shellcheck disable=SC2046
+    # shellcheck disable=SC2086
+    fullDriveSysfsPath=$(readlink -f ${rootDeviceSysfsPath}/..)
+    blockDevice=/dev/${fullDriveSysfsPath##*/}
+    currentPartitionCount=$(grep --count -E "${blockDevice#/dev/}[0-9]+" /proc/partitions)
+
+    # shellcheck disable=SC2086
+    freeSpaceStart=$(parted --script ${blockDevice} unit % print free \
+        | awk -v x=${currentPartitionCount} '$1 == x {getline; print $1}')
+    if [ -z "$freeSpaceStart" ]; then
+        info "Skipping overlay creation: there is no free space after the last partition"
+        exit 0
+    fi
+    partitionStart=$((${freeSpaceStart%.*} + 1))
+    if [ $partitionStart -eq 100 ]; then
+        info "Skipping overlay creation: there is not enough free space after the last partition"
+        exit 0
+    fi
+
+    overlayPartition=${blockDevice}$((currentPartitionCount + 1))
+
+    label=$(blkid --match-tag LABEL --output value "$rootDevice")
+    uuid=$(blkid --match-tag UUID --output value "$rootDevice")
+    if [ -z "$label" ] || [ -z "$uuid" ]; then
+        die "Overlay creation failed: failed to look up root device label and UUID"
+    fi
+}
+
+createPartition() {
+    # shellcheck disable=SC2086
+    parted --script --align optimal ${blockDevice} mkpart primary ${partitionStart}% 100%
+}
+
+createFilesystem() {
+    # shellcheck disable=SC2086
+    mkfs.${filesystem} -L ${overlayLabel} ${overlayPartition}
+
+    baseDir=/run/initramfs/create-overlayfs
+    mkdir -p ${baseDir}
+    # shellcheck disable=SC2086
+    mount -t auto ${overlayPartition} ${baseDir}
+
+    mkdir -p ${baseDir}/${live_dir}/ovlwork
+    # shellcheck disable=SC2086
+    mkdir ${baseDir}/${live_dir}/overlay-${label}-${uuid}
+
+    umount ${baseDir}
+    rm -r ${baseDir}
+}
+
+main() {
+    gatherData "$1"
+    createPartition
+    udevsettle
+    createFilesystem
+    udevsettle
+}
+
+main "$1"
--- a/test/TEST-16-DMSQUASH/test-init.sh
+++ b/test/TEST-16-DMSQUASH/test-init.sh
@@ -9,6 +9,12 @@ exec > /dev/console 2>&1
 
 echo "dracut-root-block-success" | dd oflag=direct,dsync of=/dev/disk/by-id/ata-disk_marker
 
+if grep -qF ' rd.live.overlay=LABEL=persist ' /proc/cmdline; then
+    # Writing to a file in the root filesystem lets test_run() verify that the autooverlay module successfully created
+    # and formatted the overlay partition and that the dmsquash-live module used it when setting up the rootfs overlay.
+    echo "dracut-autooverlay-success" > /overlay-marker
+fi
+
 export TERM=linux
 export PS1='initramfs-test:\w\$ '
 [ -f /etc/mtab ] || ln -sfn /proc/mounts /etc/mtab
--- a/test/TEST-16-DMSQUASH/test.sh
+++ b/test/TEST-16-DMSQUASH/test.sh
@@ -5,8 +5,9 @@ TEST_DESCRIPTION="live root on a squash filesystem"
 
 KVERSION="${KVERSION-$(uname -r)}"
 
-# Uncomment this to debug failures
-#DEBUGFAIL="rd.shell rd.debug loglevel=7"
+# Uncomment these to debug failures
+#DEBUGFAIL="rd.shell rd.debug rd.live.debug loglevel=7"
+#DEBUGTOOLS="setsid ls cat sfdisk"
 
 test_run() {
     dd if=/dev/zero of="$TESTDIR"/marker.img bs=1MiB count=1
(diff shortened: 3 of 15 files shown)


Conan-Kudo:
This looks great! I've approved the test run as you're a first-time contributor.

iammattcoleman:
Shell linting failed due to a couple instances of missing double quotes.

Conan-Kudo:
I will run this once the Fedora datacenter maintenance window is over.

Conan-Kudo (approved):
Unfortunately, we're stuck in a catch-22, and while I can verify the failing test passes locally when I build the container and use it locally, it won't fix itself in here until _after_ we merge it.

Otherwise, this looks good to merge. Thanks!

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -0,0 +1,119 @@
...
+    fi
+
+    filesystem=$(getarg rd.live.overlay.cowfs)
+    [ -z "$filesystem" ] && filesystem="ext4"
+    if [ "$filesystem" != "ext4" ] && [ "$filesystem" != "xfs" ] && [ "$filesystem" != "btrfs" ]; then
+        die "Overlay creation failed: only ext4, xfs, and btrfs are supported in the 'rd.live.overlay.cowfs' kernel parameter"
+    fi
+
+    live_dir=$(getarg rd.live.dir)
+    [ -z "$live_dir" ] && live_dir="LiveOS"
```
LaszloGombos, line 36: This line seems repeat from https://github.com/dracutdevs/dracut/blob/master/modules.d/90dmsquash-live/dmsquash-live-root.sh#L21 

Can we avoid repeating and risking making the two modules out of sync for the value of the default value with later PRs ?

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -0,0 +1,119 @@
...
+    fi
+
+    overlayLabel=${overlay#LABEL=}
+    # shellcheck disable=SC2086
+    if [ -b /dev/disk/by-label/${overlayLabel} ]; then
+        info "Skipping overlay creation: overlay already exists"
+        exit 0
+    fi
+
+    filesystem=$(getarg rd.live.overlay.cowfs)
```
LaszloGombos, line 29: rd.live.overlay.cowfs seems a new argument. It seems we should document it by updating https://github.com/dracutdevs/dracut/blob/master/man/dracut.cmdline.7.asc

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -0,0 +1,119 @@
...
+    fi
+    partitionStart=$((${freeSpaceStart%.*} + 1))
+    if [ $partitionStart -eq 100 ]; then
+        info "Skipping overlay creation: there is not enough free space after the last partition"
+        exit 0
+    fi
+
+    overlayPartition=${blockDevice}$((currentPartitionCount + 1))
+
+    label=$(blkid --match-tag LABEL --output value "$rootDevice")
```
LaszloGombos, line 82: Would it make sense here to reuse devnames from 99base/dracut-lib.sh here ? 
crypt and fs-lib modules need similar functionality and we call into devnames in those modules.

LaszloGombos (commented):
Some review comments to better align with existing modules..

test/TEST-16-DMSQUASH/test.sh
```diff
@@ -19,10 +20,16 @@ test_run() {
     "$testdir"/run-qemu \
         "${disk_args[@]}" \
         -boot order=d \
-        -append "rd.live.image rd.live.overlay.overlayfs=1 rd.live.dir=testdir root=LABEL=dracut console=ttyS0,115200n81 quiet selinux=0 rd.info rd.shell=0 panic=1 oops=panic softlockup_panic=1 $DEBUGFAIL" \
```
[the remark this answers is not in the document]
iammattcoleman: Thanks for pointing that out. I originally had created a separate `TEST-18-DMSQUASH-AUTOOVERLAY` by duplicating the entirety of this test and adding my (minor) changes. I didn't feel good about all that copypasta, so I switched to this approach.

I hadn't checked out the NFS test and that's definitely a better way to go.

I'll adjust this test to use a similar approach.

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
iammattcoleman: This script must run before `dmsquash-live-root.sh` in order to work properly and this script won't always run since it's an optional module that complements `90dmsquash-live`. So, both scripts will need a similar line, but it would be nice to have the constant ("LiveOS") in a single location.

Can you point me to some code that demonstrates sharing a constant between two related modules so I can follow the project's existing conventions?

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
iammattcoleman: Yes, it is new. I reused the parameter from Kiwi's live module (which we're hoping to deprecate & remove from Kiwi once this is merged):
https://osinside.github.io/kiwi/concept_and_workflow/customize_the_boot_process.html#boot-image-parameters

Thanks for pointing this out. I'll update the man page accordingly.

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
iammattcoleman: The approach I took was to gather all the data necessary to perform the task before making any destructive changes to the disk. That way it can avoid doing anything if it doesn't have all the information it needs.

`devnames` provides device names within `/dev`. This is gathering the root device's label and UUID for use in the `createFilesystem()` function on line 105, which creates a directory with a name that matches the default path format defined in `dmsquash-live-root.sh`. So, that's potentially another constant that should be shared between the two modules from a single place.

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
LaszloGombos: To share a constant, we would need to introduce a shared file between dmsquash-live and dmsquash-live-autooverlay. I think it make sense to limit making changes in other modules as part of this module introduction PR, so perhaps we can do this refactoring as a follow-up PR.

test/TEST-16-DMSQUASH/test.sh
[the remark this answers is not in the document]
LaszloGombos: Thank you for considering it. 
https://github.com/dracutdevs/dracut/commit/027eee902d0d7af4db45fcb7835d49990c9fafa1 just landed, which hopefully helps you here as well.

LaszloGombos:
> Unfortunately, we're stuck in a catch-22, and while I can verify the failing test passes locally

I find it a better option to just run the job under your personal github fork e.g. https://github.com/iammattcoleman/dracut/actions/workflows/manualtest.yml and post the log of at least one successful run. You might need to trigger container creation first manually under your personal github before triggering the manual test run.

test/TEST-16-DMSQUASH/test.sh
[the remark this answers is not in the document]
iammattcoleman: I've updated the test to add a separate QEMU run that uses its own initrd with my new module and modified the `init` script to only write the autooverlay marker if it's booted with the related kernel parameter. That way, the original test is unaltered.

I followed the example of the commit you linked to that updated this test script instead of using separate functions like `TEST-20-NFS`.

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
iammattcoleman: I added this to the man page. Thanks again!

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -0,0 +1,119 @@
...
+
+gatherData() {
+    overlay=$(getarg rd.live.overlay)
+    if [ -z "$overlay" ]; then
+        info "Skipping overlay creation: kernel command line parameter 'rd.live.overlay' is not set"
+        exit 0
+    fi
+    # shellcheck disable=SC2086
+    if ! str_starts ${overlay} LABEL=; then
+        die "Overlay creation failed: the partition must be set by LABEL in the 'rd.live.overlay' kernel parameter"
```
LaszloGombos, line 19: Sorry for not catching this earlier, but have you had a chance to test this with 
rd.live.overlay=none
rd.live.overlay=/dev/sdb1:persistent-overlay.img
rd.live.overlay=UUID=99440c1f-8daa-41bf-b965-b7240a8996f4

As an example according to the man page
If rd.live.overlay=none is provided as a kernel command
line option, a writable, linear Device-mapper target is
created on boot with no overlay.

It seems just packaging this new module would auto-install itself and this line would be executed (and fail with an error massage) at each boot (except when rd.live.overlay=LABEL" is used. 

I am not suggesting that this additional code would break the functionality for "rd.live.overlay=none", but perhaps there is a more inclusive way to test if overlay creation is expected. Also perhaps this should be an info and not a die level message ?

LaszloGombos (commented):
Changes all look good to me, but perhaps the module is not yet complete, see additional comments.

Also if you have a chance, can you please also update the Gentoo container and add parted ?

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
iammattcoleman: > Sorry for not catching this earlier, but have you had a chance to test this with
> rd.live.overlay=none
> rd.live.overlay=/dev/sdb1
> rd.live.overlay=UUID=99440c1f-8daa-41bf-b965-b7240a8996f4

It doesn't make sense to use this module with those settings:
* `none` would disable the use of overlay, making it pointless to include this module
* specifying an existing partition doesn't make sense because this module is meant to automatically create the partition in the free space on the boot medium
* specifying the UUID doesn't make sense because it will be created when this module formats the new partition's filesystem

> It seems just packaging this new module would auto-install itself

I was under the impression that it would only be included in the initramfs if the module was specifically listed in the configuration files or on the `dracut` command line. Did I misinterpret something?

The intent is for this module to only be included for certain builds (my use case is live ISO images booted on USB sticks) by placing a drop-in configuration file in `/etc/dracut.conf.d/` like so:
https://github.com/iammattcoleman/overlay-test/blob/fedora-36/image-description/root/etc/dracut.conf.d/dmsquash-live-autooverlay.conf

> Also perhaps this should be an info and not a die level message ?

It's a `die` level message because I figured that including this module and not properly specifying the `rd.live.overlay` value would only happen by mistake. I can't think of a scenario where someone would include this module in an initramfs but not want it to work.

LaszloGombos: > I was under the impression that it would only be included in the initramfs if the module was specifically listed in the configuration files or on the `dracut` command line.

Sorry, I missed that you're returning 255 in https://github.com/dracutdevs/dracut/pull/1991/files#diff-19cd0a77b761132c9ca0494191b45ce737857a6fc21dbed98650b528ada2b60em which will do exactly what you stated.

LaszloGombos:
LGTM, thanks for the timely interactions.

LaszloGombos (approved):

Conan-Kudo (approved):

iammattcoleman:
As I mentioned in the Matrix channel, please hold off on merging this: as @LaszloGombos's suggested, I ran the manual test in my fork and it failed:
https://github.com/iammattcoleman/dracut/actions/runs/3193095857/jobs/5211306428

While working on this, I was testing on bare metal (well, really a VM) and successfully running the test from within the `TEST-16-DMSQUASH` directory. I had not tried running the test from within the `tests` directory like the manual test GitHub Action.

After I saw the test fail in my fork, I reproduced it in a `podman` container locally. If I run `make V=1 TESTS=16 clean check` within `/dracut/tests/`, it fails. If I run `make clean setup run` within `/dracut/tests/TEST-16-DMSQUASH/`, it succeeds.

test/TEST-16-DMSQUASH/test-init.sh
```diff
@@ -9,6 +9,10 @@ exec > /dev/console 2>&1
 
 echo "dracut-root-block-success" | dd oflag=direct,dsync of=/dev/disk/by-id/ata-disk_marker
 
+if grep -qF ' rd.live.overlay=LABEL=persist ' /proc/cmdline; then
+    echo "dracut-autooverlay-success" > /overlay-marker
```
LaszloGombos, line 13: Is there a benefit of using / partition as marker and not /dev/disk/by-id/ata-disk_marker partition like other tests ?

iammattcoleman: This ensures that the overlay is actually working. The test now creates the root partition as only half the size of the image file, leaving the second half blank. When it boots, the autooverlay module creates a partition in the free space and uses it for the overlay. Writing to a file in the root filesystem lets the test verify that the autooverlay module successfully created and formatted the partition and that the dmsquash-live module used it when setting up the rootfs overlay. The test script verifies that the string `dracut-autooverlay-success` occurs in the second half of the image file, I.E. within the second partition.

LaszloGombos: Thanks, would love to see this explanation in the code as a comment :-) 
Also I was hoping that reviewing how the test is expected to work helps you figure out why the test is failing on the CI ...

iammattcoleman: I finally had a chance to add comments explaining the reasoning and look into the test failure today.

Here is TEST-16-DMSQUASH, now passing in CI in my fork using all container variants except Gentoo:
* Arch: https://github.com/iammattcoleman/dracut/actions/runs/3253893225/jobs/5341593145
* Debian: https://github.com/iammattcoleman/dracut/actions/runs/3253902964/jobs/5341613095
* Fedora: https://github.com/iammattcoleman/dracut/actions/runs/3253910178/jobs/5341627573
* OpenSuse: https://github.com/iammattcoleman/dracut/actions/runs/3253919573/jobs/5341644938

Gentoo is failing with a node-related error during the "Checkout Repository" stage that I don't understand:
https://github.com/iammattcoleman/dracut/actions/runs/3253852146/jobs/5341510695

I found this ticket that describes a similar failure, but it's specifically about a self-hosted runner and unfortunately doesn't offer any details about how they fixed it:
https://github.com/actions/checkout/issues/749

iammattcoleman: After investigating a bit more, I see that you've (@LaszloGombos) come across this issue with the Gentoo container, as well: https://github.com/dracutdevs/dracut/pull/1812#issuecomment-1263954331

It's caused by an architecture mismatch as described in this comment: https://github.com/actions/checkout/issues/334#issuecomment-1241306390

Here's an issue in actions/runner that describes how the runner uses hardcoded paths for node:
https://github.com/actions/runner/issues/2115

As suggested in the I reverted to actions/checkout@v1 and it revealed that the Gentoo container is missing dependencies still:
https://github.com/iammattcoleman/dracut/actions/runs/3254015105/jobs/5341833491

I'm fixing that now.

iammattcoleman: I fixed the issues with Gentoo.

Here is TEST-16-DMSQUASH, now passing in CI in my fork using all container variants:
* Arch: https://github.com/iammattcoleman/dracut/actions/runs/3254211434/jobs/5342208863
* Debian: https://github.com/iammattcoleman/dracut/actions/runs/3254212012/jobs/5342210044
* Fedora: https://github.com/iammattcoleman/dracut/actions/runs/3254212447/jobs/5342211082
* Gentoo: https://github.com/iammattcoleman/dracut/actions/runs/3254221018/jobs/5342224579
* OpenSuse: https://github.com/iammattcoleman/dracut/actions/runs/3254221363/jobs/5342225163

Conan-Kudo (approved):

mrc0mmand (approved):

modules.d/90dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -0,0 +1,119 @@
...
+    set -x
+fi
+
+gatherData() {
+    overlay=$(getarg rd.live.overlay)
+    if [ -z "$overlay" ]; then
+        info "Skipping overlay creation: kernel command line parameter 'rd.live.overlay' is not set"
+        exit 0
+    fi
+    # shellcheck disable=SC2086
```
mrc0mmand, line 17: Just wondering - is there any particular reason why is the shellcheck's warning suppressed instead of double-quoting the string (here and in other places as well)?

iammattcoleman: No shell dark magic going on, here. I did it for a couple reasons...
1. Based on the possible contents of this variable, there's no way it could include a space when used with this module. Feel free to add quotes. This is really just symptom of my (probably strange) personal preferences.
2. Consitency with similar lines in the rest of the file. There were other lines in the file that could not be resolved by adding quotes due to nesting and required multiple shellcheck overrides. In those cases, I opted for the more readable single line of code rather than introducing other variables.

PR #1992, opened 2022-09-28
feat(systemd-portabled): introducing the systemd-portabled module

aafeijoo-suse:
Introducing the `systemd-portabled` module.

It works with raw binary disk images contained in regular files, but it's not intended to work with directory trees, like the `systemd-sysext` module.

## Checklist
- [X] I have tested it locally with minimal squashfs images (see https://systemd.io/PORTABLE_SERVICES/#requirements-on-images) and images created with mkosi (e.g. https://github.com/systemd/portable-walkthrough).

### DIFF
--- a/modules.d/01systemd-portabled/module-setup.sh
+++ b/modules.d/01systemd-portabled/module-setup.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries \
+        portablectl \
+        "$systemdutildir"/systemd-portabled \
+        || return 1
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+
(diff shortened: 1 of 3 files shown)


LaszloGombos:
> It works with raw binary disk images contained in regular files, but it's not intended to work with directory trees, like the systemd-sysext module.

This seems quite important information. Can we capture this somewhere inside the source tree and not just as metadata in the PR ? Perhaps as a comment in modules.d/01systemd-portabled/module-setup.sh or maybe even better somewhere in the man pages ?

Why is it not intended to work with directory trees ?

What is the error message if somebody has a tree instead of raw binary disk image ?

aafeijoo-suse:
> > What is the error message if somebody has a tree instead of raw binary disk image ?
> 
> Directory trees won't be included in the initrd, just image files.

Added a warning message.

LaszloGombos:
LGTM

johannbg (approved):
LGTM

johannbg:
@LaszloGombos ping

LaszloGombos (approved):
LGTM

dtardon:
What's the use case for using portabled in initrd?

PR #1998, opened 2022-10-02
test: shutdown is not used for running tests

LaszloGombos:
This pull request changes...

## Checklist


LaszloGombos:
All test use poweroff instead of shutdown.                                                                                                                                                                                                                      
No need to install shutdown.

LaszloGombos:
Moved this work here - https://github.com/dracutdevs/dracut/pull/2011 .

PR #2001, opened 2022-10-11
90crypt: Look for dm-crypt devices in by-uuid

palmer-dabbelt:
## Changes

I know very little about Dracut, so I'm not even sure this is remotely sane.  I was just poking around trying to get my setup, which is dm-crypt/LUKS + BTRFS (in a subvolume) working.  I ended up with a grub.cfg that looks sane to me

````
linux   /vmlinuz-5.15.32-gentoo-dist root=/dev/mapper/luks-9924a2f1 ro rootflags=subvol=roots/gentoo crypt_root=UUID=cdeb23b3-62f8-4497-85f5-7a965f41b9e3 rd.luks.uuid=9924a2f1 rd.debug
````

but despite a bunch of attempts at messing around with it I couldn't get anything that would both run cryptsetup and find the resulting LUKS device.  I tried a handful of things like `partuuid` that were indicated to work on the internet, but didn't have any luck.

Happy to provide more information here, I just figured I'd send along the patch as I'm not even sure this is a bug or just a PEBKAC.

## Checklist

Fixes #

### DIFF
--- a/modules.d/90crypt/parse-crypt.sh
+++ b/modules.d/90crypt/parse-crypt.sh
@@ -166,7 +166,7 @@ else
             if [ $is_keysource -eq 0 ]; then
                 uuid=$luksid
                 while [ "$uuid" != "${uuid#*-}" ]; do uuid=${uuid%%-*}${uuid#*-}; done
-                printf -- '[ -e /dev/disk/by-id/dm-uuid-CRYPT-LUKS?-*%s*-* ] || exit 1\n' "$uuid" \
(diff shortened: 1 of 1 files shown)


LaszloGombos:
@palmer-dabbelt 

This change breaks test-13 "TEST: root filesystem on LVM on encrypted partitions of a RAID-5 ".. It is one of the test that is using crypt module. This failure needs to be investigated before proceeding with this work..

Also to pass the commit message checker, please change the title on the commit.

PR #2014, opened 2022-10-23
feat(systemd-pcrphase): introducing the systemd-pcrphase module

johannbg:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-pcrphase/module-setup.sh
+++ b/modules.d/01systemd-pcrphase/module-setup.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries "$systemdutildir"/systemd-pcrphase || return 1
+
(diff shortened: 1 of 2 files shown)


johannbg:
@aafeijoo-suse, @LaszloGombos ping for review.

aafeijoo-suse (commented):
Don't we need to install `systemd-pcrphase-sysinit.service` as well? https://github.com/systemd/systemd/commit/047273e6

johannbg:
> Don't we need to install `systemd-pcrphase-sysinit.service` as well?

Technically I dont think we need it ( if I read the docs right )  but yeah we should include it for completeness. I'll update the pr

aafeijoo-suse:
> > Don't we need to install `systemd-pcrphase-sysinit.service` as well?
> 
> Technically I dont think we need it ( if I read the docs right ) but yeah we should include it for completeness. I'll update the pr

Right, it's a bit contradictory, while the commit message says `The "sysinit" one is placed between sysinit.target and basic.target.` (so, before dracut-pre-mount.service), the man page says that the initrd phase is covered by the systemd-pcrphase-initrd.service (enter-initrd/leave-initrd strings)...

aafeijoo-suse (approved):
systemd considers sysinit as "later boot", so this is fine.

LaszloGombos (approved):
lgtm

PR #2016, opened 2022-10-23
feat(github): automating dependency updates

johannbg:
This pull request changes...

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

Fixes #

PR #2017, opened 2022-10-24
fix(github): dependabot remove ignore

johannbg:
That's rather unhelpful error msg dependabot...
Removed the ignore line since I suspect it's the name matching that is failing and dependabot files PR for changes so we can just nack those that we dont want to update.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,11 +4,11 @@ version: 2
   updates:
     - package-ecosystem: "github-actions"
       directory: "/"
-      schedule: 
-       interval: weekly
-       day: "tuesday"
+      schedule:
+        interval: weekly
+        day: "tuesday"
       commit-message:
         prefix: "chore(github):"
         include: "scope"
-      ignore:
-        - dependency-name: "Manual test"
+          #ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,11 +4,11 @@ version: 2
   updates:
```
LaszloGombos, line 4: Why are "updates" intended with 2 white spaces "  " ? Is that significant.

The example github generates does not have those indentation.

LaszloGombos (commented):
I suspect dependabot will also run in our personal forks. I hope we find a way to disable or at least make it not very annoying. Not meaning to stop this work, just a heads-up on what to expect to come.

.github/dependabot.yml
aafeijoo-suse: Right, we missed that in #2016. Check https://jsonformatter.org/yaml-validator

LaszloGombos:
Note to myself: I am also expecting an update from setup-buildx-action@v1 to setup-buildx-action@v2

johannbg:
Closing this one since I need more time to look at this.

PR #2018, opened 2022-10-26
ci: use prebuilt Linux kernel binaries for Gentoo container

LaszloGombos:
No need to rebuild the Gentoo kernel every day :-) . 

I wish I would have known about about this possibility earlier.

### DIFF
--- a/test/container/Dockerfile-Gentoo
+++ b/test/container/Dockerfile-Gentoo
@@ -3,7 +3,7 @@ FROM docker.io/gentoo/portage:latest as portage
 # kernel and its dependencies in a separate builder
 FROM docker.io/gentoo/stage3:musl as kernel
 COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
-RUN emerge -qv virtual/dist-kernel
+RUN emerge -qv sys-kernel/gentoo-kernel-bin
 
 FROM docker.io/gentoo/stage3:musl
 COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo


LaszloGombos:
@iammattcoleman - It seems lvm2 does not build for the Gentoo container - https://pipelines.actions.githubusercontent.com/serviceHosts/8c277cf6-4cc9-4b45-afc8-5c53d29d3d04/_apis/pipelines/1/runs/29410/signedlogcontent/6?urlExpires=2022-10-26T03%3A36%3A03.7355159Z&urlSigningMethod=HMACV1&urlSignature=hbLhNhGJptGMJhq84mzTyiXyZYOpBZxcCz4NRmHlO3I%3D 

I'm guessing we do not actually need this to run TEST-16. Can you please weight in on this when you have a chance ?

iammattcoleman:
> I'm guessing we do not actually need this to run TEST-16.

We do need that for the test...
* `90dmsquash-live` needs the `dmsetup` binary: https://github.com/dracutdevs/dracut/blob/master/modules.d/90dmsquash-live/module-setup.sh#L25
* the `lvm2` package provides the `dmsetup` binary on Gentoo - I had to tweak the portage configuration for the package to make it _only_ build `dmsetup` since the rest of the `lvm2` package doesn't build in the container due to the lack of kernel headers

LaszloGombos:
@iammattcoleman . Thanks for getting back so quickly.

> I had to tweak the portage configuration for the package to make it only build dmsetup

I've seen all of that and it seemed to work for me as well locally, but somehow fails on GitHub despite the portage configuration file. See error message ^^^

If we can not find a solution, I propose to link dmsetup to /bin/true to satisfy the precondition for the dmsquash-live dracut module (would not be the first time) - this is likely OK as we're not actually calling dmsetup as part of the test  ?

iammattcoleman:
> > I had to tweak the portage configuration for the package to make it only build dmsetup
> 
> I've seen all of that and it seemed to work for me as well locally, but somehow fails on GitHub despite the portage configuration file. See error message ^^^

@LaszloGombos Does a local Docker build produce the same error if you remove the `package.use` override? I don't have access to my work computer at the moment to dig through its terminal history from when I was working on getting it to build.

> If we can not find a solution, I propose to link dmsetup to /bin/true to satisfy the precondition for the dmsquash-live dracut module (would not be the first time) - this is likely OK as we're not actually calling dmsetup as part of the test ?

The `90dmsquash-live` module calls `dmsetup` in numerous locations. If the test doesn't currently hit any of those locations and do something to verify that `dmsetup` actually completed its operation successfully, then I think the test should be improved rather than hotwiring it.

> We can also just try to install the kernel headers..

It's worth a shot. Anything built with those headers would be questionable unless the kernel headers in the container match the host's kernel. If it gets `dmsetup` to successfully build, though, then header/kernel version mismatch wouldn't really matter since `dmsetup` is the only part of `lvm2` that we actually need to work.

LaszloGombos:
Added back lvm2 package as the Dockerfile itself seems to be valid.

We had some successful container builds even on GitHub action earlier - e.g. https://github.com/dracutdevs/dracut/actions/runs/3312404236/jobs/5469037068 

Regardless this PR is about switching to a different kernel package for Gentoo.

LaszloGombos:
This reduces the build time by about 2-3 hours

johannbg:
@LaszloGombos this ready?

johannbg (approved):

aafeijoo-suse (approved):

PR #2019, opened 2022-10-26
feat(multipath): install tmpfiles.d config file

mwilck:
## Changes

Recent multipath-tools (0.9.2) added a tmpfiles.d template. Install it
if present.

## Checklist

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -106,7 +106,8 @@ install() {
         /etc/xdrdevices.conf \
         /etc/multipath.conf \
         /etc/multipath/* \
-        "$config_dir"/*
+        "$config_dir"/* \
+        "$tmpfilesdir/multipath.conf"
 
     mpathconf_installed \
         && [[ $hostonly ]] && [[ $hostonly_mode == "strict" ]] && {


mwilck:
@aafeijoo-suse  for information

aafeijoo-suse (dismissed):

johannbg (changes requested):
This brings in module dependency on systemd tmpfiles module which you forgot to add.

mwilck:
> This brings in module dependency on systemd tmpfiles module which you forgot to add.

Hm, why would it? The multipath module doesn't strictly depend on systemd, and I think it should remain so. 

The `00systemd` module pulls in [systemd-tmpfiles and the related services](https://github.com/dracutdevs/dracut/blob/862ba52683834f87722cae7a6692a59d09271ec3/modules.d/00systemd/module-setup.sh#L87), even if the `systemd-tmpfiles` module is not included. On my system I can see the following in the initrd (in which `systemd-tmpfiles` is **not** included):

```
# lsinitrd /boot/initrd-5.19.12-1-default |grep tmpfiles
-rwxr-xr-x   1 root     root       100744 Sep 17 21:50 usr/bin/systemd-tmpfiles
lrwxrwxrwx   1 root     root           37 Sep 17 21:50 usr/lib/systemd/system/sysinit.target.wants/systemd-tmpfiles-setup-dev.service -> ../systemd-tmpfiles-setup-dev.service
lrwxrwxrwx   1 root     root           33 Sep 17 21:50 usr/lib/systemd/system/sysinit.target.wants/systemd-tmpfiles-setup.service -> ../systemd-tmpfiles-setup.service
-rw-r--r--   1 root     root          747 Aug 16 15:46 usr/lib/systemd/system/systemd-tmpfiles-setup-dev.service
-rw-r--r--   1 root     root          814 Aug 16 15:46 usr/lib/systemd/system/systemd-tmpfiles-setup.service
drwxr-xr-x   1 root     root            0 Sep 17 21:50 usr/lib/tmpfiles.d
-rw-r--r--   1 root     root           35 Sep 15 23:11 usr/lib/tmpfiles.d/cryptsetup.conf
-rw-r--r--   1 root     root          135 Sep 14 10:29 usr/lib/tmpfiles.d/dracut-tmpfiles.conf
-rw-r--r--   1 root     root           34 Sep 17 21:50 usr/lib/tmpfiles.d/multipath.conf
-rw-r--r--   1 root     root         1124 Sep 17 21:50 usr/lib/tmpfiles.d/systemd.conf
```

AFAICS, currently no module depends on `systemd-tmpfiles`.

johannbg (approved):
Hmm we should be doing that, as in express correct module dependency in modules and since I need to correct that we can just as well merger you PR and then have it be part of that scope and or the scope that separates every module that uses systemd and non systemd to their own modules.

johannbg:
@aafeijoo-suse It's going to be a bit of a work as in three systemd scopes ( the tmpfile module dependency fixup, the split as in systemd vs other inits, and the update of the systemd module itself ) all that needs to be done within the same release cycle to avoid breakage downstream so you can re-ack and merge

LaszloGombos (approved):
LGTM

@mwilck It looks to me that there is no test for the multipath dracut module. This module seems to be fairly complex and receive a lot of commits, so perhaps you should consider making a test. Thanks !

mwilck:
@johannbg: as he systemd module itself ships `tmpfiles.d` entries, it would probably be sufficient to just add a dependency from `systemd` to `systemd-tmpfiles` (and remove the explicit `tmpfiles` part from the `systemd` module). Either that, or fully merge the two modules, which seem to be unable to exist without each other, anyway.

johannbg:
@mwilck I got badly burned when handling sysv to systemd by such rpm package shenanigans so dependency will be correctly expressed in module.

mwilck:
@johannbg, I think you misunderstood what you meant. I am not proposing anything rpm-level. I meant to merge the dracut modules `systemd` and `systemd-tmpfiles`.

johannbg:
@mwilck how we split our modules depends on the upstream they are based upon on so if you no longer can compile systemd without tmpfile support then we can merge the module with the main systemd module.

PR #2020, opened 2022-10-26
feat(dracut.sh): allow setting output directory of UEFI executable

### LINKED ISSUE #1926: Don't always use hard-coded ``/efi/EFI/Linux/`` directory for EFISTUB
I'm using dracut and refind on Fedora and using the unified, signed EFI STUB generation. 
And Fedora already has / already created the directory ``/efi/EFI/Fedora`` on my EFI System Partition. 
So it would be nice if dracut could put the EFISTUB files into ``/efi/EFI/Fedora`` on Fedora and not into ``/efi/EFI/Linux``. 

[The lines 1125 and 1126 in dracut.sh](
https://github.com/dracutdevs/dracut/blob/master/dracut.sh#L1125): 
```
mkdir -p "$dracutsysrootdir$efidir/Linux"
outfile="$dracutsysrootdir$efidir/Linux/linux-$kernel${MACHINE_ID:+-${MACHINE_ID}}${BUILD_ID:+-${BUILD_ID}}.efi" 
```
... would have to be changed for that. 

dracut could just use the ``ID`` parameter in ``/etc/os-release`` (which is ``Fedora`` for Fedora and ``Arch`` for Arch Linux): 
```
source /etc/os-release 
distro=$ID
```

and check if the directory ``/efi/EFI/$distro/`` (``$dracutsysrootdir$efidir/$distro``) exists. 
If that directory exists, then don't use ``/efi/EFI/Linux`` (``$dracutsysrootdir$efidir/Linux``), 
but ``/efi/EFI/Fedora`` instead (``"$dracutsysrootdir$efidir/$distro``).

aafeijoo-suse:
Instead of creating the UEFI executable in a hard-coded directory based on where the EFI partition is mounted, allow the output directory to be set using the `--uefi-outdir` command line option or the `uefi_outdir` configuration option.

## Checklist

Fixes #1926

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -266,6 +266,7 @@ Creates initial ramdisk images for preloading modules
   --no-uefi             Disables UEFI mode.
   --no-machineid        Affects the default output filename of the UEFI
                          executable, discarding the <MACHINE_ID> part.
+  --uefi-outdir [DIR]   Create the UEFI executable in [DIR].
   --uefi-stub [FILE]    Use the UEFI stub [FILE] to create an UEFI executable.
   --uefi-splash-image [FILE]
                         Use [FILE] as a splash image when creating an UEFI
@@ -459,6 +460,7 @@ rearrange_params() {
             --long loginstall: \
             --long uefi \
             --long no-uefi \
+            --long uefi-outdir: \
             --long uefi-stub: \
             --long uefi-splash-image: \
             --long kernel-image: \
@@ -824,6 +826,11 @@ while :; do
         --no-reproducible) reproducible_l="no" ;;
         --uefi) uefi_l="yes" ;;
         --no-uefi) uefi_l="no" ;;
+        --uefi-outdir)
+            uefi_outdir_l="$2"
+            PARMS_TO_STORE+=" '$2'"
+            shift
+            ;;
         --uefi-stub)
             uefi_stub_l="$2"
             PARMS_TO_STORE+=" '$2'"
@@ -1073,6 +1080,7 @@ stdloglvl=$((stdloglvl + verbosity_mod_l))
 [[ $reproducible_l ]] && reproducible="$reproducible_l"
 [[ $loginstall_l ]] && loginstall="$loginstall_l"
 [[ $uefi_l ]] && uefi=$uefi_l
+[[ $uefi_outdir_l ]] && uefi_outdir="$uefi_outdir_l"
 [[ $uefi_stub_l ]] && uefi_stub="$uefi_stub_l"
 [[ $uefi_splash_image_l ]] && uefi_splash_image="$uefi_splash_image_l"
 [[ $kernel_image_l ]] && kernel_image="$kernel_image_l"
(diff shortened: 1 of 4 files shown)


dracut.sh
```diff
@@ -1107,23 +1115,26 @@ if ! [[ $outfile ]]; then
...
+                if [[ -d $dracutsysrootdir/boot/efi/EFI ]]; then
+                    uefi_outdir=/boot/efi/EFI
                 fi
             fi
-        else
-            efidir=/boot/EFI
-            if [[ -d $dracutsysrootdir/boot/efi/EFI ]]; then
-                efidir=/boot/efi/EFI
-            fi
+            uefi_outdir="$uefi_outdir/Linux"
```
LaszloGombos, line 1134: I would like to understand this line better. Why is "Linux" appended to uefi_outdir here ?

aafeijoo-suse: It's what currently dracut is doing by default since https://github.com/dracutdevs/dracut/commit/5c57209b. See https://systemd.io/BOOT_LOADER_SPECIFICATION/#type-2-efi-unified-kernel-images

aafeijoo-suse: Hmm, I think the GitHub diff is confusing you, the "/Linux" part is only appended if the user does not set a custom uefi output directory. See the snippet as a whole:

https://github.com/dracutdevs/dracut/blob/65ae8d8d0d2148a5a6e6a31adb355093c8915675/dracut.sh#L1118-L1137

LaszloGombos (approved):
LGTM

johannbg:
@aafeijoo-suse hmm what problem/usecase is this solving as in why would you ever want to place it somewhere else than where the EFI partition is mounted?

aafeijoo-suse:
> @aafeijoo-suse hmm what problem/usecase is this solving as in why would you ever want to place it somewhere else than where the EFI partition is mounted?

Not my use case, but it looks like someone else's. See #1926

johannbg:
@aafeijoo-suse yeah that's a nack from my pov. Fedora is doing it wrong. It ( along with all the other distro's ) should just follow standards here not inserting their "distro name" which makes no sense.

aafeijoo-suse:
Fair enough, if anyone else (@lnykryn @pvalena) wants to discuss this, please reopen it.

aafeijoo-suse:
> @aafeijoo-suse yeah that's a nack from my pov. Fedora is doing it wrong. It ( along with all the other distro's ) should just follow standards here not inserting their "distro name" which makes no sense.

After reading the recent https://uapi-group.org/specifications/specs/boot_loader_specification/, inserting the distribution name starts to make sense if the intention is to _share_ the same ESP and/or /boot partitions between different OSs.

johannbg:
If that's the case it sounds like a bug ( the separation should be done on file level and or simply entry level in files I would think ) from that newly formed group in it's specification in anycase we should see how well that uapi/specification will be adopted prior to rely on it. 

Given how the uapi group came about and that behind the scene systemd community is pushing it, it begs the question how it will affect none systemd based distro hence I would not be surprised that it just gets ignored with the exception of maybe fedora.

PR #2022, opened 2022-10-26
Make 03modsign module load keys to correct keyring

webczat:
Until now, 03modsign module was loading keys from /lib/modules/keys/* into the current session keyring.
This change makes it add keys to the secondary trusted keyring. This works only as long as added certificate is signed by key from the same keyring.

This pull request changes...

## Checklist

Fixes #


aafeijoo-suse:
CC @vathpela

webczat:
this issue is still valid.

johannbg:
We need someone with in-depth knowledge what a proper scope for this should look like.

aafeijoo-suse:
@msmeissn maybe you can help us to review this change.

mikhailnov:
Wow! Thanks a lot, it works! And kernel docs seem to have a mistake: https://www.kernel.org/doc/html/latest/admin-guide/module-signing.html says about adding keys into .builtin_trusted_keys, which is a child of .secondary_trusted_keys, adding into .builtin_trusted_keys did not work but adding into .secondary_trusted_keys worked.

mikhailnov:
> We need someone with in-depth knowledge what a proper scope for this should look like.

I think I should write a how to about this... There is no good doc about it, I've spent a lot of time trying to make this work

webczat:
Well, I think documentation might just be outdated.

msmeissn:
FWIW i tried to find out the standards on this, but I failed and currently have no deeper knowlegde of this.

mikhailnov:
> It might make sense to mention somewhere (e.g. code or commit message) that this feature has a dependency on the CONFIG_SECONDARY_TRUSTED_KEYRING kernel config.

We may add a check that the keyring .secondary_trusted_keys exists and print a message if it does not, I think

webczat:
there are also other common reasons for this module to fail, not only the keyring might not exist, but also it won't be able to load certificates which aren't signed with an already trusted key. However isn't keyctl failing enough? it's still a kinda special/uncommon use case.

mikhailnov:
Probably yes, I agree

webczat:
ping, anyone?

tblume (commented):
According to my information, the kernel will only load the MOK table, when booted in uefi secure boot mode.
Without secure boot, your key still won't be present in /proc/keys.
Is it desirable to bind signed module loading to secure boot?

webczat:
this pull request is not related to that. it only fixes key loading. as in loads keys to secondary trusted keyring. Which is allowed if the signing key is in the system keyring/builtin trusted keys. this is not limited to MOK keys. the key built into the kernel is also there.

tblume (approved):
Ok, thanks, I also got confirmation from our kernel developers that this is the right approach.

mikhailnov:
Thanks!

PR #2025, opened 2022-10-27
fix(dracut.sh): split drivers_dir check

### LINKED ISSUE #1608: Dracut complains that `--kmoddir /lib/modules/...` doesn't have "/lib/modules" as a parent directory.
**Describe the bug**
Dracut doesn't see `/lib/modules/` in the kmoddir and complains, but it is infact there in the path.

**Distribution used**
Arch (updated today)

**Dracut version**
055

**Init system**
systemd 249 (249.4-1-arch)

**To Reproduce**
```
> sudo dracut --kernel-image /boot/vmlinuz-linux --kmoddir /lib/modules/5.14.5-arch1-1/
dracut: -k/--kmoddir path must contain "lib/modules" as a parent of your kernel module directory,
dracut: or modules may not be placed in the correct location inside the initramfs.
dracut: was given: /lib/modules/5.14.5-arch1-1/
dracut: expected: /lib/modules/lib/modules/5.13.5-arch1-1
dracut: Please move your modules into the correct directory structure and pass the new location,
dracut: or set DRACUT_KMODDIR_OVERRIDE=1 to ignore this check.
```

**Expected behavior**
Typing a kmoddir that includes "lib/modules" should be accepted and not duplicate into "lib/modules/lib/modules"

**Additional context**
I upgraded to the latest kernel and have not yet rebooted the system. I'd like to gen an initramfs before rebooting. I wonder if dracut is confused because I'm telling it to build an image for 5.14 but am running 5.13...

aafeijoo-suse:
The directory where to look for kernel modules can be set via the `-k`/`--kmoddir` command line option or the `drivers_dir` configuration option. Its current check should be split into two different ones to avoid misleading error messages (see #1608):
- First check that its basename matches the kernel version set for the initramfs (via `--kver` or automatically set by `uname -r`).
- Second check that the parent directory of the last provided directory contains "/lib/modules/". This check was also fixed to avoid accepting other directories like "xxxlib/modules".

## Checklist

Fixes #1608

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1031,6 +1031,7 @@ stdloglvl=$((stdloglvl + verbosity_mod_l))
 ((stdloglvl < 0)) && stdloglvl=0
 
 [[ $drivers_dir_l ]] && drivers_dir=$drivers_dir_l
+drivers_dir="${drivers_dir%%+(/)}"
 [[ $do_strip_l ]] && do_strip=$do_strip_l
 [[ $do_strip ]] || do_strip=yes
 [[ $aggressive_strip_l ]] && aggressive_strip=$aggressive_strip_l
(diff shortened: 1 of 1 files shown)


LaszloGombos (approved):
LGTM

johannbg (approved):

PR #2029, opened 2022-11-01
fix(dracut-logger.sh): fix check for /dev/log

bcl:
The /dev/log check was not correct, it would continue working if /dev/log was a broken symlink. This changes the logic to be more clear -- /dev/log needs to be a socket, and writable, and logger needs to be available. If any of those conditions fail, turn off syslog and only use console logging.

This pull request changes...

## Checklist

### DIFF
--- a/dracut-logger.sh
+++ b/dracut-logger.sh
@@ -151,7 +151,7 @@ dlog_init() {
             readonly _dlogfd=15
             systemd-cat -t 'dracut' --level-prefix=true < "$_systemdcatfile" &
             exec 15> "$_systemdcatfile"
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (changes requested):
Please, fix the commit message to pass the commisery checks:

```
Error: commit_message:1:24: error: commit message's description contains blacklisted word or repeats type tag
fix(dracut-logger.sh): fix check for /dev/log
                       ^^^
commit_message:1:24: note: prefer using the imperative for verbs
```
Apart from that the patch looks good.

aafeijoo-suse (approved):

LaszloGombos (approved):
LGTM

PR #2030, opened 2022-11-02
fix(90kernel-modules): add (nonstandard) NVMe drivers

jlinton:
The M1/etc hardware has an quirky NVMe disk controller that needs a dedicated NVMe driver. Rather than picking up just that bit of hardware pick up drivers that register NVMe controllers.

This also picks up NVMe-OF style drivers, but seems like a better long term solution on Arm platforms which tend to have a lot of odd hardware.

Signed-off-by: Jeremy Linton <jlinton@redhat.com>

This pull request changes...

## Checklist

Fixes #


LaszloGombos:
@jlinton Can you please help and change the title of commit message to lowercase  ?

Error: commit_message:1:24: error: don't use title case in the description
fix(90kernel-modules): Add (nonstandard) NVMe drivers

LaszloGombos:
CC @mwilck

jlinton:
Ok, lower cased the "Add", and changed to SOB as well.

johannbg (approved):
LGTM

LaszloGombos (approved):

PR #2033, opened 2022-11-04
feat(dmsquash-live): prefer udevadm over of blkid

LaszloGombos:
det_fs wraps udevadm instead of calling blkid.

## Checklist


Conan-Kudo (approved):

johannbg:
@LaszloGombos your commit message does not accurately describe what the benefits are for making this change. The "why"

LaszloGombos:
> The "why"

udevadm is a modern replacement for blkid. udevadm  is already a hard dependency for the generated initramfs.
    
Many dracut modules (e.g. fs-lib, btrfs, mdraid) are already exclusively using udevadm.
    
dmsquash-live is one of the few remaining dracut modules that are still using blkid.

dmsquash-live in particular using blkid in 3 places and udevadm in 4 places - with no particular benefit of having two dependencies instead of just one.

Longer term I think blkid usage can be completely eliminated, but that is beyond the scope of this PR.

Hard to estimate the risk of running into different bugs and inconsistencies between using both binaries but here is one random example - https://forum.rockstor.com/t/devices-visible-to-udevadm-but-not-blkid-or-btrfs/429/10

marcosfrm:
udev uses libblkid to populate `ID_*` properties anyway. LGTM.

dtardon:
> > The "why"
> 
> udevadm is a modern replacement for blkid.

No, it's not. `udevadm info` and `blkid` have completely different purposes; they just both happen to provide a particular functionality--getting `ID_*` properties of disk devices. (Just `blkid`'s output is more flexible. E.g., there's no way to get a single property using `udevadm info`; that's why `det_fs()` has to process the output.)

(Btw, `blkid` is newer than `udevadm`. It was created in 2008 as a command line interface for `libblkid`.)

marcosfrm:
> there's no way to get a single property using `udevadm info`

There is now:

https://github.com/dracutdevs/dracut/pull/2038#issuecomment-1304881235


:)

LaszloGombos:
> udevadm is a modern replacement for blkid.

changed it to ...

udevadm is more flexible replacement for blkid.

dtardon:
> > udevadm is a modern replacement for blkid.
> 
> changed it to ...
> 
> udevadm is more flexible replacement for blkid.

That's still wrong. Neither tool is a replacement for the other. The intersection of their functionalities is quite small--the printing of (some) block device properties is just about it. (And even that they do differently: `udevadm info` reads from `systemd-udevd`'s cache, while `blkid` reads directly from `/sys` and the device's superblock.)

I'd write something like "There's no need to require another tool (`blkid`) for functionality that's also provided by `udevadm info`." or "Let's consolidate on a single tool to print block device properties." or something to that effect. IMHO The main point to convey is that there's no need to use two tools when one is enough.

LaszloGombos:
@dtardon Thanks, I changed the commit message

> The main point to convey is that there's no need to use two tools when one is enough.

Yes, this is indeed the main point.

PR #2036, opened 2022-11-04
Add `efi-secret` to qemu driver list

crobinso:
## Checklist

This adds a couple 'confidential computing' kernel modules to the qemu driver list. They have generic sounding names, but they are VM specific, and they are used in attestation flows which typical end with unlocking the luks encrypted rootfs, so they need to be available at boot time.

### DIFF
--- a/modules.d/90qemu/module-setup.sh
+++ b/modules.d/90qemu/module-setup.sh
@@ -18,5 +18,6 @@ installkernel() {
         virtio_blk virtio virtio_ring virtio_pci \
         virtio_scsi virtio_console virtio_rng virtio_mem \
         spapr-vscsi \
-        qemu_fw_cfg
+        qemu_fw_cfg \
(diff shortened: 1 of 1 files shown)


LaszloGombos:
Error: commit_message:1:13: error: don't use title case in the description
feat(qemu): Add efi_secret driver
            ^~~

Error: commit_message:1:13: error: don't use title case in the description
feat(qemu): Add sev-guest driver

crobinso:
After thinking some more, I dropped patch 2. `sev-guest` in the initrd is only useful if there's some other software in the initrd that actually communicates with it, which doesn't exist yet in distro packages. Once it does exist, it could have hooks for getting into the initrd, and would pull the `sev-guest` driver in too.

johannbg (approved):
LGTM

LaszloGombos (approved):

PR #2037, opened 2022-11-05
fix(kernel-modules): Add sysctl to initramfs to handle modprobe files

Conan-Kudo:
This pull request changes...

## Changes
`dracut-systemd` module now will incorporate `sysctl` into initramfs so that modprobe configuration files that call sysctl will function properly.

## Checklist

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -147,4 +147,5 @@ install() {
         inst_hook cmdline 01 "$moddir/parse-kernel.sh"
     fi
     inst_simple "$moddir/insmodpost.sh" /sbin/insmodpost.sh
+    inst_multiple -o sysctl
 }


davide125 (approved):

iammattcoleman (approved):

johannbg:
We already have a systemd sysctl module so adding this to dracut-systemd module should not be necessary. What usecase does this solve which is not solved by including that module?

Conan-Kudo:
The `systemd-sysctl` module doesn't use the `sysctl` binary, whereas this one will because various modprobe.d config files _can_ use it. For example:

```
./etc/modprobe.d/firewalld-sysctls.conf:install nf_conntrack /sbin/modprobe --ignore-install nf_conntrack $CMDLINE_OPTS && /sbin/sysctl --quiet --pattern 'net[.]netfilter[.]nf_conntrack.*' --system
./usr/lib/modprobe.d/50-nfs.conf:install sunrpc /sbin/modprobe --ignore-install sunrpc $CMDLINE_OPTS && { /sbin/sysctl -q --pattern sunrpc --system; exit 0; }
./usr/lib/modprobe.d/50-nfs.conf:install rpcrdma /sbin/modprobe --ignore-install rpcrdma $CMDLINE_OPTS && { /sbin/sysctl -q --pattern sunrpc.svc_rdma --system; exit 0; }
./usr/lib/modprobe.d/50-nfs.conf:install lockd /sbin/modprobe --ignore-install lockd $CMDLINE_OPTS && { /sbin/sysctl -q --pattern fs.nfs.n[sl]m --system; exit 0; }
./usr/lib/modprobe.d/50-nfs.conf:install nfsv4 /sbin/modprobe --ignore-install nfsv4 $CMDLINE_OPTS && { /sbin/sysctl -q --pattern 'fs.nfs.(nfs_callback_tcpport|idmap_cache_timeout)' --system; exit 0; }
./usr/lib/modprobe.d/50-nfs.conf:install nfs /sbin/modprobe --ignore-install nfs $CMDLINE_OPTS && { /sbin/sysctl -q --pattern fs.nfs --system; exit 0; }
```

johannbg:
If this gets added it belongs in the systemd sysctl module not in the dracut systemd module.

Conan-Kudo:
Obviously I could add it there and it'd _coincidentally_ work, but the `systemd-sysctl` module isn't where the breakage is occurring. It makes sense to me to add the binary in the module that winds up invoking the tool.

johannbg:
Arguably the modules that require sysctl be invoked in the firstplace are broken and those should be fixed so there is high probably this gets nacked

iammattcoleman:
> Arguably the modules that require sysctl be invoked in the firstplace are broken and those should be fixed so there is high probably this gets nacked

While I agree with this philosophically, the practical reality is that systemd-sysctl is not universally adopted. To try to dictate that every other distribution and creator of modprobe drop-in config files needs to use it seems like an uphill battle.

LaszloGombos:
How do you all feel about adding sysctl to the kernel-modules dracut module as an optional dependency ? 
If usage of sysctl is indeed modprobe related, than kernel-modules could be a better home for this as kernel-modules dracut module is responsible for installing the modprobe *.conf files into initrd.

johannbg:
> > Arguably the modules that require sysctl be invoked in the firstplace are broken and those should be fixed so there is high probably this gets nacked
> 
> While I agree with this philosophically, the practical reality is that systemd-sysctl is not universally adopted. To try to dictate that every other distribution and creator of modprobe drop-in config files needs to use it seems like an uphill battle.

We make the assumption that distribution that adopted systemd did so wholesale trying to support ca 250 systemd based distribution that randomly decided which component if systemd uses is madness.

johannbg:
> How do you all feel about adding sysctl to the kernel-modules dracut module as an optional dependency ? If usage of sysctl is indeed modprobe related, than kernel-modules could be a better home for this as kernel-modules dracut module is responsible for installing the modprobe *.conf files into initrd.

That's an ack from me.

Conan-Kudo:
I'm fine with that too, that's where I initially thought to do it, but when I did the search for what loaded the files, this module came up instead.

Conan-Kudo:
@LaszloGombos I just moved it over to the `kernel-modules` module now.

iammattcoleman (approved):

modules.d/90kernel-modules/module-setup.sh
```diff
@@ -147,4 +147,5 @@ install() {
         inst_hook cmdline 01 "$moddir/parse-kernel.sh"
     fi
     inst_simple "$moddir/insmodpost.sh" /sbin/insmodpost.sh
+    inst_multiple sysctl
```
LaszloGombos, line 150: I would recommend to use the "-o" argument here to make it optional dependency.

Conan-Kudo: Why? procps-ng is already a hard dependency of dracut anyway, and not having this causes modprobe config files to fail to work if they're using it.

LaszloGombos: Dependency here means a single binary included into the generated initramfs (not a package dependency).

> modprobe config files to fail 
Most examples listed are for nfs. If you want to include systctl as hard dependency (and not an optional one) than perhaps it should be done in the nfs dracut module.

Can you give an example of failure that is not nfs (ideally something that most people would run into) ? 

If an initramfs only boots from local storage (e.g. no networking), why do we need to include a new binary into everybody's generated initramfs ?

johannbg: Vendors might be using systemd only systems in which they so they might want to exclude it thus making it an optional dependency is better as @LaszloGombos is suggesting + we need to properly test this so we dont end up potentially adding some weird race condition with systemd-sysctl encase that module is enabled.

Conan-Kudo: Okay, marked as optional now.

LaszloGombos (approved):
lgtm

LaszloGombos:
Perhaps once this PR lands, we no longer need this code in the test - https://github.com/dracutdevs/dracut/commit/4a5785a8597b97957da2d8ea62e029f3da487ef8..

aafeijoo-suse (approved):

PR #2038, opened 2022-11-06
feat(fs-lib): add support for ntfs3 kernel driver

LaszloGombos:
## Checklist

### DIFF
--- a/modules.d/99fs-lib/fs-lib.sh
+++ b/modules.d/99fs-lib/fs-lib.sh
@@ -203,6 +203,11 @@ det_fs() {
         done)
     _fs=${_fs:-auto}
 
+    # prefer ntfs3 kernel driver for mouting ntfs
+    if [ "$_fs" = "ntfs" ] && load_fstype ntfs3; then
+        _fs="ntfs3"
+    fi
+
     if [ "$_fs" = "auto" ]; then
         _fs="$_orig"
     fi


LaszloGombos:
CC @marcosfrm

marcosfrm:
Less convoluted:

```sh
    livedev_fstype=$(blkid -o value -s TYPE "$livedev")
    if [ "$livedev_fstype" = "ntfs" ] && load_fstype ntfs3; then
        # prefer ntfs3 kernel driver for mouting ntfs
        livedev_fstype="ntfs3"
    fi

    ...
```

Needs `instmods ntfs3` in `module-setup.sh`, or will it be opt-in?

LaszloGombos:
> Needs `instmods ntfs3` in `module-setup.sh`, or will it be opt-in?

I hope to keep it opt-in, but lets get more input on this from others..

LaszloGombos:
Moved this piece of code to fs-lib to make ntfs3 support available for other dracut modules.

This change however makes this PR dependent on https://github.com/dracutdevs/dracut/pull/2033. Hopefully this also helps to justify https://github.com/dracutdevs/dracut/pull/2033 .

marcosfrm:
Btw, since systemd 250 (yeah, fairly new), `udevadm info` has `--property` and `--value` options. Perhaps add a TODO to simplify this in the future:

https://github.com/dracutdevs/dracut/blob/e78f71b9da29828ee4cd8444d4841ee127ef8614/modules.d/99fs-lib/fs-lib.sh#L197-L203

marcosfrm:
`det_fs()` can be simplified:

```diff
diff --git a/modules.d/99fs-lib/fs-lib.sh b/modules.d/99fs-lib/fs-lib.sh
index e05fa0c5..bd6e6c3c 100755
--- a/modules.d/99fs-lib/fs-lib.sh
+++ b/modules.d/99fs-lib/fs-lib.sh
@@ -201,12 +201,8 @@ det_fs() {
                 break
             fi
         done)
-    _fs=${_fs:-auto}

-    if [ "$_fs" = "auto" ]; then
-        _fs="$_orig"
-    fi
-    echo "$_fs"
+    echo "${_fs:-$_orig}"
 }

 write_fs_tab() {
```

Conan-Kudo:
I'm not comfortable with preferring `ntfs3` at this time. Also, I'm pretty sure that Paragon still hasn't released userspace utility code to go alongside the driver yet.

LaszloGombos:
@Conan-Kudo 

> I'm not comfortable with preferring ntfs3 at this time. 

I understand the hesitation. I do not have a strong preference on this, but note that this PR only makes a preference for ntfs3, if the kernel driver is in the generated initramfs. One way this could happen is if specifically opted in to ntfs3 using e.g. "--drivers" command line option.

Dracut would not pick up ntfs3 by default into the generated initramfs.

If however ntfs3 driver is built into the kernel, than this PR would indeed would make a switch from ntfs to ntfs3 kernel driver.

PR #2040, opened 2022-11-06
refactor(fs-lib.sh): udevadm info --query=property instead of --query…

marcosfrm:
…=env

env is a long time deprecated alias to property, not listed in the man page for more than a decade.

While at it, add TODO for udevadm info's --property and --value options.

## Checklist

Fixes #

### DIFF
--- a/modules.d/90btrfs/btrfs_device_ready.sh
+++ b/modules.d/90btrfs/btrfs_device_ready.sh
@@ -6,7 +6,7 @@ btrfs_check_complete() {
     local _rootinfo _dev
     _dev="${1:-/dev/root}"
     [ -e "$_dev" ] || return 0
-    _rootinfo=$(udevadm info --query=env "--name=$_dev" 2> /dev/null)
+    _rootinfo=$(udevadm info --query=property "--name=$_dev" 2> /dev/null)
     if strstr "$_rootinfo" "ID_FS_TYPE=btrfs"; then
         info "Checking, if btrfs device complete"
         btrfs device ready "$_dev" > /dev/null 2>&1
--- a/modules.d/90btrfs/btrfs_finished.sh
+++ b/modules.d/90btrfs/btrfs_finished.sh
@@ -6,7 +6,7 @@ btrfs_check_complete() {
     local _rootinfo _dev
     _dev="${1:-/dev/root}"
     [ -e "$_dev" ] || return 0
-    _rootinfo=$(udevadm info --query=env "--name=$_dev" 2> /dev/null)
+    _rootinfo=$(udevadm info --query=property "--name=$_dev" 2> /dev/null)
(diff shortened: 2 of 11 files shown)


LaszloGombos:
@marcosfrm thanks.. would you be interested  doing more work :-)

modules.d/90btrfs/btrfs_finished.sh:    _rootinfo=$(udevadm info --query=env "--name=$_dev" 2> /dev/null)
modules.d/90btrfs/btrfs_device_ready.sh:    _rootinfo=$(udevadm info --query=env "--name=$_dev" 2> /dev/null)
modules.d/90mdraid/mdraid-waitclean.sh:        udevinfo="$(udevadm info --query=env --name="$md")"
modules.d/90mdraid/mdraid-cleanup.sh:    udevinfo="$(udevadm info --query=env --name="$md")"
modules.d/90mdraid/mdraid_start.sh:    _udevinfo="$(udevadm info --query=env --name="${_md}")"

We could perhaps have more places call into fs-lib.sh...

It is ok to keep in the PR as is, just wanted to make a note on more call sites..

LaszloGombos (approved):

johannbg:
@marcosfrm can you expand on what you mean by when we can depend on systemd 250? 
Like other upstream we follow other upstream closely in other words we do not "support" dracut being used in some kind of downstream russian release roulette of the core/baseOS components that make up the distro doing.

When we release we make the assumption that whomever is consuming dracut is also consuming the latest kernel, systemd along with any other component that dracut uses so in that sense we already "depend" on systemd 252 and in that sense you can already make the changes you want to do.

johannbg (changes requested):
We already "depend" on systemd 250 =>
With systemd 252 being released, systemd release 250 became unsupported by upstream systemd so update the PR with the change you imply you want to do and remove that TODO section.

marcosfrm:
> @marcosfrm thanks.. would you be interested doing more work :-)

Will do.

> With systemd 252 being released, systemd release 250 became unsupported by upstream systemd so update the PR with the change you imply you want to do and remove that TODO section.

OK, but I am sure someone will complain. Let's see :).

johannbg:
> OK, but I am sure someone will complain. Let's see :).

Well if someone happens to complain we just kindly point that someone out that that's something they need to take up with the vendor/distribution they are using. 

It's not our problem that downstream has chosen to ship newer version of dracut with older version of other core/baseOS components, doing so is "unsupported" by us. ( All the fallout/workaround for such decisions is something that vendor/downstream making that decisions gets to carry as in patches, workarounds etc. ).

marcosfrm:
I will separate udevadm info's --property and --value options into another PR.

LaszloGombos (approved):
thanks

johannbg (approved):
LGTM

marcosfrm:
Should have removed "(fs-lib.sh)" from the commit title. Sorry.

PR #2041, opened 2022-11-07
ci: fix Arch container

LaszloGombos:
packages should be installed with -Syu instead of -Sy.

## Checklist

Fixes #

### DIFF
--- a/test/container/Dockerfile-Arch
+++ b/test/container/Dockerfile-Arch
@@ -8,7 +8,7 @@ LABEL RUN="docker run -it --name NAME --privileged --ipc=host --net=host --pid=h
 RUN echo 'export DRACUT_NO_XATTR=1 KVERSION=$(cd /lib/modules; ls -1 | tail -1)' > /etc/profile.d/dracut-test.sh
 
 # Install needed packages for the dracut CI container
-RUN pacman --noconfirm -Sy \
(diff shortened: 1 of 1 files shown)


LaszloGombos:
The scheduled container generation started to fail on Nov-5
https://github.com/dracutdevs/dracut/actions/runs/3399901041/jobs/5653953626

It was building on Nov-4
https://github.com/dracutdevs/dracut/actions/runs/3393451785/jobs/5640805180

LaszloGombos:
From https://wiki.archlinux.org/title/pacman

> ...do not run pacman -Sy package_name instead of pacman -Syu package_name, as this could lead to dependency issues.

johannbg (approved):

aafeijoo-suse (approved):

aafeijoo-suse:
@LaszloGombos unrelated, but Dockerfile-Gentoo took 55 minutes to finish...

LaszloGombos:
> @LaszloGombos unrelated, but Dockerfile-Gentoo took 55 minutes to finish...

That is expected. It used to be over 3 hours before https://github.com/dracutdevs/dracut/commit/11eef2cfd2349dcf0a0d872e759c6c71e85e89aa landed. 

This is only for PRs that change Dockerfiles, not for running regression tests.

PR #2045, opened 2022-11-08
fix(base): run systemctl daemon-reload after remove_hostonly_files

aafeijoo-suse:
Dependencies to systemd .device unit files are marked as "hostonly" in the special file hostonly-files. These .device unit files can be removed using the `rd.hostonly=0` command line parameter. But, since systemd v252 (https://github.com/systemd/systemd/pull/24054), the initrd-parse-etc.service service does not run `systemctl daemon-reload`, so systemd keeps waiting for these .device units to show up.
This means that if a hostonly device is not available for any reason at boot time, using `rd.hostonly=0` has no effect and the system will not boot.

## Checklist

### DIFF
--- a/modules.d/98dracut-systemd/dracut-cmdline.sh
+++ b/modules.d/98dracut-systemd/dracut-cmdline.sh
@@ -11,6 +11,7 @@ type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
 if ! getargbool 1 'rd.hostonly'; then
     [ -f /etc/cmdline.d/99-cmdline-ask.conf ] && mv /etc/cmdline.d/99-cmdline-ask.conf /tmp/99-cmdline-ask.conf
     remove_hostonly_files
+    systemctl --no-block daemon-reload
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

modules.d/99base/dracut-lib.sh
```diff
@@ -1125,6 +1125,9 @@ remove_hostonly_files() {
             [ -e "$line" ] || [ -h "$line" ] || continue
             rm -f "$line"
         done < /lib/dracut/hostonly-files
+        if [ -n "$DRACUT_SYSTEMD" ]; then
```
LaszloGombos, line 1128: Would not be modules.d/98dracut-systemd/dracut-cmdline.sh better place to do this work ? That would make the code more modular and you would not need to check if DRACUT_SYSTEMD ?

aafeijoo-suse: Yes, that makes sense. I added the check in dracut-lib just because other functions do the same.

LaszloGombos (approved):
LGTM. Thanks !

PR #2046, opened 2022-11-08
feat(dracut.sh): improve detection of kernel versions installed on the system

aafeijoo-suse:
This aims to improve the detection of kernel versions installed on the system when running dracut with the `--regenerate-all` option.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -965,13 +965,15 @@ if [[ $regenerate_all == "yes" ]]; then
     cd "$dracutsysrootdir"/lib/modules || exit 1
     if [[ $parallel != "yes" ]]; then
         for i in *; do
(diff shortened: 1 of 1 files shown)


johannbg:
Where does this path come from? You should already be in the "kernel" version at that point so how does this improve the detection of kernel versions? What's the path you used when you tested this?

johannbg:
@aafeijoo-suse sorry not following valid by what and or whom and to who or what entity do you ow the duty to submit it upstream first? yourself, your employer?  If there is a valid technical argument by Suse doing it the way it does then present that argument, if it's inline with ( established ) specifications cite that but you should know already no hacks or workarounds for downstream.

Issue #1971, opened 2022-09-02
kernel-modules-extra: dracut failed if invalid .ko files exist in modules.dep


jiayi0118:
**Describe the bug**
dracut failed inside `installkernel` in module kernel-modules-extra.

```
# dracut -f ./test.img
realpath: weak-updates/test.ko: No such file or directory
dracut: installkernel failed in module kernel-modules-extra
```

**Distribution used**
centos stream 9 (kernel version: 5.14.0-160.el9.x86_64)

**Dracut version**
dracut 057-13.git20220816.el9

**Init system**
systemd 250 (250-11.el9)

**To Reproduce**
There exist invalid files in "/lib/modules/$(uname -r)/modules.dep", whose prefix directory is listed in config files under "/run/depmod.d/", "/etc/depmod.d/" or "/lib/depmod.d/".

For example, "weak-updates" is listed in "/etc/depmod.d/dist.conf", and "weak-updates/test.ko" actually does not exist but is recorded in "/lib/modules/`uname -r`/modules.dep". 
```
# tail -n 1 /etc/depmod.d/dist.conf
search updates extra built-in weak-updates
```

Reproduce as following:
```
# echo "weak-updates/test.ko:" >> /lib/modules/$(uname -r)/modules.dep
# dracut -f ./test.img
realpath: weak-updates/test.ko: No such file or directory
dracut: installkernel failed in module kernel-modules-extra
```

**Expected behavior**
dracut gives realpath warning and aborts generating initramfs image.

**Additional context**
It is more appreciate that dracut just gives a warning message rather than aborting the process.

LaszloGombos:
@jiayi0118 If you run "depmod" before running dracut, would the issue go away ?

jiayi0118:
Yes it is, but some developers may forget to refresh it.

devkontrol:
Can you please re-file at https://github.com/dracut-ng/dracut/issues if still relevant ?

PR #2049, opened 2022-11-10
fix(kernel-modules-extra): give warning messages rather than aborting

jiayi0118:
When installing module "90kernel-modules-extra", if non-existing files are found in modules.dep, give warning message and skip install these files, rather than aborting generating initramfs #1971

This pull request changes...

## Changes
Do not abort dracut when invalid files are found in modules.dep, just gives warning messages.

## Checklist

### DIFF
--- a/modules.d/90kernel-modules-extra/module-setup.sh
+++ b/modules.d/90kernel-modules-extra/module-setup.sh
@@ -31,6 +31,7 @@ installkernel() {
     local mod_name="kernel-modules-extra"
     prinfo() { dinfo "  ${mod_name}: $*"; }
     prdebug() { ddebug "  ${mod_name}: $*"; }
+    prwarning() { dwarning "  ${mod_name}: $*"; }
 
     # Escape a string for usage as a part of extended regular expression.
     # $1 - string to escape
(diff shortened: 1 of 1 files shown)


johannbg:
What's the usecase for not exiting out if the user still needs regenerate the image after running depmod?

jiayi0118:
Sometimes due to package uninstallation, some .ko modules are removed but there are residual records in the `modules.dep`, etc. It is okay to just skipped these non-existing .ko modules.

But it may be not enough to simply skip the non-existing .ko modules, thus the warning message to rerun `depmod` is also raised to remind users of more choices.

johannbg:
@jiayi0118 why should we circumvent downstream packaging problem here upstream I mean why does not the package regenerate the module dependency list and initramfs image upon removal of said component?

jiayi0118:
@johannbg This is because `dracut` may be used in a longtime work flow, and it is a little expensive that `dracut` aborts at somewhere the problem can just be avoided. On the other hand, of course it is better the module dependency is regenerated, this is why this patch will give more instructive warning to remind of users.

johannbg:
@jiayi0118 we dont workaround downstream packaging issues here upstream and you have still not answered the question why the downstream package maintainer is not fixing his package to address this as what's technically preventing him from doing that also can you elaborate what a "longtime work flow" means?

jiayi0118:
Of course the downstream package maintainers should fix their problems when the warning is given by dracut.

"longtime work flow" means some scenes that consume long time such as system upgrade, e.g. If dracut is called at the end of the process and aborts, the whole process will be re-executed.

Anyway, from my opinion it is both reasonable that dracut whether or not abort at this problem. My major advice is that whether dracut can give a more instructive warning message to indicate users what the problem is and guide them how to fix it.

PR #2050, opened 2022-11-11
refactor(build): enable isolated compilation for each binary tool

jiayi0118:
Encapsulate the specific compilation process into the individual binary tool directories for "dracut-install", "logtee", "skipcpio" and "dracut-util".

This pull request changes...

## Changes
During building process, enter specific binary tool directory and
run "make" individually.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,48 +49,72 @@ man8pages = man/dracut.8 \
 
 manpages = $(man1pages) $(man5pages) $(man7pages) $(man8pages)
 
-.PHONY: install clean archive rpm srpm testimage test all check AUTHORS CONTRIBUTORS doc dracut-version.sh
+.PHONY: install clean archive rpm srpm testimage test all check \
+		AUTHORS CONTRIBUTORS doc dracut-version.sh \
+		clean-dracut-install clean-logtee clean-skipcpio clean-dracut-util \
+		skipcpio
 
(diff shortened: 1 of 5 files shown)


dtardon:
Please don't. Recursive make brings no advantages, only problems. If you really want to fiddle with the build system, it would make more sense to port it to something modern, like meson.

jiayi0118:
> 

My motivation is to let the build process of each binary clearer and easier for understanding and managing. But your suggestion is right, it makes more sense to shift to modern build tools directly.

dtardon:
> My motivation is to let the build process of each binary clearer and easier for understanding and managing.

IOW, you're trying to solve a problem we don't have... I mean, the whole codebase consists of 13 source files, from which 3 binaries are built. The part of Makefile that is related to C code compilation fits on a screen...

jiayi0118:
> > My motivation is to let the build process of each binary clearer and easier for understanding and managing.
> 
> IOW, you're trying to solve a problem we don't have... I mean, the whole codebase consists of 13 source files, from which 3 binaries are built. The part of Makefile that is related to C code compilation fits on a screen...

This is a matter of personal preference hahah~

In addition, perhaps in the future dracut will introduce more C or Rust code, it seams worthwhile to design a more elegant build framework.

PR #2053, opened 2022-11-11
perf(dracut.sh): do not mkdir $initdir/lib/dracut within a loop

aafeijoo-suse:
This directory is always needed to store files like `modules.txt` and `build-parameter.txt`, so always create it, but only in one place.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2010,10 +2010,11 @@ else
     done
 fi
 
+mkdir -p "${initdir}"/lib/dracut
+
 if [[ $kernel_only != yes ]]; then
     mkdir -p "${initdir}/etc/cmdline.d"
-    # shellcheck disable=SC2174
-    mkdir -m 0755 -p "${initdir}"/lib "${initdir}"/lib/dracut "${initdir}"/lib/dracut/hooks
+    mkdir -m 0755 "${initdir}"/lib/dracut/hooks
     # shellcheck disable=SC2154
     for _d in $hookdirs; do
         # shellcheck disable=SC2174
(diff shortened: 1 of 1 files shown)


dracut.sh
```diff
@@ -2010,10 +2010,13 @@ else
     done
 fi
 
+# shellcheck disable=SC2174
+mkdir -m 0755 -p "${initdir}"/lib "${initdir}"/lib/dracut
+
 if [[ $kernel_only != yes ]]; then
     mkdir -p "${initdir}/etc/cmdline.d"
     # shellcheck disable=SC2174
-    mkdir -m 0755 -p "${initdir}"/lib "${initdir}"/lib/dracut "${initdir}"/lib/dracut/hooks
```
LaszloGombos, line 2016: Why move it out from here (and not just from within the loop) ? 

> This directory is always needed 

Even if $kernel_only is set to yes  ?

aafeijoo-suse: > Why move it out from here (and not just from within the loop) ?

It's the same directory that is trying to be created inside the loop, necessary at an earlier point in the code.

> > This directory is always needed
> 
> Even if $kernel_only is set to yes ?

Yes. At least to save the [list of modules included](https://github.com/dracutdevs/dracut/blob/69e9fc84e2920b080c8726a21f9d770beed2b8f6/lsinitrd.sh#L188) and the [build arguments](https://github.com/dracutdevs/dracut/blob/69e9fc84e2920b080c8726a21f9d770beed2b8f6/lsinitrd.sh#L377).

dracut.sh
```diff
@@ -2010,10 +2010,13 @@ else
     done
 fi
 
+# shellcheck disable=SC2174
```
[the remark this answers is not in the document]
aafeijoo-suse: Thanks, I realized that there is no need for `-m` here, because the intent of the original code is to set permissions only to `/lib/dracut/hooks`.

johannbg (approved):

PR #2055, opened 2022-11-12
test(TEST-99-RPM): remove test

LaszloGombos:
Is there any interest to maintain this test ?


Conan-Kudo:
Probably would want to check with @pvalena and @lnykryn. I believe they rely on this test.

LaszloGombos:
PR that disabled this test to run as part of integration test - https://github.com/dracutdevs/dracut/pull/1061 on Feb-2021..has not been run on CI since than..

johannbg:
We are dropping package files so the answer to that would be nope but at the same time the argument can be made it can be dropped at the same time the spec file get dropped.

LaszloGombos:
The test case is failing as it is - https://github.com/LaszloGombos/dracut/actions/runs/3452741015/jobs/5762849349 

If nobody is interested maintaining the this case, than what is the value of it ?

johannbg:
@LaszloGombos zero value in it.

johannbg (approved):

aafeijoo-suse (approved):

PR #2057, opened 2022-11-12
ci: improve systemd labeling

LaszloGombos:
out of 27 systemd modules, only 4 was covered by the labeler.


johannbg:
Module labels should be named after their module name not some generic label so update the pr accordingly for the missing modules.

PR #2059, opened 2022-11-12
docs(README.md): add repology badges

LaszloGombos:
This is how it looks rendered - https://github.com/LaszloGombos/dracut/blob/badge_sdjk/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@ dracut
 
 dracut is an event driven initramfs infrastructure.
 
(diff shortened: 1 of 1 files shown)


johannbg:
That's a nack from me since repology badges are misleading

johannbg:
btw what are you hoping to gain from adding those badges, what's the benefit of having these badges?

LaszloGombos:
> version info is redundant
The version info is the latest packaged version and not the latest released version.  Packaged version can be behind latest released version.

> what's the benefit of having these badges?
 - To support the claim and quantify that dracut was designed to be a distribution independent framework and give some pointers to other people on how they can contribute and use this project.

PR #2067, opened 2022-11-17
RFC: feat(dracut): create sysusers only when the initrd is generated

### LINKED ISSUE #1830: Run systemd-sysusers when generating initrd
I was looking at https://github.com/dracutdevs/dracut/pull/1658 where we removed the Condition, so the sysusers are run every time in the initrd. I don't think this is necessary since we always take the same sysusers drop-ins and create the same users every time.
I think it should be safe to call systemd-sysusers while generating the initrd. We will move the module to the end of the queue and use systemd-sysusers --root.
This should result in smaller initrd (no need for sysusers binary in the initrd) and faster boot (no need to create the users every time).

I am willing to write a patch for this, but first, I want to check what others think.

aafeijoo-suse:
When generating the initrd, dracut copies all the sysusers dropins provided by its modules to it, and systemd-sysusers runs at every boot to always create the same users.

This patch is intended to create the system users only when the initrd is generated, without including the systemd-sysusers stuff in the initrd, as it is not necessary.

To control this functionality and allow to use the old method, added the `create_sysusers` configuration option (default to yes) and the `--no-create-sysusers` parameter.

Trying to move forward @lnykryn's idea as it seems to be stalled. RFC:
- How to toggle the current functionality, via kernel command line or configuration? using a configuration option, all the sysusers stuff can be removed from the initrd, whereas adding `ConditionKernelCommandLine` to the service, we have to keep it, although it will not run on every boot, but it can be useful if the user wants to run systemd-sysusers from the emergency shell.
- By default, should we enable the new functionality or keep the old one?

## Checklist

Fixes #1830

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -274,8 +274,10 @@ Creates initial ramdisk images for preloading modules
   --regenerate-all      Regenerate all initramfs images at the default location
                          for the kernel versions found on the system.
   -p, --parallel        Use parallel processing if possible (currently only
-                        supported --regenerate-all)
-                        images simultaneously.
+                         supported --regenerate-all)
+                         images simultaneously.
+  --no-create-sysusers  Create system users on every boot, instead of creating
+                         them only when the initramfs is generated.
   --version             Display version.
 
 If [LIST] has multiple arguments, then you have to put these in quotes.
@@ -466,6 +468,7 @@ rearrange_params() {
             --long hostonly-i18n \
             --long hostonly-nics: \
             --long no-machineid \
+            --long no-create-sysusers \
             --long version \
             -- "$@"
     )
@@ -842,6 +845,9 @@ while :; do
         --no-machineid)
             machine_id_l="no"
             ;;
+        --no-create-sysusers)
+            create_sysusers_l="no"
+            ;;
         --version)
             long_version
             exit 1
@@ -1077,6 +1083,8 @@ stdloglvl=$((stdloglvl + verbosity_mod_l))
 [[ $uefi_splash_image_l ]] && uefi_splash_image="$uefi_splash_image_l"
 [[ $kernel_image_l ]] && kernel_image="$kernel_image_l"
 [[ $machine_id_l ]] && machine_id="$machine_id_l"
+[[ $create_sysusers_l ]] && create_sysusers="$create_sysusers_l"
+[[ $create_sysusers ]] || create_sysusers=yes
 
 if ! [[ $outfile ]]; then
     if [[ $machine_id != "no" ]]; then
@@ -1927,7 +1935,7 @@ export initdir dracutbasedir \
     systemdportable systemdportableconfdir systemdsystemunitdir \
     systemdsystemconfdir systemduser systemduserconfdir \
     hostonly_cmdline loginstall tmpfilesdir tmpfilesconfdir depmodd \
-    depmodconfdir
+    depmodconfdir create_sysusers
 
 mods_to_load=""
 # check all our modules to see if they should be sourced.
@@ -2037,6 +2045,8 @@ for moddir in "$dracutbasedir/modules.d"/[0-9][0-9]*; do
     [[ $mods_to_load == *\ $_d_mod\ * ]] || continue
     if [[ $show_modules == yes ]]; then
         printf "%s\n" "$_d_mod"
+    elif [[ $create_sysusers == yes && $_d_mod == "systemd-sysusers" ]]; then
+        continue
     else
         dinfo "*** Including module: $_d_mod ***"
     fi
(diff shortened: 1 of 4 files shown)


johannbg:
Minor nitpick. You are missing a reference to where @lnykryn outlines his RFC/IDEA/SCOPE ( saves people from having to search themselves ) as well as we need something better than `--no-create-sysusers` if merged.

aafeijoo-suse:
> as we need something better than `--no-create-sysusers` if merged.

Any ideas to improve this name?

johannbg:
@aafeijoo-suse I got nothing I was hoping @LaszloGombos would chime in something better but no-$foo is something we should avoid doing.

LaszloGombos:
> To control this functionality and allow to use the old method,

In theory this is a usually a good way to move forward, but in this particular case how important is to keep the "old method" conveniently switchable ? 

If somebody really needs to old method, can they do that with existing additional command line options (e.g. "--install systemd-sysusers") ? Perhaps we just need to document that instead of carrier one more command line option ?

aafeijoo-suse:
> > To control this functionality and allow to use the old method,
> 
> In theory this is a usually a good way to move forward, but in this particular case how important is to keep the "old method" conveniently switchable ?

I don't know a specific use case, maybe a user wants to keep an alternate initrd with the systemd-sysusers stuff inside.

> If somebody really needs to old method, can they do that with existing additional command line options (e.g. "--install systemd-sysusers") ? Perhaps we just need to document that instead of carrier one more command line option ?

No, this does not work this way. If the systemd-users module is added to the initrd (because it's added with `-a` or as a dependency by other module that needs to create sysusers), with this patch the system users will be created when the initrd is generated (modifying initrd's /etc/passwd and /etc/group) and the module (and its binaries/services) won't be added to the initrd, because in principle they are not needed. That's why we need a switch to keep the old way of working.

dracut.sh
```diff
@@ -2166,6 +2176,14 @@ if [[ $kernel_only != yes ]]; then
         fi
     fi
 
+    if [[ $create_sysusers == yes ]] && type -P systemd-sysusers > /dev/null 2>&1; then
```
LaszloGombos, line 2179: @aafeijoo-suse 

Just an idea: Can you check if "systemd-sysusers" dracut module is NOT included in initrd (but still present in host) instead of a new command line option. I still think that introducing a new command line can be avoided.

aafeijoo-suse: Sorry, I don't understand. `systemd-sysusers --root="$initdir"` will only create sysusers in the initrd if some module puts sysusers conf in the initrd, but without a new command line option the old method of creating sysusers at every boot would be dropped.

LaszloGombos: > without a new command line option the old method of creating sysusers at every boot would be dropped.

Just trying to understand so that I can review. I do not see this PR changing the 01systemd-sysusers module files. Does that mean that systemd-sysusers (if included in the generated initramfs) will still copy systemd-sysusers binary into initramfs but systemd-sysusers will never be called at boot ?

aafeijoo-suse: > Just trying to understand so that I can review. I do not see this PR changing the 01systemd-sysusers module files. Does that mean that systemd-sysusers binary (_**do you mean module?**_) (if included in the generated initramfs) will still copy systemd-sysusers binary into initramfs but systemd-sysusers will never be called at boot ?

There is no need to change the systemd-sysusers module if we don't use the kernel command line approach.
```bash
    elif [[ $create_sysusers == yes && $_d_mod == "systemd-sysusers" ]]; then
        continue
```
If the systemd-sysusers module is included in the list of modules to be loaded (`$mods_to_load`) and the `$create_sysusers` switch is set to `yes` (so, create sysusers at generation time), continue to process the next module, so the `install()` method of the systemd-sysusers module is skipped. On the other hand, if the `$create_sysusers` switch is set to `no`, the `install()` method of the systemd-sysusers module will be executed.

LaszloGombos: Thanks for the explanation, it is helpful. I still feel that a new dracut command line option for this feature is a significant maitenance burdon for very little benefit. I do not think maintaining the "old way" is necessary.

A better solution would be to somehow implement this feature in the systemd-sysusers dracut module and not in dracut itself. 

It is a confusing concept for me (and maybe also for users) that a module is included but never actually installed based on some dracut command line option.

aafeijoo-suse: Thanks for you review. That's why this is an RFC, we need to find the best approach to implement this.

Another option to keep the systemd-sysusers (at least listed as installed) is moving the part of my previous comment to it. Something like:

```diff
diff --git a/modules.d/01systemd-sysusers/module-setup.sh b/modules.d/01systemd-sysusers/module-setup.sh
index 4ec48dc6..488be0e9 100755
--- a/modules.d/01systemd-sysusers/module-setup.sh
+++ b/modules.d/01systemd-sysusers/module-setup.sh
@@ -24,6 +24,10 @@ depends() {
 # Install the required file(s) and directories for the module in the initramfs.
 install() {
 
+    # Do not install systemd-sysusers stuff in the initramfs if dracut creates
+    # the system users only when it's generated.
+    [[ $create_sysusers == yes ]] && return 0
+
     inst_simple "$moddir/sysusers-dracut.conf" "$systemdsystemunitdir/systemd-sysusers.service.d/sysusers-dracut.conf"
 
     inst_multiple -o \
```

This would keep both ways of working. To remove the old method, we just have to remove the `install()` method from systemd-sysusers.

I'm not sure whether we should not keep the "old way", maybe the systemd-sysusers infrastructure might be needed in the initrd in some use case.

LaszloGombos: We can always reintroduce the "old way" if we understand better why it is needed, but why to maintain it just speculatively ? 

One option (not great) to maintain the "old way" is to have 2 modules - one for the "old way" and one for the "new way".

LaszloGombos: One more thoughts on this: The "old way" is really just installing some files into initramfs, which can be already archived by existing command line options (-I and -i) without any additional explicit dracut module. Perhaps this makes you more comfortable not maintaining the "old" way any longer.

PR #2075, opened 2022-11-21
feat(dracut-install) Add fw_devlink suppliers as module dependencies

athierry1:
## Changes

Dracut currently finds kernel module dependencies using two methods: depmod and softdeps. However, these are often insufficient on embedded systems where a lot of driver dependencies are only described in the device tree. This is often the case for low-level devices such as clocks, regulators, pinctrls, etc.

This patch series allows dracut to find those device tree dependencies, by parsing the sysfs supplier nodes populated by fw_devlink.

This slows dracut down a little bit, however. Depending on whether dracut is called with `--hostonly` or not, I measured 25-35% slowdown on the Raspberry Pi 4. According to gperftools, most of the extra time is spent in `kmod_module_new_from_lookup()`

v2. Addressed review comments
v3.
- walk the sysfs once
- make sure that if one of the suppliers in a supplier chain is not a module (ie. is built-in), the suppliers it depends on are still processed

v4. Addressed review comments

## Checklist

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -97,6 +97,12 @@ static bool arg_mod_filter_noname = false;
 static int dracut_install(const char *src, const char *dst, bool isdir, bool resolvedeps, bool hashdst);
 static int install_dependent_modules(struct kmod_list *modlist);
 
+static void item_free(char *i)
+{
+        assert(i);
+        free(i);
+}
+
 static inline void kmod_module_unrefp(struct kmod_module **p)
 {
         if (*p)
@@ -145,6 +151,18 @@ static inline void fts_closep(FTS **p)
 #define _cleanup_fts_close_ _cleanup_(fts_closep)
 
 #define _cleanup_globfree_ _cleanup_(globfree)
+
+static inline void destroy_hashmap(Hashmap **hashmap)
+{
+        void *i = NULL;
+
+        while ((i = hashmap_steal_first(*hashmap)))
+                item_free(i);
+
+        hashmap_free(*hashmap);
+}
+
+#define _cleanup_destroy_hashmap_ _cleanup_(destroy_hashmap)
 
 static size_t dir_len(char const *file)
 {
@@ -946,12 +964,6 @@ static int dracut_install(const char *orig_src, const char *orig_dst, bool isdir
         log_debug("dracut_install ret = %d", ret);
 
         return ret;
-}
-
-static void item_free(char *i)
-{
-        assert(i);
-        free(i);
 }
 
 static void usage(int status)
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -83,6 +83,7 @@ FILE *logfile_f = NULL;
 static Hashmap *items = NULL;
 static Hashmap *items_failed = NULL;
 static Hashmap *modules_loaded = NULL;
+static Hashmap *modules_suppliers = NULL;
 static regex_t mod_filter_path;
 static regex_t mod_filter_nopath;
 static regex_t mod_filter_symbol;
@@ -95,7 +96,7 @@ static bool arg_mod_filter_nosymbol = false;
 static bool arg_mod_filter_noname = false;
 
 static int dracut_install(const char *src, const char *dst, bool isdir, bool resolvedeps, bool hashdst);
-static int install_dependent_modules(struct kmod_list *modlist);
+static int install_dependent_modules(struct kmod_ctx *ctx, struct kmod_list *modlist, Hashmap *suppliers_paths);
 
 static void item_free(char *i)
 {
@@ -1515,7 +1516,111 @@ static bool check_module_path(const char *path)
         return true;
 }
 
-static int install_dependent_module(struct kmod_module *mod, int *err)
+static int find_kmod_module_from_sysfs_node(struct kmod_ctx *ctx, const char *sysfs_node, int sysfs_node_len,
+                                            struct kmod_list **modules)
+{
+        char modalias_path[PATH_MAX];
+        if (snprintf(modalias_path, sizeof(modalias_path), "%.*s/modalias", sysfs_node_len,
+                     sysfs_node) >= sizeof(modalias_path))
+                return -1;
+
+        _cleanup_close_ int modalias_file = -1;
+        if ((modalias_file = open(modalias_path, O_RDONLY | O_CLOEXEC)) == -1)
+                return 0;
+
+        char alias[page_size()];
+        ssize_t len = read(modalias_file, alias, sizeof(alias));
+        alias[len - 1] = '\0';
+
+        return kmod_module_new_from_lookup(ctx, alias, modules);
+}
+
+static int find_modules_from_sysfs_node(struct kmod_ctx *ctx, const char *sysfs_node, Hashmap *modules)
+{
+        _cleanup_kmod_module_unref_list_ struct kmod_list *list = NULL;
+        struct kmod_list *l = NULL;
+
+        if (find_kmod_module_from_sysfs_node(ctx, sysfs_node, strlen(sysfs_node), &list) >= 0) {
+                kmod_list_foreach(l, list) {
+                        struct kmod_module *mod = kmod_module_get_module(l);
+                        char *module = strdup(kmod_module_get_name(mod));
+                        kmod_module_unref(mod);
+
+                        if (hashmap_put(modules, module, module) < 0)
+                                free(module);
+                }
+        }
+
+        return 0;
+}
+
+static void find_suppliers_for_sys_node(struct kmod_ctx *ctx, Hashmap *suppliers, const char *node_path_raw,
+                                        size_t node_path_len)
+{
+        char node_path[PATH_MAX];
+        char real_path[PATH_MAX];
+
+        memcpy(node_path, node_path_raw, node_path_len);
+        node_path[node_path_len] = '\0';
+
+        DIR *d;
+        struct dirent *dir;
+        while (realpath(node_path, real_path) != NULL && strcmp(real_path, "/sys/devices")) {
+                d = opendir(node_path);
+                if (d) {
+                        size_t real_path_len = strlen(real_path);
+                        while ((dir = readdir(d)) != NULL) {
+                                if (strstr(dir->d_name, "supplier:platform") != NULL) {
+                                        if (snprintf(real_path + real_path_len, sizeof(real_path) - real_path_len, "/%s/supplier",
+                                                     dir->d_name) < sizeof(real_path) - real_path_len) {
+                                                char *real_supplier_path = realpath(real_path, NULL);
+                                                if (real_supplier_path != NULL)
+                                                        if (hashmap_put(suppliers, real_supplier_path, real_supplier_path) < 0)
+                                                                free(real_supplier_path);
+                                        }
+                                }
+                        }
+                        closedir(d);
+                }
+                strncat(node_path, "/..", 3); // Also find suppliers of parents
+        }
+}
+
+static void find_suppliers(struct kmod_ctx *ctx)
+{
+        _cleanup_fts_close_ FTS *fts;
+        char *paths[] = { "/sys/devices/platform", NULL };
+        fts = fts_open(paths, FTS_NOSTAT | FTS_PHYSICAL, NULL);
+
+        for (FTSENT *ftsent = fts_read(fts); ftsent != NULL; ftsent = fts_read(fts)) {
+                if (strcmp(ftsent->fts_name, "modalias") == 0) {
+                        _cleanup_kmod_module_unref_list_ struct kmod_list *list = NULL;
+                        struct kmod_list *l;
+
+                        if (find_kmod_module_from_sysfs_node(ctx, ftsent->fts_parent->fts_path, ftsent->fts_parent->fts_pathlen, &list) < 0)
+                                continue;
+
+                        kmod_list_foreach(l, list) {
+                                _cleanup_kmod_module_unref_ struct kmod_module *mod = kmod_module_get_module(l);
+                                const char *name = kmod_module_get_name(mod);
+                                Hashmap *suppliers = hashmap_get(modules_suppliers, name);
+                                if (suppliers == NULL) {
+                                        suppliers = hashmap_new(string_hash_func, string_compare_func);
+                                        hashmap_put(modules_suppliers, strdup(name), suppliers);
+                                }
+
+                                find_suppliers_for_sys_node(ctx, suppliers, ftsent->fts_parent->fts_path, ftsent->fts_parent->fts_pathlen);
+                        }
+                }
+        }
+}
+
+static Hashmap *find_suppliers_paths_for_module(const char *module)
+{
+        return hashmap_get(modules_suppliers, module);
+}
+
+static int install_dependent_module(struct kmod_ctx *ctx, struct kmod_module *mod, Hashmap *suppliers_paths, int *err)
 {
         const char *path = NULL;
         const char *name = NULL;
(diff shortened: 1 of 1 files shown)


LaszloGombos:
CC @nabijaczleweli to help out with the review

LaszloGombos:
@athierry1 Can you you help us and think of a why how this change can be tested ? Not asking to write the test case yourself, but maybe think of an environment or command line for testing ? Is this change makes a difference on x64 or only on arm ?

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,105 @@ static bool check_module_path(const char *path)
...
+                if (strcmp(ftsent->fts_name, "modalias") == 0) {
+                        char sysfs_node[PATH_MAX] = {0};
+                        memcpy(sysfs_node, ftsent->fts_parent->fts_path, ftsent->fts_parent->fts_pathlen);
+                        sysfs_node[ftsent->fts_parent->fts_pathlen] = '\0';
+
+                        _cleanup_kmod_module_unref_list_ struct kmod_list *list = NULL;
+                        struct kmod_list *l = NULL;
+
+                        if (find_module_from_sysfs_node(ctx, sysfs_node, &list) < 0)
+                                continue;
```
nabijaczleweli, line 1591: you keep taking FTS paths, copying them to slice them, then copying them *again* to unslice them.

you can avoid the clumsy copy here and below by getting rid of `sysfs_node` entirely and having
```c
if (find_module_from_sysfs_node(ctx, ftsent->fts_parent->fts_path, ftsent->fts_parent->fts_pathlen, &list) < 0)
```
coupled with
```c
static int find_module_from_sysfs_node(struct kmod_ctx *ctx, const char *sysfs_node, int sysfs_node_len, struct kmod_list **modules)
{
        char modalias_path[PATH_MAX] = {0};
        snprintf(modalias_path, sizeof(modalias_path), "%.*s/modalias", sysfs_node_len, sysfs_node);
        _cleanup_fclose_ FILE *modalias_file = NULL;
```
and similarly `find_suppliers_for_sys_node()` as
```c
find_suppliers_for_sys_node(ctx, suppliers, ftsent->fts_parent->fts_path, ftsent->fts_parent->fts_pathlen);
```
and
```c
find_suppliers_for_sys_node(ctx, suppliers, parent->fts_path, parent->fts_pathlen);
```
and have *that* do
```c
static void find_suppliers_for_sys_node(struct kmod_ctx *ctx, Hashmap *suppliers, const char *node_path_raw, size_t node_path_len)
{
        char node_path[PATH_MAX] = {};
        memcpy(node_path, node_path_raw, node_path_len);
        DIR *d = opendir(node_path);
        struct dirent *dir;
```

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,105 @@ static bool check_module_path(const char *path)
...
+                                if (strcmp(name, module) == 0) {
+                                        find_suppliers_for_sys_node(ctx, suppliers, sysfs_node);
+
+                                        // Also add suppliers of parent nodes
+                                        FTSENT *parent = ftsent->fts_parent;
+                                        while(parent->fts_level >= 1) {
+                                                memcpy(sysfs_node, parent->fts_path, parent->fts_pathlen);
+                                                sysfs_node[parent->fts_pathlen] = '\0';
+
+                                                find_suppliers_for_sys_node(ctx, suppliers, sysfs_node);
```
nabijaczleweli, line 1606: AFAICT this processes `ftsent->fts_parent` twice (once raw, aliased as sysfs_node, the other time in the first loop iteration)?

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,105 @@ static bool check_module_path(const char *path)
...
+                        kmod_list_foreach(l, list) {
+                                _cleanup_kmod_module_unref_ struct kmod_module *mod = kmod_module_get_module(l);
+                                const char *name = kmod_module_get_name(mod);
+
+                                if (strcmp(name, module) == 0) {
+                                        find_suppliers_for_sys_node(ctx, suppliers, sysfs_node);
+
+                                        // Also add suppliers of parent nodes
+                                        FTSENT *parent = ftsent->fts_parent;
+                                        while(parent->fts_level >= 1) {
```
nabijaczleweli, line 1602: this obfuscates the for loop which it is, probably better to write it as
```c
for(FTSENT *parent = ftsent->fts_parent; parent->fts_level >= 1; parent = parent->fts_parent)
    find_suppliers_for_sys_node(ctx, suppliers, parent->fts_path, parent->fts_pathlen);
```

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,105 @@ static bool check_module_path(const char *path)
...
+
+                        kmod_list_foreach(l, list) {
+                                _cleanup_kmod_module_unref_ struct kmod_module *mod = kmod_module_get_module(l);
+                                const char *name = kmod_module_get_name(mod);
+
+                                if (strcmp(name, module) == 0) {
+                                        find_suppliers_for_sys_node(ctx, suppliers, sysfs_node);
+
+                                        // Also add suppliers of parent nodes
+                                        FTSENT *parent = ftsent->fts_parent;
```
nabijaczleweli, line 1601: on my amd64 machine the outer-most loop yields:
```
/sys/devices/platform/PNP0C14:00/wmi_bus/wmi_bus-PNP0C14:00/0E7AF9F2-44A1-4C6F-A4B0-A7678480DA61
/sys/devices/platform/PNP0C14:00
/sys/devices/platform/reg-dummy
/sys/devices/platform/coretemp.1
/sys/devices/platform/dcdbas
/sys/devices/platform/platform-framebuffer.0
/sys/devices/platform/microcode
/sys/devices/platform/GHES.32994
/sys/devices/platform/dmi-ipmi-si.0
/sys/devices/platform/GHES.32992
/sys/devices/platform/GHES.49377
/sys/devices/platform/coretemp.0
/sys/devices/platform/ipmi_bmc.0
/sys/devices/platform/PNP0103:00
/sys/devices/platform/serial8250
/sys/devices/platform/i8042/serio0
/sys/devices/platform/i8042/serio1
/sys/devices/platform/i8042
/sys/devices/platform/GHES.32993
/sys/devices/platform/GHES.49378
/sys/devices/platform/GHES.227
/sys/devices/platform/pcspkr/input/input5
/sys/devices/platform/pcspkr
/sys/devices/platform/GHES.49376
```

Are you *sure* it's safe/sensible to process, for example, `/s/d/p/i8042` thrice? If not, mark `parent->fts_number = 1` on each iteration and make the loop condition `parent->level >= 1 && !parent->fts_number`.

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,105 @@ static bool check_module_path(const char *path)
...
+{
+        DIR *d;
+        struct dirent *dir;
+        d = opendir(node_path);
+        if (d) {
+                while ((dir = readdir(d)) != NULL) {
+                        if (strstr(dir->d_name, "supplier") != NULL) {
+                                char supplier_path[PATH_MAX] = {0};
+                                if (realpath(dir->d_name, supplier_path)) {
+                                        strncat(supplier_path, "/supplier", sizeof(supplier_path));
```
nabijaczleweli, line 1555: unchecked strncat

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,105 @@ static bool check_module_path(const char *path)
...
+
+                                        _cleanup_kmod_module_unref_list_ struct kmod_list *list = NULL;
+                                        struct kmod_list *l = NULL;
+
+                                        if (find_module_from_sysfs_node(ctx, supplier_path, &list) >= 0) {
+                                                kmod_list_foreach(l, list) {
+                                                        struct kmod_module *mod = kmod_module_get_module(l);
+                                                        char *supplier_module = strdup(kmod_module_get_name(mod));
+                                                        kmod_module_unref(mod);
+                                                        hashmap_put(suppliers, supplier_module, supplier_module);
```
nabijaczleweli, line 1565: this leaks `supplier_module` if `hashmap_put` errors

athierry1:
@LaszloGombos the change should only make a difference on systems that use a device tree (so I guess only ARM systems, maybe also RISC-V?) 
So to test it, you could emulate an environment that's using a device tree. For instance a Raspberry Pi 3 emulated with QEMU (QEMU has a `raspi3b` machine)

You could :
- boot a standard raspbian image on this VM
- find a dependency in the device tree that's going to be your testcase (for example `cam1_regulator` (`fixed` module) references `firmware:expgpio` (module `gpio-raspberrypi-exp`) in the Raspberry Pi 3 device tree, so pulling the `fixed` module with `dracut-install` should also pull the `gpio-raspberrypi-exp` module)
- make sure those two drivers are built as modules (you might need to build a new kernel for that, I think most regulators, clks, etc. are built-in in the raspbian kernel)
- call `dracut-install --kerneldir=<above kernel dir> fixed` and check if `gpio-raspberrypi-exp` is also pulled in the output folder

I could try to make it work and write a little demo script if you're interested

athierry1:
@nabijaczleweli thanks for your feedback, I'll try to submit a new version tomorrow

LaszloGombos:
> I could try to make it work and write a little demo script if you're interested

Lets continue this on a separate issue - https://github.com/dracutdevs/dracut/issues/2082

athierry1:
@nabijaczleweli I pushed a v2 addressing your feedback

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,114 @@ static bool check_module_path(const char *path)
...
+                                       struct kmod_list **modules)
+{
+        char modalias_path[PATH_MAX] = {0};
+        snprintf(modalias_path, sizeof(modalias_path), "%.*s/modalias", sysfs_node_len, sysfs_node);
+
+        _cleanup_close_ int modalias_file = -1;
+        if ((modalias_file = open(modalias_path, O_RDONLY | O_CLOEXEC)) == -1)
+                return 0;
+
+        char alias[getpagesize()];
```
nabijaczleweli, line 1528: any reason why not just `PAGE_SIZE`?

athierry1: AFAICT `PAGE_SIZE` is only declared in the kernel headers. I came across https://stackoverflow.com/questions/37897645/page-size-undeclared-c that suggested to use `getpagesize()` instead 

But I see there's also a `page_size()` function defined in `src/install/util.c` that calls `sysconf(_SC_PAGESIZE)` Maybe I should use that instead, what do you think?

nabijaczleweli: of course; POSIX requires `PAGESIZE`, glibc doesn't provide it. yeah, use that then.

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,114 @@ static bool check_module_path(const char *path)
...
+                                continue;
+
+                        kmod_list_foreach(l, list) {
+                                _cleanup_kmod_module_unref_ struct kmod_module *mod = kmod_module_get_module(l);
+                                const char *name = kmod_module_get_name(mod);
+
+                                if (strcmp(name, module) == 0) {
+                                        err = find_suppliers_for_sys_node(ctx, suppliers, sysfs_node->fts_path, sysfs_node->fts_pathlen);
+                                        if (err)
+                                                return err;
```
nabijaczleweli, line 1605: am I wrong to think this *always* overlaps with the first iteration of the loop below?

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,114 @@ static bool check_module_path(const char *path)
...
+{
+        char node_path[PATH_MAX] = {};
+        memcpy(node_path, node_path_raw, node_path_len);
+        DIR *d;
+        struct dirent *dir;
+        d = opendir(node_path);
+        if (d) {
+                while ((dir = readdir(d)) != NULL) {
+                        if (strstr(dir->d_name, "supplier") != NULL) {
+                                char supplier_path[PATH_MAX] = {0};
```
nabijaczleweli, line 1546: PATH_MAX is quite big to throw it around like this, you use the same buffer for node_path and supplier_path here.

src/install/dracut-install.c
```diff
@@ -1515,7 +1515,114 @@ static bool check_module_path(const char *path)
...
+                                                kmod_list_foreach(l, list) {
+                                                        struct kmod_module *mod = kmod_module_get_module(l);
+                                                        char *supplier_module = strdup(kmod_module_get_name(mod));
+                                                        kmod_module_unref(mod);
+
+                                                        int err = hashmap_put(suppliers, supplier_module, supplier_module);
+                                                        if (err < 0) {
+                                                                free(supplier_module);
+                                                                return err;
+                                                        }
```
nabijaczleweli, line 1568: I don't think you need to `return err` here – hashmap_put() returns -EEXIST if `supplier_module` exists in the map already (the other error is ENOMEM but dracut-install doesn't care about allocations failing anyway) so just reduce to `if (hashmap_put()) free(supplier_module);` or equivalent and it's fine.

src/install/dracut-install.c
```diff
@@ -1543,17 +1650,21 @@ static int install_dependent_module(struct kmod_module *mod, int *err)
                 _cleanup_kmod_module_unref_list_ struct kmod_list *modlist = NULL;
                 _cleanup_kmod_module_unref_list_ struct kmod_list *modpre = NULL;
                 _cleanup_kmod_module_unref_list_ struct kmod_list *modpost = NULL;
+                _cleanup_destroy_hashmap_ Hashmap *suppliers = hashmap_new(string_hash_func, string_compare_func);
                 log_debug("dracut_install '%s' '%s' OK", path, &path[kerneldirlen]);
                 install_firmware(mod);
-                modlist = kmod_module_get_dependencies(mod);
-                *err = install_dependent_modules(modlist);
+                *err = find_suppliers_for_module(ctx, suppliers, name);
```
nabijaczleweli, line 1656: if this really is *so much slower* (and that's a fair cop, you're walking a considerable hunk of sysfs for *every* module, AIUI), have you tried walking sysfs *once* and collecting all the module dependencies, then just consulting the resulting map here?

athierry1: Thanks for the suggestion, I just tried it and it seems to speed things up considerably! I'll clean it up and send a new version that walks the sysfs only once.

I also found a bug in my code: if one of the suppliers in a supplier chain is not a module (ie. is built-in), the suppliers it depends on are not processed. So I'll fix that in the next version as well.

athierry1:
Pushed a v3 that:
- walks the sysfs once
- makes sure that if one of the suppliers in a supplier chain is not a module (ie. is built-in), the suppliers it depends on are still processed

dracut-install is now a lot faster:
Executing `dracut-install -D ~/test -H --kerneldir /lib/modules/<my kernel> -o -m =drivers` on my RPi4 takes :
- 22s on the v2
- 1.26s on the v3

However, I didn't notice a speedup overall when calling dracut. It seems dracut calls dracut-install a lot of times, so the sysfs is still being walked a lot because the hashmap is recreated for each call.

athierry1:
@nabijaczleweli could you please review the latest changes when you have the chance? Thank you!

src/install/dracut-install.c
```diff
@@ -1515,7 +1516,110 @@ static bool check_module_path(const char *path)
...
+        DIR *d;
+        struct dirent *dir;
+        while (realpath(node_path, real_path) != NULL && strcmp(real_path, "/sys/devices")) {
+                d = opendir(node_path);
+                if (d) {
+                        while ((dir = readdir(d)) != NULL) {
+                                if (strstr(dir->d_name, "supplier:platform") != NULL) {
+                                        char supplier_path[PATH_MAX];
+                                        if (snprintf(supplier_path, sizeof(supplier_path), "%s/%s/supplier", real_path, dir->d_name) < sizeof(supplier_path)) {
+                                                char *real_supplier_path = realpath(supplier_path, NULL);
```
nabijaczleweli, line 1575: I've seen it when you posted it and the general approach appears sound in general. If it works it works.

You could just append to `real_path` here to not throw around a *third* PATH_MAX buffer, tho.

athierry1: Thanks for reviewing!
I could be missing something, but `real_path` needs to be unchanged across iterations of the `while ((dir = readdir(d)) != NULL)` loop, so I think there is no way around having a third PATH_MAX buffer.

nabijaczleweli: You could `size_t real_path_len = strlen(real_path)` after the `if(d)` and `snprintf(real_path + real_path_len, real_path - real_path_len, "/%s/supplier", dir->d_name)` I think?

athierry1: Oh this makes sense, thanks! I pushed a new version with this change.

athierry1:
Hi @johannbg, did you get a chance to look at this PR?

johannbg:
@athierry1 yup nothing stands out added it to the 059 milestone

athierry1:
That's great, thank you!

nabijaczleweli:
-2

johannbg:
@athierry1 mark conversation resolved when they have been resolved dont edit your PR message with "$foo. Addressed review comments" 

For those of us that are extremely time constrained we ignore reviewing and merged PR that still have open conversation on them from other reviewers since we make the assumption that issues being pointed out by other reviewers have not been addressed

johannbg (approved):
LGTM

PR #2080, opened 2022-11-23
fix(fips): do not blindly remove /boot

aafeijoo-suse:
The `mount_boot` method called from fips-noboot.sh in the pre-pivot hook blindly
executes `rm -rf /boot` if there is no `boot=` command line parameter, without
first checking that /boot is not already mounted by other means.

We hit this issue in certain specific circumstances (s390x arch, separate /boot
partition, initrd built with `--mount` to mount /boot and incorrect user
configuration without the `boot=` command line parameter), but this part of the
code must be protected anyway.

```
[   29.932607] dracut-pre-pivot[922]: rm: cannot remove '/boot/boot.readme': Read-only file system
[   29.932814] dracut-pre-pivot[922]: rm: cannot remove '/boot/System.map-5.14.21-150500.34-default': Read-only file system
[   29.932909] dracut-pre-pivot[922]: rm: cannot remove '/boot/initrd-5.14.21-150500.34-default': Read-only file system
[   29.933015] dracut-pre-pivot[922]: rm: cannot remove '/boot/image-5.14.21-150500.34-default': Read-only file system
...
[   29.960631] dracut-pre-pivot[922]: rm: cannot remove '/boot/grub2/grub.cfg': Read-only file system
[   29.961253] dracut-pre-pivot[922]: rm: cannot remove '/boot/grub2/fonts/unicode.pf2': Read-only file system
[   29.980315] dracut-pre-pivot[922]: rm: cannot remove '/boot/zipl': Read-only file system
[   29.982016] dracut-pre-pivot[923]: ln: failed to create symbolic link '/boot/boot': Read-only file system
[   29.987197] dracut-pre-pivot[912]: Checking integrity of kernel
[   30.356025] dracut-pre-pivot[912]: All initrd crypto checks done
```

## Checklist

### DIFF
--- a/modules.d/01fips/fips.sh
+++ b/modules.d/01fips/fips.sh
@@ -18,6 +18,15 @@ mount_boot() {
     boot=$(getarg boot=)
 
     if [ -n "$boot" ]; then
+        if [ -d /boot ] && ismounted /boot; then
+            boot_dev=
+            if command -v findmnt > /dev/null; then
+                boot_dev=$(findmnt -n -o SOURCE /boot)
+            fi
+            fips_info "Ignoring 'boot=$boot' as /boot is already mounted ${boot_dev:+"from '$boot_dev'"}"
+            return 0
+        fi
+
         case "$boot" in
             LABEL=* | UUID=* | PARTUUID=* | PARTLABEL=*)
                 boot="$(label_uuid_to_dev "$boot")"
@@ -47,10 +56,13 @@ mount_boot() {
         mkdir -p /boot
         fips_info "Mounting $boot as /boot"
         mount -oro "$boot" /boot || return 1
+        FIPS_MOUNTED_BOOT=1
     elif ! ismounted /boot && [ -d "$NEWROOT/boot" ]; then
         # shellcheck disable=SC2114
         rm -fr -- /boot
         ln -sf "$NEWROOT/boot" /boot
+    else
+        die "You have to specify boot=<boot device> as a boot option for fips=1"
     fi
 }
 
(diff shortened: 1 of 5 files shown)


modules.d/01fips/fips.sh
```diff
@@ -47,7 +47,7 @@ mount_boot() {
         mkdir -p /boot
         fips_info "Mounting $boot as /boot"
         mount -oro "$boot" /boot || return 1
-    elif [ -d "$NEWROOT/boot" ]; then
+    elif ! ismounted /boot && [ -d "$NEWROOT/boot" ]; then
```
LaszloGombos, line 50: What if /boot is not mounted as a separate partition but /boot directory exists as part of / (root) and it is not empty ?

LaszloGombos (approved):
LGTM

modules.d/01fips/fips.sh
LaszloGombos: I understand this is super unlikely, but what I mean to ask is what if /boot is in the initramfs (e.g. added by "dracut --include  /boot/my_file /boot/my_file_in_initramfs")

aafeijoo-suse:
Converting this to draft, because this patch only fixes one part of the problem we found with FIPS enabled in a s390 machine using a separate /boot partition.
In this case, the /boot partition is not originally mounted by dracut, but the pre-pivot hook unmounts it.
Possible fixes:
- fips.sh mounts the /boot device using a different/temporary directory.
- Reimplement the fips module to include the kernel hmac file in the initrd, so there is no need to mount /boot to check the integrity.

aafeijoo-suse:
> Converting this to draft, because this patch only fixes one part of the problem we found with FIPS enabled in a s390 machine using a separate /boot partition.
> In this case, the /boot partition is not originally mounted by dracut, but the pre-pivot hook unmounts it.

PR updated adding the patches from my colleague Raymund Will (I think he doesn't have a GitHub account), who is working on this issue with s390x.

johannbg (approved):

PR #2081, opened 2022-11-23
fix(network): for non-systemd based systems defaults to network-legacy

### LINKED ISSUE #1756: network module has transitive dependency on systemd

LaszloGombos:
## Checklist

Fixes #1756


LaszloGombos:
When network-legacy is no longer supported, we need to change the network module anyways, but until than this few lines of code expresses better what is supported currently.

johannbg:
@bengal is NM entirely dependent on dbus support or can it be used without it?

LaszloGombos:
The network-manager dracut module already explicitly depends on the dbus dracut module. If dbus dracut module can not be installed and network-manager dracut module is requested, initramfs generation would already fail.

https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/module-setup.sh#L13

This extra code is needed to make the network meta module do a better job picking defaults.

bengal:
Right, the NM module depends on D-Bus.

aafeijoo-suse (commented):
Sorry, but I don't understand this check. The `network-manager` module depends on the `dbus` module, so if the `network-manager` module is installed, it will automatically install the `dbus` module, therefore there is no need to check if it's already added to the list of modules that are going to be installed.
You may want want to check if dbus is installed on the running system, but NetworkManager requires it, so it must be and this check would not add anything.
Also, I don't see how it fixes #1756

aafeijoo-suse:
> > The network-manager module depends on the dbus module,
> 
> @aafeijoo-suse This is a change in network module not in network-manager module.
> 
> The network meta module is not that smart to know that network-manager could not be installed if dbus is not installed.
> 
> The goal is if dbus is not installed, than network meta module should pick network-legacy instead of network-manager.

I must be missing something here, the `dbus` dracut module is only included in the `$mods_to_load` or `$modules_loaded` array at the point you wrote the check if it was manually added to the initrd (`-a`,`--add`, `add_dracutmodules`) or any other dracut module adds it as a dependency (`check()` method).

aafeijoo-suse:
I see your point, but I think the checks in meta modules look for binaries/services/files on the running system, not if other dracut modules are included. So in this case, maybe the check should be `elif find_binary busctl && ...`, since `busctl` is required by the `dbus-daemon` and `dbus-broker` modules.

aafeijoo-suse:
> In theory Gentoo distro/packagers could implement an out-of-tree dbus module (dbus-without-systemd, see the systemd USE flag - https://packages.gentoo.org/packages/sys-apps/dbus) or patch the dbus-daemon module to eliminate systemd, which would actually enable Gentoo to support network-manager dracut module even on a non-systemd system.

Ok, now I understand.

The problem is this patch introduces a regression. See for example in openSUSE Leap 15.4 with NetworkManager:

Before this patch:

```
# dracut -f --stdlog 3 -a network test-before.img
# lsinitrd -m test-before.img | grep network
network-manager
network
kernel-network-modules
```

After this patch:
```
# dracut -f --stdlog 3 -a network test-after.img
# lsinitrd -m test-after.img | grep network
network-legacy
network
kernel-network-modules
```

That's because if no other module includes the dbus module, the condition is false and falls back to the legacy default.

johannbg:
> I updated the test to look for systemctl binary. I think this is a good compromise for now.

Last time I checked systemd support is build-time or runtime selectable in NM which means we cant test looking for the systemctl binary since it would fail on none systemd based distribution that use NM.

johannbg:
@LaszloGombos that just highlight that we need to create a meta module for network manager which then covers nm + systemd module and nm without systemd modules and nm with iwd ( or we have seperate iwd+nm module part of  iwd to handle that ) . 

Our modules are bound to be created based on the configuration option that are supported in the upstream that they are based upon and we need to rework our modules based on that requirement.

PR #2085, opened 2022-11-24
fix(systemd-coredump): correct systemd-coredump binary path

aafeijoo-suse:
This issue went unnoticed because the systemd module always installs the systemd-coredump binary using the right path.

https://github.com/dracutdevs/dracut/blob/fd9cd02cb21f25d32fec4cd9889457ea6361f45e/modules.d/00systemd/module-setup.sh#L35

## Checklist


LaszloGombos (approved):
LGTM

johannbg (approved):

PR #2088, opened 2022-11-25
ci: make NFS and NBD tests pass in Gentoo test containers

LaszloGombos:
This PR should enable the following tests to pass in Gentoo:
 - 20 (NFS)
 - 40 (NBD)
 - 50 (MULTINIC)

### DIFF
--- a/test/container/Dockerfile-Gentoo
+++ b/test/container/Dockerfile-Gentoo
@@ -20,12 +20,17 @@ RUN echo 'export DRACUT_NO_XATTR=1 KVERSION=$(cd /lib/modules; ls -1 | tail -1)'
 # Only install `dmsetup`: attempting to install all of lvm2 fails due to missing kernel headers.
 RUN echo 'sys-fs/lvm2 device-mapper-only -thin' > /etc/portage/package.use/lvm2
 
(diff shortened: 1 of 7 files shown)


johannbg (approved):

test/container/Dockerfile-Gentoo
```diff
@@ -20,12 +20,17 @@ RUN echo 'export DRACUT_NO_XATTR=1 KVERSION=$(cd /lib/modules; ls -1 | tail -1)'
 # Only install `dmsetup`: attempting to install all of lvm2 fails due to missing kernel headers.
 RUN echo 'sys-fs/lvm2 device-mapper-only -thin' > /etc/portage/package.use/lvm2
 
+RUN echo 'net-fs/nfs-utils -nfsv4' > /etc/portage/package.use/nfs-utils
```
aafeijoo-suse, line 23: Not being a Gentoo expert, why do you disable NFSv4? Maybe you can prepend a comment with the reasoning.

`TEST-20-NFS` tests v3 and v4, so it will fail:

https://github.com/dracutdevs/dracut/blob/037eda7ce48e60e1b14b31125616f7274dfd15cc/test/TEST-20-NFS/test.sh#L220-L221

aafeijoo-suse (approved):
Thanks for the background info.

PR #2089, opened 2022-11-25
ci: prepare to test more than two network modules

LaszloGombos:
The main motivation of this change is to prepare to enable testing all four networking dracut modules not just two.

There are five networking modules supported currently. 

## Checklist


LaszloGombos:
As a side effect, it is also much easier to read which tests are networking and which are local.
Screen Shot 2022-11-29 at 7 58 08 PM

johannbg (approved):
<sigh> This will grow complicated quite fast since many of the network solution can either be used standalone or in conjunction with one another so there will need to be alot of test cases if we intend to support all these network solution out there.

PR #2090, opened 2022-11-27
ci: remove native dracut installation from the test containers

LaszloGombos:
The installed packaged dracut version from distributions can interfere with testing latest dracut from source, which is the primary goal of these containers.

I was not able to uninstall the full dracut package in the openSUSE container, as sadly would also uninstall the mdadm package. Fedora does not have this dependency.

### DIFF
--- a/test/container/Dockerfile-Fedora-latest
+++ b/test/container/Dockerfile-Fedora-latest
@@ -51,7 +51,7 @@ RUN dnf -y install --setopt=install_weak_deps=False \
     shfmt \
     parted \
     ntfsprogs \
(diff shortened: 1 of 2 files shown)


johannbg (approved):

LaszloGombos:
> The mdadm dependency has been removed since Tumbleweed snapshot 20221203, so it should now work without the workaround. Could you try it?

@aafeijoo-suse Nice ! Tested and seem to work well. Updated the PR.

aafeijoo-suse (approved):
Thanks!

PR #2092, opened 2022-11-27
fix(load_fstype): avoid false positive searchs

FGrose:
Delimit the test strings to avoid substring matches.

### Checklist


LaszloGombos:
@FGrose Thanks for the PR.  I am thinking if there is a way to test for this. Do you have a practical example where this PR is fixing a breakage or this is more of a theoretical correction ?

FGrose:
@LaszloGombos 
>  Do you have a practical example where this PR is fixing a breakage or this is more of a theoretical correction ?

I'm working on LiveOS images where, for example, an image that needs the xfs driver is denied because of the presence of the selinuxfs driver.

LaszloGombos (approved):
LGTM

aafeijoo-suse (changes requested):
This may be working right now with our current `load_fstype` calls, but it's not 100% correct, `strstr` does not work well in this case. See:

```
# type load_fstype
load_fstype is a function
load_fstype () 
{ 
    local t n;
    t=$(printf '\t');
    n=$(printf '\n');
    strstr "$(cat /proc/filesystems)$n" "$t${2:-$1}$n" || echo "modprobe $1"
}
# cat /proc/filesystems | grep cgr
nodev	cgroup
nodev	cgroup2
# load_fstype cgroup && echo OK
OK
# load_fstype cgroup3 && echo OK
modprobe cgroup3
OK
# load_fstype cgrou && echo OK
OK
```

You can use `sed` or `while read` (I would not use `grep` to remove its dependency in #2098).

FGrose:
I force pushed a change that avoids the loss of the trailing newline during assignment.  This version passes @aafeijoo-suse's test.

FGrose:
I pushed another change that twice avoids spawning a subshell.

johannbg:
@FGrose any particular reason you are removing reviewers from your PR?

FGrose:
> @FGrose any particular reason you are removing reviewers from your PR?

@johannbg The removal was a side effect of selecting the second item in the ... submenu for the change requested by @aafeijoo-suse. It was not expected or intended. I can't remember the item's label now, but it was something affirming the requested change.  (If you enter a new change request, we might be able to debug this further.)

FGrose:
Force pushed a refactored change to use a traditional while-read loop with redirected file input.  This avoids yet another subshell and shortcuts the processing if the match is found earlier in the file stream.  Thus, 4 to 20 times more speedy in my testing.

FGrose:
@johannbg It appears that if I Re-request a review, the system removes standing requests.

aafeijoo-suse (commented):
Although you approach may be valid, I find it more complicated than it should be, and coupled with the \t between columns. IMHO, why not just something like (not tested):

```sh
load_fstype() {
    local _fs="${2:-$1}"
    local _fsmod="$1"
    while read -r nodev fs || [ -n "$nodev" ]; do
        [ -z "$fs" ] && fs="$nodev"
        [ "$fs" = "$_fs" ] && return 0
    done < /proc/filesystems
    modprobe "$_fsmod"
}
```
But it's just my point of view, if the others agree with you solution, then I'll too.

johannbg (changes requested):
You forgot to remove the debug switch in bash. 
Once done I will merge this since people can debate more or less to infinity what is the proper "approach" to resolve this.

FGrose:
@johannbg > You forgot to remove the debug switch in bash.
The `set +x` is intentional, like `debug_off` and `debug_on` in _getarg()_.  Bash and Dash both support the `local -` statement. 

> Making “-” local causes any shell options that are changed via the set command inside the function to be restored to their original values when the function returns.

This avoids unnecessary verbosity in an xtrace log.

aafeijoo-suse (approved):
Thanks, now it LGTM, and sorry for being fussy.

FGrose:
@aafeijoo-suse : Fuss appreciated.

Mehdi207-mp:
Hello, good time, I completely lost my access and applied from every site
and source. Your site is my last hope and my way to access my assets.
Please help me. Thank you.

در تاریخ چهارشنبه ۷ دسامبر ۲۰۲۲،‏ ۱۱:۰۱ Antonio Álvarez Feijoo <
***@***.***> نوشت:

Mehdi207-mp:
Hello, good time, I completely lost my access and applied from every site
and source. Your site is my last hope and my way to access my assets.
Please help me. Thank you.

در تاریخ چهارشنبه ۷ دسامبر ۲۰۲۲،‏ ۱۱:۳۰ Mehdi Por Mohamad Ali <
***@***.***> نوشت:

johannbg (approved):

johannbg:
@LaszloGombos , @aafeijoo-suse did you guys miss the share number of failing test prior to approving this or is it being intentionally ignored?

aafeijoo-suse:
> @LaszloGombos , @aafeijoo-suse did you guys miss the share number of failing test prior to approving this or is it being intentionally ignored?

I think it was related to #2107, re-running to confirm.

PR #2095, opened 2022-11-29
feat(multipath): warn if included with no multipath devices and no user conf

aafeijoo-suse:
The multipath module is always included in no-hostonly mode, but including it in the initramfs might activate multipath on the root device, which could be undesirable. This behavior can be manually fixed by explicitly excluding the root device from multipath in /etc/multipath.conf.

Therefore, issue a warning in no-hostonly mode, if there is no multipath device and the user has not configured multipath.

CC: @mwilck (bsc#1069169)

## Checklist

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -21,13 +21,23 @@ majmin_to_mpath_dev() {
 
 # called by dracut
 check() {
+    local _any_mpath_dev
+
+    for_each_host_dev_and_slaves is_mpath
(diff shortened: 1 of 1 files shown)


mwilck (commented):
This is only necessary on SUSE or openSUSE distributions, and possibly others that use multipathd's "greedy" mode by default (I am not sure if there any others).

Fedora/RHEL and most other distros don't need this, because multipathd would never grab a device that hasn't been explicitly configured as multipath device by the user beforehand.

PR #2096, opened 2022-11-29
RFC: feat(kernel-modules): exclude USB drivers in strict hostonly mode

aafeijoo-suse:
Provide a way to avoid installing all USB drivers in some use cases where they are not needed, i.e., machines without a USB bus.

## Checklist

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -35,13 +35,14 @@ installkernel() {
 
     if [[ -z $drivers ]]; then
         hostonly='' instmods \
-            hid_generic unix \
+            hid_generic unix
+
+        hostonly=$(optional_hostonly) instmods \
             ehci-hcd ehci-pci ehci-platform \
             ohci-hcd ohci-pci \
             uhci-hcd \
-            xhci-hcd xhci-pci xhci-plat-hcd
-
-        hostonly=$(optional_hostonly) instmods \
+            usbhid \
+            xhci-hcd xhci-pci xhci-plat-hcd \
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
CC: @tblume @hramrach

johannbg:
I think we need to be a bit smarter about this as in try to detect it and install it if applicable as opposed to have "optional_hostonly"

hramrach:
That's the bug - dracut is not smart about it, it always installs USB drivers no matter what.

johannbg:
@hramrach not just the usb drivers, the fix for this needs to be a part of a larger scope to address this.

hramrach:
The other class of drivers that is installed unconditionally is pcmcia which I don't build anymore so I don't really care.

I don't have the hardware to test if changes to the module list don't break systems that actually use pcmcia, either.

hramrach:
Nevemind, even the pcmcia is not unconditional, actually.

Only the USB drivers are so special that they are loaded always.

aafeijoo-suse:
> I think we need to be a bit smarter about this as in try to detect it and install it if applicable as opposed to have "optional_hostonly"

Quoting the man page: _"With "strict" mode enabled, anything not necessary for booting the local host in its current state will not be included"_.

So, more than an optional hostonly, with this patch, in strict hostonly mode the usb drivers are only installed if they are needed in any required block device, which is at least an improvement over what we have now.

I agree, the "smart" approach would require a larger scope.

johannbg (approved):
slightly better is still better and this can be reworked as a part of a larger scope at a later date...

PR #2097, opened 2022-11-30
fix(dracut.sh): use DRACUT_ARCH instead of `uname -m`

aafeijoo-suse:
DRACUT_ARCH overrides `uname -m` since https://github.com/dracutdevs/dracut/commit/a0120420

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1434,7 +1434,7 @@ if [[ ! $print_cmdline ]]; then
         fi
         unset EFI_MACHINE_TYPE_NAME
         EFI_SECTION_VMA_INITRD=0x3000000
-        case $(uname -m) in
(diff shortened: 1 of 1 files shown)


LaszloGombos (approved):
LGTM.

How about this instance ?
`modules.d/01fips/fips.sh:            if [ "$(uname -m)" = s390x ]; then`

aafeijoo-suse:
> How about this instance ? `modules.d/01fips/fips.sh: if [ "$(uname -m)" = s390x ]; then`

DRACUT_ARCH, if set, is used at generation time. The code you mention runs at boot time, so it never exists there.

johannbg (approved):
LGTM

Issue #2102, opened 2022-12-03
mmc_block.ko is not included into initrd on kernel 6.1


mikhailnov:
Kernel module mmc_block.ko is not included into initrd on kernel 6.1-rc7 (`/lib/modules/6.1.0-generic-0.rc7.1rosa2021.1-x86_64/kernel/drivers/mmc/core/mmc_block.ko.zst`), but is included on kernel 5.15 (`/lib/modules/5.15.77-generic-1rosa2021.1-x86_64/kernel/drivers/mmc/core/mmc_block.ko.zst`). This makes booting from root on /dev/mmcblk* impossible with 6.1.

I have not understood yet why this happens, but it seems that its inclusion is rather indirect on x86_64 and is done by `_blockfuncs=<...>|mmc_add_host|<...>` in `modules.d/90kernel-modules`. However I do not see that this function would have been moved between kernels 5.15 and 6.1, it is in `drivers/mmc/core/host.c` in both kernels.

Hardware information gathered on kernel 5.15: https://linux-hardware.org/?probe=3f6ebcde0a

Dracut version is at commit 5eb736

mikhailnov:
Current code in modules.d/90kernel-modules/module-setup.sh looks strange. It assumes that mmc block devices related drivers are always needed, even in hostonly mode, and that =drivers/mmc/ are needed only on embeded hardware (if I understood correctly).

mikhailnov:
If run `dracut-install --debug` for kernel 5.15:

```
dracut-install: Handling /lib/modules/5.15.79-generic-1rosa2021.1-x86_64//kernel/drivers/mmc/core/mmc_block.ko.zst
dracut-install: Module mmc_block: symbol blk_cleanup_queue matched inclusion filter
```

If run it for 6.1:
```dracut-install: Handling /lib/modules/6.1.0-generic-0.rc7.1rosa2021.1-x86_64//kernel/drivers/mmc/core/mmc_block.ko.zst
dracut-install: No symbol or path match for '/lib/modules/6.1.0-generic-0.rc7.1rosa2021.1-x86_64//kernel/drivers/mmc/core/mmc_block.ko.zst'
```

mikhailnov:
[...]

PR #2103, opened 2022-12-03
fix(90kernel-modules): fix MMC and NVMe on kernels 6.0+

mikhailnov:
Commit 6f8191fdf41 ("block: simplify disk shutdown") (https://github.com/torvalds/linux/commit/6f8191fdf41d3a53cc1d63fe2234e812c55a0092) in Linux kernels 6.0 and above renamed blk_cleanup_disk to blk_mq_destroy_queue. Adapt code to handle both cases.

Because of this the kernel module mmc_block.ko was not included into initrd on kernel 6.1:
dracut-install: No symbol or path match for '/lib/modules/6.1.0-generic-0.rc7.1rosa2021.1-x86_64//kernel/drivers/mmc/core/mmc_block.ko.zst
while on kernel 5.15 it was included:
dracut-install: Handling /lib/modules/5.15.79-generic-1rosa2021.1-x86_64//kernel/drivers/mmc/core/mmc_block.ko.zst
dracut-install: Module mmc_block: symbol blk_cleanup_queue matched inclusion filter

It made booting OS from root on /dev/mmcblk* impossible.

Just adding blk_mq_destroy_queue was not enough to make mmc_block.ko be included into initrd (maybe because call of blk_cleanup_disk() was removed and call of blk_mq_destroy_queue() was not added).
So also adding blk_mq_alloc_request which is used from mmc_block.

Fixes: https://github.com/dracutdevs/dracut/issues/2102


LaszloGombos:
@mikhailnov 

> Error: commit_message:1:24: error: commit message's description contains blacklisted word or repeats type tag
>fix(90kernel-modules): fix MMC and NVMe on kernels 6.0+
>                       ^^^
>commit_message:1:24: note: prefer using the imperative for verbs

maybe fixed ?

mikhailnov:
> maybe "fix(90kernel-modules): fixed MMC and NVMe....' ?

Done

LaszloGombos:
@mikhailnov 

> Error: commit_message:1:24: error: commit message's description contains blacklisted word or repeats type tag
> fix(90kernel-modules): fixed MMC and NVMe on kernels 6.0+
>                        ^^^^^
> commit_message:1:24: note: prefer using the imperative for verbs

Please try to find a solution (it is me not doing this) - https://github.com/dracutdevs/dracut/actions/runs/3610891530/jobs/6084978300

Perhaps "correct"

johannbg (dismissed):
You need to rephrase your commit message excluding the word `fixed` other than LGTM

mikhailnov:
@LaszloGombos thanks for rephrasing the commit message

aafeijoo-suse (approved):

aafeijoo-suse:
Related: https://github.com/redhat-plumbers/dracut-rhel9/pull/57
@pvalena is there any reason why RH didn't sent this upstream first, so all distros using dracut with the latest kernel version could benefit from this patch?

aafeijoo-suse:
> Related: [redhat-plumbers/dracut-rhel9#57](https://github.com/redhat-plumbers/dracut-rhel9/pull/57) @pvalena is there any reason why RH didn't send this upstream first, so all distros using dracut with the latest kernel version could benefit from this patch?

@johannbg what's your opinion here? I'm a bit upset, things like this should never happen with RH and the worst thing is we are only getting silence and no justification.

johannbg:
@aafeijoo-suse I can only assume the reason but how can you be upset? Any corporation or individual is free to submit patches upstream or not RH and or it's employees are not obliged to submit patches here upstream or owe upstream somekind of justification if they dont.

aafeijoo-suse:
Yeah, better forget about this.

In another vein, maybe we should add that patch, because I think this PR does not cover the case of the `nvdimm` drivers.

```
# zstd -d /lib/modules/5.14.21-150400.24.18-default/kernel/drivers/nvdimm/nd_pmem.ko.zst -o nd_pmem.ko
/lib/modules/5.14.21-150400.24.18-default/kernel/drivers/nvdimm/nd_pmem.ko.zst: 38904 bytes 
# nm nd_pmem.ko | grep blk
                 U __blk_alloc_disk
-------------->  U blk_cleanup_disk
                 U blk_freeze_queue_start
                 U blk_queue_flag_set
                 U blk_queue_logical_block_size
                 U blk_queue_max_hw_sectors
                 U blk_queue_physical_block_size
                 U blk_queue_write_cache
                 U blk_status_to_errno
                 U errno_to_blk_status

# zstd -d /lib/modules/6.1.0-3.g7f1864f-default/kernel/drivers/nvdimm/nd_pmem.ko.zst -o nd_pmem.ko
/lib/modules/6.1.0-3.g7f1864f-default/kernel/drivers/nvdimm/nd_pmem.ko.zst: 46191 bytes 
# nm nd_pmem.ko | grep blk
                 U __blk_alloc_disk
                 U blk_queue_flag_set
                 U blk_queue_logical_block_size
                 U blk_queue_max_hw_sectors
                 U blk_queue_physical_block_size
                 U blk_queue_write_cache
                 U blk_status_to_errno
                 U errno_to_blk_status
```

johannbg:
@aafeijoo-suse yeah I dont see why not but I probably should mention he did not add his patch to fedora either ( which might raise some eyebrows in Raleigh, rarely is a fix limited to RHEL ) so there might be a catch 22 for doing that

mikhailnov:
Also, it would be very nice to somehow clean up hackery like explicitly adding drivers/mmc on aarch64 only https://github.com/dracutdevs/dracut/blob/e0d57a8f5b15847a7fbae2ed3df29ec2c5d25ec7/modules.d/90kernel-modules/module-setup.sh#L77 , while relying on detecting used kernel symbols on other architectures

PR #2104, opened 2022-12-03
chore(base): remove support for <udev-176.

LaszloGombos:
This PR reverts 579238a3acd0317d7124a76dc489f19d9471fb63.

Udev 176 was released in 2012. I think it is safe to assume that dracut only supports going back 11 years and not more.

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -560,6 +560,8 @@ build_ld_cache() {
 }
 
 prepare_udev_rules() {
+    dwarn "prepare_udev_rules: deprecated and will be removed"
+
     if [ -z "$UDEVVERSION" ]; then
         UDEVVERSION=$(udevadm --version)
         export UDEVVERSION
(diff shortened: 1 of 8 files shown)


LaszloGombos:
Continuing on the quest started https://github.com/dracutdevs/dracut/pull/2033 to review blkid usage.

LaszloGombos:
CC @aidecoe

johannbg:
@LaszloGombos out of curiosity why did you add an enhancement label to a chore?

dracut-init.sh
```diff
@@ -559,46 +559,6 @@ build_ld_cache() {
     fi
 }
 
-prepare_udev_rules() {
```
aafeijoo-suse, line 562: The `prepare_udev_rules` function is also called from 3rd party modules (see for example https://github.com/leahneukirchen/hrmpf/blob/master/dracut/module-setup.sh#L22). I'd issue a deprecation warning on this function and postpone its removal.

aafeijoo-suse: I'd write it just after entering the function. Something like:

```patch
 prepare_udev_rules() {
+    dwarn "prepare_udev_rules: YOUR DEPRECATION MESSAGE HERE"
+
     if [ -z "$UDEVVERSION" ]; then
         UDEVVERSION=$(udevadm --version)
         export UDEVVERSION
```

aafeijoo-suse (approved):

Conan-Kudo (approved):

johannbg (approved):

PR #2105, opened 2022-12-04
feat(dmsquash-live-autooverlay): Add an explicit disabler argument

Conan-Kudo:
This pull request changes...

## Changes
This enables users to set up live media such that if persistence can be used if it has been set up, but it will gracefully fall back to using in-RAM overlays if no persistence volume exists.

## Checklist

### DIFF
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -1175,6 +1175,11 @@ rd.live.overlay=UUID=99440c1f-8daa-41bf-b965-b7240a8996f4
 Specifies the filesystem to use when formatting the overlay partition.
 The default is ext4.
 
+**rd.live.overlay.nocreate=**1::
+Specifies that a persistent overlay should not be automatically created.
+This option is used to allow automatic persistent overlay detection to be specified,
+while gracefully falling back to temporary overlays if nothing is found.
+
 **rd.live.overlay.size=**__<size_MiB>__::
 Specifies a non-persistent Device-mapper overlay size in MiB.  The default is
 _32768_.
--- a/modules.d/90dmsquash-live-autooverlay/create-overlay.sh
+++ b/modules.d/90dmsquash-live-autooverlay/create-overlay.sh
@@ -14,6 +14,12 @@ gatherData() {
         info "Skipping overlay creation: kernel command line parameter 'rd.live.overlay' is not set"
         exit 0
     fi
+    overlaynocreate=$(getarg rd.live.overlay.nocreate)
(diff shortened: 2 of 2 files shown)


LaszloGombos:
@Conan-Kudo How do you feel about using "rd.live.overlay.cowfs=0" syntax or something similar ? 

If not confusing, it would be useful to minimize the number of argument.

Conan-Kudo:
> @Conan-Kudo How do you feel about using "rd.live.overlay.cowfs=0" syntax or something similar ?
> 
> If not confusing, it would be useful to minimize the number of argument.

The main motivation was that I could have all the parameters needed for creating+mounting an overlay set, but then short-circuit it so it doesn't run by default if the overlay doesn't already exist and fallback to temporary overlay gracefully. That was feedback given on the mailing list, that people would like a mode where it would auto-setup the overlay if it existed, but not create one if it didn't already exist.

LaszloGombos:
>  people would like a mode where it would auto-setup the overlay if it existed, but not create one if it didn't already exist.

@Conan-Kudo Sorry if Id id not do a good job expressing myself.

`rd.live.overlay.cowfs=no`

For the sake of technical discussion (and not readability) I wanted to understand if the following argument would not just do exactly that and if not why ?

Conan-Kudo:
Basically, the reason is that I want the formatting and other steps to run _if_ the partition exists but _not_ if it doesn't. If the partition was created, then it should go through the process of using it, formatting it, etc.

LaszloGombos:
@Conan-Kudo I understand it now, thanks for your patience . 

I feel bad pressing on the number of command line arguments and bikeshedding on this, but I think it is important.

Some more ideas for syntax:
- rd.live.overlay.cowfs=ext4,skip
- rd.live.overlay.cowfs=ext4:skip

Please feel free to propose something else.

I hope to hear from @iammattcoleman as well before proceeding.

iammattcoleman:
I feel like having multiple single-purpose parameters is easier for users to understand vs. packing multiple concepts into a single parameter. It's also a bit easier to deal with on the code side.

I'm not heavily invested in that, though, so if stuffing it in as a suffix on `rd.live.overlay.cowfs` is the preferred option then I'm happy to accept that, too. If we do go that route, I think that `skip` is unclear: I'd prefer something like `nocreate` or `no-create`.

Conan-Kudo:
I'm working on this still.

Conan-Kudo:
This doesn't work exactly the way I want to yet, so I'm closing this for now. I'll reopen it once I have the behavior working as intended.

PR #2108, opened 2022-12-06
feat: improve kernel module built time checking for iscsi and squash

ryncsn:
- Refuse to install iscsi module if iscsi is not support by kernel
- Check the actual kernel module for squash.

That's two separate issue, just fix them together and introduce a new helper here.

The squash change reverts part of #1015, which converted squash module to check kernel config instead of module.
Kernel config checking won't work if kernel is installed by `make install`, it's not a distro friendly way though, but `modprobe` with `-d` should be more robust.

## Checklist

Fixes #

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -158,6 +158,9 @@ require_kernel_modules() {
     local _module_name="${moddir##*/}"
     local _ret=0
 
+    # Ignore kernel module requirement for no-kernel build
+    [[ $no_kernel == yes ]] && return 0
+
     if [[ $1 == "-m" ]]; then
         _module_name="$2"
         shift 2


ryncsn:
> Would load_fstype be a more robust way to check - see #2092 ?

Hi @LaszloGombos, sorry I didn't look at that PR carefully yesterday, now I looked at that PR again, and actually `load_fstype` is very different from what I'm trying to do here.

`load_fstype` is used to load a fs during initramfs runtime.  `require_kernel_modules` and `check_kernel_module` is used to check if a kernel module is available during build time.

ryncsn:
I've added a new commit to make require_kernel_modules return 0 if no_kernel is in use, I think this make sense since no kernel build won't install any module, so just ignore the module checking.

ryncsn:
> @ryncsn I have not spend time on understanding this PR, but my main motivation was to understand if and how you handle detecting built-in linux kernel modules.

Thanks for the review, at bulid-time, modprobe should be able to handle built-in modules, for example:
```
modinfo ipv6
name:           ipv6
filename:       (builtin)
alias:          net-pf-10
license:        GPL
file:           net/ipv6/ipv6
description:    IPv6 protocol stack for Linux
author:         Cast of dozens
parm:           disable:Disable IPv6 module such that it is non-functional (int)
parm:           disable_ipv6:Disable IPv6 on all interfaces (int)
parm:           autoconf:Enable IPv6 address autoconfiguration on all interfaces (int)
```

All kmod tools recognizes built-in modules, so does modprobe. modprobe just return 0 and exit if the module is built-in or already loaded.

About the load_fstype helper, it's working around initramfs run time issues, shouldn't be a problem for build time.

johannbg (approved):
I would have preferred seeing the helper and the module changes in separate PR's and suspect that this might lead to downstream packaging problems but those are not of our concern and hopefully will just lead to long overdue cleanup in that area if that turns out to be the case...

aafeijoo-suse (approved):

PR #2028, opened 2022-10-28
Drop dmeventd from initramfs

smoser:
devicemapper in the initramfs built on at least Ubuntu 20.04 will currently warn about

  libdevmapper-event-lvm2thin.so dlopen failed:
     libdevmapper-event-lvm2.so.2.03:
       cannot open shared object file: No such file or directory

libdevmapper-event-lvm2.so.2.03 does not have a .so link pointing to it, so it gets missed by the existing
'inst_libdir_file "libdevmapper-event-lvm*.so"'

liblvm2cmd.so.2.03 is a dependency of libdevmapper-event-lvm2.so.2.03. It must be added directly because dracut-install does not resolve library dependencies, it only works for executables. Since dm-eventd dlopen's everything there
is no executable that would cause dracut-install to resolve this.

This pull request changes...

## Checklist
- [X] I have tested it locally
- [X] I have reviewed and updated any documentation if relevant
- [X] I am providing new code and test(s) for it

Fixes #

PR #2109, opened 2022-12-07
ci: run all local tests on systemd enabled containers

LaszloGombos:
This PR runs the local tests not only on fedora but on all 4 systemd enabled containers.

This PR increases the number of test runs from 32 to 71 (more than double).This sounds quite scary but tests do run in parallel, so it does not actually increases the time to finish running all the tests.

If this for some reason becomes a problem, we can always ration and discuss what is important to run and what gives the project most test coverage, but perhaps as a start we can try with the full "local" matrix.

All tests are passing except test 17 on debian. Test 17 on debian is fixed by https://github.com/dracutdevs/dracut/pull/2028 .

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -46,7 +46,10 @@ jobs:
         strategy:
             matrix:
                 container: [
+                        "arch:latest",
+                        "debian:latest",
                         "fedora:latest",
+                        "opensuse:latest",
                 ]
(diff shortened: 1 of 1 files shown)


johannbg (approved):
More test coverage is better and we can always revisit it later if it becomes a problem.

aafeijoo-suse (approved):

PR #2111, opened 2022-12-09
cryptsetup,crypttab - Optional performance enhacement for fast ssds

wolfboxpen:
Reference: https://wiki.archlinux.org/title/Dm-crypt/Specialties#Disable_workqueue_for_increased_solid_state_drive_(SSD)_performance

adds no-read-workqueue,no-write-workqueue  options.

### DIFF
--- a/modules.d/90crypt/cryptroot-ask.sh
+++ b/modules.d/90crypt/cryptroot-ask.sh
@@ -109,12 +109,26 @@ while [ $# -gt 0 ]; do
         allow-discards)
             allowdiscards="--allow-discards"
             ;;
+        no-read-workqueue)
+            no_read_workqueue="--perf-no_read_workqueue"
+            ;;
+        no-write-workqueue)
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
First, you need to fix your commit messages, and maybe merge both into one, as I don't see a reason to split this into 2. Please read https://github.com/dracutdevs/dracut/blob/master/docs/HACKING.md#commit-messages
It should be something like: `feat(crypt): optional performance enhacement for fast ssds (nvme pci3,4,5)`

wolfboxpen:
the github interface doesn't appear to allow this, if it's a deal breaker you may close this pull request. i already made the changed to my system.

johannbg:
Well we have to first detect if the host is using solid state disks and yes incorrect commit msg is a deal breaker hence closing

wolfboxpen:
the user would have to explicitly opt into those features by specifying those flags in crypttab. detection of ssd is superfluous.

PR #2113, opened 2022-12-10
fix(dbus-*): rename dbus-* to systemd-dbus-*

LaszloGombos:
Most dracut modules follow the convention that if the module depends on systemd than the name of the module is prefix with systemd.


johannbg:
Thats a nack

PR #2115, opened 2022-12-10
RFC: feat(dracut): systemd-measure support

sammko:
Hello,
I didn't find any discussion on adding systemd-measure support to the UKI flow. Not sure if this is planned / even welcome in dracut, but I did a quick implementation for my own use. Missing documentation etc, just a quick prototype. Please lmk if something like this would be welcome and I can work on cleaning it up.

## Changes
Run systemd-measure if `--uefi` is selected and the corresponding config options are set. Package the resulting signature in the UKI.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2619,12 +2619,50 @@ if [[ $uefi == yes ]]; then
         unset uefi_splash_image
     fi
 
+    if [[ -n ${uefi_tpm2_pcr_sig_private_key} && -n ${uefi_tpm2_pcr_sig_public_key} ]]; then
+        if command -v systemd-measure &> /dev/null; then
+            systemd_measure="systemd-measure"
+        else
+            if [[ -x "${dracutsysrootdir}${systemdutildir}/systemd-measure" ]]; then
+                systemd_measure="${dracutsysrootdir}${systemdutildir}/systemd-measure"
+            else
+                dfatal "systemd-measure not found"
+                exit 1
+            fi
+        fi
+
+        dinfo "*** Computing systemd-measure TPM2 PCR signature ***"
+
+        measure_parameters="--initrd=${DRACUT_TMPDIR}/initramfs.img \
+            --linux="$kernel_image" \
+            ${uefi_osrelease:+--osrel="$uefi_osrelease"} \
+            ${uefi_cmdline:+--cmdline="$uefi_cmdline"} \
+            ${uefi_splash_image:+--splash="$uefi_splash_image"} \
+            --pcrpkey=${uefi_tpm2_pcr_sig_public_key} \
+            --bank=sha1 \
+            --bank=sha256 \
+            --private-key=${uefi_tpm2_pcr_sig_private_key} \
+            --public-key=${uefi_tpm2_pcr_sig_public_key}"
+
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
> Thanks, I find it useful. Let's see what others think.

That said, it's worth mentioning that it's still experimental, so maybe we should wait until it's stable.

_"Note: this command is experimental for now. While it is likely to become a regular component of systemd, it might still change in behaviour and interface."_

johannbg:
Looks useful to me as well.

sammko:
This may be obsoleted by ukify newly added to systemd? https://github.com/systemd/systemd/blob/6ae3bd82d04155d27488c85c63f7330cc7e144d3/NEWS#L42-L48

aafeijoo-suse:
> This may be obsoleted by ukify newly added to systemd? https://github.com/systemd/systemd/blob/6ae3bd82d04155d27488c85c63f7330cc7e144d3/NEWS#L42-L48

IMHO, just because systemd has its own tool for creating UKIs, doesn't mean that dracut should not implement this stuff when it's no longer experimental.

dracut.sh
```diff
@@ -2619,12 +2619,50 @@ if [[ $uefi == yes ]]; then
...
     fi
 
+    if [[ -n ${uefi_tpm2_pcr_sig_private_key} && -n ${uefi_tpm2_pcr_sig_public_key} ]]; then
+        if command -v systemd-measure &> /dev/null; then
+            systemd_measure="systemd-measure"
+        else
+            if [[ -x "${dracutsysrootdir}${systemdutildir}/systemd-measure" ]]; then
+                systemd_measure="${dracutsysrootdir}${systemdutildir}/systemd-measure"
+            else
+                dfatal "systemd-measure not found"
```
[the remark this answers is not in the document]
sammko: definitely -- the check in question is only executed if the config options are both set (`uefi_tpm2_pcr_sig_private_key` and `uefi_tpm2_pcr_sig_public_key`)

dracut.sh
```diff
@@ -2619,12 +2619,50 @@ if [[ $uefi == yes ]]; then
...
+        dinfo "*** Computing systemd-measure TPM2 PCR signature ***"
+
+        measure_parameters="--initrd=${DRACUT_TMPDIR}/initramfs.img \
+            --linux="$kernel_image" \
+            ${uefi_osrelease:+--osrel="$uefi_osrelease"} \
+            ${uefi_cmdline:+--cmdline="$uefi_cmdline"} \
+            ${uefi_splash_image:+--splash="$uefi_splash_image"} \
+            --pcrpkey=${uefi_tpm2_pcr_sig_public_key} \
+            --bank=sha1 \
+            --bank=sha256 \
```
Faerbit, line 2643: Hard coding sha1 and sha256 here is suboptimal. If your TPM supports a newer hash type, this will force the usage of a less secure version.
I think I get why you do it like this: If I specify a newer hash type (or none at all and rely the default, which is all hash types) on my system when using the `sign` verb, it will error out, since my TPM doesn't support it AFAICT.
So this is arguably an upstream issue, since I also don't know of an elegant way to solve this.
But this doesn't change the result: So maybe also add an config option for this too?

sammko: I think adding a config option with a default value of `sha1 sha256` or something like that would be the way to go. thanks!

PR #2116, opened 2022-12-11
test: make sure network meta module included in networking tests

LaszloGombos:
Make sure network meta module included in networking tests
    
Instead of using network-legacy dracut module as the default for networking tests, use the network dracut module.

### DIFF
--- a/test/TEST-20-NFS/test.sh
+++ b/test/TEST-20-NFS/test.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-[ -z "$USE_NETWORK" ] && USE_NETWORK="network-legacy"
-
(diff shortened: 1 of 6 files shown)


aafeijoo-suse (commented):
I'm not following what the ultimate purpose of this change will be, to select the container's default network module? could you elaborate why this would be good? Thanks.

johannbg:
Like @aafeijoo-suse I'm not following what this is supposed to achieve I mean we always need to specifically test for the specific network modules that the network meta module includes  ( in addition to having to test the network meta module itself )  and the default's might change ( for example sd-networkd is more of a sane default these days for systemd based systems )  

Note you need to resolve your conflicts

PR #2117, opened 2022-12-11
ci: introduce more networking options into test containers

LaszloGombos:
Add systemd-networked to Fedora container.
Add connman to Arch container.

This change enables manual testing of various network configurations and also allows the ci to ensure that having additional networking packages installed does not break the network meta module.

### DIFF
--- a/test/container/Dockerfile-Arch
+++ b/test/container/Dockerfile-Arch
@@ -11,7 +11,7 @@ RUN echo 'export DRACUT_NO_XATTR=1 KVERSION=$(cd /lib/modules; ls -1 | tail -1)'
 RUN pacman --noconfirm -Syu \
     linux dash strace dhclient asciidoc cpio pigz squashfs-tools \
     qemu btrfs-progs mdadm dmraid nfs-utils nfsidmap lvm2 nbd \
-    dhcp networkmanager multipath-tools vi tcpdump open-iscsi \
+    dhcp networkmanager multipath-tools vi tcpdump open-iscsi connman \
     git shfmt shellcheck astyle which base-devel glibc parted ntfs-3g && yes | pacman -Scc
 
 RUN useradd -m build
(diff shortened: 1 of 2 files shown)


test/container/Dockerfile-OpenSuse-latest
```diff
@@ -13,7 +13,7 @@ RUN dnf -y install --setopt=install_weak_deps=False \
     strace libkmod-devel gcc bzip2 xz tar wget rpm-build make git bash-completion \
     sudo kernel dhcp-client qemu-kvm /usr/bin/qemu-system-$(uname -m) e2fsprogs \
     tcpdump iproute iputils kbd NetworkManager btrfsprogs tgt dbus-broker \
-    iscsiuio open-iscsi which ShellCheck procps pigz parted squashfs ntfsprogs \
+    iscsiuio open-iscsi which ShellCheck procps pigz parted squashfs ntfsprogs wicked \
```
aafeijoo-suse, line 16: I don't know if we should spend time and resources in testing the wicked module... I doubt anyone is using it, the dracut module is incomplete (see #1670), and wicked is being phased out from openSUSE (see https://news.opensuse.org/2022/01/28/bash-systemd-libvirt-update-in-tw/).

aafeijoo-suse: I don't know if it's reason enough to drop it, I don't know how many users might be using it, but at least it's reason enough not to waste time testing it upstream.

aafeijoo-suse:
> @aafeijoo-suse let me know if you have a preference of changing openSuse container to add more option (e.g. systemd-networked or connman). For now I decided not to change the openSuse container as part of this PR.

network-legacy and network-manager are enough for now, thanks.

aafeijoo-suse (approved):

johannbg (approved):

PR #2118, opened 2022-12-11
test: do not omit dash module

LaszloGombos:
It should not be necessary to explicitly omit dash module during test runs. If this is needed, that would mean that bash is not marked as an explicit dependency for some modules.


aafeijoo-suse (approved):
BTW there is currently no module that depends on the `dash` module.

johannbg (approved):
LGTM

PR #2119, opened 2022-12-11
fix(systemd-timesynced, systemd-resolved): install system user and group

LaszloGombos:
To make use of these daemons a new system user and group needs to be created on installation.


aafeijoo-suse (commented):
This PR is not necessary, these users and groups are already installed by sysusers.d conf.

aafeijoo-suse:
> > This PR is not necessary, these users and groups are already installed by sysusers.d conf.
> 
> Thanks @aafeijoo-suse . Can some of these lines be removed than - https://github.com/dracutdevs/dracut/blob/master/modules.d/00systemd/module-setup.sh#L208 ?

Yes, the systemd module needs to be refactored. See #1370

PR #2123, opened 2022-12-12
fix(network-manager): support nm-initrd-generator under NetworkManager

LaszloGombos:
Upstream NetworkManager explicitly [supports Debian](https://github.com/NetworkManager/NetworkManager/blob/main/.gitlab-ci/debian-install.sh). This PR proposes the same for the network-manager dracut module.

Upstream NetworkManager explicitly supports prefix for libexecdir [here](https://github.com/NetworkManager/NetworkManager/blob/main/meson.build#L41) .

In [network-manager/module-setup.sh nm-initrd-generator](https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/module-setup.sh#L33) is already marked optional to accomodate different locations for nm-initrd-generator in different setups as the path is not stable between Linux installations and versions.

The file in [Debian 12](https://packages.debian.org/bookworm/amd64/network-manager/filelist) is located in `/usr/lib/NetworkManager/nm-initrd-generator` .

The file location has been in stable Debian **for about 4 years now** ([since Debian 10](https://packages.debian.org/buster/amd64/network-manager/filelist)). 

This issue impacts over 50 Linux distributions and blocks testing NetwrokManager for Debian on the upstream CI.

From [FHS 3.0](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html)

> Some previous versions of this document did not support /usr/libexec, despite it being standard practice in a number of environments. To accomodate this restriction, it became common practice to use /usr/lib instead. **Either practice is now acceptable**, but each application must choose one way or the other to organize itself.

CC @Mrfai  @bdrung

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -30,8 +30,8 @@ install() {
     inst_multiple ip sed grep
 
     inst NetworkManager
-    inst_multiple -o /usr/{lib,libexec}/nm-initrd-generator
-    inst_multiple -o /usr/{lib,libexec}/nm-daemon-helper
+    inst_multiple -o /usr/{lib,libexec}{,/NetworkManager}/nm-initrd-generator
+    inst_multiple -o /usr/{lib,libexec}{,/NetworkManager}/nm-daemon-helper
     inst_multiple -o teamd dhclient
     inst_hook cmdline 99 "$moddir/nm-config.sh"
     if dracut_module_included "systemd"; then


modules.d/40network/module-setup.sh
```diff
@@ -21,7 +21,7 @@ depends() {
             network_handler="network-wicked"
         elif [[ -e $dracutsysrootdir$systemdsystemunitdir/connman.service ]]; then
             network_handler="connman"
-        elif [[ -x $dracutsysrootdir/usr/libexec/nm-initrd-generator ]] || [[ -x $dracutsysrootdir/usr/lib/nm-initrd-generator ]]; then
+        elif find_binary NetworkManager > /dev/null; then
```
aafeijoo-suse, line 24: This or #2081?

modules.d/35network-manager/module-setup.sh
```diff
@@ -30,8 +30,8 @@ install() {
     inst_multiple ip sed grep
 
     inst NetworkManager
-    inst_multiple -o /usr/{lib,libexec}/nm-initrd-generator
-    inst_multiple -o /usr/{lib,libexec}/nm-daemon-helper
+    inst_multiple -o /usr/{lib,libexec}{,/NetworkManager}/nm-initrd-generator
+    inst_multiple -o /usr/{lib,libexec}{,/NetworkManager}/nm-daemon-helper
```
Conan-Kudo, line 34: @thaller, @lkundrak: are you folks aware that some put these binaries in `/usr/libexec/NetworkManager`? Have you considered making that the upstream default rather than dropping them directly in `/usr/libexec`?

Conan-Kudo:
> File locations for debian (and debian based distro's) - https://packages.debian.org/sid/amd64/network-manager/filelist

I fired off an email to ask about reconciling the paths with upstream NetworkManager since Debian adopted FHS 3.0 some time ago and allows `/usr/libexec`.

modules.d/40network/module-setup.sh
aafeijoo-suse: > You expect that only one of these PRs are needed but NOT both ? Or you'd like me to combine the 2 PRs into one PR ?

You're adding different checks for the same thing (NetworkManager) using different PRs... it's quite confusing. IMHO this change deserves only a single PR, and there some of the checks may be corrected or not.

johannbg:
> Different distro's place nm-initrd-generator in different locations. In network-manager/module-setup.sh nm-initrd-generator is already marked optional to accomodate different locations for nm-initrd-generator.

Hmm that's probably a historic/snug in bug as in we support standards and the effort of the people creating those standards and we do so by adhering to those standards not by support working around those standards  ( but are open for a strong technical argument for deviating from the standards if such thing exist which is a bug in the standard then ) so workarounds for downstream that deviate from the standards is something we dont support and downstream has to carry the support for it, themselves and in the progress all the effort in trying to maintain that support that deviation in otherwords we should not have to call find to find a binary in which locations should already be standardized upon. 
Us having to do so is either a bug upstream,  in the standard or religions deviation downstream ( which we dont support ).

That said like @aafeijoo-suse I'm getting a bit confused about all this work it's like you are constantly trying to refactor our code to fit your test case for debian?

bdrung (approved):
All four paths `/usr/{lib,libexec}{,/NetworkManager}` are FHS compliant and this change will support the path in Debian/Ubuntu. So +1 for this change from my side.

pvalena (approved):
LGTM.

Conan-Kudo:
FYI, [Debian 13 "Trixie" moves it to `/usr/libexec`](https://packages.debian.org/trixie/amd64/network-manager/filelist)

PR #2124, opened 2022-12-13
dmsquash-live: Improve debugging

FGrose:
refactor(dmsquash-live-root): remove unneeded exit 1 after die calls

fix(dmsquash-live-root): quote variable names in unset calls

Save rd.live.debug output accessibly in the default journal or
  `/run/initramfs/liveroot.$$.out` & ..`/livegen.$$.out`.
Export a custom trace prompt in PS4.

refactor(dracut-lib.sh debug muting): use local '-' to restore xtrace

### Checklist
- [✔] I have tested it locally -- except several functions in dracut-lib.sh
where debug_off/debug_on xtrace muting was refactored.

### DIFF
--- a/modules.d/01systemd-ac-power/module-setup.sh
+++ b/modules.d/01systemd-ac-power/module-setup.sh
@@ -22,7 +22,7 @@ depends() {
 install() {
 
     inst_rules "$moddir/99-initrd-power-targets.rules"
-    inst_simple "$systemdutildir"/systemd-ac-power
+    inst systemd-ac-power
     inst_simple "$moddir/initrd-on-ac-power.target" "$systemdsystemunitdir/initrd-on-ac-power.target"
     inst_simple "$moddir/initrd-on-battery-power.target" "$systemdsystemunitdir/initrd-on-battery-power.target"
 


Conan-Kudo (approved):

aafeijoo-suse (commented):
You need to update your branch to the HEAD version and fix the conflicts.

Some comments:
- e6488583 _refactor(dmsquash-live-root): remove unneeded exit 1 after die calls_
    - Ok
- e636e6f3 _fix(dmsquash-live-root): quote variable names in unset calls_
    - I don't know if this is necessary, unset is called without quotes in the whole project, and this will change the style only in this module.
- 7e933627 _fix(dmsquash-live): keep debug xtrace in the default journal_
    - dmsquash-live-autooverlay also saves debug output in `/tmp`, should we move both to `/run/initramfs`?
    - `rd.debug` already exports a custom trace prompt in `PS4`, but I see why you did this after reading commit 7e979afa, you want `rd.live.debug` to be independent of `rd.debug`.
- 7e979afa _refactor(dracut-lib debug muting): use local '-' to restore xtrace_
    - This was added in [bash-4.4](https://git.savannah.gnu.org/cgit/bash.git/tree/NEWS?h=bash-4.4#n66), if we merge this, we should update the minimum required version check in dracut.sh:

https://github.com/dracutdevs/dracut/blob/e3e8108eb75247249ec05eaba943c3f48637c04b/dracut.sh#L26-L30

FGrose:
@LaszloGombos, This shell code works with dash since v0.5.2 (2005-09-26).

aafeijoo-suse:
> > This shell code feature was included with dash
>
> If dash is used inside the generated initramfs to execute dracut module scripts, than why is it necessary to update the minimum required version check in dracut.sh. Would dracut.sh call any of the module code that this PR is changing ?

No, Laszlo is right, sorry for the noise.

PR #2126, opened 2022-12-14
RFC: feat(lvm): always include all drivers that LVM can use

aafeijoo-suse:
This patch adds all the kernel modules that might be needed by LVM, to avoid having to rebuild the initrd in hostonly mode after a dynamic change that requires new drivers to boot.

For example, LVM allows to dynamically convert a linear logical volume to a RAID-1 type (`lvconvert --type raid1 vg/lv`), which, in hostonly mode, will require the user to manually rebuild the initrd again to include the new RAID drivers in use, otherwise the system will fail to boot.

Identified kernel modules used by LVM (CC: @zhaohem @teigland): all in `drivers/md/*`, excluding `bcache` directory and `md-cluster.ko`.

The size increase is approx. +1MB (using zstd compression):

```
localhost:~ # ls -l initrd-6.0.7-1-default-*
-rw------- 1 root root 22675622 Dec 13 14:43 initrd-6.0.7-1-default-before
-rw------- 1 root root 23758084 Dec 14 09:38 initrd-6.0.7-1-default-after
```

## Checklist

### DIFF
--- a/modules.d/90lvm/module-setup.sh
+++ b/modules.d/90lvm/module-setup.sh
@@ -43,7 +43,7 @@ cmdline() {
 }
 
 installkernel() {
-    hostonly='' instmods dm-snapshot
+    hostonly='' dracut_instmods -o -P ".*/(bcache/|md-cluster).*" "=drivers/md"
 }
 
 # called by dracut


zhaohem:
sorry for my late reply, I give more info about this commit.

how to trigger this bug:
there is a linear LV for rootfs, then user uses lvconvert to change LV to raid1 type. but user doesn't rebuild initrd after lvconvert. then system won't boot successfully. 
the boot failure is initrd doesn't include relevant DM & MD KOs.

User may very easily forget to rebuild intird and trigger this bug when they do the lvconvert job or similar job.

This pull request is one of the solution, which adds all DM & MD KOs into intird as long as lvm is included in initrd.

Other methods may:
- modify lvm2 commands (e.g: lvconvert, lvcreate) to print rebuild initrd message for user when underlying dev is boot devices. (how to know which dev is boot devices? from mount point? from cmdline?)
- systemd-udevd will response to update initrd. (just an idea, I don't know how to handle)

johannbg:
That size increase for everyone is quite the hefty price to workaround the lack of intelligence or the forgetfulness of the few. 

We have to make the assumption that users know what they are doing when they make changes to the host they are working on. 

The fact is users forgetfulness or intelligence ( or lack there of ) cannot be workaround in code so the best solution to this problem is hinting to the user to rebuild the initramfs in the log when he makes those changes but at the same time there is no guarantee that the user remembers to read the logs or actually knows what he's doing...

Firstyear:
@johannbg Wow. I don't think it's right to be so hostile or insulting to other people like this. Empathy for other humans goes a long way, and this is not "forgetfulness" or "lack of intelligence" on their part. Most documentation about lvm online references tools like lvcreate, lvconvert etc. The top level "lvm" binary is barely if at all mentioned in most public sources. This means people may simply not be aware it exists and what it does at all - which is exactly what happened to me, and exactly why I raised this bug inside of SUSE. 

This is about usability, and consistent behaviour (per human interaction psychology) is an important principle to help people when they are interacting with a system.

In future please consider more carefully your words on public issues.

johannbg:
@Firstyear I dont see how this is supposed to be hostile or insulting to anyone. I myself often forget things as well as I'm not all wise or all knowing so my remark is as applicable to me as it is to anyone else. 
Perhaps the difference between myself and others is that I dont have unrealistic expectation that software circumvents my own shortcomings ( such as not being a subtle person ) but instead I myself take steps to remedy that myself, steps such as writing things down for the things I forget, read to expand my knowledge etc. but thank you for your concern. 

As you point out the root cause of this can simply be a upstream documentation issue as well as the command not being verbose enough as for example providing a hint to remind the users to rebuild their initramfs, which is what the upstream Logical Volume Manager community needs to look at and address.

Firstyear:
> @Firstyear I dont see how this is supposed to be hostile or insulting to anyone. I

"> lack of intelligence or the forgetfulness of the few."

This is directly insulting peoples intelligence, specifically *mine* because I was the one who encountered this problem, had an unbootable system and had to initiate emergency recovery. 

> Perhaps the difference between myself and others is that I dont have unrealistic expectation that software circumvents my own shortcomings ( such as not being a subtle person ) but instead I myself take steps to remedy that myself, steps such as writing things down for the things I forget, read to expand my knowledge etc. but thank you for your concern.

I had been a sysadmin for 7 years. I've used LVM for many of them. I sat down, read all the man pages, documented all my commands I was about to take and then executed the plan, and still got an unbootable system. Do not tell me it was lack of preparedness. 

Please, show me where on this man page it references needing to rebuild your initramfs: https://linux.die.net/man/8/lvconvert 

> As you point out the root cause of this can simply be a upstream documentation issue as well as the command not being verbose enough as for example providing a hint to remind the users to rebuild their initramfs, which is what the upstream Logical Volume Manager community needs to look at and address.

Documentation is where poor design choices go to die, because then when the user hurts themself you point at it an say "oh you should have read that". When the real failure was the system that let them make the mistake at all. 

In the world around us we have so many design safeties to protect you and everyone, because all humans are distracted and make mistakes. Microwaves stop when you open the door - they don't just keep running and we say "oh but you irradiated your hand now should have read that documentation". 

Here we have a simple case - LVM as a system allows dynamic alteration of the volume group. And yet even in "hostonly" mode the initrd is lacking what is needed to support LVM and it's dynamic nature. That is a bug.

It's not like this issue is saying "oh lets add three wifi drivers because they are common" or whatever. It's "let's make sure that when a user chooses to use LVM, the initrd reflects that choice and accepts the dynamic nature of what LVM is."

johannbg (approved):
My only concern here is the size increase of the image.

teigland:
We should look into making lvm commands aware of when they are changing the root LV.  Knowing this, the commands could print messages (and maybe use a confirmation prompt) about updating initrd.  There is a similar issue if the user enables the devices file (/etc/lvm/devices/system.devices) in the initrd.  If a new device is added to system.devices and the root LV is changed to begin using that device, then the initird needs to be updated to include the new system.devices (or the root LV can't be activated.)

zhaohem:
> We should look into making lvm commands aware of when they are changing the root LV. Knowing this, the commands could print messages (and maybe use a confirmation prompt) about updating initrd. There is a similar issue if the user enables the devices file (/etc/lvm/devices/system.devices) in the initrd. If a new device is added to system.devices and the root LV is changed to begin using that device, then the initird needs to be updated to include the new system.devices (or the root LV can't be activated.)

In my knowledge, check & analyze kernel cmdline maybe a good way to know if system uses lvm2 as rootfs.

i.e.
my machine cmdline: root=/dev/mapper/system-root
then" dmsetup info" show:

```
Name:              system-root
State:             ACTIVE
Read Ahead:        1024
Tables present:    LIVE
Open count:        1
Event number:      0
Major, minor:      254, 1
Number of targets: 2
UUID: LVM-qbDc0hY8fKQVt0YekW26deKZ3WwCHwfvPdcVUUZ1Pc5pneVXAfrEEcSyYmRqTWRc
```

we could see UUID starting from "LVM-", and I see some lvm2 functions use this format style to estimate lvm uuid. (e.g: match_dm_uuid, register_device@dmeventd_vdo.c).

aafeijoo-suse:
@LaszloGombos ping

PR #2127, opened 2022-12-14
feat(overlayfs): add new overlayfs module to dracut.spec

LaszloGombos:
This is a follow-up to 40dd5c90e0efcb9ebaa9abb42a38c7316e9706bd


johannbg (approved):
Let's drop the spec file in 059...

johannbg:
@aafeijoo-suse  ping

aafeijoo-suse (approved):
> Let's drop the spec file in 059...

ok then

PR #2129, opened 2022-12-15
feat(network-wicked): remove module

LaszloGombos:
wicked is being phased out from openSUSE
https://news.opensuse.org/2022/01/28/bash-systemd-libvirt-update-in-tw/

I think having less networking option would help the project focus on quality.

From my limited amount of testing, running tests in the openSUSE container with network-wicked both NFS (test 20) and NBD (test 40) are failing. In the same openSUSE container these tests are passing both with network-manager and with network-legacy.

Since the network meta module prefers network-wicked over  network-manager and with network-legacy - by default - just installing wicked on the host system would likely result in networking not working in the generated initramfs.


Conan-Kudo (commented):
This makes sense to me, as I didn't want this module in the first place. However, I'd like a transition timeline from @aafeijoo-suse before we yank it completely.

aafeijoo-suse:
> This makes sense to me, as I didn't want this module in the first place. However, I'd like a transition timeline from @aafeijoo-suse before we yank it completely.

If this is accepted upstream, I'll send an email to the Factory mailing list with the warning notice. I don't think any other distro ships wicked, so the timeline can be managed downstream. If 058 is released too early, I'll wait for 059 and keep the module only in openSUSE.

johannbg:
> I think having less networking option would help the project focus on quality.
We should be able to strip all the way down to only support iwd and sd-networkd if we want to limit the amount of available network support upstream.

johannbg:
> If 058 is released too early, I'll wait for 059 and keep the module only in openSUSE.

Since I was planning on trying to make a release in the next couple of days what is too early?

johannbg:
> This makes sense to me, as I didn't want this module in the first place.

Interesting given that you weren't actively involved with the project nor expressed your "opinion" in the review process when Daniel made the PR for wicked being included, where did you express your opinion that you did not want this module included in the first place? Somewhere in the (open)suse community?

aafeijoo-suse:
> Since I was planning on trying to make a release in the next couple of days what is too early?

Then I'd merge this for 059, so I do not have to carry it downstream only.

Conan-Kudo:
> > This makes sense to me, as I didn't want this module in the first place.
> 
> Interesting given that you weren't actively involved with the project nor expressed your "opinion" in the review process when Daniel made the PR for wicked being included, where did you express your opinion that you did not want this module included in the first place? Somewhere in the (open)suse community?

I was working with @danimo on the entire upstreaming process. Again, just because my name isn't written on every commit doesn't mean I didn't do work on it. I worked with him to eliminate 900 patches from dracut and upstream most of it.

Stop being a jerk.

Conan-Kudo:
> > This makes sense to me, as I didn't want this module in the first place. However, I'd like a transition timeline from @aafeijoo-suse before we yank it completely.
> 
> If this is accepted upstream, I'll send an email to the Factory mailing list with the warning notice. I don't think any other distro ships wicked, so the timeline can be managed downstream. If 058 is released too early, I'll wait for 059 and keep the module only in openSUSE.

Fedora and Bottlerocket OS ship Wicked in the distribution. I packaged `wicked` for Fedora as a favor to the Amazon Linux/Bottlerocket teams (HT @stewartsmith and @nmeyerhans) for their integration of Wicked downstream, but Bottlerocket has decided to move away from Wicked now (https://github.com/bottlerocket-os/bottlerocket/issues/2449) and Amazon Linux 2022 doesn't ship it at all.

I will be retiring Wicked from Fedora shortly after this module is removed from Dracut upstream.

aafeijoo-suse (approved):

PR #2130, opened 2022-12-17
fix(dmsquash-live): allow other fstypes

LaszloGombos:
This PR provides a test case that currently fails. It than fixes the dmsquash-live module which than make the test case to pass.

The test is booting from an ntfs partition. The fix is to mount "other fstypes" as "auto" if dracut does not recognize the fstype. This than enables booting from ntfs with the ntfs3 driver without assuming that ntfs3 driver is preferred or specifically adding any kind of ntfs3 support code into dracut.

The test case introduced fails without additional fixes in the dmsquash-live module.
Link to the log of the failed run https://github.com/dracutdevs/dracut/actions/runs/3717705436/jobs/6305328592

> 2022-12-17T01:09:26.6103722Z [   18.783055] dracut: FATAL: Failed to mount block device of live image: Missing NTFS support
> 2022-12-17T01:09:26.6112994Z [   18.784772] dracut: Refusing to continue

This error message is coming from: https://github.com/dracutdevs/dracut/blob/master/modules.d/90dmsquash-live/dmsquash-live-root.sh#L136

## Checklist

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
+type det_fs > /dev/null 2>&1 || . /lib/fs-lib.sh
 
 command -v unpack_archive > /dev/null || . /lib/img-lib.sh
 
@@ -61,7 +62,7 @@ get_check_dev() {
 # Find the right device to run check on
 check_dev=$(get_check_dev "$livedev")
 # CD/DVD media check
-[ -b "$check_dev" ] && fs=$(blkid -s TYPE -o value "$check_dev")
+[ -b "$check_dev" ] && fs=$(det_fs "$check_dev")
 if [ "$fs" = "iso9660" -o "$fs" = "udf" ]; then
     check="yes"
 fi
@@ -110,7 +111,7 @@ if [ -f "$livedev" ]; then
     esac
     [ -e /sys/fs/"$fstype" ] || modprobe "$fstype"
 else
-    livedev_fstype=$(blkid -o value -s TYPE "$livedev")
+    livedev_fstype=$(det_fs "$livedev")
     if [ "$livedev_fstype" = "squashfs" ]; then
         # no mount needed - we've already got the LiveOS image in $livedev
         SQUASHED=$livedev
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
So what happens in the case that ntfs-3g and ntfs3 are on the system?

Conan-Kudo (approved):

iammattcoleman (approved):

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -110,7 +111,7 @@ if [ -f "$livedev" ]; then
     esac
     [ -e /sys/fs/"$fstype" ] || modprobe "$fstype"
 else
-    livedev_fstype=$(blkid -o value -s TYPE "$livedev")
+    livedev_fstype=$(det_fs "$livedev" "auto")
```
aafeijoo-suse, line 114: Equal to `det_fs "$livedev"`:

https://github.com/dracutdevs/dracut/blob/aca51203609954b38b41ea303a4376cbce061150/modules.d/99fs-lib/fs-lib.sh#L190

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -61,7 +62,7 @@ get_check_dev() {
 # Find the right device to run check on
 check_dev=$(get_check_dev "$livedev")
 # CD/DVD media check
-[ -b "$check_dev" ] && fs=$(blkid -s TYPE -o value "$check_dev")
+[ -b "$check_dev" ] && fs=$(det_fs "$check_dev" "")
```
aafeijoo-suse, line 65: Equal to `det_fs "$check_dev"`, and the default is "auto"

aafeijoo-suse (commented):
> changed the title ~~feat(dmsquash-live): allow other fstypes~~ fix(dmsquash-live): allow other fstypes

Now a PR with one feat and two test commits is a bug fix?

aafeijoo-suse (approved):
> As an example downstream maintainers might ignore all "feat" commits for consideration, but they might consider all the "fix" commits to cherry-pick.

I wouldn't do that. A feature (provide the "auto" fallback to allow other fs types) is not a bug fix.
Anyway, that's just a naming issue, the patch looks good.

Conan-Kudo (approved):

aafeijoo-suse:
Closing in favor of #2413

PR #2132, opened 2022-12-19
Guard kernel install hook

swsnr:
## Changes

Only run our kernel-install if the initrd generator is set to dracut, and if the layout asks for a conventional bls layout.

In particular do not run this hook if the initrd generator is `mkinitcpio`, as used by Arch's standard mkinitpcio tool, and do not run this hook if the layout is set to uki, i.e. if the user asks for a unified kernel image to be generated.

## Checklist

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -5,6 +5,13 @@ KERNEL_VERSION="$2"
 BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
 
+# Only generate a standard initrd if the installation layout asks for it.
+# In particular do not generate an initrd file if the layout is "uki", i.e. asks
+# for a unified kernel image.
+[[ "$KERNEL_INSTALL_LAYOUT" = "bls" ]] || exit 0
+# Only run dracut if asked to do so.
+[[ "$KERNEL_INSTALL_INITRD_GENERATOR" = "dracut" ]] || exit 0
+
 # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
 # So, let's skip to create initrd.
 if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
--- a/install.d/51-dracut-rescue.install
+++ b/install.d/51-dracut-rescue.install
@@ -8,6 +8,13 @@ BOOT_DIR_ABS="${3%/*}/0-rescue"
 KERNEL_IMAGE="$4"
 
 
+# Only generate a standard initrd if the installation layout asks for it.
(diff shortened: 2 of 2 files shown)


swsnr:
I am not sure how to add tests for this hook.

I'm also unsure whether the check for the "bls" layout might be overly strict.  Should perhaps rather check that the layout is not `uki`?

aafeijoo-suse:
See related discussion: #1825 #1691

swsnr:
@aafeijoo-suse I've only skimmed those threads but oh dear… 🙄  I'd have been nice to have dracut play nice with `kernel-install` but I guess this is a lost cause.   So much fuss about two lines of bash 🙈 

I'll just disable the hooks in my configuration 🤷🏿

swsnr:
That said, could dracut  at least check the layout perhaps, please?  The current hook always generates a standard initrd image, but I'd like to use kernel-install to generate a (signed) UKI with a custom hook.

johannbg:
We wont support setups where multiple initramfs generation tools are being used on the same host or an initrd is being provided/generated by something other than dracut. 

If there is no intention of using dracut ( or any other initramfs generation tool for that matter ) or for dracut  ( or any other initramfs generation tool for that matter ) being used on the host system, it should not be installed there in the first place.

Any workarounds a.k.a hacks to support such scenarios is something that downstream that intends to support such scenarios has to carry ( thus support ) themselves and quite frankly how this came about ( mkosi/systemd/redhat ) is the most gross misuse of project/company being in positions of power in the linux ecosystem and the attack on dracut the project itself and defamation campaign on the people working when it no longer became relevant to them for the sole purpose of pushing the adoption of mkosi/systemd, and their self created "standard" is the worst I have ever seen and experienced in my lifetime, in the history of (f)oss/*nix which spans decades now and I strongly suggest for other upstream initramfs project writers/maintainers to not support this both from technical and ethical reasons. 

Anyway rant aside it literally makes no sense from a technical standpoint to support this hence closing.

swsnr:
Oh dear, you got up on the very wrong side of your bed today, didn't you? 😉It's just two lines of bash, dear 🤷 Vented now? 🤗
 
So let's discuss this technically 🙂

Because technically, I think those variables are part of the kernel-install plugin interface. I'd say if you care to provide kernel-install, then do it right 🙂

Otherwise, just don't do it at all: If you don't like kernel-install for whatever reason, why not just do a proper cut and remove the kernel-install plugin entirely?

And perhaps you could read my comments again? I _do_ use dracut, but I'd like to use it with the --uefi flag which the current plugin doesn't offer as far as I can see. 

I figured changing the layout setting for kernel-install and adding your own custom plugin for that new layout was the way to go, hence this change.

PR #2134, opened 2022-12-20
Revert "fix(35network-manager): avoid restarting NetworkManager"

lkundrak:
This reverts commit 77630365aed201a729c73a9ffda0733a75f3fee4.

Its commit message states:

```
  On EL8.3 the NetworkManager keep restarting even if it exits successfully
  while waiting for Clevis to unlock. This patch ensures NetworkManager runs
  only once in initrd.
```

Yes; NetworkManager is run multiple times, so that it's able to configure interfaces that haven't been seen previously (because bus was slow to scan or device took time to initialize).

It's not clear what problem was the original commit trying to fix. I suspect there was no problem, just a misunderstanding.

### DIFF
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -2,10 +2,6 @@
 
 type source_hook > /dev/null 2>&1 || . /lib/dracut-lib.sh
 
-if [ -e /tmp/nm.done ]; then
-    return
-fi
-
 if [ -z "$DRACUT_SYSTEMD" ]; then
     # Only start NM if networking is needed
     if [ -e /run/NetworkManager/initrd/neednet ]; then


lkundrak:
This has also caused a regression in RHEL-9, where a initqueue/online hook from installer's driver disk handling brings the network down temporarily: https://bugzilla.redhat.com/show_bug.cgi?id=2051588

aafeijoo-suse:
You need to change your commit message to `revert(network-manager): avoid restarting NetworkManager` to pass the commissery check.

https://github.com/dracutdevs/dracut/actions/runs/3740372996/jobs/6348672133

ly4096x:
Hi, the original commit was because the connection establishing takes time in my case (it's an upstream problem), and frequent restarting caused the connection to fail before it makes it.

lkundrak:
> Hi, the original commit was because the connection establishing takes time in my case (it's an upstream problem), and frequent restarting caused the connection to fail before it makes it.
>
> Also, network-based Clevis auto-unlock requires time to unlock the disk. Restarting NM in the middle breaks the connection and caused Clevis fails to unlock.

That doesn't make sense to me. Restarting NetworkManager shouldn't cause any connection to fail.

Also note that with NetworkManager running with systemd, running the hook doesn't cause anything to restart and is required to run in order to mount netroot.

johannbg (approved):
LGTM

aafeijoo-suse (approved):

PR #2139, opened 2022-12-22
fix(dbus-broker): add missing sockets.target.wants/dbus.socket

aafeijoo-suse:
If `dbus.socket` is not started automatically in the initrd, all binaries that require dbus to work fail.

E.g.:
```
sh-5.2# networkctl 
Failed to connect system bus: No such file or directory
sh-5.2# resolvectl 
sd_bus_open_system: No such file or directory
sh-5.2# systemctl status dbus.socket
○ dbus.socket - D-Bus System Message Bus Socket
     Loaded: loaded (/usr/lib/systemd/system/dbus.socket; static)
     Active: inactive (dead)
   Triggers: ● dbus-broker.service
     Listen: /run/dbus/system_bus_socket (Stream)
sh-5.2# systemctl start dbus.socket
sh-5.2# networkctl 
IDX LINK   TYPE     OPERATIONAL SETUP     
  1 lo     loopback carrier     unmanaged
  2 enp1s0 ether    routable    configured

2 links listed.
sh-5.2# resolvectl 
Global
       Protocols: +LLMNR +mDNS -DNSOverTLS DNSSEC=no/unsupported
resolv.conf mode: stub
     DNS Servers: 8.8.8.8 1.1.1.1

Link 2 (enp1s0)
Current Scopes: DNS LLMNR/IPv4 LLMNR/IPv6
     Protocols: +DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
   DNS Servers: 192.168.122.1
```

## Checklist

### DIFF
--- a/modules.d/06dbus-broker/module-setup.sh
+++ b/modules.d/06dbus-broker/module-setup.sh
@@ -54,6 +54,8 @@ install() {
         "$systemduser"/dbus-broker.service \
         "$systemdsystemunitdir"/dbus.socket \
         "$systemduser"/dbus.socket \
+        "$systemdsystemunitdir"/sockets.target.wants/dbus.socket \
+        "$systemduser"/sockets.target.wants/dbus.socket \
         "$systemdsystemunitdir"/dbus.target.wants \
         busctl dbus-broker dbus-broker-launch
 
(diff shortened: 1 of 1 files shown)


johannbg:
So I have been wondering that we should be include our own type units and use those instead of sourcing the ones from the host system. ( sed-ing type units is an clear indicator we should be including our own ) 
For example if an distribution is still using  <= dbus-broker-31 this module is broken and since your are poking this you could update the module with those units  as well.

johannbg (approved):

aafeijoo-suse:
> So I have been wondering that we should be include our own type units and use those instead of sourcing the ones from the host system. ( sed-ing type units is an clear indicator we should be including our own ) For example if an distribution is still using <= dbus-broker-31 this module is broken and since your are poking this you could update the module with those units as well.

Ok, I'll do this in a separate PR, gathering all the modules that are modifying host units.

johannbg:
> Ok, I'll do this in a separate PR, gathering all the modules that are modifying host units.

Unfortunately I think we need to do this ( as in ship our own units ) for all systemd based modules since that's the only way we can ensure consistency in the initramfs generation across downstream distributions.

PR #2146, opened 2022-12-22
feat: instead of modifying systemd host units, ship our own units

aafeijoo-suse:
Asked in comment https://github.com/dracutdevs/dracut/pull/2139#issuecomment-1362771590

## Checklist


aafeijoo-suse:
> We probably end up having to do this for all modules that use systemd and have type units

Ok, let's start with this first.

PR #2154, opened 2022-12-28
fix(crypt-gpg): Do not use always --card-status

### LINKED ISSUE #2004: Boot with crypt-gpg stuck at password prompt
**Describe the bug**
I am using crypt-gpg to decrypt a luks partition that contains a volume group with the root partition. After the latest upgrade the boot is stuck at the command prompt, as entering the correct password just displays the password prompt again.

I only upgrade the kernel sporadically, so the previous update is a while back (see below for latest working version.

**Distribution used**
I am using Fedora 36.

**Dracut version**
dracut-057-3.fc36
dracut-056-1.fc36 (tested by downgrading & running dracut -f --kver ...)
linux-5.19.15-201.fc36.x86_64

Latest working versions:
dracut-056-1.fc36
linux-5.18.19-200.fc36.x86_64
initramfs generated on August 21, 2022

**Dracut config**
```
add_dracutmodules+=" resume crypt-gpg "
filesystems+=" fat msdos vfat "
# https://bugzilla.redhat.com/show_bug.cgi?id=1754055
omit_dracutmodules+=" systemd systemd-initrd dracut-systemd ifcfg network "
```

**Init system**
systemd, but it is excluded in dracut (see config above) for crypt-gpg to work

**To Reproduce**
1. When the system boots up, a password prompt is presented: Password (XXX on /dev/sdYY for /dev/sdZZ)
2. After entering the (correct) password, the prompt appears again without any error message.
    1. This is subjective, but I think it appears too quicky to have tried to decrypt the partition.
4. After repeating for 3 times, the prompt disappears, but the boot does not continue.
    1. A couple of messages are displayed (No key available with this passphrase (multiple times)), but nothing that stood out to me as being specific enough for debugging.

**Expected behavior**
After entering the correct password, the partition is decrypted and boot continues normally.

**Additional context**
I just tried a fresh install of fedora server and ran into the same issues there.

I have backported some dracut patches in the past, so I cannot 100% guarantee that the last working version above was not somehow modified, although I don't think it was.

LinkTed:
This pull request changes...

## Changes
If no GPG smart card is used then the card-status breaks the decryption of the keyfile. Therefore, use the card-status only if a smart card is used to decrypt the keyfie.

## Checklist

Fixes #2004


aafeijoo-suse (commented):
You need to fix your commit message: https://github.com/dracutdevs/dracut/actions/runs/3796692882/jobs/6457013439

Other than that, the patch looks good.

aafeijoo-suse (approved):

PR #2155, opened 2022-12-29
fix(dracut.sh): --sysroot option broken if global variables not set in conf

aafeijoo-suse:
When `$dracutsysrootdir` is set and references an existing directory, the internal global variables are never set, unless their value is assigned via configuration file (first reported in stalled PR #1975).

The simple fix would be turn all:

```sh
[[ -d $dracutsysrootdir$var ]] || var=value
```
into:
```sh
[[ -z $var || ! -d $dracutsysrootdir$var ]] && var=value
```

But this patch also includes a refactoring, removing duplicate code using the same `set_global_var` function to set and export global variables, and improving readability.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1389,18 +1389,76 @@ esac
 
 abs_outfile=$(readlink -f "$outfile") && outfile="$abs_outfile"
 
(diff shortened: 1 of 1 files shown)


johannbg (approved):
Nice cleanup

Issue #2034, opened 2022-11-04
/usr/bin/lsinitrd: line 366: zstd: command not found


FGrose:
When running the following command with dracut 057-4.fc38 in Fedora-Workstation-Live-x86_64-Rawhide-20221103.n.0.iso
`$ sudo lsinitrd /run/initramfs/live/images/pxeboot/initrd.img -m`
I receive this output:
```
Image: /run/initramfs/live/images/pxeboot/initrd.img: 17M
========================================================================
/usr/bin/lsinitrd: line 366: zstd: command not found
Version: 

dracut modules:
/usr/bin/lsinitrd: line 187: zstd: command not found
========================================================================
```
When I chroot into Fedora 36, with dracut 057-3.fc36,  I receive the expected output:
`# lsinitrd run/initramfs/live/images/pxeboot/initrd.img -m`
```
Image: run/initramfs/live/images/pxeboot/initrd.img: 17M
========================================================================
Version: 

dracut modules:
dash
nss-softokn
i18n
btrfs
dm
dmsquash-live
kernel-modules
kernel-modules-extra
rootfs-block
terminfo
udev-rules
virtiofs
biosdevname
usrmount
base
fs-lib
img-lib
shutdown
========================================================================
```
The initrd.img was built in Fedora 36 with these arguments:
`# lsinitrd run/initramfs/live/images/pxeboot/initrd.img | grep Arguments:`
```
Arguments:  --kver '6.1.0-0.rc3.20221102git8f71a2b3f435.29.fc38.x86_64' --hostonly --nomdadmconf --nolvmconf --zstd --add 'dmsquash-live dash virtiofs' --omit 'plymouth bash network-manager network ifcfg cifs iscsi nfs dracut-systemd systemd systemd-initrd memstrack' --no-early-microcode --force --add-drivers 'f2fs overlay nvme' -M
```
and successfully boots.
(I've modified `../90dmsquash-live/module-setup.sh` to allow hostonly building.)
When building in Fedora 38 Rawhide, I receive this message:
`dracut: cannot execute compression command 'zstd', falling back to default`

aafeijoo-suse:
This is a bug in Fedora, you should report it there. Here we cannot control which packages are included in each distribution that uses dracut.

PR #2156, opened 2022-12-31
Save the configuration of an image in the image.

FGrose:
Save the configuration files used to build an image in
the initrd image and incorporate them in --rebuild mode.
This facilitates image rebuilding and supplements the
build-parameter.txt record of the dracut command line
arguments.

Allow the --rebuild [output file] to be optional, as
documented. Update and fix the documentation of --rebuild,
which is hidden on some online versions of the manual page.

Also, drop a redundant existence check for config files
in the configuration directory.

### Checklist
- [✔] I have reviewed and updated the documentation.

Also, check for missing compressor software packages
and provide improved feedback to the user.  Avoid the
confusion experienced in #2034.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -79,11 +79,15 @@ Creates initial ramdisk images for preloading modules
 
   --kver [VERSION]      Set kernel version to [VERSION].
   -f, --force           Overwrite existing initramfs file.
-  [OUTPUT_FILE] --rebuild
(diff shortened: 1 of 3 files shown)


johannbg:
Ugh we need to fix/modernize our files and directory structure preferably prior to merging this then use something like a file ending with .build for this.

PR #2157, opened 2022-12-31
install: Do not undef _FILE_OFFSET_BITS

kraj:
_FILE_OFFSET_BITS is a feature test macro to determine largefile support. Usually its set to 64 on systems supporting LFS. Its also needed to be set to 64 for supporting 64bit time_t on glibc on 32bit systems. If its undefined explicitly, then 64bit time_t can not be enabled.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

This pull request changes...

## Checklist

Fixes #


kraj:
> @kraj
> 
> > If its undefined explicitly, then 64bit time_t can not be enabled.
> 
> Can you please elaborate why this is a problem exactly that 64bit time_t can not be enabled ? What fails on what distro ? Do you have a problem building this code ?

Yes, it fails to build on yocto based distros which enable 64bit time_t on 32bit architectures.

> 
> CC @nabijaczleweli

aafeijoo-suse (commented):
This was added in https://github.com/dracutdevs/dracut/commit/794b2d2c to use `fts.h` with the glibc versions that were in use at the time, but it was fixed later in https://sourceware.org/git/?p=glibc.git;a=commit;h=8b7b7f75

I think it's safe to remove the undef, but I'll try to test it next week.

kraj:
> > Running Commisery for dracutdevs/dracut, Pull Request #2157
> > Error: commit_message:1:1: error: use of type tag that's neither 'feat', 'fix' nor whitelisted (build, chore, ci, docs, perf, refactor, revert, style, test, improvement)
> > error: don't use title case in the description
> > install: Do not undef _FILE_OFFSET_BITS
> 
> @kraj Please change the commit title to
> 
> > fix(install): do not undef _FILE_OFFSET_BITS

done

aafeijoo-suse (approved):

PR #2158, opened 2023-01-01
fix(dmsquash-live): live:/dev/*

LaszloGombos:
fix a bug introduced in 71cfa2e22

make sure root is set correctly for both of the following cases
 - "root=live:/dev/*"
 - "rd.live.image root=/dev/*"

### DIFF
--- a/modules.d/90dmsquash-live/parse-dmsquash-live.sh
+++ b/modules.d/90dmsquash-live/parse-dmsquash-live.sh
@@ -34,6 +34,7 @@ case "$liveroot" in
         rootok=1
         ;;
     live:/dev/*)
+        root="live:${root#live:}"
         rootok=1
         ;;
     live:/*.[Ii][Mm][Gg] | /*.[Ii][Mm][Gg])


johannbg (approved):

aafeijoo-suse (commented):
@LaszloGombos maybe I'm missing something here. If the root cmdline parameter is `root=live:/dev/whatever` then
`liveroot=live:/dev/whatever`

https://github.com/dracutdevs/dracut/blob/10cf8e46f43a7a2094e4d861c1296db6af0e0fa1/modules.d/90dmsquash-live/parse-dmsquash-live.sh#L12-L14

so `root="live:${root}"` makes `root=live:live:/dev/whatever"`

modules.d/90dmsquash-live/parse-dmsquash-live.sh
```diff
@@ -34,6 +34,7 @@ case "$liveroot" in
         rootok=1
         ;;
     live:/dev/*)
+        root="live:${root#live:}"
```
aafeijoo-suse, line 37: Why remove the `live:` prefix to add it back? maybe the absence of this line wasn't not a bug from the beginning.

aafeijoo-suse (approved):

PR #2164, opened 2023-01-09
fix(dracut-systemd): remove unused argument

aafeijoo-suse:
The `generator_wait_for_dev` function of the dracut rootfs systemd generator only uses the first argument. Moreover, `RDRETRY` is unset at this point.

## Checklist

### DIFF
--- a/modules.d/98dracut-systemd/rootfs-generator.sh
+++ b/modules.d/98dracut-systemd/rootfs-generator.sh
@@ -104,7 +104,7 @@ esac
 GENERATOR_DIR="$1"
 
 if [ "$rootok" = "1" ]; then
-    generator_wait_for_dev "${root#block:}" "$RDRETRY"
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
Same CI failure as #2160

johannbg (approved):

aafeijoo-suse:
> If RDRETRY only used inside dracut-initqueue.sh (as this PR suggests), than perhaps you could also consider removing the export from https://github.com/dracutdevs/dracut/blob/master/modules.d/98dracut-systemd/dracut-initqueue.sh#L17 (and perhaps from https://github.com/dracutdevs/dracut/blob/master/modules.d/99base/init.sh#L165 ).

Hm, I'm not sure about that.. Also `main_loop` is exported and not used outside of the initqueue loop, I believe this was done this way to make these vars visible within initqueue scripts.

PR #2166, opened 2023-01-12
test(UEFI): add test case for UEFI boot

LaszloGombos:
Test case for UEFI boot.

Include OVMF package in containers if it is not already present. 

Install efi sub into Gentoo container using a multi-stage build step from a non-musl layer.
                                                                                                                                                                                                             This PR is limited to a new test case. No changes were made to dracut itself. 

Since test container changes are only taken into use after the PR is landed and after the daily container generation finished running, I had to run the test for the following containers in my fork where the container changes are already in place.

UEFI test passing on Debian - https://github.com/LaszloGombos/dracut/actions/runs/3919765472
UEFI test passing on Gentoo - https://github.com/LaszloGombos/dracut/actions/runs/3919773364
UEFI test passing on Fedora - https://github.com/dracutdevs-extra/dracut-canary/actions/runs/5222390021

For the rest of the test containers, the test is already passing as part of this PR CI run - see below.

In summary, this UEFI test case passes on all 5 test containers (Arch, Debian, Fedora, Gentoo, OpenSuse).


johannbg:
Note that we only stick to the 5 parent distributions ( Arch,Debian, Gentoo, Fedora,OpenSuse  ) or a unique distribution unrelated to the 5 parents ( For example Void might fall under such category since I dont think it's based on any of them  ) for test coverage since we are not going to open the flood gates and end up trying to run test cases for gazillion variants of downstream linux distributions which more or less only exist because forked off due to some "artistic" differences so we only have two options a) work with ( in this case Gentoo ) community to sort out the regression caused by the downstream change or b) drop the test case for it and let downstream just handle it themselves in their own ci testing infrastructure.

aafeijoo-suse:
> This PR is a "Test case for UEFI boot" and not "test dracut on the Alpine distro".
> 
> It seem the project is not interested to have a test case maintained for UEFI boot.

I disagree, this PR adds `TEST-18-UEFI` and `Dockerfile-Alpine`, which didn't exist before, two different things in a single commit. I think both Johann and I were only commenting on the Alpine part.

aafeijoo-suse:
> > Add an Alpine container to continue testing dracut on musl to resolve the test coverage regression
> 
> Do you have an alternative suggestion how to maintain UEFI testing and testing musl for the project without introducing Alpine ? Sorry if I missed your earlier proposal on this.

If the point is to test all the C compilers, we can look for more distributions that support uclibc, bionic, etc. 

I'm just kidding, IMHO I think we should only focus on testing the functionality that dracut provides, so the new UEFI test is great.

aafeijoo-suse:
> @aafeijoo-suse I noticed a lot of PR for UEFI generation. The project would benefit from a UEFI test case.
> 
> Do you have a proposal how could we add this test to the CI ?

Without using `systemd-boot`, but directly creating a new EFI boot entry and booting from it?

johannbg:
I think both me and @@aafeijoo-suse have been quite clear if a test cant be performed on one or all of the 5 parent distribution. ( Arch, Debian,Fedora,Gentoo, Opensuse) we wont be including it regardless what that test covers. So if the test cant be performed on one ( or all ) of the 5 distro's above then that's obviously a nack.

aafeijoo-suse:
> > Without using systemd-boot, but directly creating a new EFI boot entry and booting from it?
> 
> As it is described in the PR description,
> 
> > systemd-boot package is required to install efi stubs" and not to create new EFI boot entry.

`systemd-boot` is required (in some distros) to install the `systemd-stub`. Use other UEFI kernel boot stub.

aafeijoo-suse (approved):
Well done, thanks!

aafeijoo-suse:
https://github.com/dracutdevs/dracut/actions/runs/5213739276/jobs/9420156754?pr=2166#step:4:53 -> Test 18 failing on Fedora since https://github.com/dracutdevs/dracut/compare/e5ce4302162717647b19ed2853545c543523f4d2..2391c318c74e871d73e20d793aaba3fdf060b9fb ?

lnykryn (approved):

PR #2173, opened 2023-01-19
fix(network-manager): on machines with systemd, start the online hook…

lnykryn:
…s from nm-dispatcher

So there are three steps we need to do when a NIC appears: 1) We want udev to do its stuff
2) We want NetworkManager to setup the interface
3) We want to run the online hooks

On machines without systemd this works just fine. nm-run is called after udev settles and it starts NM in the oneshoot mode and call the hooks. Everything is done nicely synchronously.

On machines with systemd, NM is started as a daemon earlier and listens to udev events. When it gets the event that device appeared it starts the configuration.

But we only run the hooks once after NM told us that it finished the setting up everything. But if an additional device appears later, we will not start the related hooks again.

So let's use NM's dispatcher to start the hooks.

This pull request changes...

## Checklist
- [N/A ] I have reviewed and updated any documentation if relevant

### DIFF
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -1,68 +1,34 @@
 #!/bin/bash
 
 type source_hook > /dev/null 2>&1 || . /lib/dracut-lib.sh
+type nm_call_hooks > /dev/null 2>&1 || . /lib/nm-lib.sh
 
-if [ -z "$DRACUT_SYSTEMD" ]; then
-    # Only start NM if networking is needed
-    if [ -e /run/NetworkManager/initrd/neednet ]; then
-        for i in /usr/lib/NetworkManager/system-connections/* \
-            /run/NetworkManager/system-connections/* \
-            /etc/NetworkManager/system-connections/* \
-            /etc/sysconfig/network-scripts/ifcfg-*; do
-            [ -f "$i" ] || continue
-            /usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
-            break
-        done
-    fi
+if [ -e /tmp/nm.done ]; then
+    return
+fi
+
+# Only start NM if networking is needed
+if [ -e /run/NetworkManager/initrd/neednet ]; then
+    for i in /usr/lib/NetworkManager/system-connections/* \
+        /run/NetworkManager/system-connections/* \
+        /etc/NetworkManager/system-connections/* \
+        /etc/sysconfig/network-scripts/ifcfg-*; do
+        [ -f "$i" ] || continue
+        /usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
+        break
+    done
 fi
 
 if [ -s /run/NetworkManager/initrd/hostname ]; then
     cat /run/NetworkManager/initrd/hostname > /proc/sys/kernel/hostname
 fi
 
-kf_get_string() {
-    # NetworkManager writes keyfiles (glib's GKeyFile API). Have a naive
-    # parser for it.
-    #
-    # But GKeyFile will backslash escape certain keys (\s, \t, \n) but also
-    # escape backslash. As an approximation, interpret the string with printf's
-    # '%b'.
-    #
-    # This is supposed to mimic g_key_file_get_string() (poorly).
-
-    v1="$(sed -n "s/^$1=/=/p" | sed '1!d')"
-    test "$v1" = "${v1#=}" && return 1
-    printf "%b" "${v1#=}"
-}
-
-kf_unescape() {
-    # Another layer of unescaping. While values in GKeyFile format
-    # are backslash escaped, the original strings (which are in no
-    # defined encoding) are backslash escaped too to be valid UTF-8.
-    # This will undo the second layer of escaping to give binary "strings".
-    printf "%b" "$1"
-}
-
-kf_parse() {
-    v3="$(kf_get_string "$1")" || return 1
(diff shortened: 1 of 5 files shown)


lnykryn:
There are two potential problems with this approach.

* Environment might not be set correctly in the online hooks.
* nm.done was previously created after NM finished and the hooks finished; now it just waits for NM

I don't see the second point as not a problem, but I may be wrong.

Anyway, I just wanted to post this PoC to get feedback.

bengal:
There are some integration tests failures that must be investigated; apart from that the approach looks good to me. 

> nm.done was previously created after NM finished and the hooks finished; now it just waits for NM

I initially suggested it, but at a second thought I'm not sure it's actually a problem. Online and netroot hooks probably can be called after the initqueue is started.

aafeijoo-suse:
> There are some integration tests failures that must be investigated; apart from that the approach looks good to me.

Tests fedora/network-manager/20 and 40 were already intermittently failing before this change.

lnykryn:
In the latest version, I got rid of nm.done completely on machines with systemd.
It was pointless, since we always created it in nm-wait-online-initrd.service which is run before initqueue.

lnykryn:
So the failed tests were not a false-positive. There were two bugs in the code. First, we need to source the dracut-status before we start the online hooks, and I had to move the start of the dispatcher to earlier because there was some weird race when using nbd.

johannbg (approved):
LGTM

johannbg:
@lnykryn just a heads up at some point Fedora/RHEL will have to carry any /etc/sysconfig/$foo downstream since it's Fedora/RHEL specific. Back in the day I was actually planning on removing it ( via feature process in Fedora ) since it should not be needed anymore in atleast Fedora but you know bureaucracy + politics...

dtardon:
> > @lnykryn just a heads up at some point Fedora/RHEL will have to carry any /etc/sysconfig/$foo downstream since it's Fedora/RHEL specific. Back in the day I was actually planning on removing it ( via feature process in Fedora ) since it should not be needed anymore in atleast Fedora but you know bureaucracy + politics...
> 
> In context of network, it should not be an issue. In fedora, we are slowly migrating to native NM configs instead of ifcfg files.

Btw, one of the blockers for this is dracut. Modules `45ifcfg` and `80cms` both create ifcfg files.

aafeijoo-suse (commented):
Sorry, you have to merge the changes added in #2134.
Other than that, everything looks good and works fine, thanks.

johannbg:
> Btw, one of the blockers for this is dracut. Modules `45ifcfg` and `80cms` both create ifcfg files.

You probably mean rhel not dracut since it's not hard justifying dropping these modules here upstream even as is ( without suitable replacement ).

modules.d/35network-manager/nm-run.sh
```diff
@@ -1,68 +1,34 @@
...
-        for i in /usr/lib/NetworkManager/system-connections/* \
-            /run/NetworkManager/system-connections/* \
-            /etc/NetworkManager/system-connections/* \
-            /etc/sysconfig/network-scripts/ifcfg-*; do
-            [ -f "$i" ] || continue
-            /usr/sbin/NetworkManager --configure-and-quit=initrd --no-daemon
-            break
-        done
-    fi
+if [ -e /tmp/nm.done ]; then
```
aafeijoo-suse, line 6: #2134 removed:

```bash
if [ -e /tmp/nm.done ]; then
    return
fi
```
from `nm-run.sh`

aafeijoo-suse (changes requested):

bengal:
If the command line contains a hostname (for example `ip=192.168.50.201::192.168.50.1:24:myhostname:eth0:none`), the generator writes the hostname to `/run/NetworkManager/initrd/hostname` and expects `nm-run.sh` to set it in the system. Now `nm-run.sh` no longer runs, and that hostname is not set; we need to find another way to set it, perhaps the first time the dispatcher runs, or directly from the generator.

lnykryn:
Sorry, I still plan to work on this.

bengal:
I think there are 2 issues still open:

 - the current version of the PR doesn't set the hostname for command line e.g. `ip=192.168.50.201::192.168.50.1:24:myhostname:eth0:none`. That can be solved by something like:
```diff
--- a/modules.d/35network-manager/nm-wait-online-initrd.service
+++ b/modules.d/35network-manager/nm-wait-online-initrd.service
@@ -6,11 +6,12 @@ Before=network-online.target
 Before=dracut-initqueue.service
 ConditionPathExists=/run/NetworkManager/initrd/neednet

 [Service]
 Type=oneshot
 ExecStart=/usr/bin/nm-online -s -q -t 3600
+ExecStartPost=-/bin/sh -c '[ -s /run/NetworkManager/initrd/hostname ] && cat /run/NetworkManager/initrd/hostname > /proc/sys/kernel/hostname'
 RemainAfterExit=yes

 [Install]
 WantedBy=initrd.target
 WantedBy=network-online.target
```

  Or maybe the generator should directly set the hostname in kernel instead of writing `/run/NetworkManager/initrd/hostname".


 - there is a race condition. The "up" dispatcher script isn't guaranteed to finish before the "network-online" event. So, it can happen that the initqueue starts and there is nothing to do because the dispatcher script didn't run yet. We need to add a synchronization mechanism; for example, NetworkManager should expose a "busy" flag that is set to 1 when scripts are running. nm-online should gain a new option to wait for the flag to settle, and `nm-wait-online-initrd.service` should use it in addition to `-s`.

PR #2177, opened 2023-01-31
fix(network-manager): add "After" dependency on dbus.service

bengal:
## Changes

During shutdown, there is no ordering dependency between the nm-initrd service and the D-Bus daemon, and so the latter can be stopped before. This causes issues to NetworkManager, especially when team interfaces are present because NM will see teamd dropping from the bus and will try to reactivate the connection.

Add a `After` dependency to make sure the D-Bus daemon is stopped after NM on shutdown.

## Checklist


aafeijoo-suse (approved):
Thanks!

fpokryvk:
Hello, NetworkManager QE here, I can confirm this change makes VLAN over team tests stable in RHEL9.

johannbg (approved):
LGTM

PR #2180, opened 2023-02-02
Improvements for multinic test with NetworkManager

bengal:
This pull request hopefully fixes the 50-multinic test that used to randomly fail with the network-manager module.

## Checklist

### DIFF
--- a/test/TEST-50-MULTINIC/test.sh
+++ b/test/TEST-50-MULTINIC/test.sh
@@ -63,7 +63,7 @@ client_test() {
     # shellcheck disable=SC2034
     declare -i disk_index=0
     qemu_add_drive_args disk_index disk_args "$TESTDIR"/marker.img marker
-    cmdline="$cmdline rd.net.timeout.dhcp=3"
+    cmdline="$cmdline rd.net.timeout.dhcp=30"
 
     # Invoke KVM and/or QEMU to actually create the target filesystem.
     "$testdir"/run-qemu \
--- a/test/TEST-50-MULTINIC/test.sh
+++ b/test/TEST-50-MULTINIC/test.sh
@@ -169,7 +169,7 @@ test_client() {
     # bridge, where only one interface is actually connected
     client_test "MULTINIC bridging" \
         00 01 02 \
-        "root=nfs:192.168.50.1:/nfs/client ip=bridge0:dhcp  bridge=bridge0:enp0s1,enp0s5,enp0s6" \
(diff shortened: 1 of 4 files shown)


.github/workflows/integration.yml
```diff
@@ -72,7 +72,7 @@ jobs:
                         "30",
                         "35",
                         "40",
-                        # "50", # times out
+                        "50",
```
aafeijoo-suse, line 75: But I'm not sure if we should re-enable test 50, although it's working now with the NetworkManager, it still fails with the network-legacy module:

https://github.com/dracutdevs/dracut/actions/runs/4076254978/jobs/7075949251

test/TEST-50-MULTINIC/test.sh
```diff
@@ -365,7 +362,7 @@ test_setup() {
     # Make server's dracut image
     "$basedir"/dracut.sh -l -i "$TESTDIR"/overlay / \
         -m "dash rootfs-block debug kernel-modules watchdog qemu network network-legacy" \
-        -d "af_packet piix ide-gd_mod ata_piix ext3 sd_mod nfsv2 nfsv3 nfsv4 nfs_acl nfs_layout_nfsv41_files nfsd e1000 i6300esb ib700wdt" \
```
aafeijoo-suse, line 368: Multiple `modprobe: FATAL: Module ib700wdt not found in directory /lib/modules/6.1.9-200.fc37.x86_64`:

https://github.com/dracutdevs/dracut/actions/runs/4076254978/jobs/7075949251#step:4:386

`ib700wdt` module required by: https://github.com/dracutdevs/dracut/blob/master/modules.d/04watchdog/watchdog.sh

aafeijoo-suse (commented):
Thanks for you help here!

test/TEST-50-MULTINIC/test.sh
aafeijoo-suse: > Not sure if this is in scope for this PR, but we could consider removing ib700wdt from watchdog module.

Indeed, out of the scope of this PR.

bengal: Okay, I'll file a separate MR for the watchdog changes.

.github/workflows/integration.yml
bengal: Okay

bengal:
I have removed the watchdog changes and left the test disabled for now since there are still some failures. Nevertheless, I think these changes should be merged because they improve things.

aafeijoo-suse (approved):

PR #2181, opened 2023-02-02
fix(network,dbus): improve dependency checking

LaszloGombos:
This PR removes distribution dependent logic from the code.

Decouple implementation details of the backend dependencies from the meta module.

Call [check_module](https://github.com/dracutdevs/dracut/blob/master/dracut-init.sh#L985) for each submodule instead of reimplementing sub-package dependency checking inside the meta module itself.

[check_module](https://github.com/dracutdevs/dracut/blob/master/dracut-init.sh#L985) calls the module's [check()](https://github.com/dracutdevs/dracut/blob/master/dracut-init.sh#L784) function from [module-setup.sh](https://github.com/dracutdevs/dracut/blob/master/dracut-init.sh#L779) to check if it can be installed or not.

Add missing NetworkManager binary to the network-manager dracut module check. 

Fixes: #1756

### DIFF
--- a/modules.d/09dbus/module-setup.sh
+++ b/modules.d/09dbus/module-setup.sh
@@ -19,7 +19,7 @@ depends() {
         fi
     done
 
-    if find_binary dbus-broker &> /dev/null; then
+    if check_module "dbus-broker"; then
         echo "dbus-broker"
         return 0
     else
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -2,7 +2,7 @@
 
 # called by dracut
 check() {
-    require_binaries sed grep || return 1
+    require_binaries sed grep NetworkManager || return 1
 
     # do not add this module by default
     return 255
--- a/modules.d/40network/module-setup.sh
+++ b/modules.d/40network/module-setup.sh
@@ -17,11 +17,11 @@ depends() {
     done
 
     if [ -z "$network_handler" ]; then
-        if [[ -e $dracutsysrootdir$systemdsystemunitdir/connman.service ]]; then
+        if check_module "connman"; then
             network_handler="connman"
-        elif [[ -x $dracutsysrootdir/usr/libexec/nm-initrd-generator ]] || [[ -x $dracutsysrootdir/usr/lib/nm-initrd-generator ]]; then
+        elif check_module "network-manager"; then
             network_handler="network-manager"
-        elif [[ -x $dracutsysrootdir$systemdutildir/systemd-networkd ]]; then
+        elif check_module "systemd-networkd"; then
             network_handler="systemd-networkd"
         else
             network_handler="network-legacy"


johannbg:
What do you think you solve by checking for the existence of the module but not the binaries?

johannbg:
What are you trying to solve here? If a binary is installed then the relevant module that depends on that binary should be installed as well so there should not be any need for both checking the module and the binary

johannbg:
>Decouple means separate or disengage. It means if one of the sub modules needs different binary check, than we should not have to keep changing the meta module - we should be just changing the module code itself that needs different binaries.

No we should not. 

We do not want to add complexity the sub modules themselves, we want to simplify them and to do so we need introduce more sub modules not add complexity to existing one which will result in more bugs not less and will be harder to dissect.

For example when upstream make incompatible changes to their binary between releases that a module depends upon we deal with that by introducing a new sub module of said component and turn the existing module into a meta module, this provides a migration path downstream for upstream without disrupting user configuration and a *clean* package dependency downstream as well ( if required ).  

What's the use case for a module suddenly needing a different binary check if not above?

ahesford:
This approach seems to me the cleanest, most natural, most easily maintained, most obvious and simplest way to allow meta-modules to delegate their tasks to specific backends. Let the specific backend modules encapsulate the knowledge about which binaries (and other system conditions) are necessary for proper functioning and leave the meta-module to ask only whether the backend(s) it wishes to include report that they will function. What is not to like about this?

Void Linux has been bitten by the half-assed check in the network meta-module that uses the existence of some executable as evidence that a backend module should be preferred when that module is non-functional without systemd. With this fix, the problem goes away. The backend module can be omitted from the Void package, which allows the meta-module to properly detect its absence and move on to the next choice.

I understand that dracut is a product of Red Hat, but the world is not solely composed of Red Hat employees interesting in dogfooding their distribution. Simple fixes like this broaden the utility of the software without hinderance.

lnykryn:
I agree with @LaszloGombos. I hate all the checks based on the existence of a file or checking the presence of some switch in --help outputs. So, if we want to de-duplicate these checks, I am all for it. 

Honestly, I would drop these checks altogether and tell downstreams and users to set their preferred implementation via drop-in config instead of automagically determining the proper setup. But that is because in RHEL and Fedora we have the resources to write such defaults. I understand that smaller distros are glad if things work out of the box.

Before giving a proper review, I need to dig into one thing that does not sound right to me (but that is probably not related to the patch but to my ignorance)

Why do we also have this check https://github.com/dracutdevs/dracut/blob/master/modules.d/40network/module-setup.sh#L12 ? Should that not be enough? We run the network module after the submodules.

lnykryn:
Yep, you are right. So overall, LGTM.

lnykryn (approved):

Conan-Kudo (approved):

johannbg:
> This approach seems to me the cleanest, most natural, most easily maintained, most obvious and simplest way to allow meta-modules to delegate their tasks to specific backends. Let the specific backend modules encapsulate the knowledge about which binaries (and other system conditions) are necessary for proper functioning and leave the meta-module to ask only whether the backend(s) it wishes to include report that they will function. What is not to like about this?

The main problem with the network meta module is that the network submodules can be used alongside eachother at the same time and the only way to deal with that is in the meta module itself.

If the argument is we expect downstream to handle this ( via config or packaging or by something else ) we can just as well drop the meta modules and  just make the assumption that downstream will take care of this...

pvalena:
To me this sounds great, as it adds an abstraction layer instead of hardcoded checks in various places. This way you can change the module check in one place, and it's also not dependent on the module implementation itself. I cannot see any disadvantage.

PR #2182, opened 2023-02-03
fix(url-lib.sh): nfs_already_mounted() with trailing slash in nfs path

mmatsuya:
nfs_already_mounted() doesn't work when the installation ISO and kickstart file on a same NFS share are specified with inst.repo and inst.ks boot parameter as below.

   inst.repo=nfs:192.168.1.1:/home/data/rhel9.iso inst.ks=nfs:192.168.1.1:/home/data/ks.cfg

   NOTE: /home/data is configured for nfs share on 192.168.1.1

One problem is a file (not a directory) was passed into nfs_already_mounted(). nfs_already_mounted() is the function to judge if the given directory is already mounted. So, filepath should be passed in nfs_fetch_url().

The other problem is about the trailing slash in the nfs path in /proc/mounts.

The /proc/mounts has an entry after nfs mount of inst.repo.

   192.168.1.1:/data/   /run/install/isodir  nfs  ro,relatime,<snip>

In this case, nfs_already_mounted() returns "/run/install/isodir//home/data/ks.cfg" wrongly. The following is from the log.

[   14.556279] localhost.localdomain dracut-initqueue[1282]: ///lib/url-lib.sh@156(nfs_fetch_url): nfs_already_mounted 192.168.122.1 /home/data/ks.cfg
[   14.556279] localhost.localdomain dracut-initqueue[1282]: ///lib/url-lib.sh@137(nfs_already_mounted): local server=192.168.122.1 path=/home/data/ks.cfg s= p=
...
[   14.654966] localhost.localdomain dracut-initqueue[1282]: ///lib/url-lib.sh@140(nfs_already_mounted): '[' 192.168.122.1 = 192.168.122.1 ']'
[   14.654966] localhost.localdomain dracut-initqueue[1282]: ///lib/url-lib.sh@141(nfs_already_mounted): '[' /home/data/ks.cfg = /home/data/ ']'
[   14.654966] localhost.localdomain dracut-initqueue[1282]: ///lib/url-lib.sh@143(nfs_already_mounted): str_starts /home/data/ks.cfg /home/data/
[   14.654966] localhost.localdomain dracut-initqueue[1282]: ///lib/dracut-lib.sh@51(str_starts): '[' ks.cfg '!=' /home/data/ks.cfg ']'
[   14.654966] localhost.localdomain dracut-initqueue[1282]: ///lib/url-lib.sh@144(nfs_already_mounted): echo /run/install/isodir//home/data/ks.cfg
...
[   14.658069] localhost.localdomain dracut-initqueue[934]: //lib/url-lib.sh@156(nfs_fetch_url): mntdir=/run/install/isodir//home/data/ks.cfg

This function doesn't expect the trailiing slash of the nfs path in /proc/mounts, so it should be removed before processing it.

This pull request changes...

## Checklist

Fixes #
- RHBZ#2166727


mmatsuya:
Oh, sorry. I missed your comment. I will stop test it here.

mmatsuya:
> Also, the goal is to make sure that the code you changing gets executed by the test, it does not need to replicate your exact scenario.

Was it enough to run nfs_already_mounted() and check the result without replicating the problem?
Was it ok even if the test doesn't check the fix of this problem?
nfs_already_mounted() depends on the nfs mount entries in /proc/mounts, and the trailing slash of the nfs path is required to replicate the problem. So, I was trying to make the entry in /proc/mounts by mounting another share from client. if the above question is yes, the change for test case should be simpler:)

mmatsuya:
As a result, I added a test to check the behavior of nfs_fetch_url() into TEST-20-NFS. nfs_already_mounted() is called in it. I didn't make any additional nfs mounts, but I just used the existing mount for the nfs root. /nfs/client/root/fetchfile was located on the nfs server. And if a nfs client can fetch the file with nfs_already_mounted(), the result is written into marker2.img. And, the result in the marker2.img is checked in client_test().

Hopefully, this meets what you expected.

johannbg (approved):
LGTM

PR #2184, opened 2023-02-03
NVMeoF / TCP boot support

mwilck:
## Changes

Add support for booting from NVMe over Fabrics according to the [NVM Express Boot Specification v1.0](https://nvmexpress.org/developers/nvme-boot-specification/) to dracut. Note that this is orthogonal to boot from NVMeoF over Fibre Channel, which is already supported by dracut (using features exclusive of the FC transport). The NVMe-oF Boot Spec is designed to be transport-agnostic (but v1.0 contains support for the TCP transport only).

The spec defines an ACPI table called "NBFT", in which the system firmware stores information about network configuration and NVMe subsystems used for booting (similar to the iBFT table for iSCSI). The code in this PR reads the NBFT table (parsing JSON as printed by nvme-cli), transforms it into `ip=` parameters that dracut's network setup code understands, and connects to the NVMe subsystems after network setup is complete. The code is modeled after the respective code for IBFT boot in the `95iscsi` module.

This PR depends on https://github.com/linux-nvme/nvme-cli/pull/1791, which implements the nvme-cli subcommands `nvme show-nbft` and `nvme connect-nbft`. Therefore this PR **should not be merged** before the syntax and semantics of the nvme subcommands have been consolidated (i.e. before the nvme-cli PR has been merged). In spite of that, I submit this PR here as RFC, to make the community aware of the feature. While it's expected that the nvme-cli command-line API will change somewhat, the fact that the NBFT content is printed in JSON format will not change, and the structure of the JSON itself will probably change only slightly, if at all. If the API changes in the nvme-cli PR, this PR will be adapted .

## Checklist
- [ x] I have tested it locally
- [ x] I have reviewed and updated any documentation if relevant

## Notes

It is difficult to get one's hands on a sample setup for testing and playing around with boot over NVMeoF. A proof-of-concept setup using the Linux NVMe target (`nvmet`) and qemu is under preparation and will be published soon under https://github.com/timberland-sig.

### DIFF
--- a/modules.d/95nvmf/module-setup.sh
+++ b/modules.d/95nvmf/module-setup.sh
@@ -50,7 +50,7 @@ depends() {
 # called by dracut
 installkernel() {
     instmods nvme_fc lpfc qla2xxx
-    hostonly="" instmods nvme_tcp nvme_fabrics
+    hostonly="" instmods nvme_tcp nvme_fabrics 8021q
 }
 
 # called by dracut
--- a/modules.d/95nvmf/module-setup.sh
+++ b/modules.d/95nvmf/module-setup.sh
@@ -2,7 +2,7 @@
 
 # called by dracut
 check() {
-    require_binaries nvme || return 1
+    require_binaries nvme jq || return 1
     [ -f /etc/nvme/hostnqn ] || return 255
     [ -f /etc/nvme/hostid ] || return 255
 
@@ -25,17 +25,27 @@ check() {
         [[ $trtype == "fc" ]] || [[ $trtype == "tcp" ]] || [[ $trtype == "rdma" ]]
     }
 
+    has_nbft() {
+        local f found=
+        for f in /sys/firmware/acpi/tables/NBFT*; do
+            [ -f "$f" ] || continue
+            found=1
+            break
+        done
+        [[ $found ]]
+    }
+
     [[ $hostonly ]] || [[ $mount_needs ]] && {
         pushd . > /dev/null
         for_each_host_dev_and_slaves is_nvmf
         local _is_nvmf=$?
         popd > /dev/null || exit
         [[ $_is_nvmf == 0 ]] || return 255
-        if [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ]; then
-            if [ ! -f /etc/nvme/discovery.conf ]; then
-                echo "No discovery arguments present"
-                return 255
-            fi
+        if [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ] \
+            && [ ! -f /etc/nvme/discovery.conf ] \
+            && [ ! -f /etc/nvme/config.json ] && ! has_nbft; then
+            echo "No discovery arguments present"
+            return 255
         fi
     }
     return 0
@@ -126,8 +136,9 @@ install() {
     inst_multiple ip sed
 
     inst_script "${moddir}/nvmf-autoconnect.sh" /sbin/nvmf-autoconnect.sh
+    inst_script "${moddir}/nbftroot.sh" /sbin/nbftroot
 
-    inst_multiple nvme
+    inst_multiple nvme jq
     inst_hook cmdline 92 "$moddir/parse-nvmf-boot-connections.sh"
     inst_simple "/etc/nvme/discovery.conf"
     inst_simple "/etc/nvme/config.json"
--- a/modules.d/95nvmf/module-setup.sh
+++ b/modules.d/95nvmf/module-setup.sh
@@ -130,6 +130,7 @@ install() {
     inst_multiple nvme
     inst_hook cmdline 92 "$moddir/parse-nvmf-boot-connections.sh"
     inst_simple "/etc/nvme/discovery.conf"
+    inst_simple "/etc/nvme/config.json"
     inst_rules /usr/lib/udev/rules.d/71-nvmf-iopolicy-netapp.rules
     inst_rules "$moddir/95-nvmf-initqueue.rules"
     dracut_need_initqueue
--- a/modules.d/95nvmf/nvmf-autoconnect.sh
+++ b/modules.d/95nvmf/nvmf-autoconnect.sh
@@ -1,5 +1,54 @@
-#!/bin/bash
+#!/bin/sh
+# Argument $1 is "settled", "online", or "timeout", indicating
+# the queue from which the script is called.
+# In the "timeout" case, try everything.
+# Otherwise, try options according to the priorities below.
 
-[ -f /sys/class/fc/fc_udev_device/nvme_discovery ] || exit 1
-echo add > /sys/class/fc/fc_udev_device/nvme_discovery
+[ "$RD_DEBUG" != yes ] || set -x
+
+if [ "$1" = timeout ]; then
+    [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ] \
+        || echo add > /sys/class/fc/fc_udev_device/nvme_discovery
+    /usr/sbin/nvme connect-all
+    exit 0
+fi
+
+NVMF_HOSTNQN_OK=
+[ ! -f "/etc/nvme/hostnqn" ] || [ ! -f "/etc/nvme/hostid" ] || NVMF_HOSTNQN_OK=1
+
+# Only nvme-cli 2.5 or newer supports the options --nbft and --no-nbft
+# for the connect-all command.
+# Make sure we don't use unsupported options with earlier versions.
+NBFT_SUPPORTED=
+# shellcheck disable=SC2016
+/usr/sbin/nvme connect-all --help 2>&1 | sed -n '/[[:space:]]--nbft[[:space:]]/q1;$q0' \
+    || NBFT_SUPPORTED=1
+
+if [ -e /tmp/nvmf-fc-auto ] && [ "$NVMF_HOSTNQN_OK" ] \
+    && [ -f /sys/class/fc/fc_udev_device/nvme_discovery ]; then
+    # prio 1: cmdline override "rd.nvmf.discovery=fc,auto"
+    echo add > /sys/class/fc/fc_udev_device/nvme_discovery
+    exit 0
+fi
+if [ "$NBFT_SUPPORTED" ] && [ -e /tmp/valid_nbft_entry_found ]; then
+    # prio 2: NBFT
+    /usr/sbin/nvme connect-all --nbft
+    exit 0
+fi
+if [ -f /etc/nvme/discovery.conf ] || [ -f /etc/nvme/config.json ] \
+    && [ "$NVMF_HOSTNQN_OK" ]; then
+    # prio 3: configuration from initrd and/or kernel command line
+    # We can get here even if "rd.nvmf.nonbft" was given, thus use --no-nbft
+    if [ "$NBFT_SUPPORTED" ]; then
+        /usr/sbin/nvme connect-all --no-nbft
+    else
+        /usr/sbin/nvme connect-all
+    fi
+    exit 0
+fi
+if [ "$NVMF_HOSTNQN_OK" ] \
+    && [ -f /sys/class/fc/fc_udev_device/nvme_discovery ]; then
+    # prio 4: no discovery entries, try NVMeoFC autoconnect
+    echo add > /sys/class/fc/fc_udev_device/nvme_discovery
+fi
 exit 0
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -898,6 +898,15 @@ NVMf
 **rd.nonvmf**::
     Disable NVMf
 
+**rd.nvmf.nonbft**::
+    Disable connecting to targets from the NVMe Boot Firmware Table. Without
+    this parameter, NBFT connections will take precedence over _rd.nvmf.discover_.
+
+**rd.nvmf.nostatic**::
+    Disable connecting to targets that have been statically configured when
+    the initramfs was built. Targets specified with rd.nvmf.discover on the
+    kernel command line will still be tried.
+
 **rd.nvmf.hostnqn=**__<hostNQN>__::
     NVMe host NQN to use
 
--- a/modules.d/95nvmf/nbftroot.sh
+++ b/modules.d/95nvmf/nbftroot.sh
@@ -1,0 +1,5 @@
+#! /bin/sh
+# This script is called from /sbin/netroot
+
+/sbin/nvmf-autoconnect.sh online
+exit 0
--- a/modules.d/95nvmf/parse-nvmf-boot-connections.sh
+++ b/modules.d/95nvmf/parse-nvmf-boot-connections.sh
@@ -17,11 +17,223 @@
 # specify any discover parameters for FC.
 #
 
-type is_ip > /dev/null 2>&1 || . /lib/net-lib.sh
+command -v getarg > /dev/null || . /lib/dracut-lib.sh
+command -v is_ip > /dev/null || . /lib/net-lib.sh
+
+## Sample NBFT output from nvme show-nbft -H -s -d -o json
+# [
+#   {
+#     "filename":"/sys/firmware/acpi/tables/NBFT",
+#     "host":{
+#       "nqn":"nqn.2014-08.org.nvmexpress:uuid:d6f07002-7eb5-4841-a185-400e296afae4",
+#       "id":"111919da-21ea-cc4e-bafe-216d8372dd31",
+#       "host_id_configured":0,
+#       "host_nqn_configured":0,
+#       "primary_admin_host_flag":"not indicated"
+#     },
+#     "subsystem":[
+#       {
+#         "index":1,
+#         "num_hfis":1,
+#         "hfis":[
+#           1
+#         ],
+#         "transport":"tcp",
+#         "transport_address":"192.168.100.216",
+#         "transport_svcid":"4420",
+#         "subsys_port_id":0,
+#         "nsid":1,
+#         "nid_type":"uuid",
+#         "nid":"424d1c8a-8ef9-4681-b2fc-8c343bd8fa69",
+#         "subsys_nqn":"timberland-01",
+#         "controller_id":0,
+#         "asqsz":0,
+#         "pdu_header_digest_required":0,
+#         "data_digest_required":0
+#       }
+#     ],
+#     "hfi":[
+#       {
+#         "index":1,
+#         "transport":"tcp",
+#         "pcidev":"0:0:2.0",
+#         "mac_addr":"52:54:00:4f:97:e9",
+#         "vlan":0,
+#         "ip_origin":63,
+#         "ipaddr":"192.168.100.217",
+#         "subnet_mask_prefix":24,
+#         "gateway_ipaddr":"0.0.0.0",
+#         "route_metric":0,
+#         "primary_dns_ipaddr":"0.0.0.0",
+#         "secondary_dns_ipaddr":"0.0.0.0",
+#         "dhcp_server_ipaddr":"",
+#         "this_hfi_is_default_route":1
+#       }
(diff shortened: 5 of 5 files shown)


mwilck:
Adding @aafeijoo-suse , @johnmeneghini, @igaw, @hreinecke for information.

aafeijoo-suse (commented):
So far it looks good, as I've followed your development process and already reviewed it internally over the past months. Thanks for your work here.

Just two comments:
- As you said, we must wait until `nvme-cli` consolidates the syntax of the commands used here.
- It'd be great to add a new test for the `nvmf` module in the future (similar to `TEST-30-ISCSI`), but that shouldn't block this PR, as we don't have a test for the current implementation at this time, and the `nvme-cli` changes need to be available from our CI.

modules.d/95iscsi/module-setup.sh
```diff
@@ -161,7 +161,7 @@ installkernel() {
     local _funcs='iscsi_register_transport'
 
     instmods bnx2i qla4xxx cxgb3i cxgb4i be2iscsi qedi
-    hostonly="" instmods iscsi_tcp iscsi_ibft crc32c iscsi_boot_sysfs
+    hostonly="" instmods iscsi_tcp iscsi_ibft crc32c iscsi_boot_sysfs 8021q
```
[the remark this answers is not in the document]
mwilck: > Can you please help to understand why are you also need to change the iscsi module ?

NVMeoF / NBFT and iSCSI  / iBFT boot are very similar concepts. I used the iSCSI code as template for what I did for NVMeoF.  While doing that, I ran into some issues which obviously affect the iSCSI code, too.

> Is this change in the scope of this PR or perhaps should land as a separate PR ?

The iSCSI changes are in separate patches, so I could easily separate them out if you prefer. Still, they are closely related to the nvmf changes, and I figured it might make sense to submit both together.

mwilck:
Updated with some minor IPv6 fixes (again, one for iSCSI, too).

I will squash these fixes before this is merged. For now I guess keeping them separate makes the review easier.

mwilck:
@LaszloGombos, I have created #2192 now, which separates out the non-nvmf related changes from this PR.
I am rebasing this PR onto #2192.

mwilck:
I rebased onto #2192 now, and added another commit (91e1ee9), which enables a retry logic for nvme connects.

aafeijoo-suse (commented):
Thanks Martin, you need to fix the `shfmt` issues in `nvmf-autoconnect.sh` and `parse-nvmf-boot-connections.sh`:

https://github.com/dracutdevs/dracut/actions/runs/4136435631/jobs/7150292822#step:4:56

Also, I assume the part of commit https://github.com/dracutdevs/dracut/commit/91e1ee90589ef5423c734d1b2c1ac81453ace857 message "_Also, make sure that the initqueue script doesn't call exit()._" is outdated, right?

I see the `nvmf-autoconnect.sh` never checks if the connection is already established, right? What if it runs again in this case, have you tried that? I'm just pointing this out, because `initqueue --online` without `--onetime` will run the `nvmf-autoconnect.sh` script for **_every_** network interface configured on the kernel command line, and also `initqueue --settled` and `--timeout` can run after the connection is set.

Other than that, the changes look good.

mwilck:
> Thanks Martin, you need to fix the `shfmt` issues in `nvmf-autoconnect.sh` and `parse-nvmf-boot-connections.sh`:
> 
> https://github.com/dracutdevs/dracut/actions/runs/4136435631/jobs/7150292822#step:4:56

Argh. Why do I keep forgetting to run it? Sorry.

> Also, I assume the part of commit [91e1ee9](https://github.com/dracutdevs/dracut/commit/91e1ee90589ef5423c734d1b2c1ac81453ace857) message "_Also, make sure that the initqueue script doesn't call exit()._" is outdated, right?
> 
> I see the `nvmf-autoconnect.sh` never checks if the connection is already established, right? 

No, that's currently impossible. It would be _extremely hard and error-prone_ to get such a test right in a shell script. `nvme connect-nbft` does check this, though, and won't make another connection attempt for already-established connections. I have made a [remark](https://github.com/linux-nvme/nvme-cli/pull/1791#discussion_r1101275433) 
on the nvme-cli PR, asking for a more graceful error handling in the special case that all NBFT-specified connections are already established.

> What if it runs again in this case, have you tried that? 

Yes, it does nothing but prints a (currrently misleading) error message and exits with error status. Anyway, note that the NBFT can include multiple connections and it's possible that some are established while others are not. `nvme connect-nbft` does "the right thing" in this case (tries to connect to the non-yet-connected targets and skips others), but it isn't easy to reflect this in the exit status. Admittedly, test coverage for such scenarios is very limited currently.

> I'm just pointing this out, because `initqueue --online` without `--onetime` will run the `nvmf-autoconnect.sh` script for **_every_** network interface configured on the kernel command line, and also `initqueue --settled` and `--timeout` can run after the connection is set.

If this happens we'll see a lot of warnings from nvme-cli. But we will work on fixing that on the nvme-cli side.

The thing is, the single attempt that the original code is doing won't be sufficient in all cases, and it's better to see a few error messages than to fail booting. As this code evolves, we may be able to do this more elegantly in the future, without the need to retry blindly. But as I said, iSCSI does it the same way, and it has been around for more than a decade.

mwilck:
Pushed shfmt fixes.

aafeijoo-suse:
> If this happens we'll see a lot of warnings from nvme-cli. But we will work on fixing that on the nvme-cli side.

Thanks for clarifying.

> Pushed shfmt fixes.

Sorry to bother you again with this minor format requirement, but it's still failing for `nvmf-autoconnect.sh`:

https://github.com/dracutdevs/dracut/actions/runs/4143068641/jobs/7164486549#step:4:15

johannbg:
We probably want to have this in a separated new NVMeoF module also @mwilck I notice that you are poking the network legacy module which we are planning on dropping so my question to you is how much do you think it's being used in the wild since downstream should have switched to iwd, connman,networkd,nm by now?

mwilck:
> We probably want to have this in a separated new NVMeoF module

I don't think that's a good idea. It would cause the nvmf and nvmeof modules to possibly  conflict. There's is really just nvmeof (NVMe over Fabrics). Fibre Channel (which is what the current `nvmf` module supports) is just one of multiple NVMeoF transports. By integrating NVMeoF / TCP into the `nvmf` module, we are able to eliminate possible conflicts.  My implementation in the PR still gives precendence to the NVMeoF / FC boot method.

I agree that the name`nvmf` is not ideal for a transport-independent NVMeoF module, but I wouldn't want to change the existing name.

> @mwilck I notice that you are poking the network legacy module 

The only changes to `35legacy` are those which you acked on #2192 already. Sorry for not making this clear.

> which we are planning on dropping so my question to you is how much do you think it's being used in the wild since downstream should have switched to iwd, connman,networkd,nm by now

My distro (SUSE/openSUSE) is still using network-legacy, and will continue to do so for the life time of SLE15 at least. At least that's what I expect. I designed the NVMeoF changes such that they just set cmdline parameters like `ip=`, so that they should (in principle, at least) work with any network backend.

aafeijoo-suse:
> > @mwilck I notice that you are poking the network legacy module
> > which we are planning on dropping so my question to you is how much do you think it's being used in the wild since downstream should have switched to iwd, connman,networkd,nm by now
> 
> My distro (SUSE/openSUSE) is still using network-legacy, and will continue to do so for the life time of SLE15 at least (EOL currently planned for 2031). At least that's what I expect. I designed the NVMeoF changes such that they just set cmdline parameters like `ip=`, so that they should (in principle, at least) work with any network backend.

Yes, that's a recurring discussion. We (SUSE) will not be removing the `network-legacy` module anytime soon. I think it's also the only option for any other non-systemd distro for now (because the `network-manager` module depends on systemd : #1756).

While the `network-legacy` module is still present upstream, I'd suggest keeping it up to date, even more if the fixes are minor like the one submitted in this PR.

johannbg:
> > > @mwilck I notice that you are poking the network legacy module
> > > which we are planning on dropping so my question to you is how much do you think it's being used in the wild since downstream should have switched to iwd, connman,networkd,nm by now
> > 
> > 
> > My distro (SUSE/openSUSE) is still using network-legacy, and will continue to do so for the life time of SLE15 at least (EOL currently planned for 2031). At least that's what I expect. I designed the NVMeoF changes such that they just set cmdline parameters like `ip=`, so that they should (in principle, at least) work with any network backend.
> 
> Yes, that's a recurring discussion. We (SUSE) will not be removing the `network-legacy` module anytime soon. I think it's also the only option for any other non-systemd distro for now (because the `network-manager` module depends on systemd : #1756).
> 
> While the `network-legacy` module is still present upstream, I'd suggest keeping it up to date, even more if the fixes are minor like the one submitted in this PR.

Yeah sure I just want to keep invested party aware that it's eventually going away.
The reality is that it wont happen until we introduce iwd ( and wifi support in the process ) and iwd wont be introduced until a) I or someone else finish it ( which wont happen for me until at best at easter vacation that is if I get one ) and b) it survives switch root which requires upstream changes to iwd which I have not even begun to discuss with upstream iwd.

johannbg (changes requested):
@mwilck can you remove the relevant bit's in this PR that got merged in #2192 otherwise LGTM

mwilck:
> @mwilck can you remove the relevant bit's in this PR that got merged in #2192 otherwise LGTM

I'll have to rebase this PR to current master, which I'll have to coordinate in the Timberland group.

Also, please **do not merge** this yet. As I said above, the `nvme-cli` command line API isn't finalized yet and will certainly change.

johannbg:
@mwilck for work in progress the workflow here is to put the pull requests in a draft mode preferably with WIP: <foo> in the subject line to further emphasize this is an work in progress. Once the PR is ready for review/merging it's moved out of the draft status so you should move the PR to a draft  ( or I can do so for you ) and once the relevant changes have settled down upstream you can move the PR out of the draft status.

mwilck:
@johannbg , I added "WIP" now. If anything else is to be done to properly assign draft status, feel free to do so.

mwilck:
Merged and squashed https://github.com/timberland-sig/dracut/pull/10, and rebased.

This is ready for review now.

modules.d/95nvmf/nvmf-autoconnect.sh
```diff
@@ -1,5 +1,54 @@
-#!/bin/bash
+#!/bin/sh
+# Argument $1 is "settled", "online", or "timeout", indicating
```
aafeijoo-suse, line 2: Only "timeout" is taking into account, the other parameters are only for debugging purposes, right?

modules.d/95nvmf/nvmf-autoconnect.sh
```diff
@@ -1,5 +1,54 @@
...
+# Otherwise, try options according to the priorities below.
 
-[ -f /sys/class/fc/fc_udev_device/nvme_discovery ] || exit 1
-echo add > /sys/class/fc/fc_udev_device/nvme_discovery
+[ "$RD_DEBUG" != yes ] || set -x
+
+if [ "$1" = timeout ]; then
+    [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ] \
+        || echo add > /sys/class/fc/fc_udev_device/nvme_discovery
+    /usr/sbin/nvme connect-all
```
aafeijoo-suse, line 12: In the timeout case, are you ignoring the possible `rd.nvmf.nonbft` on purpose?

modules.d/95nvmf/module-setup.sh
```diff
@@ -25,14 +25,24 @@ check() {
...
+
     [[ $hostonly ]] || [[ $mount_needs ]] && {
         pushd . > /dev/null
         for_each_host_dev_and_slaves is_nvmf
         local _is_nvmf=$?
         popd > /dev/null || exit
         [[ $_is_nvmf == 0 ]] || return 255
         if [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ]; then
-            if [ ! -f /etc/nvme/discovery.conf ]; then
+            if [ ! -f /etc/nvme/discovery.conf ] && ! has_nbft; then
```
aafeijoo-suse, line 45: Shouldn't we also check here if `/etc/nvme/config.json` does not exist?

modules.d/95nvmf/nvmf-autoconnect.sh
mwilck: yes. In the timeout case, we want to activate all connections we are possibly aware of.

modules.d/95nvmf/nvmf-autoconnect.sh
mwilck: Yes. The argument allows the script to figure out the type of event it was being called for. Currently "timeout" is the only event type that gets a special treatment.

I realize that the value "online" actually doesn't occur any more, because we're using the netroot handler for that purpose. I'll remove it from the comment to avoid confusion.

modules.d/95nvmf/module-setup.sh
mwilck: Ooops. Thanks a lot.

mwilck: fixed with 9fd4901

mwilck:
@aafeijoo-suse, your issues should be addressed now.

modules.d/95nvmf/module-setup.sh
aafeijoo-suse: sorry, our "friend" `shfmt` is complaining: https://github.com/dracutdevs/dracut/actions/runs/5265185676/jobs/9517488413?pr=2184

modules.d/95nvmf/nvmf-autoconnect.sh
aafeijoo-suse: ok, thanks for your feedback

modules.d/95nvmf/module-setup.sh
mwilck: hm, I could swear I ran this locally. Anyway, hopefully fixed with 0a4d7f9

aafeijoo-suse (approved):
LGTM. Thanks!

lnykryn (approved):
LGTM

modules.d/95nvmf/module-setup.sh
```diff
@@ -25,17 +25,27 @@ check() {
         [[ $trtype == "fc" ]] || [[ $trtype == "tcp" ]] || [[ $trtype == "rdma" ]]
     }
 
+    has_nbft() {
+        local f found=
+        for f in /sys/firmware/acpi/tables/NBFT*; do
+            [ -f "$f" ] || continue
```
pvalena, line 31: This checks only for regular files. Is that intentional?

You could check for `-e` (existence) instead.

modules.d/95nvmf/module-setup.sh
```diff
@@ -25,17 +25,27 @@ check() {
...
+            found=1
+            break
+        done
+        [[ $found ]]
+    }
+
     [[ $hostonly ]] || [[ $mount_needs ]] && {
         pushd . > /dev/null
```
pvalena, line 39: I know this is not part of the PR, but does anyone know why `pushd .` is here? I think this is basically a no-op.

pvalena: Is it just checking that `for_each_host_dev_and_slaves` does not run `popd`? I'd love to know the reasoning behind this implementation.

modules.d/95nvmf/module-setup.sh
```diff
@@ -25,17 +25,27 @@ check() {
...
         popd > /dev/null || exit
         [[ $_is_nvmf == 0 ]] || return 255
-        if [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ]; then
-            if [ ! -f /etc/nvme/discovery.conf ]; then
-                echo "No discovery arguments present"
-                return 255
-            fi
+        if [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ] \
+            && [ ! -f /etc/nvme/discovery.conf ] \
+            && [ ! -f /etc/nvme/config.json ] && ! has_nbft; then
```
pvalena, line 46: Just a note: same usage of `-f` vs `-e` as stated above.

modules.d/95nvmf/nvmf-autoconnect.sh
```diff
@@ -1,5 +1,54 @@
...
+# the queue from which the script is called.
+# In the "timeout" case, try everything.
+# Otherwise, try options according to the priorities below.
 
-[ -f /sys/class/fc/fc_udev_device/nvme_discovery ] || exit 1
-echo add > /sys/class/fc/fc_udev_device/nvme_discovery
+[ "$RD_DEBUG" != yes ] || set -x
+
+if [ "$1" = timeout ]; then
+    [ ! -f /sys/class/fc/fc_udev_device/nvme_discovery ] \
```
pvalena, line 10: Just a note: same usage of `-f` vs `-e` as stated above.

pvalena (approved):
Just minor nitpicks, otherwise LGTM.

johnmeneghini:
pushd/popd are bash builtin commands that allow the user to change directories (cd) by pushing and popping them as though they 
were on a stack.  I use this bash command all of the time.

Say `man pushd`

/John

aafeijoo-suse:
Thank you all for your work here.

pvalena:
@johnmeneghini Off course, it's not about the command itself, but about that full logic:

```
pushd .  
some_external_function
popd || exit
```

I.e. changing to current directory & exiting if popd fails.... what does logic this aim to capture/prevent/achieve?

modules.d/95nvmf/module-setup.sh
mwilck: ok.

modules.d/95nvmf/module-setup.sh
mwilck: It is indeed not part of this PR, therefore I'd like to close this conversation. 

AFAICS, the `pushd/popd` code (like lots of the initial nvmf code) has been copied from `95iscsi/module-setup.sh`, where it had been added by @haraldh in commit b093aa2 ("beautified shell code") 10y ago, without further explanation. Perhaps at that time `for_each_host_dev_and_slaves` could change the working directory? I don't know.

modules.d/95nvmf/module-setup.sh
mwilck: `-f` is correct here. We wouldn't want to account for non-regular files in by the respective names.

modules.d/95nvmf/nvmf-autoconnect.sh
mwilck: Again, `-f` is correct here. If this existed and was anything else but a regular file, we'd have to ignore it.

modules.d/95nvmf/module-setup.sh
pvalena: Right, folders, I didn't think of that - but It might be symlink or other file-like type as well f.e.; but if only regular files are expected/accepted, I'm fine with that.

modules.d/95nvmf/module-setup.sh
pvalena: Right, good to know the origin of it. Thanks for the response!

modules.d/95nvmf/nvmf-autoconnect.sh
pvalena: Ack, thanks!

modules.d/95nvmf/module-setup.sh
mwilck: Yes, only regular files should be accepted. 

Note that `[` (aka `test`) dereferences symlinks for all flags except `-h` and `-L`. Thus if `/sys/firmware/acpi/tables/NBFT*` was a symlink to some existing regular file, this test wouldn't skip it, which looks wrong. But this case would indicate a severely broken kernel. I believe we can ignore it. In the worst case, we'd include the NBFT code in the initramfs on a broken system.

PR #2192, opened 2023-02-08
Fixes for iSCSI boot & IPv6

mwilck:
This pull request fixes a few issues I encountered while testing network boot with IPv6.
It prepares a final submission of #2184 (it contains the parts of #2184 which are not related to the nvmf module).

## Checklist
- [x]  I have reviewed and updated any documentation if relevant

### DIFF
--- a/modules.d/35network-legacy/module-setup.sh
+++ b/modules.d/35network-legacy/module-setup.sh
@@ -15,7 +15,8 @@ depends() {
 
 # called by dracut
 installkernel() {
-    return 0
+    # arping depends on af_packet
+    hostonly='' instmods af_packet
 }
 
 # called by dracut
--- a/modules.d/40network/net-lib.sh
+++ b/modules.d/40network/net-lib.sh
@@ -727,7 +727,6 @@ wait_for_ipv6_dad() {
     while [ $cnt -lt $timeout ]; do
         [ -n "$(ip -6 addr show dev "$@")" ] \
             && [ -z "$(ip -6 addr show dev "$@" tentative)" ] \
-            && { ip -6 route list proto ra dev "$@" | grep -q ^default; } \
             && return 0
         [ -n "$(ip -6 addr show dev "$@" dadfailed)" ] \
             && return 1
--- a/modules.d/40network/net-lib.sh
+++ b/modules.d/40network/net-lib.sh
@@ -311,7 +311,7 @@ ibft_to_cmdline() {
                 [ -e "${iface}"/hostname ] && read -r hostname < "${iface}"/hostname
                 if [ "$family" = "ipv6" ]; then
                     if [ -n "$ip" ]; then
-                        [ -n "$prefix" ] || prefix=64
+                        [ -n "$prefix" ] || prefix=128
                         ip="[${ip}]"
                         mask=$prefix
                     fi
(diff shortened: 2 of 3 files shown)


mwilck:
Adding @aafeijoo-suse and @tblume.

I expect that d814192 might be controversial. Adding @pallotron here, the author of c603419, which is reverted by that patch. Perhaps he can explain what waiting for a route was supposed to achieve. IMO it is rather obvious that waiting for a non-.tentative address and waiting for a default route are not the same thing. I can see that in particular for remote targets, waiting for a route to be established might make sense. But it is a different thing then waiting for DAD (duplicate address detection) to complete, and should be done separately.

aafeijoo-suse:
> I expect that [d814192](https://github.com/dracutdevs/dracut/commit/d814192ac5aba1d8ab4276e0d9cfceeb065d17af) might be controversial. Adding @pallotron here, the author of [c603419](https://github.com/dracutdevs/dracut/commit/c603419030136570b5944dc4620f62d07b9e82bb), which is reverted by that patch. Perhaps he can explain what waiting for a route was supposed to achieve. IMO it is rather obvious that waiting for a non-.tentative address and waiting for a default route are not the same thing. I can see that in particular for remote targets, waiting for a route to be established might make sense. But it is a different thing then waiting for DAD (duplicate address detection) to complete, and should be done separately.

Just for reference: #214

mwilck:
> Just for reference: https://github.com/dracutdevs/dracut/pull/214

Thanks. I think our code is confusing matters somewhat. IPv6 address configuration takes place in multiple steps:

 1. DAD (duplicate address detection): this step is mandatory. This is what `wait_for_ipv6_dad()` is supposed to wait for. This step is mandatory. After it's complete, link-local peers would already be reachable.
 2. Router solicitation / waiting for Router Advertisement (RA). *This step is optional*. An RA might never be received, and even if it is, there is not guarantee that it will contain a default route. This happens if the route lifetime in the RA is set to 0; it's the equivalent of an isolated network, and what libvirt does for isolated IPv6 networks. _The current test for a default route is broken_.
 3. DHCPv6 solicit/advertise/request/reply in the DHCPv6 case.

The idea of #214 is (I think) that even after step 1. completes, certain network peers may not be reachable yet, and if we proceed too fast, we might fail setting up a connection. One option here is to simply retry, as we do with other steps, too. Another would be to check `ip -6 route get $TARGET_IP` before we actually initiate some connection. But that's not part of the interface setup; it should be done in the later steps that attempt an actual connection.

pallotron:
Hi guys,

That was a long time ago, [#214](https://github.com/dracutdevs/dracut/pull/214) has some context.
In our production network we use IPv6 only, and static address assignment and back then we could not reach out to the network to download Anaconda's `squashfs.img` because we were not waiting for RA messages coming from closest router to setup routes (specifically default gateway).

That said, that was 6y ago, I suppose networking management is better now?
We stopped using dracut long ago in our production environment, so I no longer have any stake in this anymore.

You can do as you please, it won't have impact on us.

pallotron:
> The idea of https://github.com/dracutdevs/dracut/pull/214 is (I think) that even after step 1. completes, certain network peers may not be reachable yet, and if we proceed too fast, we might fail setting up a connection.

yeah I believe that was it.

johannbg (approved):
LGTM

aafeijoo-suse (approved):
Thanks!

PR #2194, opened 2023-02-09
fix(systemd): do not include in the initrd units configured not to run there

aafeijoo-suse:
Do not include in the initrd systemd units configured not to run there with `ConditionPathExists=!/etc/initrd-release`, as they just fill the log with entries like:

```
systemd[1]: <UNIT_DESCRIPTION> was skipped because of an unmet condition check (ConditionPathExists=!/etc/initrd-release).
```

References:
- https://github.com/systemd/systemd/commit/fe7f113c
- https://github.com/systemd/systemd/commit/7ef09e20

## Checklist


johannbg:
FYI  there is "systemd's upstream" views how initrds should be used then there is the reality how they are actually being used in the real world so you cant just blindly follow upstream systemd changes ( which is one of the reason we arguably should be shipping our own type units as opposed to be sourcing it from the host system ).

johannbg (approved):

PR #2206, opened 2023-02-15
WIP: feat(crypt): remove crypt keyfile before entering emergency shell

tblume:
## Changes

There is work in progress for grub to write keyfiles for decrypting an encrypted root file systems into the initrd.
This is done in order to avoid having to type the decryption password twice, one time for grub and another time in the initrd.
Also user provided keyfiles in the initrd provide an attack vector, because it is easy to arrive at the emergency shell and the system root password could be guessed.
Therefore, dracut should per default remove all crypt keyfiles from the initrd before entering the emergency shell while leaving the option to keep them if desired.

The patch for grub is not yet upstream, and I'm open for a discussion whether removing user provided keyfiles is really desireable.
Therefore I provide this pr as WIP.

## Checklist

Fixes #

security concerns

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -259,6 +259,8 @@ Creates initial ramdisk images for preloading modules
   --printsize           Print out the module install size.
   --sshkey [SSHKEY]     Add SSH key to initramfs (use with ssh-client module).
   --logfile [FILE]      Logfile to use (overrides configuration setting).
+  --keep-cryptkeyfiles  Don't remove keyfiles for decryption when entering the
+                        emergency shell (use with crypt module)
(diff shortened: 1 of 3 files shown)


johannbg:
@tblume we need more generic approach to this ( read as not limiting this to systemd based systems ) also when that grub patch(es) have been submitted to upstream grub can you link that PR to this PR

LaszloGombos:
>  dracut should per default remove all crypt keyfiles from the initrd before entering the emergency shell 

I am trying to understand the attack vector here. 

If the user has access to emergence shell, wouldn't the user also have access to change kernel command line options and just append e.g. init=/bin/sh ?

PR #2207, opened 2023-02-15
fix(dracut-install): fix possible infinite recursion with suppliers

athierry1:
During search for fw_devlink suppliers, it's possible to encounter a situation where supplier A depends on supplier B, and supplier B has a parent node that depends on supplier A. This leads to an infinite recursion.

To fix this, make sure suppliers are only processed once.

## Checklist

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -83,6 +83,7 @@ static Hashmap *items = NULL;
 static Hashmap *items_failed = NULL;
 static Hashmap *modules_loaded = NULL;
 static Hashmap *modules_suppliers = NULL;
+static Hashmap *processed_suppliers = NULL;
 static regex_t mod_filter_path;
 static regex_t mod_filter_nopath;
 static regex_t mod_filter_symbol;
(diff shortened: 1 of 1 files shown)


src/install/dracut-install.c
```diff
@@ -1667,6 +1668,11 @@ static int install_dependent_modules(struct kmod_ctx *ctx, struct kmod_list *mod
         const char *supplier_path;
         Iterator i;
         HASHMAP_FOREACH(supplier_path, suppliers_paths, i) {
+                if (check_hashmap(processed_suppliers, supplier_path))
+                        continue;
+
+                hashmap_put(processed_suppliers, strdup(supplier_path), strdup(supplier_path));
```
aafeijoo-suse, line 1674: Couldn't this 2 `strdup` calls be simplified to just one using a temporary `char*` variable?

```c
char *path = strdup(supplier_path);
hashmap_put(processed_suppliers, path, path);
```

Other than that, your patch looks good, thanks.

athierry1: Fixed, thank you!

aafeijoo-suse (approved):

johannbg (approved):
LGTM

PR #2210, opened 2023-02-18
Multiple source for dracut configuration directory

sshedi:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -154,8 +154,9 @@ Creates initial ramdisk images for preloading modules
   -q, --quiet           Decrease verbosity level.
   -c, --conf [FILE]     Specify configuration file to use.
                          Default: /etc/dracut.conf
-  --confdir [DIR]       Specify configuration directory to use *.conf files
-                         from. Default: /etc/dracut.conf.d
+  --confdir [LIST]      Specify a space separated list of configuration
+                         directories to use *.conf files from.
+                         Default: /etc/dracut.conf.d
   --tmpdir [DIR]        Temporary directory to be used instead of default
                          ${TMPDIR:-/var/tmp}.
   -r, --sysroot [DIR]   Specify sysroot directory to collect files from.
@@ -668,7 +669,7 @@ while :; do
             shift
             ;;
         --confdir)
-            confdir="$2"
+            confdirs_l=("$2")
             PARMS_TO_STORE+=" '$2'"
             shift
             ;;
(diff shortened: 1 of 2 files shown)


dtardon:
Out of curiosity: what's the use case for this?

sshedi:
Let's say we have a config file `fips.conf` which is at `/etc/dracut.conf.d` and kernel has it's own set of configs unde `/lib/modules/<version>/dracut.conf.d/kernel.conf`

And it is possible to have more than one flavor of kernel, like realtime, generic, secure etc.

While creating initrd I need a way to include kernel's own custom config from modules directory and fips.conf from default location. Meaning, I want fips configs to be included in all flavors along with kernel's own custom configs.

With this change I can do:
```
dracut --confdir "/etc/dracut.conf.d /lib/modules/<ver>/dracut.conf.d" ...
```

Now dracut consumes configs from both locations, fips.conf is a common config for all kernel flavors in this example, so I want to maintain a single copy of it.

Same applies for config files.

So, thought of making this change and keeping it generic, users can specify any number of config files and locations to consume configs from. Hope this is a valid case. If there is any other way to do it, please let me know.

johannbg:
I dont see a valid usecase to support this

sshedi:
> I dont see a valid usecase to support this

I will try to explain it wrt to our system (Photon OS 3.0)

We have a `fipsify` package https://github.com/vmware/photon/blob/3.0/SPECS/fipsify/fipsify.spec#L37

This creates a conf file under `/etc/dracut.conf.d` and the fips modules should be a part of all kernel flavors present in the system.

And each kernel provides it's own drivers list and the list is unique.
https://github.com/vmware/photon/blob/3.0/SPECS/linux/linux.spec#L927
https://github.com/vmware/photon/blob/3.0/SPECS/linux/linux-secure.spec#L433

Currently we are using a work around to get this done.

Now I'm trying to create a conf file for each kernel under `<modules-dir>/dracut.conf.d` and trying to include both fips conf (common to all flavors) and kernel conf of individual kernel from modules dir while creating initrd of the specific kernel.

If I keep everything under `/etc/dracut.conf.d` all the configs will be included at once and will result in wrong driver list.

So, it should be possible to use multiple config dirs while creating initrd. Please let me know if there is any other way to achieve this.

sshedi:
> Strictly speaking, it seems to me that your use case does not need changes to`--conf`, it only needs changes to `--confdir` . Can you please confirm ? Did you change ` --conf` just for completeness ?

Yes

aafeijoo-suse:
I don't see the need for this change either. If a package wants to add its own custom dracut configuration, it can create the .conf files under `/etc/dracut.conf.d`, as many packages are doing nowadays.

sshedi:
> I am open to consider this change, but only for directory argument and not for conf files.
> 
> I think conf file argument is somewhat confusing and we should not improve it or add complexity to it. See also #1974
> 
> We already have 10+ arguments that take a list instead of a single entry. This is a niche use case but it is also a low risk, low maintenance and small change.
> 
> I think having support for conf dir under `/etc` and `/usr` and `/run` is reasonable - which would require supporting more than one dir for config.

Agreed. Having support for multiple conf dirs & files is redundant. Dropped the multiple conf file change. Thanks.

sshedi:
What if user already has some conf files under `/usr/lib/dracut/dracut.conf.d` and want's those configs to take effect during initrd creation?

PR #2221, opened 2023-02-21
fix(udev-rules): remove old edd_id extra rules

### LINKED ISSUE #2009: remove legacy udev rules
50-udev.rules
61-persistent-storage-edd.rules
70-persistent-net.rules
95-late.rules

See also https://github.com/dracutdevs/dracut/pull/1965/files

LaszloGombos:
The `61-persistent-storage-edd.rules` rule has been removed upstream in [2012](https://github.com/systemd/systemd/commit/4774868ccabd76c3d208343026f1c6e57094642b)

Fixes https://github.com/dracutdevs/dracut/issues/2009 (partially)

### DIFF
--- a/modules.d/95udev-rules/module-setup.sh
+++ b/modules.d/95udev-rules/module-setup.sh
@@ -38,7 +38,6 @@ install() {
         60-cdrom_id.rules \
         60-pcmcia.rules \
         60-persistent-storage.rules \
-        61-persistent-storage-edd.rules \
         64-btrfs.rules \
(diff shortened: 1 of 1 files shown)


johannbg:
Can you make a single PR with all your udev rules removal and add to that list 40-redhat.rules since we dont carry downstream distribution specific udev rules anymore (  redhat is no exception to that rule )  as well as drop the eudev specific rule since we dont support "forks" either.

PR #2222, opened 2023-02-21
fix(base): log the full udev database in rdsosreport

LaszloGombos:
`blkid -o udev` is deprecated since [2012](https://github.com/util-linux/util-linux/commit/24d741d88a31863e82feb3f3a4516d7df006378b)

`udevadm info --export-db` logs the full content of the udev database.

### DIFF
--- a/modules.d/99base/rdsosreport.sh
+++ b/modules.d/99base/rdsosreport.sh
@@ -28,7 +28,7 @@ cat /proc/self/mountinfo
 cat /proc/mounts
 
 blkid
-blkid -o udev
+command -v udevadm > /dev/null 2> /dev/null && udevadm info --export-db
 
 ls -l /dev/disk/by*
 


aafeijoo-suse (commented):
I miss `ID_FS_PARTUUID` and `ID_FS_PARTLABEL` in the `udevadm` output.

modules.d/99base/rdsosreport.sh
```diff
@@ -27,8 +27,8 @@ done
 cat /proc/self/mountinfo
 cat /proc/mounts
 
-blkid
-blkid -o udev
+command -v blkid > /dev/null 2> /dev/null && blkid
```
aafeijoo-suse, line 30: Why this extra `command -v blkid`? don't we always require `util-linux`?

aafeijoo-suse: I only pointed out that the `command -v blkid` check is not necessary, but the `command -v udevadm` is required here.

aafeijoo-suse (approved):

PR #2233, opened 2023-02-24
fix(dmsquash-live): restore compatibility with earlier releases

### LINKED ISSUE #2232: New 90overlayfs module does not set up the overlay in three cases where it happened before (breaks Fedora/RHEL installer image boot)
**Describe the bug**
In dracut 058/059, with commit https://github.com/dracutdevs/dracut/commit/8caaad4fc2d75982eb87f5ebc72a4c276986f756 and follow-up https://github.com/dracutdevs/dracut/commit/40dd5c90e0efcb9ebaa9abb42a38c7316e9706bd , overlayfs setup was moved from being done in-line at the end of `dmsquash-live-root.sh` to being part of a separate module. There were four codepaths in which overlayfs setup was reached in `dmsquash-live-root.sh`, but with the move to a module, only one of those paths now results in overlayfs setup happening. On the other three paths, it no longer happens, and boot fails.

**Distribution used**
Fedora 38 and Rawhide.

**Dracut version**
058 and 059.

**Init system**
systemd.

**To Reproduce**
Boot a Fedora installer image built with dracut 058/059 without the offending commits reverted. They were reverted in https://koji.fedoraproject.org/koji/buildinfo?buildID=2157455 and we caught this before it reached a nightly compose, so you would have to build your own image using the older dracut-059-1 build; the images built by openQA (which caught this bug) have been garbage-collected now. Affected images get stuck in a loop, showing the error "mount: /sysroot: special device LiveOS_rootfs does not exist."

**Expected behavior**
The image should boot normally.

**Additional context**
I did a detailed root cause of the problem in [the downstream bug](https://bugzilla.redhat.com/show_bug.cgi?id=2172269). Basically, if you look at `dmsquash-live-root.sh`, the overlayfs setup previously happened near the end of the file, any time the variable `$overlayfs` was a non-zero-length string. There were four paths on which that happened: one where it was set to "yes" (when `rd.live.overlay.overlayfs` is on the cmdline), and three where it was set to "required" (two inside the block that starts `if [ -z "$setup" -a -n "$devspec" -a -n "$pathspec" -a -n "$overlay" ]; then`, and one in the block that starts `if [ -e /run/initramfs/live/${live_dir}/${squash_image} ]; then`).

When the overlayfs setup was moved to a module, only the first of these paths was respected: the new module only actually does the overlayfs setup if the cmdline parameter set. In all other cases it exits after doing nothing. So on those three other paths, the overlayfs is not set up and boot will likely fail. Fedora's installer images go down the squashfs path, I don't know in what situations the other two paths are hit.

For now in Fedora we have reverted the relevant commits. @lnykryn [suggested](https://bugzilla.redhat.com/show_bug.cgi?id=2172269#c5) having `dmsquash-live-root.sh` do `echo "rd.live.overlay.overlayfs=1" > /etc/cmdline.d/dracut-need-overlay.conf` on the affected paths, which...I guess could work, but seems very hacky. I was kinda assuming there must be a better way to do this, some canonical way for such a script to signal to a module that its action is required?

LaszloGombos:
Follow-up to https://github.com/dracutdevs/dracut/pull/1934

Fixes #2232

CC @FGrose @Conan-Kudo @AdamWill

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -403,6 +403,10 @@ fi
 
 ROOTFLAGS="$(getarg rootflags)"
 
+if [ "$overlayfs" = required ]; then
+    echo "rd.live.overlay.overlayfs=1" > /etc/cmdline.d/dmsquash-need-overlay.conf
+fi
+
 if [ -n "$overlayfs" ]; then
     if [ -n "$FSIMG" ]; then
         mkdir -m 0755 -p /run/rootfsbase


AdamWill:
I can test this, but one thought: would it be worth a comment explaining why this happens? It's the kind of thing that might look odd to someone reading over the code in a year...

lnykryn:
@LaszloGombos I had the same idea about how to fix this, but I have no clue if this cover all use cases.

AdamWill:
The fix does look good for the Fedora case in [openQA testing](https://openqa.stg.fedoraproject.org/tests/2609594).

Conan-Kudo (approved):

lnykryn (approved):

pvalena:
LGTM.

aafeijoo-suse (approved):
#2240 is creating some noise, let's fix this first.

PR #2237, opened 2023-02-24
fix(dracut.sh): kmoddir does not handle trailing /

### LINKED ISSUE #2236: dracut --kmoddir fails on paths with traling `/`
**Describe the bug**
There's an error in recently added commit https://github.com/dracutdevs/dracut/pull/2025/commits/2c2bd868cfa4c61d7f4792c163262a24f6e02f54#diff-99bfb8fe311d6a59318f685a5e3401b59293b577968aa08cc4ff7320a163bb28R1034

The code added in the commit only works in interactive mode:

```
$ cat example.sh
set -x

drivers_dir=/root/test/tmp/lib/modules/6.2.0-63.fc39.x86_64+debug/
drivers_dir="${drivers_dir%%+(/)}"
```

Note the trailing `/` which the last line of the example.sh was supposed to remove:
```
$ bash example.sh
+ drivers_dir=/root/test/tmp/lib/modules/6.2.0-63.fc39.x86_64+debug/
+ drivers_dir=/root/test/tmp/lib/modules/6.2.0-63.fc39.x86_64+debug/
```

Only way I was able to trigger *the correct* behaviour was an interactive mode:
```
$ bash -i < example.sh
stty: 'standard input': Inappropriate ioctl for device
$ set -x
$
$ drivers_dir=/root/test/tmp/lib/modules/6.2.0-63.fc39.x86_64+debug/
+ drivers_dir=/root/test/tmp/lib/modules/6.2.0-63.fc39.x86_64+debug/
$ drivers_dir="${drivers_dir%%+(/)}"
+ drivers_dir=/root/test/tmp/lib/modules/6.2.0-63.fc39.x86_64+debug
$ exit
```

**Distribution used**
Fedora Rawhide

**Dracut version**
059

**Init system**
systemd

**To Reproduce**

Add trailing `/` to kmoddir:
```
$ dracut -f --kmoddir /lib/modules/`uname -r`/
dracut: drivers_dir path set via -k/--kmoddir must contain "/lib/modules/" as a parent of your kernel module directory,
dracut: or modules may not be placed in the correct location inside the initramfs.
dracut: was given: /lib/modules/6.1.12-200.fc37.x86_64/
dracut: expected: /lib/modules/6.1.12-200.fc37.x86_64/lib/modules/6.1.12-200.fc37.x86_64
dracut: Please move your modules into the correct directory structure and pass the new location,
dracut: or set DRACUT_KMODDIR_OVERRIDE=1 to ignore this check.
```

**Expected behavior**
Trailing / removed.

**Additional context**

pvalena:
Fixes https://github.com/dracutdevs/dracut/issues/2236

+(/) only works in bash interactive mode. Using any non-interactive bash run simply doesn't remove anything.

The solution is copied from dracut-lib.sh, trim().

## Changes

Trailing / for kmoddir variable are now handled properly.

## Checklist

Fixes #2236

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1033,7 +1033,7 @@ stdloglvl=$((stdloglvl + verbosity_mod_l))
 ((stdloglvl < 0)) && stdloglvl=0
 
 [[ $drivers_dir_l ]] && drivers_dir=$drivers_dir_l
-drivers_dir="${drivers_dir%%+(/)}"
+drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
 [[ $do_strip_l ]] && do_strip=$do_strip_l
 [[ $do_strip ]] || do_strip=yes
 [[ $aggressive_strip_l ]] && aggressive_strip=$aggressive_strip_l


pvalena:
Note: I've checked for similar code and found only:
```
dracut-functions.sh:    local p=${1//+(\/)//}
modules.d/90mdraid/module-setup.sh:            -e '/(RUN|IMPORT\{program\})\+?="[[:alpha:]/]*mdadm[[:blank:]]+(--incremental|-I)[[:blank:]]+(--export )?(\$env\{DEVNAME\}|\$tempnode|\$devnode)/d' \
```

pvalena:
RHBZ for reference: https://bugzilla.redhat.com/show_bug.cgi?id=2173100#

aafeijoo-suse (approved):
My bad, thanks for the fix.

dracut.sh
```diff
@@ -1033,7 +1033,7 @@ stdloglvl=$((stdloglvl + verbosity_mod_l))
 ((stdloglvl < 0)) && stdloglvl=0
 
 [[ $drivers_dir_l ]] && drivers_dir=$drivers_dir_l
-drivers_dir="${drivers_dir%%+(/)}"
+drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
```
dtardon, line 1036: Alternatively, you could call `normalize_path()` on the dir. That also replaces multiple `/` inside the path by a single one, but I don't think that is a problem...

pvalena: Oh, I didn't know that exists. Probably a better solution.

PR #2011, opened 2022-10-18
test: improve infrastructure and coverage

LaszloGombos:
Series of commits to improve testing infrastructure.
None of these commits changes any of the dracut functionality for existing modules.
Goal is to improve portability of tests and code coverage.
It is also a goal to move more execution to dracut command line to improve code coverage.
Sharing more code between test cases would allow to remove about 50 lines of source code from each test case.

Changes are to to TEST-03-USR-MOUNT only as part of this PR to demonstrate the benefits and confirm that there is no regression. Plan to change other TEST as a follow-up PR. 

Picked TEST-03-USR-MOUNT as it is one of our simplest and most portable test that is most unlikely to interfere with other open PRs.

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [X] I am providing new code and test(s) for it

PR #2238, opened 2023-02-25
test(RAID): make test 10 use the test dracut modules

LaszloGombos:
Same as https://github.com/dracutdevs/dracut/pull/2011 but for test 10.

This PR is a refactor, it should not change how the test is executed.

### DIFF
--- a/test/TEST-10-RAID/test.sh
+++ b/test/TEST-10-RAID/test.sh
@@ -7,7 +7,6 @@ TEST_DESCRIPTION="root filesystem on an encrypted LVM PV on a RAID-5"
 #DEBUGFAIL="rd.break rd.shell rd.debug debug"
 test_run() {
     declare -a disk_args=()
-    # shellcheck disable=SC2034
     declare -i disk_index=0
     qemu_add_drive_args disk_index disk_args "$TESTDIR"/marker.img marker
     qemu_add_drive_args disk_index disk_args "$TESTDIR"/raid-1.img raid1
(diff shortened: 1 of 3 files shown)


test/TEST-10-RAID/test.sh
```diff
@@ -26,59 +25,26 @@ test_run() {
...
     # create an initramfs that will create the target root filesystem.
     # We do it this way so that we do not risk trashing the host mdraid
     # devices, volume groups, encrypted partitions, etc.
     "$basedir"/dracut.sh -l -i "$TESTDIR"/overlay / \
-        -m "bash crypt lvm mdraid kernel-modules qemu" \
+        -m "test-makeroot bash crypt lvm mdraid kernel-modules" \
         -d "piix ide-gd_mod ata_piix ext2 sd_mod" \
         --nomdadmconf \
+        -I "mke2fs grep" \
+        -i ./create-root.sh /lib/dracut/hooks/initqueue/01-create-root.sh \
```
aafeijoo-suse, line 47: Although you didn't change the `create-root.sh` part, and it's working as it was before, I can't understand who or what is adding `cryptsetup` in a systemd initrd... the `crypt` module only adds `systemd-cryptsetup`...

aafeijoo-suse: Now I see, the first two calls use `-m` (`test-root` and `test-makeroot`), while the last one uses `-a` (`test`), so systemd is only included there... kind of tricky.

aafeijoo-suse (approved):

PR #2244, opened 2023-02-28
fix(multipath): check for all required binaries

LaszloGombos:
Introduce a new pattern to list required binaries once and use it both at check() and install() time.

I picked the multipath module as multipath module is actively developed, so I hope to get some meaningful feedback on this new pattern before extending it to more modules.

Fixed depends() call.

Removed dmsetup from here as it is already checked in the dm module.

As a result of this PR, if e.g. multipathd is not installed on the host check() would now fail early (instead of install).

The goal is NOT to make check() perfect, but simply make it better. Some tests for the environment will remain only inside the install() function.

CC @mwilck

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+binaries() {
+    # if there's no multipath binary, no go.
+    echo \
+        kpartx \
+        multipath \
+        multipathd
+}
+
 is_mpath() {
     local _dev=$1
     [ -e /sys/dev/block/"$_dev"/dm/uuid ] || return 1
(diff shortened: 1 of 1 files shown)


mwilck:
The correct way would probably be to make it depend on `dracut_module_included "systemd"`. But I am unsure if this value is already available in the `check()` stage.

mwilck:
Sorry, I have another question. Why do you need a shell function here? Wouldn't it be sufficient to simply set a variable?

```
BINARIES="multipath multipahtd ..."
```

mwilck:
AFAIK, The reason that `check()` etc are functions is that they are called from the main dracut binary. But `binaries()` seems to be called only from functions in the same `module_setup.sh` file.

But I would suggest that @johannbg or some other core developer comment on this.

PR #2250, opened 2023-02-28
fix(net-lib): add a new dracut module called net-lib

### LINKED ISSUE #2195: net-lib dracut module
Move out dhcp-root.sh  ifname-genrules.sh   net-lib.sh  netroot.sh from the network meta module to net-lib dracut module.

LaszloGombos:
Motivations:
 - meta modules (such as dbus or network) should not include any scripts. meta modules should only have `module-setup.sh` files as it has been also pointed out during earlier PR reviews and discussions.
 - fix dependencies. As an example `dracut -m network-manager` produces a broken initramfs that this PR fixes.
 - follow the wise people before me with good modular design as it has been proven for e.g. fs-lib, url-lib modules

## Checklist

Fixes #2195


lnykryn (approved):

lnykryn (approved):

modules.d/45net-lib/module-setup.sh
```diff
@@ -0,0 +1,20 @@
...
+    return 255
+}
+
+depends() {
+    echo base
+    return 0
+}
+
+install() {
+    inst_script "$moddir/netroot.sh" "/sbin/netroot"
```
aafeijoo-suse, line 13: `connman` is installing exactly the same `netroot.sh` script.

```
~/src/dracut/pr-2250> diff -s modules.d/45net-lib/netroot.sh modules.d/35connman/netroot.sh 
Files modules.d/45net-lib/netroot.sh and modules.d/35connman/netroot.sh are identical
```

After this PR, the following line can be removed from connman's `install()`, because `net-lib` is always installed as a dependency:

https://github.com/dracutdevs/dracut/blob/52351cfa049a9594d539e6a5337d591e8039ab80/modules.d/35connman/module-setup.sh#L28

modules.d/01systemd-networkd/module-setup.sh
```diff
@@ -22,7 +22,7 @@ check() {
 depends() {
 
     # This module has external dependency on other module(s).
-    echo dbus kernel-network-modules systemd-sysusers
+    echo dbus kernel-network-modules systemd-sysusers net-lib
```
aafeijoo-suse, line 25: Why does `systemd-networkd` requires `net-lib`?

modules.d/95nfs/nfs-lib.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/sh
 
 type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
-. /lib/net-lib.sh
```
aafeijoo-suse, line 4: Why did you remove this?

https://github.com/dracutdevs/dracut/blob/52351cfa049a9594d539e6a5337d591e8039ab80/modules.d/95nfs/nfs-lib.sh#L28

aafeijoo-suse:
> @aafeijoo-suse if you have time, can you please help and check if your concerns are resolved on this PR. Thank you !

My main concern is how many setups will break with this change, [I thought that you wanted to be conservative](https://github.com/dracutdevs/dracut/pull/2314#issuecomment-1552330085). Just doing a quick search on GitHub:
- https://github.com/flatcar/bootengine/blob/cc0fdec0cc6c5692acac95a928984ea8a5eb8f08/dracut/03flatcar-network/module-setup.sh#L46
- https://github.com/xcat2/confluent/blob/809099c8f80febb1937244fff17f6bd8f1364c7a/imgutil/el9/dracut/install#L28
- https://github.com/awmacpherson/deboot/blob/582c33e6ef2bc8b537a2f6c086ce276819350d8c/rootfs/mkroot.sh#L42
- https://github.com/alpinelinux/aports/blob/61691bc12240a7ed40637b784c6e2723eb38f3ad/community/dracut/APKBUILD#L39

PR #2253, opened 2023-03-01
fix(dmsquash-live): handle relative pathspec

### LINKED ISSUE #579: dracut.cmdline(7): document overlay persistence requirements
`do_live_overlay()` in `dmsquash-live-root.sh` makes assumptions about persistence command-line arguments:

- rd.live.overlay's `pathspec` must start with /

because `devspec` is mounted on `/run/initramfs/overlayfs` and `pathspec` is expanded as `/run/initramfs/overlayfs$pathspec`. [source](https://github.com/dracutdevs/dracut/blob/RHEL-8/modules.d/90dmsquash-live/dmsquash-live-root.sh#L128)

- the overlay image must contain directories named overlayfs and ovlwork

[source](https://github.com/dracutdevs/dracut/blob/RHEL-8/modules.d/90dmsquash-live/dmsquash-live-root.sh#L145)

LaszloGombos:
This change is required to make the following canonical example work that is mentioned in the [documentation](https://github.com/dracutdevs/dracut/blob/master/man/dracut.cmdline.7.asc):

> rd.live.overlay=/dev/sdb1:persistent-overlay.img

This example has been in the dracut documentation since [2015](https://github.com/dracutdevs/dracut/commit/a1b4efe6a722ac3754d8a48109575d11d5ccf61b).

At the same time, improve the documentation with a more specific example.

## Checklist

Fixes #579 (partially)

CC @FGrose @Conan-Kudo @fabiand

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -150,6 +150,8 @@ do_live_overlay() {
 
     if [ -z "$pathspec" -o "$pathspec" = "auto" ]; then
         pathspec="/${live_dir}/overlay-$l-$u"
+    elif ! str_starts "$pathspec" "/"; then
+        pathspec=/"${pathspec}"
     fi
     devspec=${overlay%%:*}
 
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -1157,7 +1157,7 @@ used.
 [listing]
 .Examples
 --
-rd.live.overlay=/dev/sdb1:persistent-overlay.img
+rd.live.overlay=/dev/sdb1:/persistent-overlay.img
 rd.live.overlay=UUID=99440c1f-8daa-41bf-b965-b7240a8996f4
 --
 


Conan-Kudo (approved):

aafeijoo-suse (commented):
> This change is required to make the following canonical example work that is mentioned in the documentation

Are we sure that the example in the man page (`pathspec` not starting with / ) is correct?

If the answer is yes, wouldn't it be easier to just fix when `pathspec` is assigned and doesn't start with / , to avoid the "//" case?

https://github.com/dracutdevs/dracut/blob/aca51203609954b38b41ea303a4376cbce061150/modules.d/90dmsquash-live/dmsquash-live-root.sh#L140

modules.d/90dmsquash-live/dmsquash-live-root.sh
```diff
@@ -145,7 +145,7 @@ do_live_overlay() {
         pathspec="/${live_dir}/overlay-$l-$u"
     elif strstr "$overlay" ":"; then
         # pathspec specified, extract
-        pathspec=${overlay##*:}
+        pathspec=/${overlay##*:}
```
aafeijoo-suse, line 148: I would only prepend / if `pathspec` does not start with /.
With your code, `rd.live.overlay=/dev/disk/by-uuid/XXXX:/path/to/overlay.img` ends with `pathspec=//path/to/overlay.img`

aafeijoo-suse (commented):
> > Are we sure that the example in the man page (`pathspec` not starting with / ) is correct?
> 
> I think the example in the documentation can be improve. I added the improvement in the doc to this PR as well.
> 
> Regardless this example has been in the documentation for 7 years, and maintaining compatibility with this example seems to be simple enough.
 
What I meant was I haven't found any example of `pathspec` without starting with /, because it refers to a file path.

aafeijoo-suse (approved):

PR #2254, opened 2023-03-03
chore: remove git2spec.pl, it is no longer used

LaszloGombos:
This tool was introduced in combination of [rpm package testing](https://github.com/dracutdevs/dracut/commit/0e6a94b32) which is no longer supported in the tree here upstream.


aafeijoo-suse (approved):

tblume (approved):
OK

PR #2260, opened 2023-03-04
test: convert more tests use the test dracut modules

LaszloGombos:
Same as https://github.com/dracutdevs/dracut/pull/2011 but for more tests.

Follow-up to https://github.com/dracutdevs/dracut/pull/2238 .

The last [commit](https://github.com/dracutdevs/dracut/pull/2260/commits/a8238596239f2958af59f5fad9c6d3114e719d09) in the PR is the first networking test (iscsi test) converted using this infrastructure, which includes not only client but server initramfs and rootfs generation. 

After this reviewed, I plan to convert over more networking tests.

This PR is a refactor, it should not change how the tests are executed. 

This PR removes about 400 lines of code while maintaining existing functionality.

### DIFF
--- a/test/TEST-11-LVM/finished-false.sh
+++ b/test/TEST-11-LVM/finished-false.sh
@@ -1,2 +1,0 @@
-#!/bin/sh
-exit 1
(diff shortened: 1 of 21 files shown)


aafeijoo-suse:
I encourage you to update each test using a separate PR, otherwise this is unhandable (at least for me).

PR #2264, opened 2023-03-05
fix(rootfs-block): remove support for [no]readonlyroot and fastboot 

### LINKED ISSUE #1948: document or remove undocumented arguments
rd.distroinit 
readonlyroot
noreadonlyroot
fastboot

LaszloGombos:
This PR removes a command line option for a core dracut module.

"readonlyroot" and "noreadonlyroot" command line option are no longer supported (not sure if they actually ever fully worked). 

Use "ro" or "rw" instead.

I am open to keep these arguments and document them instead of removing them, but since there is a much more prevalent command line alternative, I do not think these are used."

Original commit from 2010: https://github.com/dracutdevs/dracut/commit/3871942d13b9175b8143f1ebbd48a211a007c7d2

Fixes #1948 (partially)

### DIFF
--- a/modules.d/95rootfs-block/mount-root.sh
+++ b/modules.d/95rootfs-block/mount-root.sh
@@ -30,22 +30,13 @@ mount_root() {
         fsck_ask_err
     done
 
-    READONLY=
     fsckoptions=
     if [ -f "$NEWROOT"/etc/sysconfig/readonly-root ]; then
         # shellcheck disable=SC1090
         . "$NEWROOT"/etc/sysconfig/readonly-root
     fi
 
-    if getargbool 0 "readonlyroot=" -y readonlyroot; then
-        READONLY=yes
-    fi
-
-    if getarg noreadonlyroot; then
-        READONLY=no
-    fi
-
-    if [ -f "$NEWROOT"/fastboot ] || getargbool 0 fastboot; then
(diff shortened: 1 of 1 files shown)


modules.d/95rootfs-block/mount-root.sh
```diff
@@ -30,21 +30,12 @@ mount_root() {
         fsck_ask_err
     done
 
-    READONLY=
     fsckoptions=
     if [ -f "$NEWROOT"/etc/sysconfig/readonly-root ]; then
```
aafeijoo-suse, line 34: Old RH pre-systemd stuff: https://access.redhat.com/solutions/2196881
I'd remove the `/etc/sysconfig/readonly-root` part as well.

modules.d/95rootfs-block/mount-root.sh
```diff
@@ -30,21 +30,12 @@ mount_root() {
...
 
-    if getargbool 0 "readonlyroot=" -y readonlyroot; then
-        READONLY=yes
-    fi
-
-    if getarg noreadonlyroot; then
-        READONLY=no
-    fi
-
     if [ -f "$NEWROOT"/fastboot ] || getargbool 0 fastboot; then
```
aafeijoo-suse, line 39: I'd also remove the `fastboot` part, it provides the same functionality as `rd.skipfsck`.

PR #2266, opened 2023-03-06
fix(zipl): do not use deprecated `blkid -o udev`

aafeijoo-suse:
`blkid -o udev` is deprecated since https://github.com/util-linux/util-linux/commit/24d741d8, replace it with `udevadm info`.

Issue spotted in #2222

## Checklist


LaszloGombos (approved):
LGTM. Thanks

PR #2269, opened 2023-03-07
fix(overlayfs): allow hostonly and nfs support

LaszloGombos:
Check for overlay kernel module support.

Execute mount-overlayfs later in the boot process. Without this change `dracut  -a "overlayfs nfs"` does not work as expected.

## Checklist

Fixes #

### DIFF
--- a/modules.d/90overlayfs/module-setup.sh
+++ b/modules.d/90overlayfs/module-setup.sh
@@ -14,6 +14,7 @@ installkernel() {
 }
 
 install() {
-    inst_hook mount 01 "$moddir/mount-overlayfs.sh"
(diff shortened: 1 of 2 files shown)


aafeijoo-suse (commented):
You need to fix the `shfmt` format:

https://github.com/dracutdevs/dracut/actions/runs/4350868266/jobs/7601984774?pr=2269

The "allow hostonly" part is ok, but it has nothing to do with "Execute mount-overlayfs later in the boot process as well.". You should split that into another commit and give the proper reason for that (why does not it work as expected?).

aafeijoo-suse (approved):
> I improved the test case to test for this scenario, so now everybody can reproduce and understand the failure and also this test case would guard for future regression of this use-case.

Ok, thanks!

PR #2271, opened 2023-03-07
fix(dracut.sh): handle /etc/machine-id empty or uninitialized

### LINKED ISSUE #2267: Do not use a machine-id value of uninitialized
**Describe the bug**
systemd supports writing `uninitialized` to `/etc/machine-id`, which can be used for `systemd-firstboot` instead of removing `/etc/machine-id`. Dracut does not check the contents of this file to populate `MACHINE_ID`, so the value `unitialized` is used

**Distribution used**
Debian bookworm (12)

**Dracut version**
`059`

**Init system**
systemd

**To Reproduce**
```
echo "uninitialized" > /etc/machine-id
dracut -f
```

**Expected behavior**
Dracut behaves as if there was no `machine-id` set, i.e. setting `MACHINE_ID="Default"`.

**Additional context**

aafeijoo-suse:
Handle the case where the user tries to generate the initrd after explicitly resetting the `/etc/machine-id` file.

## Checklist

Fixes #2267

### DIFF
--- a/dracut-initramfs-restore.sh
+++ b/dracut-initramfs-restore.sh
@@ -19,8 +19,9 @@ SKIP="$dracutbasedir/skipcpio"
 
 if [[ -d /efi/Default ]] || [[ -d /boot/Default ]] || [[ -d /boot/efi/Default ]]; then
     MACHINE_ID="Default"
-elif [[ -f /etc/machine-id ]]; then
+elif [[ -s /etc/machine-id ]]; then
     read -r MACHINE_ID < /etc/machine-id
+    [[ $MACHINE_ID == "uninitialized" ]] && MACHINE_ID="Default"
 else
     MACHINE_ID="Default"
 fi
--- a/dracut.sh
+++ b/dracut.sh
@@ -1087,8 +1087,9 @@ if ! [[ $outfile ]]; then
             || [[ -d "$dracutsysrootdir"/boot/Default ]] \
             || [[ -d "$dracutsysrootdir"/boot/efi/Default ]]; then
             MACHINE_ID="Default"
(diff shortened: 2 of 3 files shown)


Cornelicorn:
I guess this should also be adjusted in `dracut-initramfs-restore.sh` and `lsinitrd.sh`.

johannbg:
> > I guess this should also be adjusted in `dracut-initramfs-restore.sh` and `lsinitrd.sh`.
> 
> Patching the initrd creation to handle this kind of user action is already a moot use case, but debugging on shutdown and querying its content is one step ahead... ok, it can be done for completeness, but I doubt that someone would hit this issue.

Users expect tools to behave the same thus someone will always hit this issue and eventually create an issue report as a result of that so we gain nothing being incomplete and in general being incomplete is considered bad practice.

That said the handling of machine-id has been subjected to changes in sd so we need to take a careful look at how it is being handled in sd now and how *we* should treat it which may or may not align with what sd ( now ) does.

aafeijoo-suse:
> > > I guess this should also be adjusted in `dracut-initramfs-restore.sh` and `lsinitrd.sh`.
> > 
> > 
> > Patching the initrd creation to handle this kind of user action is already a moot use case, but debugging on shutdown and querying its content is one step ahead... ok, it can be done for completeness, but I doubt that someone would hit this issue.
> 
> Users expect tools to behave the same thus someone will always hit this issue and eventually create an issue report as a result of that so we gain nothing being incomplete and in general being incomplete is considered bad practice.

Ok, also patched both scripts.

> That said the handling of machine-id has been subjected to changes in sd so we need to take a careful look at how it is being handled in sd now and how we should treat it which may or may not align with what sd ( now ) does.

AFAIK, its possible values are empty, non-existent, or "uninitialized" for now.

PR #2274, opened 2023-03-10
skip generating the initrd when installing an unified kernel image

kraxel:
## Changes

The KERNEL_INSTALL_IMAGE_TYPE variable is set by newer kernel-install versions.  In case it is set to "uki" skip initrd generation.

The related systemd update is here:
https://github.com/systemd/systemd/commit/3d5f0bfe4e72fdc4d8f8d65f96dc5501dfed8a64

## Checklist

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -11,6 +11,11 @@ if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
     exit 0
 fi
 
+# unified kernel images already have an initrd included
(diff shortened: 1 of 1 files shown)


johannbg:
We do not support environments where multiple initramfs generators are installed and or environments that dracut is not actually being used.

kraxel:
> We do not support environments where multiple initramfs generators are installed and or environments that dracut is not actually being used.

Care to explain why the install script checks for an existing initrd then?
See https://github.com/dracutdevs/dracut/commit/bc140a0b299d6f873f6d583d85cbdbc40afbc964

PR #2277, opened 2023-03-13
fix(dracut.sh): use dynamically uefi's sections offset

### LINKED ISSUE #2275: UKI section offsets are hard-coded
Looking at `dracut.sh` I can see that dracut is still hard-coding PE section offsets despite the warning in v252 release notes. This will cause overlapping PE sections at runtime should the stub's sections grow too much (our ci has been bitten by this, which is why ukify came about). It also fails to mark the `.linux` section as executable (on x86 it may be run in-place if the kernel is too old). Furthermore, a future systemd release may have EFI binaries with a non-zero image base. These hard-coded offsets will result in broken images that will be refused by firmware.

You could fix all this and gain extra features by using ukify, but I guess there is a case of [NIH](https://github.com/dracutdevs/dracut/issues/2191) going on.

keentux:
* Uefi section are creating by `objcopy` with hardcoded sections offset. This patch allow to have a dynamic set of offsets between each part of the efi file, needed to create an UKI. Offsets are simply calculated so no sections overlap, see https://wiki.archlinux.org/title/Unified_kernel_image#Manually
Moreover, efi stub file's header is parsed to apply the correct offsets according to the section alignment factor. More information here: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-windows-specific-fields-image-only
* Remove EFI_SECTION_VMA_INITRD, no need anymore as initrd section offset dynamically calculated.

## Checklist

Fixes #2275

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1023,3 +1023,26 @@ get_dev_module() {
     fi
     echo "$dev_drivers"
 }
+
+# Check if file is in PE format
+pe_file_format() {
+    if [[ $# -eq 1 ]]; then
+        local magic
+        magic=$(objdump -p "$1" \
+            | awk '{if ($1 == "Magic"){print strtonum("0x"$2)}}')
+        magic=$(printf "0x%x" "$magic")
+        # 0x10b (PE32), 0x20b (PE32+)
+        [[ $magic == 0x20b || $magic == 0x10b ]] && return 0
+    fi
+    return 1
+}
+
+# Get the sectionAlignment data from the PE header
+pe_get_section_align() {
+    local align_hex
+    [[ $# -ne "1" ]] && return 1
+    [[ $(pe_file_format "$1") -eq 1 ]] && return 1
+    align_hex=$(objdump -p "$1" \
+        | awk '{if ($1 == "SectionAlignment"){print $2}}')
+    echo "$((16#$align_hex))"
+}
--- a/dracut.sh
+++ b/dracut.sh
@@ -1506,7 +1506,6 @@ if [[ ! $print_cmdline ]]; then
             exit 1
         fi
         unset EFI_MACHINE_TYPE_NAME
-        EFI_SECTION_VMA_INITRD=0x3000000
         case "${DRACUT_ARCH:-$(uname -m)}" in
             x86_64)
                 EFI_MACHINE_TYPE_NAME=x64
@@ -1516,8 +1515,6 @@ if [[ ! $print_cmdline ]]; then
                 ;;
             aarch64)
                 EFI_MACHINE_TYPE_NAME=aa64
-                # aarch64 kernels are uncompressed and thus larger, so we need a bigger gap between vma sections
-                EFI_SECTION_VMA_INITRD=0x4000000
                 ;;
             *)
                 dfatal "Architecture '${DRACUT_ARCH:-$(uname -m)}' not supported to create a UEFI executable"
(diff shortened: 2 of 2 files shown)


dracut.sh
```diff
@@ -2467,29 +2464,43 @@ if [[ $uefi == yes ]]; then
         fi
     fi
 
+    offset=$(objdump -h "$uefi_stub" | awk 'NF==7 {size=strtonum("0x"$3);\
```
aafeijoo-suse, line 2467: Just a minor nitpick, I'd save the output of `objdump -h "$uefi_stub"` to an intermediate variable to check its return code and exit with a dfatal message if it's non-zero.

aafeijoo-suse (commented):
Thanks for looking into this!

dracut.sh
keentux: Thank you for your feedback.
I add a protection on the result of the command, to be sure that an offset is really taken.

medhefgo:
I forgot to mention that the offsets should really be aligned to the PE binary's SectionAlignment header field.

keentux:
> I forgot to mention that the offsets should really be aligned to the PE binary's SectionAlignment header field.

Thanks for that clarification. I am going to look on at.

keentux:
On my last changes, during the UEFI generation, it will parse the initial PE file's header (the stub) checking if it is a good PE format and catch the value of `sectionAlignment` . According to the PE documentation, each section should start on a factor of the value of `sectionAlignment`. 
@medhefgo What do you think about that. Should it be good up to you ?

dracut-functions.sh
```diff
@@ -1023,3 +1023,24 @@ get_dev_module() {
     fi
     echo "$dev_drivers"
 }
+
+# Check if file is in PE format
+pe_file_format() {
+    if [[ $# -eq 1 ]]; then
+        dos_header=$(hexdump -s 0 -n 64 -e'64/1 "%_c"' "$1")
+        signature=$(hexdump -s 128 -n 4 -e'4 "%_c"' "$1")
```
medhefgo, line 1031: This is just wrong. The PE header can reside virtually anywhere. You have to examine the offset in the DOS header to find the PE header.
You should build the latest master of systemd to test this as its PE header starts at offset 64 and your code spits out the wrong alignment.

Why not use `objdump -p`? Or better yet: why not just use ukify and get rid of senseless duplication of effort and gain support for more arches in the process?

keentux: Thank you for your feedback, indeed `objdump -p` is cleaner !
Changes applied

dracut-functions.sh
```diff
@@ -1023,3 +1023,25 @@ get_dev_module() {
...
 }
+
+# Check if file is in PE format
+pe_file_format() {
+    if [[ $# -eq 1 ]]; then
+        local magic
+        magic=$(objdump -p "$1" \
+            | awk '{if ($1 == "Magic"){print strtonum("0x"$2)}}')
+        # 0x10b (PE32), 0x20b (PE32+)
+        [[ -n $magic && ($magic == 0x20b || $magic == 0x10b) ]] && return 0
```
aafeijoo-suse, line 1034: `magic` contains a decimal value and it's being compared to hex values.

`magic=$(printf "0x%x" $magic)` before the check?

keentux: Thanks for this feedback.
Indeed, I thought, in the condition, hexa will be convert in decimal. I am going to update this.

keentux: Perfect! Update done.

aafeijoo-suse (approved):

johannbg (approved):
LGTM

PR #2281, opened 2023-03-20
Fix handling of `quiet` in loginit

bmwiedemann:
`dracut-lib.sh` can [set `DRACUT_QUIET=no`](https://github.com/dracutdevs/dracut/blob/e9b477423/modules.d/99base/dracut-lib.sh#L459)
and passes this to `loginit`

This pull request changes...

## Changes
* handling of `quiet` in `loginit`

## Checklist

Fixes #

### DIFF
--- a/modules.d/99base/loginit.sh
+++ b/modules.d/99base/loginit.sh
@@ -18,6 +18,6 @@ while read -r line || [ -n "$line" ]; do
     fi
     echo "<31>dracut: $line" >&5
     # if "quiet" is specified we output to /dev/console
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (commented):
You need to fix your commit message, please change it to something like `fix(base): correct handling of quiet in loginit`:

https://github.com/dracutdevs/dracut/actions/runs/4466552657/jobs/7845152049?pr=2281#step:4:8

Other than that, the patch looks good (it does what it was doing before, which is a bit contradictory to the `if "quiet" is specified we output to /dev/console` comment).

PR #2287, opened 2023-03-23
fix(debug xtrace): failures with mksh and with rd.live.debug

FGrose:
Drop _debug_on()_ _debug_off()_ project-wide as setting xtrace is only
local in mksh.  Use `set -x` directly instead. And use the shell's
`local -` feature to automate the restoration of local xtrace
option changes on function returns.

This fixes the bug with `rd.live.debug` where the _debug_on()_ call
in the first _getarg()_ call would fail to restore xtracing because
`RD_DEBUG` was not set. This also allows `rd.live.debug` to be
independent of `rd.debug` and eliminates 18 _debug_off/debug_on_ calls.

This feature is present in bash and dash and not needed in mksh.
In mksh, `local -` (`-` alone) is an alias for `typeset -p` (synonymous
with `declare -p` in bash). Use `local -` with additional arguments to
avoid this alternate behavior in mksh.

This code works with dash since v0.5.2 (2005-09-26) and with bash
since v4.4 (2016-09-15).

Replace the nulling of _debug_on()/debug_off()_ functions in
TEST-98-GETARG by overriding `set +x` in a temporary working copy of
the relevant code to aid in debugging a _getarg()_ failure.

### Checklist
- [✔] I have provided new testing code to replace the dropped functions.

### DIFF
--- a/modules.d/90crypt/parse-crypt.sh
+++ b/modules.d/90crypt/parse-crypt.sh
@@ -3,7 +3,7 @@
 type crypttab_contains > /dev/null 2>&1 || . /lib/dracut-crypt-lib.sh
 
 _cryptgetargsname() {
-    debug_off
+    set +x
     local _o _found _key
     unset _o
     unset _found
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -19,14 +19,6 @@ if [ -z "$PREFIX" ]; then
     [ -d /run/log ] || mkdir -p -m 0755 /run/log
 fi
 
-debug_off() {
-    set +x
-}
-
-debug_on() {
-    [ "$RD_DEBUG" = "yes" ] && set -x
-}
-
 # returns OK if $1 contains literal string $2 (and isn't empty)
 strstr() {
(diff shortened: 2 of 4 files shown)


FGrose:
> What happens when [/bin/sh points to mksh](https://github.com/dracutdevs/dracut/issues/1530) ?

With, `ln -sf mksh /bin/sh`, if module mksh is not requested, _resolve_deps()_ will test the wrong file for script installs with a shebang, `#!/bin/sh` or `#!/usr/bin/sh`.

Pull Request #2359 addresses this bug.

FGrose:
@LaszloGombos > What happens when [/bin/sh points to mksh](https://github.com/dracutdevs/dracut/issues/1530) ?  -- bug 2

Independent of this code, when these conditions apply,
1. host has `ln -sf mksh /bin/sh`, and
2. if a shell program is NOT specified, and
3. if NO modules are called with a dependency for the `99base` module (where _bash_ would be specified), then

_resolve_deps()_ will install _mksh_, as the source for `/bin/sh` as in Script installs with a shebang.

Pull Request #2362 addresses this bug.

Issue #2175, opened 2023-01-26
multipath boot hangs with dracut-059


aafeijoo-suse:
**Describe the bug**
Multipath is broken due to https://github.com/dracutdevs/dracut/commit/02e646fc7ec91e1fbaa0f2097f35781ae41da937

**Distribution used**
openSUSE Tumbleweed

**Dracut version**
059-358-g8ecd6e83

**Init system**
systemd 252.4

**To Reproduce**

1. Create a multipath setup
2. Try to boot the system

**Additional context**
After creating an initrd without including `multipathd.socket`, the system boots.
The definition of `multipathd.service` does not suggest that the inclusion is wrong:

https://github.com/dracutdevs/dracut/blob/3c16c76fb51683cd70c3643b122685b7cdb4178c/modules.d/90multipath/multipathd.service#L7

https://github.com/dracutdevs/dracut/blob/3c16c76fb51683cd70c3643b122685b7cdb4178c/modules.d/90multipath/multipathd.service#L28

I'm not sure if this only affects openSUSE Tumbleweed, or if others are affected as well.

Edit: https://github.com/dracutdevs/dracut/commit/02e646fc7ec91e1fbaa0f2097f35781ae41da937 reverted in 059+suse.360.g2e0ed5f7, so multipath works again in openSUSE Tumbleweed.

aafeijoo-suse:
CC: @cthbleachbit, what was your reasoning behind your change, only the multipathd.service definition? did you try it?

aafeijoo-suse:
Good, thanks for your feedback. I just wanted to know if this is a SUSE-only issue.

PR #2010, opened 2022-10-18
90multipath: install multipathd.socket

cthbleachbit:
## Changes

This pull request makes `90multipath` module install `multipathd.socket` from system paths along side with the systemd service unit (which lists the socket as a dependency). Without the socket unit file, systemctl invoked by dracut may report `failed to enable unit` during generation, and the resulting initrd will not have a usable multipathd at early boot.

## Checklist
- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

Fixes #

PR #2290, opened 2023-03-25
fix(multipath): enable service but disable socket

### LINKED ISSUE #2289: initrd is blocked because of conflict between multipathd.service  and initrd-cleanup.service
**Describe the bug**
initrd is blocked after `Finished dracut pre-pivot and cleanup hook` and initrd-cleanup.service is not activated.

image

**Distribution used**
openEuler

**Dracut version**
059

**Init system**
systemd

**To Reproduce**
Install multipath-tools in the system, and use `dracut --no-hostonly` to generate the initramfs. And there should be at least a multipath disk in the system. 

**Expected behavior**
initrd-cleanup.service should be activated and dracut continues switching root

**Additional context**
In previous commit 02e646fc7ec91e1fbaa0f2097f35781ae41da937, `multipathd.socket` is injected into initramfs when multipath module is added. And in the hook script `multipathd-needshutdown.sh` in cleanup, `multipath -l -v1` is called. If there is a multipath block discovered, `multipathd.service` will be activated by  `multipathd.socket`. However, `multipathd.service` is configured to be before and be conflict with `initrd-cleanup.service`, which results in `initrd-cleanup.service` can not be activated and initrd is blocked.

cthbleachbit:
## Changes

This fixes a hang in the cleanup hook: running multipath binary triggers activation of `multipathd.service` after it is (and must remain) stopped as dracut prepares to switch root in `initrd-cleanup.service`.

An earlier fix from PR #2010 was based on an old systemd behavior where missing socket will prevent the service from being enabled. This does not seem to be the case anymore. For completeness, I'm disabling the socket alone after enabling service.

At the end of the day we should probably remove `Also=multipathd.socket` from `multipathd.service` and upsteam this change. The whole point of having a socket enabled separately from the service is that systemd can start the service as-needed. Binding enable/disable status together would not be very sane.

## Checklist

Fixes #2289, #2175

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -137,7 +137,6 @@ install() {
             inst_simple "${moddir}/multipathd-configure.service" "${systemdsystemunitdir}/multipathd-configure.service"
             $SYSTEMCTL -q --root "$initdir" enable multipathd-configure.service
         fi
-        inst_simple "${systemdsystemunitdir}/multipathd.socket"
         inst_simple "${moddir}/multipathd.service" "${systemdsystemunitdir}/multipathd.service"
         $SYSTEMCTL -q --root "$initdir" enable multipathd.service
     else
--- a/modules.d/90multipath/multipathd.service
+++ b/modules.d/90multipath/multipathd.service
@@ -25,4 +25,3 @@ TasksMax=infinity
 
 [Install]
 WantedBy=sysinit.target
-Also=multipathd.socket


mwilck:
See https://github.com/opensvc/multipath-tools/issues/65

I think we should get rid of the entire `multipathd-needshutdown` logic. It isn't needed any more.

mwilck:
I don't know why #2210 was made because I wasn't involved, and there was no bug report linked.
I suppose the issue was related to udev; when multipath is running from udev rules it might try to access the socket.
But the startup sequence for multipathd has been changed such that it starts before udev ist starting, so that *should* not be a problem any more.

I don't think we should rely on socket activation for multipathd in the initrd.

mwilck:
> I suppose the issue was related to udev

That can't be the case; `multipath -u` and `multipath -U` as run from udev rules don't access the socket.

Another thing worth trying here would be to run multipath with the `D` option. This avoids that multipath tries to connect to the multipathd socket.

cthbleachbit:
> I don't think we should rely on socket activation for multipathd in the initrd.

Absolutely not. The only reason that the socket got involved is due to it declared in `Also=` from the service. This cause systemd to sync status to the socket unit when enabling/disabling the service. On a hindsight the service probablty should not even have this `Also=multipathd.socket` declared.

The `-D` option doesn't seem to be documented in the manual nor in the command help prints as of multipath-tools 0.9.3...

mwilck:
> > I don't think we should rely on socket activation for multipathd in the initrd.
> 
> Absolutely not. The only reason that the socket got involved is due to it declared in `Also=` from the service. This cause systemd to sync status to the socket unit when enabling/disabling the service. On a hindsight the service probablty should not even have this `Also=multipathd.socket` declared.

Well, as I said above, either multipathd must be running, or socket activation must be enabled, otherwise multipath activation in the initrd will fail. The best idea would be to start multipathd early in the intird.

The `Also=` line has been added 10y ago and never been questioned. You are right that `Also=` is wrong here: If multipathd is enabled, enabling the socket unit makes little sense. Likewise, the socket shouldn't be disabled just because someone disables the service. If socket activation is desired, the socket should be enabled and service should be disabled.

> The `-D` option doesn't seem to be documented in the manual nor in the command help prints as of multipath-tools 0.9.3...

Right. It's an undocumented feature :-) . It isn't meant for end users. But it might be useful for dracut. It would *not* make a difference for `multipath -u` though (see above). I need to double-check this again.

cthbleachbit:
> either multipathd must be running, or socket activation must be enabled

So dracut enables multipathd.service - fine on its own. Enabling service enables `multipathd.socket` as well because the service has `Also=multipathd.socket`.

mwilck:
> > either multipathd must be running, or socket activation must be enabled
> 
> So dracut enables multipathd.service - fine on its own. Enabling service enables `multipathd.socket` as well because the service has `Also=multipathd.socket`.

Understood. The `Also=` is wrong indeed.

mwilck:
Just sent a patch to `dm-devel@redhat.com` for the upstream service unit.

mwilck:
What I said about `-D` above was wrong, too, sorry.

`-D` only has an effect for map creation (`multipath`) or flushing (`multipath -f $map`, `multiapth -F`).
But the dracut scripts in question only run `multipath -l`, which doesn't connect to the socket.

So it remains unclear to me why the socket is activated.

jiayi0118:
This patch works in my system.

LGTM

cthbleachbit:
I'd guess that dracut needs to account for older versions of mp-tools.

Maybe I should just revert #2010 and remove `Also=` for the copy of `multipathd.service` in dracut's repo as well.

mwilck:
> IIUC, if Martin's patch removing the `Also=` from `multipathd.service` is accepted, I'd just simply revert [02e646f](https://github.com/dracutdevs/dracut/commit/02e646fc7ec91e1fbaa0f2097f35781ae41da937)

It has got positive review and is scheduled to be merged with https://github.com/opensvc/multipath-tools/pull/64

cthbleachbit:
Following upstream. My old fix reverted and `Also=` line removed.

aafeijoo-suse:
> Using `Also=$X.socket` in a service unit `$X` makes sense _if and only if the service unit has no `[Install]` section_. In that case, `systemctl enable $X.service` will install the socket unit _instead_ of the service unit, which makes sense for a daemon that has no purpose except serving the socket. But **multipathd** serves many more purposes, and therefore using `Also=multipathd.socket` is wrong.

Thanks for this background explanation Martin, sometimes systemd is too tricky...

PR #2293, opened 2023-03-27
fix(base): do not quote $CLINE in the `set` command

konfetka1989:
## Changes

- fix(base): do not quote $CLINE in the `set` command

## Checklist


aafeijoo-suse (commented):
Thanks, your fix is good, but you need to prepend `# shellcheck disable=SC2086`  to both lines to make github checks happy.

konfetka1989:
Added `# shellcheck disable=SC2086` lines. Thanks.

aafeijoo-suse (approved):

PR #2296, opened 2023-03-29
perf(dracut.sh): do not call find_mp_fstype twice

aafeijoo-suse:
Instead, save the return value in a temporary variable.

## Checklist


LaszloGombos (approved):
lgtm

PR #2298, opened 2023-03-29
feat(lsinitrd): notify user on missing compressor

FGrose:
Provide a notice to user if the required compressor package is missing. 
This rather than failing silently.

### Checklist


aafeijoo-suse (changes requested):
Please, simplify the message to just:

`echo "Need '${CAT%% *}' to unpack the initramfs."`

PR #2303, opened 2023-04-06
fix(memstrack): change memstrack report hook during kdump

liutgnu:
Previously memstrack report is attached to the cleanup hook. It is OK if memstrack is enabled for the 1st kernel. However if we want to memstrack the 2nd kernel during kdump process, the cleanup hook may not get triggered before the kdump finishes vmcore dumping and forcing reboot. As a result, the log collected by memstrack analyze service will not be outputted.

For example in fedora, the kdump-capture.service is wanted by initrd.target, same as memstrack.service. After the successful vmcore dumping, it will force reboot by 'reboot -f' by default [1], which will not wait for cleanup hook to finish.

In this patch, we will replace memstrack report cleanup hook with memstrack-report.service in kdump case, and make it wanted by initrd.target. So memstrack report will be triggered earlier.

[1]: https://src.fedoraproject.org/rpms/kexec-tools/blob/rawhide/f/dracut-kdump.sh#_29

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/99memstrack/module-setup.sh
+++ b/modules.d/99memstrack/module-setup.sh
@@ -20,9 +20,15 @@ install() {
     inst "/bin/memstrack" "/bin/memstrack"
 
     inst "$moddir/memstrack-start.sh" "/bin/memstrack-start"
-    inst_hook cleanup 99 "$moddir/memstrack-report.sh"
-
     inst "$moddir/memstrack.service" "$systemdsystemunitdir/memstrack.service"
-
     $SYSTEMCTL -q --root "$initdir" add-wants initrd.target memstrack.service
+
+    pid=$$
+    if grep -wq "kdumpbase" <<< $(cat /proc/"$pid"/cmdline | tr '\0' ' '); then
+        inst "$moddir/memstrack-report.sh" "/bin/memstrack-report"
+        inst "$moddir/memstrack-report.service" "$systemdsystemunitdir/memstrack-report.service"
+        $SYSTEMCTL -q --root "$initdir" add-wants initrd.target memstrack-report.service
+    else
+        inst_hook cleanup 99 "$moddir/memstrack-report.sh"
+    fi
 }
--- a/modules.d/99memstrack/memstrack-report.service
+++ b/modules.d/99memstrack/memstrack-report.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Memstrack Report Service
+DefaultDependencies=no
+Before=dracut-cmdline.service systemd-udevd.service local-fs-pre.target
+After=memstrack.service
+IgnoreOnIsolate=true
+ConditionKernelCommandLine=|rd.memdebug=4
+ConditionKernelCommandLine=|rd.memdebug=5
+
+[Service]
+Type=forking
+ExecStart=/bin/memstrack-report
+StandardInput=null
+StandardOutput=journal+console
+StandardError=journal+console


liutgnu:
Hi @ryncsn , Please help review the patch, there is one point which I'm not quite certain. By this patch, in kdump case, memstrack.service and memstrack-report.service and kdump-capture.service are all wanted by initrd.target. Though memstrack.service starts ahead of memstrack-report.sh, however the time interval for memstrack from start to stop is shorter than before(cleanup hook), maybe not enough to cover the whole kdump-capture.service. I don't know whether it will affect the memstrack data collection, maybe the longer time for memstrack to collect data, the better analyze/report it can give? If it is true, which stage other than initrd.target do you think is better in this case? Thanks!

modules.d/99memstrack/module-setup.sh
```diff
@@ -20,9 +20,15 @@ install() {
...
 
     inst "$moddir/memstrack-start.sh" "/bin/memstrack-start"
-    inst_hook cleanup 99 "$moddir/memstrack-report.sh"
-
     inst "$moddir/memstrack.service" "$systemdsystemunitdir/memstrack.service"
-
     $SYSTEMCTL -q --root "$initdir" add-wants initrd.target memstrack.service
+
+    pid=$$
+    if grep -wq "kdumpbase" <<< $(cat /proc/"$pid"/cmdline | tr '\0' ' '); then
```
ryncsn, line 27: kdumpbase is a out of tree dracut module, it may not be a good idea to be used a keyword here.

modules.d/99memstrack/module-setup.sh
```diff
@@ -20,9 +20,15 @@ install() {
...
+        inst "$moddir/memstrack-report.sh" "/bin/memstrack-report"
+        inst "$moddir/memstrack-report.service" "$systemdsystemunitdir/memstrack-report.service"
+        $SYSTEMCTL -q --root "$initdir" add-wants initrd.target memstrack-report.service
```
ryncsn, line 30: This makes initrd.target want both memstrack-report.service and memstrack.service? Won't this make memstrack stop right after getting started?

ryncsn:
Hi Tao, thanks for reporting such issue.

A few ideas after a quick review... 

To track the memory usage, memstrack should get started as early as possible and end as late as possible. initrd.target is a bit too late indeed.

Will sysinit.target be more suitable? Or maybe we can start memstrack use a dracut-cmdline hook? That should be easy to do and make it start much earlier than now.

Ending by starting a report.service in initrd.target may end memstrack too early I think, as systemd may decide to start the report.service very early. I'm not sure how to end the service properly, since if kdump capture service will intercept the normal booting progress and call "reboot -f", there seems no stable hooks to be used. Will it work to just add a ExecStop command for the memstrack.service unit? so when ever it get stopped, it will dump the report.

modules.d/99memstrack/module-setup.sh
liutgnu: Yes indeed, memstrack will stop right after getting started, so the time interval for memstrack to live is very short. I do captured some data, however I'm not very certain if the longer time for memstrack to live, the better report it can give. From your comments, I agree this solution is no longer appropriate.

modules.d/99memstrack/module-setup.sh
liutgnu: I didn't find a good way for dracut to determine whether it is making a normal initramfsdisk for 1st kernel, or making a kdump initramfsdisk for 2nd kernel. My original thought was to check the cmdline which invokes dracut. For kdump case, it will have string as "--add 'kdumpbase'", so it can act as a sign. I agree kdumpbase is out of tree for dracut, it may not be suitable for other linux distros other than fedora. Maybe putting the check in kexec-tools side, and patch dracut modules.d/99memstrack/module-setup.sh during its install stage is better?

liutgnu:
Hi Kasong,

> Hi Tao, thanks for reporting such issue.
> 
> A few ideas after a quick review...
> 
> To track the memory usage, memstrack should get started as early as possible and end as late as possible. initrd.target is a bit too late indeed.
> 
Agreed

> Will sysinit.target be more suitable? Or maybe we can start memstrack use a dracut-cmdline hook? That should be easy to do and make it start much earlier than now.
> 
I'm OK with the 2 targets, as long as it starts earlier. My only concern is the consumption of memory. The occupation of memstrack may shorten the available memory for other components, I will give it a try to see if OOM is more likely to occur.

> Ending by starting a report.service in initrd.target may end memstrack too early I think, as systemd may decide to start the report.service very early. I'm not sure how to end the service properly, since if kdump capture service will intercept the normal booting progress and call "reboot -f", there seems no stable hooks to be used. Will it work to just add a ExecStop command for the memstrack.service unit? so when ever it get stopped, it will dump the report.

Yeah, the ending target for memstrack is a problem. I will give ExecStop a try, thanks for the suggestion! In addtion, I'm thinking of another solution. When memstrack enabled in 2nd kernel, after kdump finishes its job, it will call a normal reboot. So we can keep memstrack end hooking on cleanup. If normal reboot fails, it will fall back to "reboot -f". How do you think?

PR #2305, opened 2023-04-11
fix(dracut-functions.sh): convert mmcblk to the real kernel module name

liutgnu:
In some x86_64 platforms such as Intel Elkhartlake, an issue of missing
    necessary modules due to udevadm drivers field unmatch the real kernel module
    name is found:
    
      $ udevadm info -a /dev/block/179:1
    
      looking at parent device '/devices/pci0000:00/0000:00:1a.0/mmc_host/mmc0/mmc0:0001':
        KERNELS=="mmc0:0001"
        SUBSYSTEMS=="mmc"
        DRIVERS=="mmcblk"
        ....
    
    The DRIVERS field, aka mmcblk will be given to instmods to install the
    corresponding mmc_block.ko kernel module. However mmc_block.ko cannot be
    selected by string mmcblk, as a result, mmc_block.ko cannot be installed
    in hostonly-mode strict, which will fail to bootup the machine such as in
    kdump cases:
    
      $ /usr/lib/dracut/dracut-install -D /var/tmp --kerneldir /lib/modules/$(uname -r)/ -m mmcblk
      dracut-install: Failed to find module 'mmcblk'
    
    In this patch, we will convert the string mmcblk to mmc-block, so the
    kernel module can be successfully loaded.
    
    Signed-off-by: Tao Liu <ltao@redhat.com>

## Checklist

Fixes #

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -983,13 +983,30 @@ block_is_netdevice() {
     block_is_nbd "$1" || block_is_iscsi "$1" || block_is_fcoe "$1"
 }
 
+# convert the driver name given by udevadm to the corresponding kernel module name
+get_module_name() {
+    local dev_driver
+    while read -r dev_driver; do
+        case "$dev_driver" in
+            mmcblk)
+                echo "mmc_block"
+                ;;
+            *)
+                echo "$dev_driver"
+                ;;
+        esac
+    done
+}
+
 # get the corresponding kernel modules of a /sys/class/*/* or/dev/* device
 get_dev_module() {
     local dev_attr_walk
     local dev_drivers
     local dev_paths
     dev_attr_walk=$(udevadm info -a "$1")
-    dev_drivers=$(echo "$dev_attr_walk" | sed -n 's/\s*DRIVERS=="\(\S\+\)"/\1/p')
+    dev_drivers=$(echo "$dev_attr_walk" \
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (commented):
Thanks for detecting this.

First, could you remove the "90" from the commit message? Although the commisery check does not fail, we should not include module numbers.

IMHO I'd fix this in the `get_dev_module` function, because it's an `udevadm` issue and this function is called in more places. Maybe directly, maybe by adding something like a `get_module_name` helper function, using a `case` instead of a single `if`, to be ready for other cases.

liutgnu:
> Thanks for detecting this.
> 
Thanks for reviewing the patch.

> First, could you remove the "90" from the commit message? Although the commisery check does not fail, we should not include module numbers.
> 
Agreed.

> IMHO I'd fix this in the `get_dev_module` function, because it's an `udevadm` issue and this function is called in more places. Maybe directly, maybe by adding something like a `get_module_name` helper function, using a `case` instead of a single `if`, to be ready for other cases.

Agreed, thanks a lot for the suggestion! I will update the patch and send v2.

liutgnu:
Hi, v2 is ready, sorry for the number of force-pushes. Thanks for the reviewing!

aafeijoo-suse (approved):
LGTM, thanks!

johannbg (approved):
LGTM

PR #2306, opened 2023-04-17
fix(memstrack): change memstrack startup hook

liutgnu:
Previously memstrack start is wanted by initrd.target, which is a pretty late stage in bootup. To have a good memory allocation record and a good analysis report, memstrack should start as early and finish as late as possible.

This patch mainly has 2 modifications:

1) We will start memstrack at the cmdline hook of dracut, and end memstrack
   at the cleanup hook. So the memstrack.service file is no longer needed.

2) If memstrack works in 1st kernel, we will use vwarn instead of cat to
   output the content of /.memstrack, becasue it is a wrapper of stderr,
   and more stable to have outputs in console.

   If memstrack works in 2nd kernel, aka in kdump case, we will not print
   /.memstrack here, the output will be handled in kexec-tools side [1].
   The reason is, for example in fedora, after vmcore dumping, kdump will
   use 'reboot -f' by default, which leaves no stable target to print
   /.memstrack. So the output is better handled in kexec-tools side.

[1]: https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/thread/YXSXU2A25QGHZHUMG3WH6H7ARH2L7NWS/

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/99memstrack/memstrack-report.sh
+++ b/modules.d/99memstrack/memstrack-report.sh
@@ -14,4 +14,10 @@ else
     done
 fi
 
-cat /.memstrack
+if ! [ -e /proc/vmcore ]; then
+    if [ -e /.memstrack ]; then
+        IFS= vwarn < /.memstrack
+    else
+        warn 'No memstrack log generated!'
+   fi
+fi


liutgnu:
Hi @ryncsn, please see if this solution is better than v1? Thanks!

ryncsn:
Hi Tao, looking at the dracut boot process:
https://github.com/dracutdevs/dracut/blob/master/man/dracut.bootup.7.asc

It seems actually having a dedicated service file like before will make memstrack start even earlier since we have 
> Before=dracut-cmdline.service
In memstrack.service.

Sorry my previous comment about using cmdline hook seems actually not a good idea... my bad.

For the stop and report part, about the vwarn change, it adds a `Warning: ` prefix for everyline, and does stderr have better stability than stdout?

Just a few thought after a brief review, how do you think?

liutgnu:
> Hi Tao, looking at the dracut boot process: https://github.com/dracutdevs/dracut/blob/master/man/dracut.bootup.7.asc
> 
> It seems actually having a dedicated service file like before will make memstrack start even earlier since we have
> 
> > Before=dracut-cmdline.service
> > In memstrack.service.
> 
> Sorry my previous comment about using cmdline hook seems actually not a good idea... my bad.
> 
OK, I will keep the previous memstrack.service implementation. I didn't notice the ' Before=dracut-cmdline.service' neither...

> For the stop and report part, about the vwarn change, it adds a `Warning: ` prefix for everyline, and does stderr have better stability than stdout?

Yes, stderr is more stable than stdout, and we do will have a 'Warning:' prefix for everyline, however I think it won't matter much since we are in a debug environment... I have tested with "cat /.memstrack 1>&2", however it didn't work stably as vwarn(). So I just leave vwarn() here.
> 
> Just a few thought after a brief review, how do you think?

liutgnu:
I will close the PR and submit v3. Thanks!

PR #2309, opened 2023-04-19
dracut: add support for one only module groups

airlied:
This supports the case where only a single module from a group will be put into the initramfs to keep initramfs sizes down where firmware files are getting large.

The kernel will denote a group like this:

```
MODULE_FIRMWARE_GROUP_ONLY_ONE("ga106-gsp");
MODULE_FIRMWARE_GROUP_LIST("nvidia/ga106/gsp/gsp-5303002.bin,nvidia/ga106/gsp/gsp-5258902.bin")
MODULE_FIRMWARE("nvidia/ga106/gsp/gsp-5258902.bin");
MODULE_FIRMWARE("nvidia/ga106/gsp/gsp-5303002.bin");
MODULE_FIRMWARE_GROUP_ONLY_ONE("ga106-gsp");
```

dracut will then scan for these group tags. It will use the group list inside the group only tags to and ignore the legacy firmware lines (which will be left in place for old dracut).  It will search for the first fw in the group list and work it's way down.

The kernel driver will then handle whatever it ends up finding.

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -1391,13 +1391,55 @@ static int install_firmware_fullpath(const char *fwpath)
         return ret;
 }
 
+static bool try_install_single_firmware(struct kmod_module *mod,
+                                        const char *value, bool in_group_one_only)
+{
+        bool group_only_one_found = false;
+        int ret = -1;
+        char **q;
+
+        STRV_FOREACH(q, firmwaredirs) {
+                _cleanup_free_ char *fwpath = NULL;
+
+                _asprintf(&fwpath, "%s/%s", *q, value);
+
+                if (strpbrk(value, "*?[") != NULL
+                    && access(fwpath, F_OK) != 0) {
+                        size_t i;
+                        _cleanup_globfree_ glob_t globbuf;
+
+                        glob(fwpath, 0, NULL, &globbuf);
+                        for (i = 0; i < globbuf.gl_pathc; i++) {
+                                if (group_only_one_found)
+                                        break;
+                                ret = install_firmware_fullpath(globbuf.gl_pathv[i]);
+                                if (ret != 0) {
+                                        if (!in_group_one_only)
+                                                log_info("Possible missing firmware %s for kernel module %s", value,
+                                                         kmod_module_get_name(mod));
+                                } else if (in_group_one_only)
+                                        group_only_one_found = true;
+
+                        }
+                } else {
+                        ret = install_firmware_fullpath(fwpath);
+                        if (ret != 0) {
+                                if (!in_group_one_only)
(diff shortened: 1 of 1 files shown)


airlied:
The kernel patch is:
https://lore.kernel.org/all/20230419043652.1773413-1-airlied@gmail.com/T/#u

mupuf:
> The kernel patch is: https://lore.kernel.org/all/20230419043652.1773413-1-airlied@gmail.com/T/#u

Given that some firmwares may come with bug/security fixes, shouldn't the order of module firmwares inside a group be specified to mean something? As in, the first firmware specified would be the least wanted/oldest, and the newest one would be the preferred?

airlied:
> > The kernel patch is: https://lore.kernel.org/all/20230419043652.1773413-1-airlied@gmail.com/T/#u
> 
> Given that some firmwares may come with bug/security fixes, shouldn't the order of module firmwares inside a group be specified to mean something? As in, the first firmware specified would be the least wanted/oldest, and the newest one would be the preferred?

It is defined by how the module info section is populated. It just happens that it's populated in reverse, so the last firmware you specify in the kernel is the first one dracut tries.

I can't think of anyway to define it better that would be backwards compatible.

mupuf:
> It is defined by how the module info section is populated. It just happens that it's populated in reverse, so the last firmware you specify in the kernel is the first one dracut tries.
> 
> I can't think of anyway to define it better that would be backwards compatible.

I think the code and the logic is sound. Sure, the interface is quirky, but what matters is that you document in what order the fw should be listed in the kernel drivers' source code, and the expected result in the module info section. As for dracut, please add a comment about the expected ordering from the module info, so that the interface is clear.

This should make it easier on both kernel and initramfs developers to know what's the expectation, reducing the chances of screwups on either sides that would silently lead to differences between drivers and distros. WDYT?

mupuf:
Thanks @airlied!

Conan-Kudo (approved):

airlied:
I also assume the procedure would be to get sign off on the kernel side first before landing this.

aafeijoo-suse (commented):
> I also assume the procedure would be to get sign off on the kernel side first before landing this.

Yes, thanks for your patch.

Meanwhile, you can fix your commit message (https://github.com/dracutdevs/dracut/actions/runs/4740463846/jobs/8440368788?pr=2309) and some formatting issues (https://github.com/dracutdevs/dracut/actions/runs/4740463843/jobs/8440368768?pr=2309)

airlied:
I've updated this, and new kernel patches are 

https://lore.kernel.org/linux-modules/20230704025322.2623556-1-airlied@gmail.com/T/#t

PR #2312, opened 2023-04-19
fix(url-lib): remove junk before path to ca-bundle

mikhailnov:
grep -F --binary-files=text -z .crt extracts junk before the path to the ca-bundle from /usr/lib64/libcurl.so.4 from RPM lib64curl4-7.86.0-1.x86_64 on ROSA Linux rosa2021.1.

Such a path is invalid, remove all junk before first slash (/) to make the path be valid.

libcurl.so.4: https://file-store.rosalinux.ru/download/ff55031fdd35038563dbc09f65040e7dbc9026fb
dracut log: https://file-store.rosalinux.ru/download/560b42b8f3b0819194a53c984d94fd9df773fbdc

### DIFF
--- a/modules.d/45url-lib/module-setup.sh
+++ b/modules.d/45url-lib/module-setup.sh
@@ -34,7 +34,9 @@ install() {
             if ! [[ $_nssckbi ]]; then
                 read -r -d '' _nssckbi < <(grep -F --binary-files=text -z libnssckbi "$_lib")
             fi
-            read -r -d '' _crt < <(grep -E --binary-files=text -z "\.(pem|crt)" "$_lib" | sed 's/\x0//g')
+            # Last sed expression removes all symbols before first '/' to clean up junk (https://askubuntu.com/a/438393)
+            # Junk may contain a line break
+            read -r -d '' _crt < <(grep -E --binary-files=text -z "\.(pem|crt)" "$_lib" | sed -e 's/\x0//g' -e 's,^[^/]*/,/,' | grep '^/' | head -n1)
             [[ $_crt ]] || continue
             [[ $_crt == /*/* ]] || continue
             if [[ -e $_crt ]]; then


aafeijoo-suse (commented):
I failed to reproduce this with Fedora Rawhide:

```
# read -r -d '' _crt < <(grep -E --binary-files=text -z "\.(pem|crt)" "/usr/lib64/libcurl.so.4.8.0" | sed 's/\x0//g')
# echo "$_crt"
/etc/pki/tls/certs/ca-bundle.crt
```

If this is a distro-specific packaging issue, maybe it should be fixed there.

mikhailnov:
@aafeijoo-suse Sorry for late responce, I missed an email about your reply. Grepping text from binary files in this way cannot guarantee anything, it may depend on compiler, linker flags, binutils version etc., so I would not consider it to be a packaging issue. `--with-ca-bundle=%{_openssldir}/certs/ca-bundle.crt` is defined in curl configure values.

```
$ rpm -E %optflags
-O2 -fomit-frame-pointer -gdwarf-4 -Wstrict-aliasing=2 -pipe -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fPIC -fstack-protector-strong --param=ssp-buffer-size=4 -m64 -mtune=generic
```

Maybe you can set the same optflags in SUSE and get the same result...

mikhailnov:
Here is junk that I get:

image

mikhailnov:
It would be better to use strings(1) which give a nice result without hackery:

$ strings /usr/lib64/libcurl.so.4.8.0 | grep ca-bundle
/etc/pki/tls/certs/ca-bundle.crt

but binutils may be not installed

aafeijoo-suse:
> It would be better to use strings(1) which give a nice result without hackery:
> 
> $ strings /usr/lib64/libcurl.so.4.8.0 | grep ca-bundle /etc/pki/tls/certs/ca-bundle.crt
> 
> but binutils may be not installed

I like this idea, it's a clean solution. We already use `strip` from `binutils` (or better `eu-strip` from `elfutils`, we require this package in SUSE), so we could use `strings`/`eu-strings` here... but let's see what other people think.

mikhailnov:
The only usage of `strings` is in lsinitrd.sh

mikhailnov:
What about strings vs eu-strings?

First try eu-strings, then strings, then current hackery?

PR #2310, opened 2023-04-19
fix(btrfs): add missing cmdline function

aafeijoo-suse:
Otherwise `dracut --print-cmdline` does not print the command line option added by the btrfs module.

## Checklist
- [X] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #2313, opened 2023-04-19
fix(watchdog-modules): add missing cmdline function

aafeijoo-suse:
Otherwise `dracut --print-cmdline` does not print the command line option added by the `watchdog-modules` module.

## Checklist

### DIFF
--- a/modules.d/04watchdog-modules/module-setup.sh
+++ b/modules.d/04watchdog-modules/module-setup.sh
@@ -10,32 +10,48 @@ depends() {
     return 0
 }
 
-# called by dracut
-install() {
-    return 0
-}
-
-installkernel() {
-    local -A _drivers
-    local _wdtdrv
+get_watchdog_drivers() {
+    local _wd _wdtdrv
 
     for _wd in /sys/class/watchdog/*; do
         ! [ -e "$_wd" ] && continue
         _wdtdrv=$(get_dev_module "$_wd")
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
> @aafeijoo-suse I noticed you have other similar PR as well, but perhaps best to ask this question on this PR.
> 
> From the description of this command line option ([1743473](https://github.com/dracutdevs/dracut/commit/1743473b2b7c1fb09b9eddc11d33fb3269877529)), it is not clear to me that it should print out "all" command line options, even the "internal" one
> 
> > Print the kernel command line for the given disk layout
> 
> Not sure what is the motivation for these PRs - perhaps consistency.
> 
> I am wondering when and how would users use this feature.

I was asked how to differentiate the parameters added to the command line (via bootloader) from those added by dracut (/etc/cmdline.d), and I thought the way to do this is using `dracut --print-cmdline`, but I found the list is incomplete (I'm using btrfs - #2310, so yes, the other 2 PRs are for consistency).

Excerpt from the man page:

```
module-setup.sh: cmdline()
This function should print the kernel command line options needed to boot the current machine setup
```

Seeing how other modules are implemented, not only the disk layout is included in the `cmdline()` method. E.g.:

https://github.com/dracutdevs/dracut/blob/a62e895db9510f0fc4c47ee81b1436096eca4d64/modules.d/90multipath/module-setup.sh#L43-L49

PR #2314, opened 2023-04-21
RFC: perf(base): do not unconditionally process cmdline with each `getcmdline` call

aafeijoo-suse:
At boot, every time a script calls a `getarg` function, the `getcmdline` function is called, which parses /etc/cmdline, /etc/cmdline.d/*.conf and /proc/cmdline.

Although a call only takes a few ns, this is quite inefficient due to the large number of times a `getarg` function is called. Only to show this number, I added a warning every time `getcmdline` is called in two vms, one with a minimal setup and other using the NetworkManager to get internet connection at boot (in both cases the content returned by `getcmdline` is always the same, i.e., no hook script is adding new content at boot). These are the results:

```
localhost:/home/dev # journalctl -b | grep -c getcmdline
43
nm:/home/dev # journalctl -b | grep -c getcmdline
67
```

This patch adds a new function `setcmdline`, which parses command line options, exports the `CMDLINE` variable (required by `dracut-util`), and persists its content in a temporary file, so that different processes can access it (e.g. systemd generators).

With this change, if a hook script adds new content in /etc/cmdline.d at boot, it must explicitly call `setcmdline` to update the content of the `CMDLINE` variable.

Let's compare the performance when calling 67 times the old version and the new one (with the new version, `setcmdline` is called only once, no cmdline changes at boot):

```
sh-5.2# hyperfine -S'bash --norc' '. /lib/dracut-lib.sh ; for ((i=1;i<=67;i++)); do getcmdline_v0; done' '. /lib/dracut-lib.sh ; for ((i=1;i<=67;i++)); do getcmdline_v1; done'
Benchmark 1: . /lib/dracut-lib.sh ; for ((i=1;i<=67;i++)); do getcmdline_v0; done
  Time (mean ± σ):       4.7 ms ±   0.6 ms    [User: 3.9 ms, System: 1.1 ms]
  Range (min … max):     4.3 ms …  11.6 ms    380 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: The first benchmarking run for this command was significantly slower than the rest (6.6 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.

Benchmark 2: . /lib/dracut-lib.sh ; for ((i=1;i<=67;i++)); do getcmdline_v1; done
  Time (mean ± σ):       1.1 ms ±   0.3 ms    [User: 1.2 ms, System: 0.2 ms]
  Range (min … max):     0.8 ms …   2.4 ms    991 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  '. /lib/dracut-lib.sh ; for ((i=1;i<=67;i++)); do getcmdline_v1; done' ran
    4.44 ± 1.29 times faster than '. /lib/dracut-lib.sh ; for ((i=1;i<=67;i++)); do getcmdline_v0; done'
```

## Checklist

### DIFF
--- a/modules.d/35connman/cm-config.sh
+++ b/modules.d/35connman/cm-config.sh
@@ -4,6 +4,7 @@ type cm_generate_connections > /dev/null 2>&1 || . /lib/cm-lib.sh
 
 if [ -n "$netroot" ] || [ -e /tmp/net.ifaces ]; then
     echo rd.neednet >> /etc/cmdline.d/connman.conf
+    setcmdline
 fi
 
 if getargbool 0 rd.debug -d -y rdinitdebug -d -y rdnetdebug; then
--- a/modules.d/35connman/cm-lib.sh
+++ b/modules.d/35connman/cm-lib.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-type getcmdline > /dev/null 2>&1 || . /lib/dracut-lib.sh
+type getargbool > /dev/null 2>&1 || . /lib/dracut-lib.sh
 
 cm_generate_connections() {
     if getargbool 0 rd.neednet; then
--- a/modules.d/35network-legacy/parse-ip-opts.sh
+++ b/modules.d/35network-legacy/parse-ip-opts.sh
@@ -43,6 +43,7 @@ BOOTDEV=$(getarg bootdev=)
 if [ -n "$NEEDBOOTDEV" ] && getargbool 1 rd.neednet; then
     #[ -z "$BOOTDEV" ] && warn "Please supply bootdev argument for multiple ip= lines"
     echo "rd.neednet=1" > /etc/cmdline.d/dracut-neednet.conf
+    setcmdline
     info "Multiple ip= arguments: assuming rd.neednet=1"
 else
     unset NEEDBOOTDEV
@@ -131,6 +132,7 @@ for p in $(getargs ip=); do
             "$(expr substr "$dev" 11 2)" \
             "$(expr substr "$dev" 13 2)" \
             >> /etc/cmdline.d/80-enx.conf
+        setcmdline
     fi
 done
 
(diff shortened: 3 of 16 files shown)


aafeijoo-suse:
> > This patch adds a new function `setcmdline`, which parses command line options, exports the `CMDLINE` variable (required by `dracut-util`), and persists its content in a temporary file, so that different processes can access it (e.g. systemd generators).
> 
> Just a thought - can we avoid creating some of the /etc/cmdline.d files ?
> 
> As an example change
> 
> ```
> echo rd.neednet >> /etc/cmdline.d/connman.conf
> setcmdline
> ```
> 
> to
> 
> ```
> echo rd.neednet | setcmdline
> ```
> 

We could, but I don't know if it's a good idea. If we remove these conf files, it would be more difficult to debug from the emergency shell which module adds command line parameters.

> > With this change, if a hook script adds new content in /etc/cmdline.d at boot, it must explicitly call setcmdline to update the content of the CMDLINE variable.
> 
> Would this break out-of-tree dracut modules ?

AFAIK the only documented use of /etc/cmdline.d to inject custom command line options is during the build process, not dynamically at boot. If an external dracut module is doing this (e.g. [openSUSE kdump](https://github.com/openSUSE/kdump/blob/eed2330577c0e1a4b392e0cc4c6227ca93be9d5b/dracut/kdump-boot.sh#L19)), then yes, it will need to adapt its code. But I think this should not be an obstacle to fix this inefficient code.

FGrose:
This will also clean up debug xtrace logs.

PR #2316, opened 2023-04-26
perf(dracut-init.sh): optimize inst_rule_initqueue function

aafeijoo-suse:
There is no need to grep inside each rule if the `/lib/dracut/need-initqueue` flag is already set.

## Checklist


LaszloGombos (approved):
LGTM.
Not sure if this is faster, but should not be slower.

PR #2317, opened 2023-04-26
perf(dracut): optimize get_cpu_vendor function

aafeijoo-suse:
This function performs two calls to `grep` and its result is always the same, so save it into a `cpu_vendor` variable.

## Checklist


LaszloGombos (approved):
LGTM

PR #2320, opened 2023-04-27
fix(dracut-init.sh): correct check in `is_qemu_virtualized` function

aafeijoo-suse:
Do not redirect `systemd-detect-virt` to /dev/null, otherwise, the `vm` variable is always empty. This function was working only thanks to the following /sys check.

## Checklist


tblume (approved):
LGTM

PR #2324, opened 2023-05-01
fix(dracut-functions): avoid calling grep with PCRE (-P)

### LINKED ISSUE #2323: Call to grep PCRE (-P Perl regex) blocked by SELinux policy
**Describe the bug**
During initramfs generation the call to grep in `get_maj_min()` in `dracut-functions.sh` (https://github.com/dracutdevs/dracut/blob/master/dracut-functions.sh#L247) generates many SELinux `execmem` errors. Depending on the SELinux policy `grep` may be blocked from executing mapped memory.  `grep`'s `-P` flag enables Perl Compabible RegEx (PCRE). This requires JIT'd Perl code which in turn requires executing code out of mapped memory.

This is not a fatal error since the fallback path (i.e., cache miss) still works as expected, but it generates a significant number of SELinux/audit messages.

**Distribution used**
[CBL-Mariner](https://github.com/microsoft/CBL-Mariner/tree/2.0/SPECS/dracut)

**Dracut version**
055

**Init system**
systemd

**To Reproduce**
I can put together in-depth repro steps if needed, but since it involves setting up a specific SELinux policy it's a bit non-trivial.
```bash
# Running in a CBL-Mariner 2.0 image with SELinux enabled. Depending on exact policy this will likely occur in other SELinux distros as well.

# Enable SELinux in enforcing mode
setenforce 1
getenforce
# >>> Enforcing
mkinitrd
# >>> Output omitted, but call is eventually made to 'dracut $dracut_args --add-drivers "$basicmodules" "${kernel_cfg[@]}" "$target" "$kernel"'

# Check for errors
journalctl -b | grep execmem
# >>> Apr 17 15:58:49 mariner-vm audit[4256]: AVC avc:  denied  { execmem } for  pid=4256 comm="grep" scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=process permissive=0
# >>> (repeats ~20x similar)

ausearch -m avc -ts recent
# >>> ----
# >>> time->Fri Apr 28 18:48:52 2023
# >>> type=PROCTITLE msg=audit(1682707732.438:104): proctitle=67726570002D6D31002D6F50005E2F6465762F73646132205C4B5C532B24002F7661722F746D702F6472616375742E48326664614E2F6D616A6D696E5F6361636865
# >>> type=SYSCALL msg=audit(1682707732.438:104): arch=c000003e syscall=9 success=no exit=-13 a0=0 a1=10000 a2=7 a3=22 items=0 ppid=22110 pid=22111 auid=1000 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=pts0 ses=5 comm="grep" exe="/usr/bin/grep" subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null)
# >>> type=AVC msg=audit(1682707732.438:104): avc:  denied  { execmem } for  pid=22111 comm="grep" scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=process permissive=0
# >>> ----
# >>> (repeats ~20x similar)
```
Decoded `proctitle=` is `grep-m1-oP^/dev/sda2 \K\S+$/var/tmp/dracut.H2fdaN/majmin_cache`

**Expected behavior**
Dracut should make efforts to avoid generating basic SELinux errors where possible

**Additional context**
While not quite as elegant I believe this should result in the same cache behavior without requiring JIT'd Perl:
```diff
diff --git a/dracut-functions.sh b/dracut-functions.sh
--- dracut-functions.sh
+++ dracut-functions.sh
@@ -243,9 +243,9 @@
     local _majmin
     local _out
 
     if [[ $get_maj_min_cache_file ]]; then
-        _out="$(grep -m1 -oP "^$1 \K\S+$" "$get_maj_min_cache_file")"
+        _out="$(grep -m1 -oE "^$1 \S+$" "$get_maj_min_cache_file" | awk '{print $NF}')"
     fi
 
     if ! [[ "$_out" ]]; then
         _majmin="$(stat -L -c '%t:%T' "$1" 2> /dev/null)"

```

dmcilvaney:
Invoking grep in Perl mode requires JIT'ing the Perl regex. This can run into issues with SELinix policy which will generally try to limit use of execmem in general purpose scripts/tools. This occurs since the JIT'd code will live in executable memory.

The PCRE only '\K' command in the Perl REGEX can be replaced by a call to awk instead.

## Changes
- Swap the use of `\K` Perl regex to a separate call to `awk` when doing device cache lookups to avoid using executable memory in grep which can generate SELinux permission errors.

## Checklist

Fixes #2323

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -244,7 +244,7 @@ get_maj_min() {
     local _out
 
     if [[ $get_maj_min_cache_file ]]; then
-        _out="$(grep -m1 -oP "^$1 \K\S+$" "$get_maj_min_cache_file")"
+        _out="$(grep -m1 -oE "^$1 \S+$" "$get_maj_min_cache_file" | awk '{print $NF}')"
     fi
 
     if ! [[ "$_out" ]]; then


aafeijoo-suse (commented):
Thanks for reporting this issue and providing a patch!

Could you please change your commit message to make the commisery check happy (https://github.com/dracutdevs/dracut/actions/runs/4852990662/jobs/8659360926?pr=2324) to something like: `fix(dracut-functions): avoid calling grep with PCRE (-P)`

Other than that, the patch looks good.

dmcilvaney:
@aafeijoo-suse Commit message updated.

aafeijoo-suse (approved):

PR #2329, opened 2023-05-08
ci: add gawk and more test dependencies to Debian container

LaszloGombos:
Allow more test coverage for manual test runs.

- tgt is needed for test 30 and 35
- nbd-server is needed for test 40
- gawk dependency has been introduced by https://github.com/dracutdevs/dracut/commit/f32e95bcadbc5158843530407adc1e7b700561b1
- install dracut instead of avoid installation of initramfs-tools to match actual dracut usage environment for testing
- remove workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=962300 as it is now fixed on Debian 12.

No additional test will run as part of CI, this PR only impacts test container generation.

### DIFF
--- a/test/container/Dockerfile-Debian
+++ b/test/container/Dockerfile-Debian
@@ -3,7 +3,10 @@ FROM docker.io/debian:latest
 MAINTAINER https://github.com/dracutdevs/dracut
 
 # Install needed packages for the dracut CI container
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (approved):
> gawk dependency has been introduced by https://github.com/dracutdevs/dracut/commit/f32e95bcadbc5158843530407adc1e7b700561b1

Yes, `mawk` is installed by default on Debian and `strtonum` is not defined. Thanks!

PR #2332, opened 2023-05-10
fix(dracut-lib): use poweroff instead of halt

lnykryn:
Using halt will keep the HW still powered on. This can be confusing for users. The screen is blank, but for example, the fan is still running, and they have no idea what is going on. So let's call poweroff.

### DIFF
--- a/modules.d/98dracut-systemd/dracut-emergency.sh
+++ b/modules.d/98dracut-systemd/dracut-emergency.sh
@@ -39,7 +39,7 @@ else
     export hook="shutdown-emergency"
     warn "$action has failed. To debug this issue add \"rd.shell rd.debug\" to the kernel command line."
     source_hook "$hook"
-    [ -z "$_emergency_action" ] && _emergency_action=halt
+    [ -z "$_emergency_action" ] && _emergency_action=poweroff
 fi
 
 /bin/rm -f -- /.console_lock
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -447,7 +447,7 @@ die() {
     fi
 
     if [ -n "$DRACUT_SYSTEMD" ]; then
-        systemctl --no-block --force halt
+        systemctl --no-block --force poweroff
     fi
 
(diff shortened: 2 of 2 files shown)


aafeijoo-suse (commented):
I see this was done on purpose in https://github.com/dracutdevs/dracut/commit/e2ecc2d8, unfortunately we don't have access to https://bugzilla.redhat.com/show_bug.cgi?id=1053655 :)
Maybe to (at least) see some info on the screen of what caused the `die()`, instead of power off the display?

LaszloGombos (approved):
LGTM

I like the idea of adding a few (30?) seconds delay before poweroff for the user to see what has happened on the screen and what is about to happen, but I would not wait for user input before issuing the shutdown. This could be done as a follow-up patch.

There is no very good solution here and poweroff is less confusing than halt. For some remote/unattended setup this could even damage the HW.. perhaps poweroff at least safer for the HW.

aafeijoo-suse (commented):
This would also need to be changed:

https://github.com/dracutdevs/dracut/blob/86c8a5a7c2573645e67537fb9975efab808d42c9/modules.d/98dracut-systemd/dracut-emergency.sh#L42

IMHO if we replace `halt` with `poweroff`, we should add the delay proposed by @rmetrich, so that the message passed to `die` can at least be seen, otherwise that message would be meaningless.

lnykryn:
Sure, we can do a follow-up with sleep. I think thirty seconds is a bit too much; ten should be enough.

AdamWill:
This is a problem for https://bugzilla.redhat.com/show_bug.cgi?id=2246410 . See also https://github.com/dracutdevs/dracut/pull/2545 .

Ten seconds seems waaaaay too short to me. In general, boot takes a while; people get bored and go to do something else while it's happening. For the specific case of media check, it takes longer. You can't assume someone will stay glued to the screen during the process such that they will catch a message that appears for ten seconds before the system suddenly shuts down.

I'd think ten minutes *at least* is the flat minimum to be *reasonably* sure someone caught the message.

PR #2339, opened 2023-05-12
feat(systemd-oomd): introducing the systemd-oomd module

aafeijoo-suse:
Introducing the `systemd-oomd` module.

## Checklist

### DIFF
--- a/modules.d/01systemd-oomd/module-setup.sh
+++ b/modules.d/01systemd-oomd/module-setup.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries \
+        oomctl \
+        "$systemdutildir"/systemd-oomd \
+        || return 1
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+
+}
+
+# Module dependency requirements.
+depends() {
+
+    # This module has external dependency on other module(s).
+    echo dbus systemd-sysusers
+    # Return 0 to include the dependent module(s) in the initramfs.
+    return 0
+
+}
+
+# Install the required file(s) and directories for the module in the initramfs.
+install() {
+
+    inst_multiple -o \
+        "$dbussystem"/org.freedesktop.oom1.conf \
+        "$dbussystemservices"/org.freedesktop.oom1.service \
+        "$systemdutildir"/systemd-oomd \
+        "$systemdsystemunitdir"/systemd-oomd.service \
+        "$systemdsystemunitdir/systemd-oomd.service.d/*.conf" \
+        "$systemdsystemunitdir"/systemd-oomd.socket \
+        "$systemdsystemunitdir/systemd-oomd.socket.d/*.conf" \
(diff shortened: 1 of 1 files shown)


tblume:
> What would be a reason for a user to include systemd-oomd module into initramfs and how would initramfs handle out of memory condition (e.g. start rescue shell?) ? Thanks !

oom situations occurred for me when using the livenet module.
The kernel started killing processes and the boot just hung without continuing to the emergency shell.
Reason is that liveimages need some additional memory for getting extracted and booted.
We even mention that in the dracut.cmdline manpage for the case that the liveimage is writable.

However, systemd-oomd needs some memory itself and might slightly increase the memory demand .
I'm working on an alternative approach, e.g. a memory check, that compares the size of the live image with the available RAM and fails to the emergency shell when this is below a certain number.
Still, I'm unsure whether this would be flexible enough for all dracut usecases.

So, what approach would be the preferred one?

aafeijoo-suse:
With this module, `ManagedOOM*` systemd service properties are handled before switching root.
It is not intended to be a universal solution to OOM problems, but it may be useful depending on the use case.
Also, this module would not be automatically added to the initrd, only if the user/conf wants it.

E.g., from the emergency shell (`rd.break=cmdline`):
```
sh-5.2# systemctl cat psi-test
# /etc/systemd/system/psi-test.service
[Unit]
Description=PSI test

[Service]
ExecStart=tail /dev/zero
ManagedOOMMemoryPressure=kill
ManagedOOMMemoryPressureLimit=10%
sh-5.2# systemctl start psi-test
sh-5.2# systemctl status psi-test         
× psi-test.service - PSI test
     Loaded: loaded (/etc/systemd/system/psi-test.service; static)
     Active: failed (Result: oom-kill) since Fri 2023-05-12 14:24:06 UTC; 29s ago
   Duration: 442ms
    Process: 608 ExecStart=tail /dev/zero (code=killed, signal=KILL)
   Main PID: 608 (code=killed, signal=KILL)
        CPU: 485ms

May 12 14:24:06 fedora systemd[1]: Started psi-test.service - PSI test.
May 12 14:24:06 fedora systemd[1]: psi-test.service: A process of this unit has been killed by the OOM killer.
May 12 14:24:06 fedora systemd[1]: psi-test.service: Main process exited, code=killed, status=9/KILL
May 12 14:24:06 fedora systemd[1]: psi-test.service: Failed with result 'oom-kill'.
```

tblume:
Ok, if we add add ManagedOOM*  to dracut-initqueue.service, it could probably address the issue I've mentioned.
In that case, I guess systemd-oomd is the better solution.
Still, we might want to make sure that the system enters the emergency shell if some processes are killed, since the boot process is most probably affected by the kill.

aafeijoo-suse:
> Ok, if we add add ManagedOOM* to dracut-initqueue.service, it could probably address the issue I've mentioned. In that case, I guess systemd-oomd is the better solution. Still, we might want to make sure that the system enters the emergency shell if some processes are killed, since the boot process is most probably affected by the kill.

I'm not sure if using this in dracut systemd services would be a good idea, because it would not be a generic fix and it'll only work if PSI is enabled in the kernel. E.g., SUSE kernels:

```
# zgrep PSI /proc/config.gz 
CONFIG_PSI=y
CONFIG_PSI_DEFAULT_DISABLED=y
```

So, you can continue working on your initial idea :)

PR #2342, opened 2023-05-16
fix(systemd-sysext): handle confexts and correct extensions search path

aafeijoo-suse:
- Support to confexts added in https://github.com/systemd/systemd/pull/26013 (instead of working on `/usr` and `/opt`, confext will extend only `/etc`).

- `/usr/lib/extensions/` not supported since https://github.com/systemd/systemd/commit/de862276

- In addition to these two fixes, add only raw binary disk images contained in regular `.raw` files, because dracut does not have a helper function to add directories recursively.

## Checklist

### DIFF
--- a/modules.d/01systemd-sysext/module-setup.sh
+++ b/modules.d/01systemd-sysext/module-setup.sh
@@ -26,20 +26,34 @@ depends() {
 # Install the required file(s) and directories for the module in the initramfs.
 install() {
 
+    # It's intended to work only with raw binary disk images contained in
+    # regular files, but not with directory trees.
+
     inst_multiple -o \
-        "/usr/lib/extensions/*" \
+        "/usr/lib/confexts/*.raw" \
+        "/var/lib/confexts/*.raw" \
+        "/var/lib/extensions/*.raw" \
+        "/etc/extension-release.d/extension-release.*" \
         "/usr/lib/extension-release.d/extension-release.*" \
+        "$systemdsystemunitdir"/systemd-confext.service \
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
> > * Support to confexts added in [confext: extension of sysext systemd/systemd#26013](https://github.com/systemd/systemd/pull/26013)
> 
> This link is not working for me. Perhaps [systemd/systemd#24864](https://github.com/systemd/systemd/issues/24864) ?

No, the previous link is ok (https://github.com/systemd/systemd/pull/26013), maybe you ran into temporary GitHub issues. 

> I assume a purpose of this module is to load a systemd sysext extension in initramfs.
> 
> What would be a practical typical scenario where a boot would require loading systemd sysext extension during the initramfs phase and not after the chroot ?

Excerpt:

The ‘confext’ concept follows the same principle as the [sysext](https://www.freedesktop.org/software/systemd/man/systemd-sysext.html) functionality. sysext allows to extend the host filesystem (or a unit’s) at runtime with signed images that add new files to /usr and /opt, using OverlayFS. In the same fashion, confext will extend /etc (and only /etc).

- Call using the same `systemd-sysext` binary: https://github.com/systemd/systemd/blob/5f655ffb1557aa9b20b3db7bc6fe28fe3b9c34cd/src/sysext/sysext.c#L1079

- `systemd-confext.service` ordered before `initrd-switch-root.target`: https://github.com/systemd/systemd/blob/5f655ffb1557aa9b20b3db7bc6fe28fe3b9c34cd/units/systemd-confext.service#L23-L24

Issue #1780, opened 2022-04-09
Improve error handling for rd.live.ram=1 when image does not fit into RAM


fenixleon:
**Describe the bug**
Passing the parameters `vmlinuz initrd=initrd.img root=live:CDLABEL=Fedora-WS-Live-35-1-2 rd.live.image rd.live.check rd.live.ram=1` or `vmlinuz initrd=initrd.img root=live:CDLABEL=Fedora-WS-Live-35-1-2 rd.live.image rd.live.check rd.live.ram` results with the system halting. The system starts the boot process with no errors and returns `The media check is complete, the result is: PASS`, followed by `Copying live image to RAM`, `(this may take a minute)` and ends with a system halt message.

Passing the parameters `vmlinuz initrd=initrd.img root=live:CDLABEL=Fedora-WS-Live-35-1-2 rd.live.image rd.live.ram nomodeset rhgb` ends with the same issue of the system halting.

Starting Fedora Workstation Live 35 boots successfully.

**Distribution used**
Fedora 35 Workstation

**Dracut version**
055-5.fc35

**Init system**
systemd 249 (v249.4-2.fc35)

**To Reproduce**
- Plug USB into laptop/PC
- Press tab
- Add `rd.live.ram=1` or `rd.live.ram` parameters

**Expected behavior**
- System to boot cleanly
- Ability to remove USB and for the system to run in RAM only.

**Additional context**
kernel - 5.14.10-300.fc35.x86_64
architecture - x86_64 64 bits
machine - HP Pavilion dv6 Notebook PC
BIOS or Legacy mode only. No EFI/UEFI support

johannbg:
Hmm sounds more like the live image has dropped the script that supported this ( the boot to ram option on live images got removed like a decade ago but the workaround rd.live.ram kept in place ), @Conan-Kudo are you aware of any changes that involve removed the rd.live.ram support on live images in Fedora?

fenixleon:
@johannbg, the [DRACUT.CMDLINE(7) Manual Page](https://github.com/dracutdevs/dracut/blob/a7a4b76c4ad5794f5f8a24ecd5b495f1512d05f7/man/dracut.cmdline.7.asc#booting-live-images) suggests that it's still a dracut kernel command line option.

johannbg:
@fenixleon that it's an available kernel commandline option does not mean that the script that are/were? used to make use of that available kernel command line option are still present on the live image that is being create in Fedora. 
[...]

PR #2344, opened 2023-05-17
RFC: feat(livenet): add memory size check depending on live image size

tblume:
For writeable live images, the memory must hold the complete image. That means it must be at least the size of the image plus some RAM necessary to run processes.
With too small RAM, the OOM Killer steps in and makes the boot hang. This patch lets the system go into the emergency shell instead.

The minimum RAM of 512Mb is just arbitrarily choosen. We could set any other number that appears appropriate instead. Alternatively, we could add a new parameter rd.min.ram or such.

This pull request changes...

## Changes

livenet module

## Checklist

Fixes #
OOM in livenetboot with too low memory.

### DIFF
--- a/modules.d/90livenet/livenetroot.sh
+++ b/modules.d/90livenet/livenetroot.sh
@@ -16,6 +16,13 @@ netroot="$2"
 liveurl="${netroot#livenet:}"
 info "fetching $liveurl"
 
+if getargbool 0 'rd.writable.fsimg'; then
+
+    imgsize=$(($(curl -sIL "$liveurl" | sed -n 's/Content-Length: *\([[:digit:]]*\).*/\1/p') / (1024 * 1024)))
+
+    check_live_ram $imgsize
+fi
+
 imgfile=
 #retry until the imgfile is populated with data or the max retries
 i=1
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -1117,6 +1117,13 @@ NOTE: There must be enough free RAM available to hold the complete image.
 +
 This method is very suitable for diskless boots.
 
+**rd.minmem=**__<megabyte>__::
+Specify minimum free RAM in MB after copying a live disk image into memory.
+The default is 1024.
++
+This parameter only applies together with the parameters rd.writable.fsimg
+or rd.live.ram.
+
 **root=**live:__<url>__::
 Boots a live image retrieved from __<url>__.  Requires the dracut 'livenet'
 module.  Valid handlers: __http, https, ftp, torrent, tftp__.
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -314,6 +314,8 @@ if [ -e /run/initramfs/live/${live_dir}/${squash_image} ]; then
 fi
 if [ -e "$SQUASHED" ]; then
     if [ -n "$live_ram" ]; then
+        imgsize=$(($(stat -c %s -- $SQUASHED) / (1024 * 1024)))
+        check_live_ram $imgsize
         echo 'Copying live image to RAM...' > /dev/kmsg
         echo ' (this may take a minute)' > /dev/kmsg
         dd if=$SQUASHED of=/run/initramfs/squashed.img bs=512 2> /dev/null
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -1151,3 +1151,28 @@ load_fstype() {
     done < /proc/filesystems
     modprobe "$1"
 }
+
+# parameter: size of live image
+# calls emergency shell if ram size is too small for the image
+check_live_ram() {
+    minmem=$(getarg rd.minmem)
+    minmem=${minmem:-1024}
+    imgsize=$1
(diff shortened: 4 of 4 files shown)


Conan-Kudo:
> Would it be acceptable to set 1G as default lower limit and add an boot parameter that could modify this?

Yes, I think that's reasonable. Distributions will want to set their own minimum limits, and giving a way to do that makes sense.

tblume:
> > Yes, my patch would cover #1780.
> 
> I am not sure. #1780 is not using livenet, just local ISO.
>
> CC @Conan-Kudo

Indeed, I overlooked this, sorry.
Still, the calculation of subtracting the live image size from the main memory should work for the writable case.
So, there should probably be a similar code for the dmsquash-live module.
And with an additional parameter to specify the minimum leftover RAM, this can be adjusted to the particular needs.
I will work on that.

aafeijoo-suse (commented):
Thanks Thomas! Some things I miss:

- Could you add some doc about the new parameter `rd.minmem.liverw` in `dracut.cmdline.7`, close to `rd.writable.fsimg`?
- All the calls to `curl` are encapsulated in functions within `url-lib.sh`, maybe the best approach is create a new function there with the code that retrieves the value of `imgsize`. What do you think?
- Some kind of error handling for getting `imgsize`. What happens if the status code of the `curl` request is not 200? If the `liveurl` address cannot be reached, we get the error `bash: / (1024 * 1024): syntax error: operand expected (error token is "/ (1024 * 1024)")`. Another example that can be easily fixed, if it returns 302 (redirect), the headers does not contain `Content-Length`. E.g. with our OBS:
```
sh-5.2# curl -I "https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live.aarch64-ALP.iso"
HTTP/2 302 
date: Wed, 21 Jun 2023 13:34:58 GMT
server: Apache
location: https://ftp.gwdg.de/pub/opensuse/repositories/systemsmanagement%3A/Agama%3A/Staging/images/iso/agama-live.aarch64-ALP.iso
vary: Accept,COUNTRY
content-type: application/x-iso9660-image

sh-5.2# curl -I "https://ftp.gwdg.de/pub/opensuse/repositories/systemsmanagement%3A/Agama%3A/Staging/images/iso/agama-live.aarch64-ALP.iso"
HTTP/1.1 200 OK
Server: nginx
Date: Wed, 21 Jun 2023 13:36:04 GMT
Content-Type: application/octet-stream
Content-Length: 991166464
Last-Modified: Fri, 16 Jun 2023 14:09:36 GMT
Connection: keep-alive
ETag: "648c6d20-3b140000"
Strict-Transport-Security: max-age=63072000; includeSubDomains; preload
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Accept-Ranges: bytes
```

In this case, the `-L` option does the trick:

```
sh-5.2# curl -IL "https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live.aarch64-ALP.iso"
HTTP/2 302 
server: Apache
date: Wed, 21 Jun 2023 13:46:55 GMT
vary: Accept,COUNTRY
location: https://ftp.gwdg.de/pub/opensuse/repositories/systemsmanagement%3A/Agama%3A/Staging/images/iso/agama-live.aarch64-ALP.iso
content-type: application/x-iso9660-image

HTTP/1.1 200 OK
Server: nginx
Date: Wed, 21 Jun 2023 13:46:56 GMT
Content-Type: application/octet-stream
Content-Length: 991166464
Last-Modified: Fri, 16 Jun 2023 14:09:36 GMT
Connection: keep-alive
ETag: "648c6d20-3b140000"
Strict-Transport-Security: max-age=63072000; includeSubDomains; preload
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Accept-Ranges: bytes
```

tblume:
Thanks for the feedback.
I've updated the PR according to your recommendations.

modules.d/90livenet/livenetroot.sh
```diff
@@ -16,6 +16,13 @@ netroot="$2"
 liveurl="${netroot#livenet:}"
 info "fetching $liveurl"
 
+if getargbool 0 'rd.writable.fsimg'; then
+
+    imgsize=$(($(curl -IL "$liveurl" | sed -n 's/Content-Length: *\([[:digit:]]*\).*/\1/p') / (1024 * 1024)))
```
aafeijoo-suse, line 21: You accidentally deleted the `-s` from the `curl` call, otherwise it also prints the progress:

```
sh-5.2# curl -IL "$liveurl" | sed -n 's/Content-Length: *\([[:digit:]]*\).*/\1/p'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0  885M    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
928514048
sh-5.2# curl -sIL "$liveurl" | sed -n 's/Content-Length: *\([[:digit:]]*\).*/\1/p'
928514048
```

Other than that, the patch looks good, thanks!

aafeijoo-suse (approved):

modules.d/90livenet/livenetroot.sh
tblume: No, that was an mistake.
Thanks for spotting!
->fixed

Conan-Kudo (approved):

PR #2355, opened 2023-05-21
fix(udev-rules): remove firmware.rules

### LINKED ISSUE #2009: remove legacy udev rules

Henrik66:
## Checklist

Fixes #2009


modules.d/95udev-rules/module-setup.sh
```diff
@@ -68,10 +67,7 @@ install() {
         "${udevdir}"/ata_id \
         "${udevdir}"/cdrom_id \
         "${udevdir}"/create_floppy_devices \
-        "${udevdir}"/firmware.sh \
         "${udevdir}"/firmware \
```
aafeijoo-suse, line 70: Isn't the `/lib/udev/firmware` binary also deprecated since https://github.com/systemd/systemd/commit/be2ea723b1d0 ?

aafeijoo-suse (approved):

PR #216, opened 2017-04-05
correct pallotron's name

pallotron:

PR #2360, opened 2023-05-21
fix(DRACUT_INSTALL): test variable properly, show failing input

FGrose:
Test for part/word counts greater than or equal to 1 and command excluding ' --debug', at least.

### Checklist

Note: `command -v "${DRACUT_INSTALL/ --debug/}"` may still fail with other arguments in DRACUT_INSTALL.
At least the example at line [#216](https://github.com/FGrose/dracut/blob/e6687df3feb3ac406221cb902eecacb34f574446/dracut-init.sh#L216) will work with an absolute path.

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -220,8 +220,8 @@ for i in $DRACUT_INSTALL; do
     DRINSTALLPARTS=$((DRINSTALLPARTS + 1))
 done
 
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (commented):
The purpose of this test is only to validate `DRACUT_INSTALL` when its value is guessed and set previously in `dracut-init.sh`, and not if a user assigns an incorrect value. I would not change this part, because it's impossible to know what a user can set here.

PR #2362, opened 2023-05-22
fix(mksh): implicit install requires printf

### LINKED ISSUE #1530: printf not installed

FGrose:
Check for and install _printf_, when missing, and _mksh_ is `/bin/sh`.

When these conditions apply,
1. host has `ln -sf mksh /bin/sh`, and
2. if a shell program is NOT specified, and
3. if NO modules are called with a dependency for the `99base` module (where _bash_ would be specified), then

_resolve_deps()_ will install _mksh_, as the source for `/bin/sh` as in Script installs with a shebang.

### Checklist

Fixes #1530; finishes #1534.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2080,6 +2080,15 @@ for ((i = 0; i < ${#include_src[@]}; i++)); do
     fi
 done
 
+[[ $(readlink "$initdir"/bin/sh) == mksh ]] && [[ ! -f "$initdir"/bin/printf ]] && {
+    # Handle mksh's dependency on printf when mksh is installed implicitly.
+    require_binaries printf || {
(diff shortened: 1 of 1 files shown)


FGrose:
@FGrose > Can you think of a solution to move the fix out of dracut.sh into mksh dracut module instead ?
@LaszloGombos 
Superseded by #2368.

PR #2365, opened 2023-05-26
fix(dracut.sh): correct path to UEFI stub on split-usr systems

floppym:
systemd always installs the UEFI stub in ${prefix}/lib/systemd/boot/efi.

On split-usr systems, systemdutildir is ${rootprefix}/lib/systemd, which makes dracut look in the wrong place.

Instead, grab 'prefix' from systemd.pc and store it in 'systemd_prefix'.

Bug: https://bugs.gentoo.org/765208

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1392,12 +1392,13 @@ esac
 abs_outfile=$(readlink -f "$outfile") && outfile="$abs_outfile"
 
 # Helper function to set global variables
-# set_global_var <pkg_config> <var> <value[:check_file]> [<value[:check_file]>] ...
+# set_global_var <pkg_config> <pkg_var[:exported_var]> <value[:check_file]> [<value[:check_file]>] ...
 set_global_var() {
     local _pkgconfig="$1"
-    local _var="$2"
+    local _pkgvar="${2%:*}"
+    local _var="${2#*:}"
     [[ -z ${!_var} || ! -d ${dracutsysrootdir}${!_var} ]] \
-        && export "$_var"="$(pkg-config "$_pkgconfig" --variable="$_var" 2> /dev/null)"
+        && export "$_var"="$(pkg-config "$_pkgconfig" --variable="$_pkgvar" 2> /dev/null)"
     if [[ -z ${!_var} || ! -d ${dracutsysrootdir}${!_var} ]]; then
         shift 2
         if (($# == 1)); then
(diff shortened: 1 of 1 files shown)


floppym:
The failing integration tests seem rather flaky. I don't think it's related to this PR.

aafeijoo-suse (commented):
I'm not sure if this patch is necessary... testing in Gentoo the current upstream code (before applying this patch) shows:

```
# set_global_var "systemd" "systemdutildir" "/lib/systemd:/lib/systemd/systemd-udevd" "/usr/lib/systemd:/usr/lib/systemd/systemd-udevd"
# echo $systemdutildir
/lib/systemd
```

If `systemdutildir` contains the `/usr` prefix, all the systemd global variables that use `systemdutildir` would be incorrect, not just the EFI stub location.

floppym:
As the Gentoo maintainer of dracut, I assure you the patch is necessary.

The EFI stub is not installed in systemdutildir. Many other systemd components are installed there, however.

floppym:
> If systemdutildir contains the /usr prefix, all the systemd global variables that use systemdutildir would be incorrect, not just the EFI stub location.

My patch does not add /usr to `systemdutildir`, so I don't understand this comment.

aafeijoo-suse:
> > If systemdutildir contains the /usr prefix, all the systemd global variables that use systemdutildir would be incorrect, not just the EFI stub location.
> 
> My patch does not add /usr to `systemdutildir`, so I don't understand this comment.

Quoting https://github.com/dracutdevs/dracut/pull/2365#pullrequestreview-1466934212 :

> testing in Gentoo the current upstream code (**before applying this patch**)

I was trying to say that with the current code `systemdutildir` is equal to `/lib/systemd`, at least with my Gentoo test vm.
In this case, if your patch does not add `/usr` to `systemdutildir`, isn't it doing the same thing? 

```bash
-            uefi_stub="$dracutsysrootdir${systemdutildir}/boot/efi/linux${EFI_MACHINE_TYPE_NAME}.efi.stub"
+            uefi_stub="$dracutsysrootdir${systemd_prefix}/lib/systemd/boot/efi/linux${EFI_MACHINE_TYPE_NAME}.efi.stub"
```

floppym:
Yes, systemdutildir is supposed to be equal to /lib/systemd.

The EFI stub is NOT installed in systemdutildir, however. It is installed in /usr/lib/systemd.

floppym:
> Ok, sorry, after checking systemd's meson.build now I get this `prefixdir` vs `rootprefixdir` issue. Unfortunately, `bootlibdir` is not exposed in `systemd.pc`, because I'd prefer a fix similar to what was done for `tmpfilesdir`:

Agreed. Unfortunately, adding bootlibdir to systemd.pc just makes the required logic more complex at this point since we would still need a fallback for old systemd versions.

floppym:
I force-pushed a new commit with the suggested changes. However, this github PR doesn't seem to be updating for some reason.

aafeijoo-suse (approved):
Thanks and sorry for the initial confusion.

PR #2374, opened 2023-06-04
feat(resume): also consider resume= in the cmdline for hibernation

### LINKED ISSUE #924: Resume does not work (potential #cee5853 regression?)
Referencing this bug which did not get any love at all:
https://bugzilla.redhat.com/show_bug.cgi?id=1842279

Fedora 32, x64 fresh installation. Using LVM + LUKS (that is, LUKS encrypted LVM) that contains rootfs + swap partitions.
This might be potentially happening since https://github.com/dracutdevs/dracut/pull/715

After a fresh Fedora install in the above indicated setup, /sys/power/resume is "0:0", which causes the dracut setup scripts to not install the required resume-from-disk machinery to recover from hibernation. This behaves like a loop since the lack of a defined swap partiton results in "dracut -f" generating ramdisks that do no contain the kernel arg "resume=foo" and so forth.

IMHO the script should gracefully handle the case where there's no swap partition defined for resume but one is available.

Please assume good intent. I'm not very familiar with Dracut nor Linux booting system so I could be wrong. However more users seem to experience this, suggesting there's indeed a bug and it is relatively easy to solve (see Bugzilla)

Henrik66:
Move getcmdline from dracut-lib.sh to dracut-dev-lib.sh to make it available on the host as well.

## Checklist

Fixes #924

### DIFF
--- a/modules.d/95resume/module-setup.sh
+++ b/modules.d/95resume/module-setup.sh
@@ -2,6 +2,9 @@
 
 # called by dracut
 check() {
+    # shellcheck source=dracut-dev-lib.sh
+    . "${dracutsysrootdir}${dracutbasedir}"/modules.d/99base/dracut-dev-lib.sh
+
     swap_on_netdevice() {
         local _dev
         for _dev in "${swap_devs[@]}"; do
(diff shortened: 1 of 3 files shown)


YHNdnzj:
Please also check #2438.

martinwhitaker:
@LaszloGombos , sorry for the delay in responding. I've reviewed and tested this PR and confirmed it does the same job as the revised code I suggested in PR #2160. I would recommend updating the comment in module-setup.sh, as I did.

modules.d/95resume/module-setup.sh
```diff
@@ -2,6 +2,9 @@
 
 # called by dracut
 check() {
+    # shellcheck source=dracut-dev-lib.sh
+    . "${dracutsysrootdir}${dracutbasedir}"/modules.d/99base/dracut-dev-lib.sh
```
aafeijoo-suse, line 6: Do we really need to move the boot-specific `getcmdline` function to `dracut-dev-lib.sh` and source it only for this? AFAIK `/etc/cmdline` and  `/etc/cmdline.d` are a dracut specific thing (ignored by systemd, and therefore by `systemd-hibernate-resume-generator`, see https://github.com/systemd/systemd/issues/22935) available only in the initrd (not on the host).

If the only concern of this PR is fixing the bug introduced by #2160, I'd just prefer @martinwhitaker's patch ( https://github.com/dracutdevs/dracut/pull/2160#issuecomment-1464496727), removing `/etc/cmdline` and  `/etc/cmdline.d` from the grep.

aafeijoo-suse: Maybe @Henrik66 understands it, because he didn't ask. Or is it another PR thought up by you and submitted by him so you can count your approval for your own PR?

boot-specific => it's a function provided by `dracut-lib.sh`, script which is included and used in the initrd only (at boot) by the scripts called from there (e.g. hooks).

Henrik66:
> I'd just prefer @martinwhitaker's patch ( https://github.com/dracutdevs/dracut/pull/2160#issuecomment-1464496727), removing /etc/cmdline and /etc/cmdline.d from the grep.

Since I do not fully understand all the possible side effects, I think it is best to close this PR. I hope others will help out.

Issue #2372, opened 2023-06-04
Silence the output of hardlinking files by default


LaszloGombos:
Only report the output of [hardlinking files](https://github.com/dracutdevs/dracut/blob/master/dracut.sh#L2085) if `--stdlog 5 `  or `--stdlog 6 ` and not by default.

PR #2375, opened 2023-06-04
feat(dracut-init.sh): do not print by default if a module cannot be installed

Henrik66:
Also do not print by default if udev rule is missing it is not a warning.

I had this idea to work on this after working on https://github.com/dracutdevs/dracut/issues/2372 .

## Checklist


aafeijoo-suse:
> I think these are good points to discuss, but for clarity memstrack info is still printed by default after this PR and modules themselves can individually still print messages - https://github.com/dracutdevs/dracut/blob/master/modules.d/99memstrack/module-setup.sh#L5

memstrack is just an example of a module with a check() method returning 0 or 1, I think its `dinfo` messages should be `ddebug` if this PR is accepted.

Henrik66:
I updated the PR just to change the log level for `Skipping udev rule`. Thanks.

aafeijoo-suse (approved):

PR #2379, opened 2023-06-07
fix(fcoe-uefi): exit early on empty vlan

pvalena:
Exit early in case `get_fcoe_boot_vlan` exits with error or just an empty string, instead of producing invalid config entry.

## Changes

Just handling a corner-case when vlan is empty; this should skip to another vlan if present.

## Checklist

### DIFF
--- a/modules.d/95fcoe-uefi/parse-uefifcoe.sh
+++ b/modules.d/95fcoe-uefi/parse-uefifcoe.sh
@@ -9,19 +9,22 @@ print_fcoe_uefi_conf() {
     mac=$(get_fcoe_boot_mac "$1")
     [ -z "$mac" ] && return 1
     dev=$(set_ifname fcoe "$mac")
-    vlan=$(get_fcoe_boot_vlan "$1")
-    if [ "$vlan" -ne "0" ]; then
-        case "$vlan" in
-            [0-9]*)
-                printf "%s\n" "vlan=$dev.$vlan:$dev"
-                dev="$dev.$vlan"
-                ;;
-            *)
-                printf "%s\n" "vlan=$vlan:$dev"
-                dev="$vlan"
-                ;;
-        esac
-    fi
+    vlan=$(get_fcoe_boot_vlan "$1") || return 1
+    case "$vlan" in
+        "0") ;;
+
+        '')
+            return 1
+            ;;
+        [0-9]*)
+            printf "%s\n" "vlan=$dev.$vlan:$dev"
+            dev="$dev.$vlan"
+            ;;
+        *)
+            printf "%s\n" "vlan=$vlan:$dev"
+            dev="$vlan"
+            ;;
+    esac
     # fcoe=eth0:nodcb
     printf "fcoe=%s\n" "$dev:nodcb"
     return 0


mwilck:
The code in question is obviously broken. But the fix is broken, too. Instead of combining several `if` statements and a `case` statement, the handling of  `$vlan` should be done in a single `case` clause.

@LaszloGombos, sorry I saw this too late. Please reopen. I have no permission to do this.

mwilck:
@pvalena, could you update the PR as indicated above?

modules.d/95fcoe-uefi/parse-uefifcoe.sh
```diff
@@ -10,6 +10,9 @@ print_fcoe_uefi_conf() {
     [ -z "$mac" ] && return 1
     dev=$(set_ifname fcoe "$mac")
     vlan=$(get_fcoe_boot_vlan "$1")
+    if [ -z "$vlan" ]; then
```
mwilck, line 13: Please move this into the case statement below, and simplify the `if` ... `case` section.

pvalena:
> @pvalena, could you update the PR as indicated above?

Will do. Not sure why I didn't cover it in the `case`...

pvalena:
@mwilck PTAL.

mwilck:
Thanks! Can you elmininate the `if` statement entirely, please?

pvalena:
Sure thing... this should have identical behaviour.

pvalena:
Just `shfmt` whitespace change... fixed.

mwilck (approved):
Looks good, thanks!

aafeijoo-suse (commented):
Can `get_fcoe_boot_vlan` return 1

https://github.com/dracutdevs/dracut/blob/6acfecae572fb457115b276b5b64d9424ad5187b/modules.d/99uefi-lib/uefi-lib.sh#L141-L144

after printing some value?

https://github.com/dracutdevs/dracut/blob/6acfecae572fb457115b276b5b64d9424ad5187b/modules.d/99uefi-lib/uefi-lib.sh#L147-L150

i.e., in the next loop iteration. If the answer is "no", the patch is ok.

mwilck:
> Can `get_fcoe_boot_vlan` return 1

@aafeijoo-suse, I don't understand the question. Neither the current code nor the patched code check the return value of `get_fcoe_boot_vlan()`, so the case that you are concerned about has never been taken into account. The empty string is not a valid VLAN identifier and as such, exiting early is justified.

I can't see that Pavel's patch would cause a regression.

If your intention is to improve this code further, you might as well stumble  upon the `[0-9]*)` clause in the case statement, which is unclean because it would accept something like `5foo`, which is not a valid VLAN number.

aafeijoo-suse:
> > Can `get_fcoe_boot_vlan` return 1
> 
> @aafeijoo-suse, I don't understand the question. Neither the current code nor the patched code check the return value of `get_fcoe_boot_vlan()`, so the case that you are concerned about has never been taken into account. The empty string is not a valid VLAN identifier and as such, exiting early is justified.

I just read what he tries to solve in the commit message: "_Exit early in case get_fcoe_boot_vlan **exits with error** or just an empty string_".
Of course, he is already improving the current code.

mwilck:
I guess we could write
```
    vlan=$(get_fcoe_boot_vlan "$1") || return 1
    case $§vlan in
    ...
```
I feel [uncomfortable wrt the return value of assignments](https://unix.stackexchange.com/questions/270828/how-is-the-return-status-of-a-variable-assignment-determined) in general, but this should work.

aafeijoo-suse:
> I guess we could write
> 
> ```
>     vlan=$(get_fcoe_boot_vlan "$1") || return 1
>     case $vlan in
>     ...
> ```
> 
> I feel [uncomfortable wrt the return value of assignments](https://unix.stackexchange.com/questions/270828/how-is-the-return-status-of-a-variable-assignment-determined) in general, but this should work.
> 
> @aafeijoo-suse, would you be fine with that?

Sure. Or just changing the commit message would be enough :)

mwilck:
@pvalena would you change the code as indicated above?

pvalena:
Sure, let's enhance it further. Sorry about the forgotten commit message :) will clear things up.

pvalena:
Rebased, fixed as suggested. Sorry it has taken so long :).

aafeijoo-suse (approved):
> Rebased, fixed as suggested. Sorry it has taken so long :).

No problem, it would have been enough to amend the commit message.

lnykryn (approved):

mwilck (approved):
LGTM

lnykryn (approved):

PR #2382, opened 2023-06-09
fix(multipath): do not install by default

LaszloGombos:
## Checklist


mwilck:
What's the purpose of this PR? In hostonly mode, the `check()` function does the right thing. In non-hostonly mode, multipath must be included.

mwilck:
Ah, this was the CI thing. Still, it's weird that this would be necessary.

PR #2386, opened 2023-06-11
perf(fs-lib): create /etc/fstab.empty runtime instead of buildtime

### LINKED ISSUE #2371: /etc/fstab.empty is created and not used
dracut creates /etc/fstab.empty for each generated initramfs and this does not seem to be used for anything. Get rid of it.

Henrik66:
## Checklist

Fixes #2371


LaszloGombos (approved):
It make sense to me to only create this empty file when it is actually needed.

Due to ramfs' design, at least one page will be allocated for one file however small the file is. On machine with large page size, this will become worse and waste too many memory.

PR #2392, opened 2023-06-11
fix(kernel-modules): add firewire-sbp2 driver

### LINKED ISSUE #2388: Add firewire-sbp2 to initramfs
See https://unix.stackexchange.com/questions/716608/booting-on-firewire-ieee1394-broken-in-fedora

Henrik66:
## Checklist

Fixes #2388


aafeijoo-suse (commented):
This tries to address a very specific issue in a distro configured with `hostonly=yes` by default, so a simply `instmods` is not going to add the driver, and when building in no-hostonly mode this driver is already installed... so I'm not sure what we gain here.

PR #2400, opened 2023-06-14
fix(dracut.sh): Make uki's reproducible

glance-:
If the user asks for the dracut output to be reproducible, we should ensure objcopy produces a reproducible uki to.

# Changes

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2519,6 +2519,7 @@ if [[ $uefi == yes ]]; then
     fi
 
     if objcopy \
+        ${DRACUT_REPRODUCIBLE:+--enable-deterministic-archives --preserve-dates} \
         ${uefi_osrelease:+--add-section .osrel="$uefi_osrelease" --change-section-vma .osrel=$(printf 0x%x "$uefi_osrelease_offs")} \
         ${uefi_cmdline:+--add-section .cmdline="$uefi_cmdline" --change-section-vma .cmdline=$(printf 0x%x "$uefi_cmdline_offs")} \
         ${uefi_splash_image:+--add-section .splash="$uefi_splash_image" --change-section-vma .splash=$(printf 0x%x "$uefi_splash_offs")} \


aafeijoo-suse (commented):
Sorry, I'm not an expert on the subject. Does this patch intend to set only the `Time/Date` field of the PE header to 01-01-1970? or is there any other timestamp that I'm missing?

glance-:
I'm currently on vacation and can't really dig into this, but from what my memory serves, diffoscope on the two uki's produced with and without this patched made sense. I can look deeper into this if it's needed when I'm back at work in a couple of weeks.

glance-:
The bits that diffoscope show's me when comparing using --disable-deterministic-archives vs --enable-deterministic-archives --preserve-dates:

```--- non-deterministic.efi
+++ deterministic.efi
├── objdump
│ @@ -4,15 +4,15 @@
│  start address 0x0000000000004000
│  
│  Characteristics 0x206
│       executable
│       line numbers stripped
│       debugging information removed
│  
│ -Time/Date            Mon Sep 18 09:01:03 2023
│ +Time/Date            Thu Jan  1 00:00:00 1970
│  Magic                        020b    (PE32+)
│  MajorLinkerVersion   2
│  MinorLinkerVersion   40
│  SizeOfCode           000000000000c000
│  SizeOfInitializedData        00000000048b6400
│  SizeOfUninitializedData      0000000000000000
│  AddressOfEntryPoint  0000000000004000
│ @@ -25,15 +25,15 @@
│  MajorImageVersion    0
│  MinorImageVersion    0
│  MajorSubsystemVersion        0
│  MinorSubsystemVersion        0
│  Win32Version         00000000
│  SizeOfImage          070edc00
│  SizeOfHeaders                00000400
│ -CheckSum             048c6a7b
│ +CheckSum             048cf3a3
│  Subsystem            0000000a        (EFI application)
│  DllCharacteristics   00000000
│  SizeOfStackReserve   0000000000000000
│  SizeOfStackCommit    0000000000000000
│  SizeOfHeapReserve    0000000000000000
│  SizeOfHeapCommit     0000000000000000
│  LoaderFlags          00000000
│ @@ -4194295,8 +4194295,8 @@
│   32bab2f:    cltd
│   32bab30:    imul   $0xa03c368c,%esp,%edx
│   32bab36:    and    %cl,0x493d06fc(%rax)
│   32bab3c:    sub    %cl,%cl
│   32bab3e:    test   %ebx,0x56236a29(%rbx)
│   32bab44:    ss pop %rax
│   32bab46:    mov    $0xc7,%ch
│ -[ Too much input for diff (SHA256: f36208e32d08c79280feef99ce2673296c8e8b6dd057e1100157e2f2b76063b3) ]
│ +[ Too much input for diff (SHA256: 19ce42db79b4033a7eebe250ad193913f4d1d4339deb332314268854433fece1) ]
```

So sure, someone might like to have a timestamp from the build time in the PE binary, but I'd preferred to have it reproducible instead.

The "Too much input for diff" shows that there might be other things that's not shown.

aafeijoo-suse (approved):

PR #2404, opened 2023-06-16
fix(multipath): explicitly check if `hostonly_cmdline` is `yes`

aafeijoo-suse:
`hostonly_cmdline` can be `no`, so the current check is invalid.

## Checklist


aafeijoo-suse:
Closing in favor of #2412

PR #2035, opened 2022-11-04
kernel install plugins: several fixlets

aafeijoo-suse:
- `shfmt -s` reformat
- `shellcheck`
- Remove call to undefined `usage` function in `dracut-rescue.install`
- Honor `KERNEL_INSTALL_VERBOSE=1`
- Minor improvements related to `/etc/os-release` in `dracut-rescue.install`
- Honor `KERNEL_INSTALL_CONF_ROOT`

## Checklist
- [X] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #2405, opened 2023-06-17
 install.d/*.install: Fix kernel install with uefi=yes

AndrewAmmerlaan:
This pull request fixes installing a kernel with uefi=yes in dracut config and layout=uki in kernel/install.conf.

Before this change this would fail because dracut needs a `--kernel-image` argument if uefi=yes. Additionally, if `kernel-install` has defined a `KERNEL_INSTALL_STAGING_AREA` for us we put the generated initrd/uki.efi there so the actual install can be handled by the appropriate plugins. This simplifies the code, otherwise we would need some extra logic that is already present in `90-loaderentry.install` and `90-uki-copy-install`. 

If we are dealing with some unknown `KERNEL_INSTALL_LAYOUT` or `KERNEL_INSTALL_STAGING_AREA` is not defined we revert to the old behaviour for backwards compatibility. This can happen if we have defined some custom `KERNEL_INSTALL_LAYOUT` or if we aren't using systemd's `kernel-install` (e.g. Gentoo's `installkernel-gentoo` for use on non-systemd systems).

## Changes

- If kernel-install has defined a staging area for us (KERNEL_INSTALL_STAGING_AREA) install generated initrd/uki.efi there. The actual install is then handled by `90-loaderentry.install` or `90-uki-copy-install`.

- Also skip regeneration if an uki.efi already exists in the kernel image directory.

- Pass --kernel-image to dracut, this is required to generate an uki (uefi=yes)

- Don't fsfreeze a tmpfs, this does not work (KERNEL_INSTALL_STAGING_AREA is often on a tmpfs)

- For `51-dracut-rescue.install` we simply add `--no-uefi`. Getting this to work with uki's requires some additional work here and/or in `90-uki-copy-install`. `--no-uefi` is a quick fix for now to ensure that install will at least not error out if `dracut_rescue_image=yes`, `uefi=yes` and `layout=uki`.

## Checklist

Signed-off-by: Andrew Ammerlaan <andrewammerlaan@gentoo.org>

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2594,6 +2594,9 @@ freeze_ok_for_fstype() {
         zfs)
             return 1
             ;;
+        tmpfs)
+            return 1
+            ;;
         btrfs)
             freeze_ok_for_btrfs "$outfile"
             ;;
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -12,18 +12,18 @@ if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
 fi
 
 # Mismatching the install layout and the --uefi/--no-uefi opts just creates a mess.
-if [[ "$KERNEL_INSTALL_LAYOUT" == "uki" && -n "$KERNEL_INSTALL_STAGING_AREA" ]]; then
+if [[ $KERNEL_INSTALL_LAYOUT == "uki" && -n $KERNEL_INSTALL_STAGING_AREA ]]; then
     BOOT_DIR_ABS="$KERNEL_INSTALL_STAGING_AREA"
     IMAGE="uki.efi"
     UEFI_OPTS="--uefi"
-elif [[ "$KERNEL_INSTALL_LAYOUT" == "bls" && -n "$KERNEL_INSTALL_STAGING_AREA" ]]; then
+elif [[ $KERNEL_INSTALL_LAYOUT == "bls" && -n $KERNEL_INSTALL_STAGING_AREA ]]; then
     BOOT_DIR_ABS="$KERNEL_INSTALL_STAGING_AREA"
     IMAGE="initrd"
     UEFI_OPTS="--no-uefi"
 else
     # No layout information, use users --uefi/--no-uefi preference
     UEFI_OPTS=""
-    if [[ -d "$BOOT_DIR_ABS" ]]; then
+    if [[ -d $BOOT_DIR_ABS ]]; then
         IMAGE="initrd"
     else
         BOOT_DIR_ABS="/boot"
@@ -32,9 +32,10 @@ else
 fi
 
 ret=0
+
 case "$COMMAND" in
     add)
-        if [[ "$IMAGE" == "uki.efi" ]]; then
+        if [[ $IMAGE == "uki.efi" ]]; then
             IMAGE_PREGENERATED=${KERNEL_IMAGE%/*}/uki.efi
         else
             IMAGE_PREGENERATED=${KERNEL_IMAGE%/*}/initrd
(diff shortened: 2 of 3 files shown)


AndrewAmmerlaan:
I fixed the commit message as instructed by the CI bot.

I don't think the remaining failures are caused by the changes here, please let me know if this is not the case.

aafeijoo-suse (approved):
I'd have loved to land #2035 cleanup first, but...
LGTM, thanks!

AndrewAmmerlaan:
The next release of systemd will ship with a new kernel-install plugin: `60-ukify.install` [1] and an accompanying new `KERNEL_INSTALL_UKI_GENERATOR` variable. In anticipation of this I have added another commit respecting this variable and `KERNEL_INSTALL_INITRD_GENERATOR`.

@aafeijoo-suse I took the liberty of basing my fixes upon your earlier fixes, I hope you don't mind.

Let me know what you think

[1] https://github.com/systemd/systemd/blob/main/src/kernel-install/60-ukify.install.in

aafeijoo-suse:
> @aafeijoo-suse I took the liberty of basing my fixes upon your earlier fixes, I hope you don't mind.

On the contrary, thanks for that!

> The next release of systemd will ship with a new kernel-install plugin: `60-ukify.install` [1] and an accompanying new `KERNEL_INSTALL_UKI_GENERATOR` variable. In anticipation of this I have added another commit respecting this variable and `KERNEL_INSTALL_INITRD_GENERATOR`.

Oops, the `KERNEL_INSTALL_INITRD_GENERATOR` variable caused heated arguments in the past (#1825)... Personally, I don't have any problem with that, because these scripts are just plugins following the spec of the `kernel-install` provider (systemd). But dracut's main upstream admin won't allow it. So, if you want this PR to be merged, I would skip both variables for now.

AndrewAmmerlaan:
> Oops, the `KERNEL_INSTALL_INITRD_GENERATOR` variable caused heated arguments in the past (#1825)... Personally, I don't have any problem with that, because these scripts are just plugins following the spec of the `kernel-install` provider (systemd). But dracut's main upstream admin won't allow it. So, if you want this PR to be merged, I would skip both variables for now.

Alright, lets avoid repeating that argument again then. Though after the next systemd release if the users config file would enable ukify as the uki generator then `60-ukify.install` will fail to find an initrd to include in the uki since we didn't generate one (we generated an uki instead). Effectively this would make it impossible to combine dracut with ukify. Which is probably not that big of an issue since I don't think there is any advantage to having dracut generate the initrd and then let ukify combine things into a uki, over having dracut just generate the uki.

In any case we can always just patch this downstream in Gentoo if the need arises. The important bit is that this plugin shouldn't break if dracut.conf specifies `uefi=yes` so lets focus on this first.

AndrewAmmerlaan:
> could you run shfmt -w -s install.d/50-dracut.install install.d/51-dracut-rescue.install?

Now that is a cool trick, I will be using that more often. Thanks!

aafeijoo-suse:
> > could you run shfmt -w -s install.d/50-dracut.install install.d/51-dracut-rescue.install?
> 
> Now that is a cool trick, I will be using that more often. Thanks!

We use it here just so we can check our scripts with https://github.com/luizm/action-sh-checker on our CI. These install.d scripts were not formatted properly...

aafeijoo-suse (commented):
I just tried this for the non-uki layout and it's failing, could you please double-check? it works after explicitly setting `--kver`.

```diff
-            "$BOOT_DIR_ABS/$IMAGE" \
-            "$KERNEL_VERSION"
+            --kver "$KERNEL_VERSION" \
+            "$BOOT_DIR_ABS/$IMAGE"
```

AndrewAmmerlaan:
> I just tried this for the non-uki layout and it's failing, could you please double-check? it works after explicitly setting `--kver`.

That is weird, for me it works either way (with or without explicitly setting `--kver`).

I made the change you suggested anyway because it can't hurt to set this explicitly.

aafeijoo-suse (approved):
> That is weird, for me it works either way (with or without explicitly setting --kver).
> I made the change you suggested anyway because it can't hurt to set this explicitly.

Hmm, no idea... but I'm not going to dig into that, and as you said `--kver` does not hurt.
Thanks!

aafeijoo-suse (approved):

AndrewAmmerlaan:
:partying_face: Thanks for merging!

Issue #2378, opened 2023-06-07
dbus in the initrd just like this is a bad idea


poettering:
So I recently noticed that dracut nowadays embeds dbus in the initrd. That's a bad idea and leads to deadlocks like this one: https://bugzilla.redhat.com/show_bug.cgi?id=1976653#c4

This then led people to submit this PR: https://github.com/bus1/dbus-broker/pull/271

Which is wrong too.

D-Bus simply isn't ready to run in the initrd, because it doesn't allow to schedule at which point in time services shall become activatable. activatable services are activatable during the entire runtime of the bus and if something talks via the bus to a service like that too early it will hang if the activation request cannot be fulfilled yet because the system is not booted up far enough for this to work. The result of that are deadlocks.

So, I am sorry, but the dbus support in dracut should never have been merged like this.

This can be solved properly, but requires some engineering in dbus. See: https://github.com/bus1/dbus-broker/issues/259

But that hasn't materialized yet (though dbus-broker upstreams agree this makes sense). But until this running dbus in the initrd, is just prone to issues like the mentioned.

dbus-broker PR 271 is bogus if you ask me, because it rearranges an upstream unit file to match a downstream change in a unit file that dracut made. Specifically, dracut's dbus support patches out some deps in dbus.socket. But if so it should really patch out the same deps in the broker unit file too, and not expect upsrream dbus-rboker to adjust for that...

dvdhrm:
> dracut has support for D-Bus since Nov 2020 (#900), I wonder how 2.5 years later this problem shows up. @teg @dvdhrm could you share your opinion on this?

I think Lennart did a splendid job in explaining the race-condition regarding activation of dbus services. The underlying cause is a lack of APIs to add dbus-activatable-services at runtime. Hence, systemd has to treat all dbus-services activatable once dbus is running, which can easily conflict with configurations specified in the respective service units.

poettering:
it does not.

jmpolom:
Fedora has opted to remove the network-legacy module from dracut in Fedora 40 effectively forcing NetworkManager into the initrd if you need early boot networking. This seems like an actual problem given what's detailed here and no resolution to the issue described? I don't think anyone wants an initrd that deadlocks/hangs.

PR #2406, opened 2023-06-17
fix(bluetooth): warn user instead of icluding it by default

Henrik66:
Bluetooth module can result in a known race condition at boot. Discourage its usage by default.

See https://github.com/dracutdevs/dracut/issues/2378#issuecomment-1584699033

## Checklist


AdamWill:
LGTM.

lnykryn (approved):
Although I would rather see the Bluetooth support in Linux get better, I agree that it should be opt-in now.

aafeijoo-suse:
> @aafeijoo-suse - are you also Ok with this PR ?

I have mixed feelings... I don't know how many distros other than Fedora are affected (at least we never got any complaints with any SUSE distro), but after #2378, maybe it's the safe approach for now.

PR #2410, opened 2023-06-19
btrfs: Fix --no-kernel builds

Werkov:
initrd built without modules would fail when it attempts to load a kernel module. (Use case scenario: kernel with built-in module.)

This pull request changes...

## Checklist

Fixes #


Werkov:
Thanks for the pointers. Rebased on top of #2310. (Consequentially, I couldn't use the `install`/`installkernel` distinction and I kept the predicate for whole cmdline generation.)

aafeijoo-suse:
> Thanks for the pointers. Rebased on top of #2310. (Consequentially, I couldn't use the `install`/`installkernel` distinction and I kept the predicate for whole cmdline generation.)

I was thinking about this:

```diff
diff --git a/modules.d/90btrfs/module-setup.sh b/modules.d/90btrfs/module-setup.sh
index 3a047053..e893c2bf 100755
--- a/modules.d/90btrfs/module-setup.sh
+++ b/modules.d/90btrfs/module-setup.sh
@@ -32,6 +32,8 @@ cmdline() {
 # called by dracut
 installkernel() {
     instmods btrfs
+
+    printf "%s\n" "$(cmdline)" > "${initdir}/etc/cmdline.d/00-btrfs.conf"
 }
 
 # called by dracut
@@ -56,6 +58,4 @@ install() {
 
     inst_multiple -o btrfsck btrfs-zero-log
     inst "$(command -v btrfs)" /sbin/btrfs
-
-    printf "%s\n" "$(cmdline)" > "${initdir}/etc/cmdline.d/00-btrfs.conf"
 }
```

Regardless, the `multipath` module would need a similar fix.

aafeijoo-suse:
> Just last week I was also poking around a similar problem. I think the actual root cause is maybe that error from modprobe should not be fatal or maybe should be handled better at https://github.com/dracutdevs/dracut/blob/master/modules.d/90kernel-modules/parse-kernel.sh#L16.

That is only non-systemd code:

https://github.com/dracutdevs/dracut/blob/2ff7b4702c2098f7db63b0ffb118da0b5c0723c1/modules.d/90kernel-modules/module-setup.sh#L149-L151

Werkov:
@aafeijoo-suse Regarding your suggestion -- I'm (mis?)lead by the "called by dracut" comment -- if it was only `install`/`installkernel` who calls the function it'd be fine but I saw `cmdline` called also from other parts to synthesize the cmdline -- in that case, `--no-kernel` should be honored too, IMO.

@LaszloGombos That sounds more systematic indeed, however, can dracut _runtime_ access `no_kernel` value? (In contrast to dracut _build_ time. A kernel-dependent initrd may still treat modprobe errors fatal.)

aafeijoo-suse:
> @aafeijoo-suse Regarding your suggestion -- I'm (mis?)lead by the "called by dracut" comment -- if it was only `install`/`installkernel` who calls the function it'd be fine but I saw `cmdline` called also from other parts to synthesize the cmdline -- in that case, `--no-kernel` should be honored too, IMO.

That is a different issue, the option `--print-cmdline` produces the same output without caring about the `--kernel-only`/`--no-kernel` options. I think that is out of the scope of this PR, not only the `btrfs` module is affected, and your proposed patch checking the value of `no_kernel` is not producing the expected results in all the cases.

E.g., missing `rd.driver.pre=btrfs` from `--print-cmdline`:

```
# dracut -f -a btrfs --stdlog 3 test.img
# lsinitrd test.img | grep btrfs.conf
-rw-r--r--   1 root     root           21 Apr 26 14:43 etc/cmdline.d/00-btrfs.conf
# dracut --print-cmdline test.img
 resume=UUID=68501f45-2145-498a-babc-eda0932047b9 root=UUID=13b9ef45-6df6-49de-98f9-d93ae6f8d353 rootfstype=btrfs rootflags=rw,relatime,discard=async,space_cache=v2,subvolid=266,subvol=/@/.snapshots/1/snapshot,subvol=@/.snapshots/1/snapshot
```

aafeijoo-suse (approved):
Thanks!

aafeijoo-suse:
I realized that we only solved the `--no-kernel` issue, but `rd.driver.pre` is still added if the kernel has btrfs built-in, right?

Maybe we can add a workaround similar to what the `multipath` module [does](https://github.com/dracutdevs/dracut/blob/7ed765dd23e4c5616c82d1cbf8b4dbceaafc7647/modules.d/90multipath/module-setup.sh#L44-L48):

```diff
diff --git a/modules.d/90btrfs/module-setup.sh b/modules.d/90btrfs/module-setup.sh
index 5d881332..13606ae4 100755
--- a/modules.d/90btrfs/module-setup.sh
+++ b/modules.d/90btrfs/module-setup.sh
@@ -26,7 +26,9 @@ depends() {
 cmdline() {
     # Hack for slow machines
     # see https://github.com/dracutdevs/dracut/issues/658
-    printf " rd.driver.pre=btrfs"
+    if grep -m 1 -q -w btrfs /proc/modules; then
+        printf " rd.driver.pre=btrfs"
+    fi
 }

```

Werkov:
Yes, built-in module + manual `--no-kernel` is exactly my scenario. Users who build module-full initrd but with btrfs built-in may be still affected.

(This seems like stacking hacks on top of each other, I wonder why modprobe can't deal with module dependencies better (i.e. without explicit preload of "leaf" modules).)

PR #2411, opened 2023-06-19
feat(35-network): add rd.net.dhcp6.stateless to run dhclient with -S

fayak:
## Changes

- add `rd.net.dhcp6.stateless=0` to run dhclient -6 -S if IP is to be provided by SLAAC only
- Reshape dhclient-script to allow `/tmp/net.$netif.resolvconf` write if we have a `RENEW6` (which is obtained by `dhclient -S -6`) so we can still have /etc/resolv.conf in the end

## Checklist

Fixes #

### DIFF
--- a/modules.d/35network-legacy/dhclient-script.sh
+++ b/modules.d/35network-legacy/dhclient-script.sh
@@ -75,11 +75,25 @@ setup_interface() {
     [ -n "$hostname" ] && echo "echo ${hostname%."$domain"}${domain:+.$domain} > /proc/sys/kernel/hostname" > /tmp/net."$netif".hostname
 }
 
+setup_namesrv6() {
+    namesrv=$new_dhcp6_name_servers
+    domain=$new_domain_name
+    # get rid of control chars
+    search=$(printf -- "%s" "$new_dhcp6_domain_search" | tr -d '[:cntrl:]')
+    if getargbool 1 rd.peerdns; then
+        [ -n "${search}${domain}" ] && echo "search $search $domain" > /tmp/net."$netif".resolv.conf
+        if [ -n "$namesrv" ]; then
+            for s in $namesrv; do
+                echo nameserver "$s"
+            done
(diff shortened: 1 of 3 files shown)


modules.d/35network-legacy/ifup.sh
```diff
@@ -478,7 +478,11 @@ for p in $(getargs ip=); do
                 ;;
             dhcp6)
                 load_ipv6
-                do_dhcp -6
+                if getargbool 0 rd.net.dhcp6.stateless; then
+                    do_dhcp -6 -S
+                else
+                    do_dhcp -6
+                fi
```
aafeijoo-suse, line 485: There are 2 more calls to `do_dhcp -6`:

https://github.com/dracutdevs/dracut/blob/33a66ed04bdc30eccb172a0cd6dcc36d9d74f825/modules.d/35network-legacy/ifup.sh#L487

https://github.com/dracutdevs/dracut/blob/33a66ed04bdc30eccb172a0cd6dcc36d9d74f825/modules.d/35network-legacy/ifup.sh#L541

Shouldn't the `rd.net.dhcp6.stateless` option also be checked there?

modules.d/35network-legacy/dhclient-script.sh
```diff
@@ -75,11 +75,25 @@ setup_interface() {
...
+                echo nameserver "$s"
+            done
+        fi >> /tmp/net."$netif".resolv.conf
+    fi
+}
+
 setup_interface6() {
     domain=$new_domain_name
     # get rid of control chars
     search=$(printf -- "%s" "$new_dhcp6_domain_search" | tr -d '[:cntrl:]')
```
aafeijoo-suse, line 96: Is the `search` variable still needed in `setup_interface6`?

aafeijoo-suse:
BTW, please remove the prefix `35-` from the module name in the commit message.

PR #2418, opened 2023-06-28
chore(gentoo.conf): remove examples to avoid confusion

### LINKED ISSUE #2415: remove unused config examples
https://github.com/dracutdevs/dracut/tree/master/dracut.conf.d

The examples that are used can be kept.

As an example, I do not think anybody is using the gentoo examples. Not even gentoo packaging.

Henrik66:
These examples are not used and not maintained and changing them would not have an impact on Gentoo.

Fixes #2415


floppym:
I wasn't even aware these files existed. Let's remove them. Thanks!

floppym:
I did a little digging, and it looks like I stopped using the "example" config files back in 2017. So definitely safe to remove these.

https://github.com/gentoo/gentoo/commit/90bb5611b5aa9279ff3b0d4f5c25240bf4b0f9db

aafeijoo-suse (approved):

PR #2420, opened 2023-07-05
feat(systemd-rfkill): remove module

aafeijoo-suse:
The rfkill service and socket provided by systemd will not be activated in the initrd since https://github.com/systemd/systemd/commit/49c55abc

## Checklist


tblume (approved):
LGTM

PR #2421, opened 2023-07-05
fix(man): add missing initrd-root-device.target to flow chart

aafeijoo-suse:
This target was added since https://github.com/dracutdevs/dracut/commit/d4efc0ae and is ordered between `basic.target` and `sysroot.mount`, see `man bootup.7` (`dracut-pre-mount.service` starts after).

## Checklist


tblume (approved):
LGTM

PR #2426, opened 2023-07-12
dracut.sh: add --sbat option to add sbat policy to UKI

esposem:
Take existing .sbat section from the uefi stub and merge it with vmlinux .sbat (if it exists) and user-provided .sbat file using the new --sbat option.

For some reasons, --update-section in objcopy does not resize the .sbat section, so remove the section from the stub and add it to the UKI as new one, to avoid having incomplete SBAT strings.

This pull request changes...

## Checklist

Fixes #


esposem:
Not sure what `TEST: root filesystem on LVM on encrypted partitions of a RAID-5   [FAILED]` has to do with my patch. Any help?

aafeijoo-suse:
Nice thread: https://lore.kernel.org/lkml/7DF38657-99C7-4C88-8835-7EE28E82C829@zytor.com/T/#u

esposem:
> Nice thread: https://lore.kernel.org/lkml/7DF38657-99C7-4C88-8835-7EE28E82C829@zytor.com/T/#u

I am the author of that patch. We will probably not end up having anything in kernel upstream, but distros might add their own SBAT section to the vmlinux binary. So IMHO this PR is still needed, also because:
- sbat will be optional
- ukify (the systemd tool that builds UKI the same way as dracut does) also supports it. So if you want to maintain a minimum of consistency with the other tools, you need this.

dracut.sh
```diff
@@ -2453,6 +2464,24 @@ fi
...
 umask 077
 
+sbat_default="sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md"
+sbat_out=$DRACUT_TMPDIR/uki.sbat
+
+clean_sbat_string() {
+    inp=$1
+    temp=$DRACUT_TMPDIR/temp.sbat
+    sed "s|$sbat_default||g" "$inp" > "$temp"
+    awk -i inplace 'NF' "$temp"
```
aafeijoo-suse, line 2474: Is the `awk` required? Would not be enough `sed -i "/${sbat_default//\//\\/}/d" "$temp"` ?

dracut.sh
```diff
@@ -272,6 +272,10 @@ Creates initial ramdisk images for preloading modules
                         Use [FILE] as a splash image when creating an UEFI
                          executable. Requires bitmap (.bmp) image format.
   --kernel-image [FILE] Location of the kernel image.
+  --sbat [FILE]         Location of the SBAT string to be added to .sbat.
```
aafeijoo-suse, line 275: Why do you we need to save the SBAT string to a file? this option should be just `--sbat [STRING]` or `[TEXT]`, what you like most.

aafeijoo-suse (commented):
Please, also document the new option in `dracut.8.asc` and `dracut.conf.5.asc`.

dracut.sh
esposem: I am not bash expert, but your line doesn't seem to do what I wanted. I need to get rid of the unnecessary newlines that are left when removing SBAT_DEFAULT, and I found online that `awk` does the job :) Feel free to suggest any alternative command line, as long as it works I have no issue in changing the command.

dracut.sh
esposem: Usually the sbat string is relatively long (see SBAT_DEFAULT), and not ideal to write by hand or paste in console. I think a file is better, especially if it is extracted from some other PE already.

dracut.sh
```diff
@@ -2453,6 +2464,24 @@ fi
 
 umask 077
 
+sbat_default="sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md"
+sbat_out=$DRACUT_TMPDIR/uki.sbat
+
+clean_sbat_string() {
+    inp=$1
+    temp=$DRACUT_TMPDIR/temp.sbat
```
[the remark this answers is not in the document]
aafeijoo-suse: and for clean too: https://github.com/dracutdevs/dracut/pull/2426#discussion_r1270349195

dracut.sh
aafeijoo-suse: The `sed` I suggested removes lines matching the pattern (no new empty lines are left).

```
localhost:~ # objcopy -O binary --only-section=.sbat /usr/lib/systemd/boot/efi/linuxx64.efi.stub stub.sbat
localhost:~ # cat stub.sbat 
sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md
systemd,1,The systemd Developers,systemd,253,https://systemd.io/
systemd.opensuse,1,The openSUSE Project,systemd,253.1-1.2,mailto:security@suse.de
localhost:~ # sbat_default="sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md"
localhost:~ # sed -i "/${sbat_default//\//\\/}/d" stub.sbat
localhost:~ # cat stub.sbat
systemd,1,The systemd Developers,systemd,253,https://systemd.io/
systemd.opensuse,1,The openSUSE Project,systemd,253.1-1.2,mailto:security@suse.de
```

dracut.sh
aafeijoo-suse: Hmm... I'd prefer not to use a file for this, but let's see what others think.

dracut.sh
esposem: you are absolutely right, sorry for the confusion. I'll use your command instead of awk

dracut.sh
esposem: ok

aafeijoo-suse: FWIW, if your main concern handling the string as a command line option, it could also be passed using the `sbat=` variable in a conf file, so both needs would be covered.

esposem: @aafeijoo-suse  please check if now is better

aafeijoo-suse (commented):
Other than that, it looks good from my side. Just one minor thing I forgot last time, could you please add the bash completion for the new `--sbat` option after https://github.com/dracutdevs/dracut/blob/master/shell-completion/bash/dracut#L49

esposem:
See if it makes sense for you :+1:

aafeijoo-suse:
> See if it makes sense for you +1

Have you added the new changes to the commit? I can see the [diff](https://github.com/dracutdevs/dracut/compare/1ffcb0c54e5777c39ea5cfd18ca1a2b860047f93..6b8a4eed296c36af5cab470767a4ead8014a416d) but the [PR](https://github.com/dracutdevs/dracut/pull/2426/files) was not updated...
btw you can reword the `and user-provided .sbat file` part of the commit message :)

esposem:
done

aafeijoo-suse (approved):
Thanks!

PR #2427, opened 2023-07-12
ci: install util-linux-systemd and systemd-boot into openSUSE container

aafeijoo-suse:
- `findmnt` is installed by `util-linux-systemd`. E.g.:

https://github.com/dracutdevs/dracut/actions/runs/5424796254/jobs/10097178325#step:4:53

- `/usr/lib/systemd/boot/efi/linuxx64.efi.stub` is installed by `systemd-boot`. E.g.:

https://github.com/dracutdevs/dracut/actions/runs/5424796254/jobs/10097178325#step:4:55

## Checklist

### DIFF
--- a/test/container/Dockerfile-OpenSuse-latest
+++ b/test/container/Dockerfile-OpenSuse-latest
@@ -9,5 +9,5 @@ RUN dnf -y install --setopt=install_weak_deps=False \
     sudo kernel dhcp-client qemu-kvm /usr/bin/qemu-system-$(uname -m) e2fsprogs \
     tcpdump iproute iputils kbd NetworkManager btrfsprogs tgt dbus-broker \
     iscsiuio open-iscsi which ShellCheck shfmt procps pigz parted squashfs ntfsprogs \
-    multipath-tools \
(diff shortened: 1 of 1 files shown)


tblume (commented):
https://github.com/dracutdevs/dracut/blob/master/test/TEST-18-UEFI/test.sh#L74
 expects the file in  /usr/lib/gummiboot but the systemd-boot package installs it to:
/usr/lib/systemd/boot/efi.
Is the ci taking care of this?

aafeijoo-suse:
> https://github.com/dracutdevs/dracut/blob/master/test/TEST-18-UEFI/test.sh#L74 expects the file in /usr/lib/gummiboot but the systemd-boot package installs it to: /usr/lib/systemd/boot/efi. Is the ci taking care of this?

That is to explicitly change the default path of the uefi stub (via `--uefi-stub` option).

https://github.com/dracutdevs/dracut/blob/dbdab2d87fa49c7fdd08b274d8a21e6046360cad/dracut.sh#L1528-L1530

tblume (approved):
I see, thanks.

LGTM

PR #2428, opened 2023-07-13
chore(Makefile): improve AUTHORS and CONTRIBUTORS targets

Henrik66:
Make make rules more generic to be able to use them not just in manual release process but also in CI release process.

This PR should not have an observable effect on any of the release generation steps. This is a refactor to reuse more code.


Henrik66:
Still a draft until I find time to fix: https://github.com/Henrik66/dracut/actions/runs/5571418110

GitHub Releases requires a tag

Henrik66:
The script now export the computed release to Github Action. It also detects if there is a new commit since last release that would warrant a new release generation.

lnykryn (approved):

PR #2429, opened 2023-07-14
fix(test): running tests no longer requires to be root

### LINKED ISSUE #1147: test: support non-root unit tests
While working on unit tests for Dracut's cpio utilities I was hoping to be able to run them as a non-root user. This doesn't appear to be currently possible, due to the `$EUID" == "0"` assertions scattered throughout.
It'd be nice if the requirement for root could be specified on a per-test basis via the `test_check()` hook. In the case of non-root execution, tests requiring root could be skipped.

Henrik66:
After https://github.com/dracutdevs/dracut/commit/fd9cd02cb21f25d32fec4cd9889457ea6361f45e, tests no longer requires to be root. Being able to run VMs to run tests is required, but that should not imply root user.

Fixes https://github.com/dracutdevs/dracut/issues/1147 and https://bugs.gentoo.org/298014

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,7 +270,6 @@ endif
 endif
 
 check: all syncheck
-	@[ "$$EUID" == "0" ] || { echo "'check' must be run as root! Please use 'sudo'."; exit 1; }
 	@$(MAKE) -C test check
 
 testimage: all
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -253,27 +253,27 @@ For the testsuite to pass, you will have to install at least the software packag
 mentioned in the `test/container` Dockerfiles.
 
 ```
-$ sudo make clean check
+$ make clean check
 ```
 
 in verbose mode:
 ```
-$ sudo make V=1 clean check
+$ make V=1 clean check
 ```
 
 only specific test:
 ```
-$ sudo make TESTS="01 20 40" clean check
+$ make TESTS="01 20 40" clean check
 ```
 only runs the 01, 20 and 40 tests.
 
 debug a specific test case:
 ```
 $ cd TEST-01-BASIC
-$ sudo make clean setup run
+$ make clean setup run
 ```
 ... change some kernel parameters in `test.sh` ...
 ```
-$ sudo make run
+$ make run
 ```
 to run the test without doing the setup.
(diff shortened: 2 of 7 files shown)


LaszloGombos (approved):
Thanks.

I will close https://github.com/dracutdevs/dracut/pull/2217 in favor of this PR.

aafeijoo-suse (commented):
FWIW, this line would also have to be removed:

https://github.com/dracutdevs/dracut/blob/6af3fcfd7f699ec4bb3f1e9ac71b7a9ebb9804b3/test/Makefile#L4

Henrik66:
> FWIW, this line would also have to be removed:
> 
> https://github.com/dracutdevs/dracut/blob/6af3fcfd7f699ec4bb3f1e9ac71b7a9ebb9804b3/test/Makefile#L4

Thanks @aafeijoo-suse for the review.
Unfortunately, I do not understand what else needs to be removed as the line 4 of test/Makefile is already removed in this PR - see https://github.com/dracutdevs/dracut/pull/2429/files#diff-740cb5a1689091cb894445e46683c255e39689ccba1c6c63d8a8841c8df8817dL4 . Can you help me understand please ! Thanks !

aafeijoo-suse:
> > FWIW, this line would also have to be removed:
> > https://github.com/dracutdevs/dracut/blob/6af3fcfd7f699ec4bb3f1e9ac71b7a9ebb9804b3/test/Makefile#L4
> 
> Thanks @aafeijoo-suse for the review. Unfortunately, I do not understand what else needs to be removed as the line 4 of test/Makefile is already removed in this PR - see https://github.com/dracutdevs/dracut/pull/2429/files#diff-740cb5a1689091cb894445e46683c255e39689ccba1c6c63d8a8841c8df8817dL4 . Can you help me understand please ! Thanks !

Sorry, I missed that change somehow... it's all good

bdrung:
+1 from my side. I tested the patch from this PR plus https://github.com/dracutdevs/dracut/pull/2484 successfully for the tests 01 02 03 04 on Debian and Ubuntu.

One caveat: The kernel on Ubuntu is only readable by root. So either make it readable or run as root. See https://launchpad.net/bugs/759725

bdrung (changes requested):
Please remove `sudo` from the Docker files in `test/container/`.

bdrung (approved):

LaszloGombos (approved):

PR #2436, opened 2023-07-20
fix(dracut.sh): use gawk for strtonum

thesamesam:
This pull request changes dracut.sh and dracut-functions.sh to use `gawk` instead of `awk` because we're using `strtonum`.

strtonum is a gawkism and is not available in all awks, e.g. mawk. Use gawk to avoid failure.

Fixes: f32e95bcadbc5158843530407adc1e7b700561b1

## Checklist
- [N/A] I am providing new code and test(s) for it

Fixes #

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1047,7 +1047,7 @@ pe_file_format() {
     if [[ $# -eq 1 ]]; then
         local magic
         magic=$(objdump -p "$1" \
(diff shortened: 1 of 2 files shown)


thesamesam:
cc @keentux

aafeijoo-suse (approved):
LGTM, thanks.

PR #2442, opened 2023-07-23
fix(rngd): install system service file

pvalena:
as there's no reason to keep a copy; there shouldn't be any modifications.

Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=2181541

## Changes

Install system `rngd.service` file instead of a custom (bundled) one.

## Checklist


LaszloGombos (approved):
lgtm. Thanks

aafeijoo-suse (commented):
The service shipped with `rng-tools` is not ordered `Before=systemd-udevd.service` and has a install section with `WantedBy=multi-user.target`. Are you sure this change works?

CC: @rmetrich

pvalena:
> The service shipped with `rng-tools` is not ordered `Before=systemd-udevd.service` and has a install section with `WantedBy=multi-user.target`. Are you sure this change works?
> 
> CC: @rmetrich

Thanks for checking, let me test various systems.

pvalena:
FTR this does not work atm. as `/etc/sysconfig/rngd` is not installed.

aafeijoo-suse:
> FTR this does not work atm. as `/etc/sysconfig/rngd` is not installed.

Is that RH specific? https://github.com/nhorman/rng-tools/blob/master/rngd.service

pvalena:
> > FTR this does not work atm. as `/etc/sysconfig/rngd` is not installed.
> 
> Is that RH specific? https://github.com/nhorman/rng-tools/blob/master/rngd.service

Unfortunately, yes, it seems we ship it https://src.fedoraproject.org/rpms/rng-tools/blob/rawhide/f/rngd.service | https://gitlab.com/redhat/centos-stream/rpms/rng-tools/-/blob/c9s/rngd.service

pvalena:
_(Sorry about all of those pushes; I somehow thought I'd resolve it without testing first.)_

pvalena:
Tested: https://gist.github.com/pvalena/d790ab99deb06dff8acd271239ce28b3

pvalena:
Added to commit message:
```
    In case there are args stored in a separate file (Fedora and alike),
    it needs to be supplied too, but without the option to change the user.
```

pvalena:
This is still pending merge.

PR #2443, opened 2023-07-23
feat(hwdb): install hwdb on demand when module is needed

pvalena:
## Changes

Adding a module to install hwdb. Further extensions might make only selected part of hwdb installable, to save space. The module is not included by default.

Including the module adds 2MB of compressed data (on Fedora, the file has 12MB).

hwdb is needed in case of custom HW, like a keyboard/mouse or various interfaces.

Original PR: https://github.com/dracutdevs/dracut/pull/1681
Downstream Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1968118

Please note that `hwdb.bin` is similarly installed by `systemd-udevd`, which could instead pull in this module as dependency.

## Checklist

### DIFF
--- a/modules.d/95hwdb/module-setup.sh
+++ b/modules.d/95hwdb/module-setup.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+check() {
+    return 255
+}
+
+# called by dracut
+install() {
+    local hwdb_bin
+
+    # systemd-hwdb ships the file in /etc, with /usr/lib as an alternative.
+    # The alternative location is preferred, as we can consider it being user
+    # configuration.
+    hwdb_bin="${udevdir}"/hwdb.bin
+
+    if [[ ! -r "${hwdb_bin}" ]]; then
+        hwdb_bin="${udevconfdir}"/hwdb.bin
+    fi
+
+    if [[ $hostonly ]]; then
+        inst_multiple -H "${hwdb_bin}"
+    else
+        inst_multiple "${hwdb_bin}"
+    fi
+}


dtardon:
> Note that the downstream issue would not be resolved unless somehow hwdb.d/20-net-ifname.hwdb also gets installed into the initramfs.

No, that's not needed. `.hwdb` files are not used directly; they must first be compiled into `hwdb.bin`, which is a binary format optimized for fast lookup. The compilation (`udevadm hwdb --update`) is out of scope of udevd. E.g., in `rpm` world it is likely done in `%post` (or `%posttrans`) of any package that installs a `.hwdb` file. Hence, dracut only needs to install `hwdb.bin` file(s).

modules.d/95hwdb/module-setup.sh
```diff
@@ -0,0 +1,16 @@
...
+check() {
+    return 255
+}
+
+# called by dracut
+install() {
+    inst_multiple -o "${udevdir}"/hwdb.bin
+
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o "${udevconfdir}"/hwdb.bin
```
dtardon, line 14: Hmm, but this doesn't fix anything, does it? You only install `/etc/udev/hwdb.bin` for `hostonly`, but the downstream issue is for a generic image (installer).

pvalena:
> > Note that the downstream issue would not be resolved unless somehow hwdb.d/20-net-ifname.hwdb also gets installed into the initramfs.
> 
> No, that's not needed. `.hwdb` files are not used directly; they must first be compiled into `hwdb.bin`, which is a binary format optimized for fast lookup. The compilation (`udevadm hwdb --update`) is out of scope of udevd. E.g., in `rpm` world it is likely done in `%post` (or `%posttrans`) of any package that installs a `.hwdb` file. Hence, dracut only needs to install `hwdb.bin` file(s).

Exactly. I've just confirmed: https://gist.github.com/pvalena/873bed9e9eb749f127ca7f60c330dd3d

modules.d/95hwdb/module-setup.sh
pvalena: You're right, I've set this to Draft because I had to test. I've rewritten it to work correctly now :).

modules.d/95hwdb/module-setup.sh
```diff
@@ -0,0 +1,26 @@
...
+    return 255
+}
+
+# called by dracut
+install() {
+    local hwdb_bin
+
+    # systemd-hwdb ships the file in /etc, with /usr/lib as an alternative.
+    # Therefore consider this location as preferred for configuration.
+    hwdb_bin="${udevdir}"/hwdb.bin
```
dtardon, line 15: `${udevdir}/hwdb.bin` is already unconditionally installed by `01systemd-udevd`, so it can be ignored here.

pvalena: Right, I was rather thinging on systemd-udevd depending on this module, which might handle hwdb in more complex way; also not install multiple ones unnecessarily. Please also note that the install is not optional (`-o`) intentionally; and this module is not enabled by default. Does that make sense?

There's also an additional benefit of this module working correctly without systemd-udevd :) in case that's ever needed.

dtardon: > There's also an additional benefit of this module working correctly without systemd-udevd :)

That's a benefit without any value. ATM, HWDB is configuration for `systemd-udevd`, just like `.rules` files. It _could_ be used for other things, but it's not.

dtardon: > > There's also an additional benefit of this module working correctly without systemd-udevd :) in case that's ever needed.
> 
> This makes sense to me. This situation is very similar to udev rules - they are installed both by `systemd-udevd` and `udev-rules` .

Uff. Does `udev-rules` even work on modern systems? The list of `.rules` files seem to be rather short. (OTOH, it contains stuff that makes no sense in initrd, like `70-uaccess.rules` or `71-seat.rules`.)

pvalena: I'd say if it's not breaking anything, let's keep it (and this also verifies `hwdb.bin` is present / installed, as udevd module simply runs opt-install).

pvalena:
```
+    # The alternative location is preferred, as we can consider it being user
+    # configuration.
```
IOW first looks in `/usr` and then in `/etc`. I don't know what `systemd-hwdb` behaviour is, it might be best to mimic it. Either way I see having `hwdb.bin` in both locations (or installing it) undesirable.

PR #2445, opened 2023-07-26
ci: match comment with code

Henrik66:
Properly document the time container generation runs daily.

I think best to fix the comment (or perhaps remove the comment) if it is no longer matching the code.


.github/workflows/container.yml
```diff
@@ -1,7 +1,7 @@
 name: Container
 on:
     schedule:
-        -  cron: '30 11 * * *'   # every day at 4:40
+        -  cron: '30 11 * * *'   # every day at 11:30 GMT
```
aafeijoo-suse, line 4: GMT or UTC?

Henrik66: You are right @aafeijoo-suse . It should be UTC. I changed it.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

aafeijoo-suse (approved):
Thanks!

PR #2447, opened 2023-07-26
ci: change the order of lint jobs as lint_shell is more likely to fail

Henrik66:
ci: change the order of lint jobs as lint_shell is more likely to fail..

Now the lint_shell test result is "above the scroll" on the page.

Screen Shot 2023-07-29 at 2 49 15 PM

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,21 +7,6 @@ on:
     branches: [ master ]
 
 jobs:
-  lint-c:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v3
-
-      - name: install tools
-        run: sudo apt-get install astyle
-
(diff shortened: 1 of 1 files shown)


Henrik66:
After PR CI is failing as expected

Screen Shot 2023-07-26 at 11 28 48 AM

aafeijoo-suse (approved):
Maybe this approach is fine, but AFAIK we were able to identify an entry with `Lint / lint-shell` so far, the problem is that both Lint entries are not ordered in the checklist.

Henrik66:
Screen Shot 2023-07-28 at 11 49 25 PM

I think I was wrong. Both job failures are reported without changing the code. It is just hard to find the failure for lint-shell as it is usually at the bottom.

@LaszloGombos  What should I do now ? Should I close this PR ?

PR #2450, opened 2023-07-26
fix(release): generate dracut-version.sh as part of the release

Henrik66:
Include version information inside the generated release archive.

## Checklist


aafeijoo-suse (commented):
hmm... the thing with the `dracut-version.sh` script is that some distros add [its own bits](https://src.fedoraproject.org/rpms/dracut/blob/rawhide/f/dracut.spec#_212) after the upstream main release number to track custom modifications, but maybe this is useful for other distros that just want the main upstream version.

bdrung:
With my Debian/Ubuntu packaging hat on, this pull request is fine for me. The alternative is to make setting the version easier (see https://github.com/dracutdevs/dracut/pull/2488). The third alternative would be to allow setting the full version via the configure script (that way packagers could include the minor packaging version).

dracut-version.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+# shellcheck disable=SC2034
+DRACUT_VERSION=059
```
bdrung, line 3: How about hard-code `DRACUT_VERSION` in `dracut.sh`? This would save more lines of code and prevents mismatching `dracut.sh` and `dracut-version.sh`.

aafeijoo-suse: If it can be overridden if `dracut-version.sh` exists, that looks good by me. But I wouldn't change how `dracut-version.sh` can be built, it's useful for some distros out there.

bdrung: Let's talk code. Here is the proposal: https://github.com/dracutdevs/dracut/pull/2500

aafeijoo-suse: See https://github.com/dracutdevs/dracut/pull/2500#pullrequestreview-1594996990

aafeijoo-suse (changes requested):
I'll test it properly when I have some time, but regardless you have to `chmod 755 dracut-version.sh`

```
pr-2450> ls -l dracut*.sh
-rwxr-xr-x 1 afeijoo users  3693 Aug 25 08:40 dracut-catimages.sh
-rwxr-xr-x 1 afeijoo users 31793 Aug 25 08:40 dracut-functions.sh
-rwxr-xr-x 1 afeijoo users  3599 Aug 25 08:40 dracut-initramfs-restore.sh
-rwxr-xr-x 1 afeijoo users 39298 Aug 25 08:40 dracut-init.sh
-rwxr-xr-x 1 afeijoo users 13457 Aug 25 08:40 dracut-logger.sh
-rwxr-xr-x 1 afeijoo users 94001 Aug 25 08:40 dracut.sh
-rw-r--r-- 1 afeijoo users    57 Aug 25 08:40 dracut-version.sh
```

aafeijoo-suse (approved):
> I'll test it properly when I have some time

All good, this does not interfere with distros that update `dracut-version.sh` on their .spec file. Thanks!

PR #2451, opened 2023-07-27
fix(base): correct handling of quiet in loginit

aafeijoo-suse:
`dracut-lib.sh` [sets `DRACUT_QUIET` to `yes` or `no`](https://github.com/dracutdevs/dracut/blob/a804945f27a0ccc2f69ae694599b1afec2afe8b1/modules.d/99base/dracut-lib.sh#L460-L470) and [passes this to `loginit`](https://github.com/dracutdevs/dracut/blob/a804945f27a0ccc2f69ae694599b1afec2afe8b1/modules.d/99base/init.sh#L99).

Follow up to #2281 with the commit message fixed.

## Checklist

### DIFF
--- a/modules.d/99base/loginit.sh
+++ b/modules.d/99base/loginit.sh
@@ -18,6 +18,6 @@ while read -r line || [ -n "$line" ]; do
     fi
     echo "<31>dracut: $line" >&5
     # if "quiet" is specified we output to /dev/console
-    [ -n "$QUIET" ] || echo "dracut: $line"
+    [ "$QUIET" = "yes" ] || echo "dracut: $line"
     echo "$line" >&6
 done


LaszloGombos (approved):
Alternatively, one can force push to https://github.com/bmwiedemann/dracut/tree/loginit and fix the commit message.

This would allow keeping the contribution credit to the original author and would help with the reviews also.

See also https://github.com/dracutdevs/dracut/pull/1759#issuecomment-1291320074

aafeijoo-suse:
> This would allow keeping the contribution credit to the original author and would help with the reviews also.

The commit is still authored by Bernhard (see the git log), I only created a new PR importing his patch, which is the same as force pushing his branch (AFAIK).

aafeijoo-suse:
> > which is the same as force pushing his branch (AFAIK).
> 
> It is the same in terms of commit history (if and when it lands), but it is not the same in terms of our processes.

For me this is the only thing that really matters.

PR #2454, opened 2023-07-29
test: increase test VM memory from 512M to 1024M to avoid OOM killer

Henrik66:
512 MB might not be enough for some of the more complex tests with systemd. Increase it to 1GB.


aafeijoo-suse (approved):
Ouch! [the oom killer in fedora](https://pagure.io/fedora-workstation/issue/358)... let's see if this helps

PR #2456, opened 2023-07-29
test(NFS): run either nfsv4 or nfsv3 tests

### LINKED ISSUE #2455: modernize nfs testing
Currently 45 min is not enough to finish nfs tests. This is mainly because the test is trying to test both nfsv3 and nfsv4.

Instead of testing both nfsv4 and nfsv3 all the time, I suggest that the test checks which nfs version is available and it only tests one version. This would still allow testing both version if somebody really want to but on modern containers it would only test nfsv4, which would allow to cut testing time significantly.

Henrik66:
Run only nfsv4 tests unless nfs4 is not available 

## Checklist

Fixes #2455

### DIFF
--- a/test/TEST-20-NFS/test.sh
+++ b/test/TEST-20-NFS/test.sh
@@ -213,8 +213,12 @@ test_run() {
         return 1
     fi
 
-    test_nfsv3 \
-        && test_nfsv4
+    # run either nfsv4 or nfsv3 tests
+    if grep -q '\-nfs4' /etc/portage/package.use/nfs-utils 2> /dev/null; then
+        test_nfsv4
+    else
+        test_nfsv3
+    fi
 
     ret=$?
 


LaszloGombos (approved):
LGTM. Thanks. This helps having passing CI without compromising test coverage significantly.

test/TEST-20-NFS/test.sh
```diff
@@ -213,8 +213,12 @@ test_run() {
         return 1
     fi
 
-    test_nfsv3 \
-        && test_nfsv4
+    # run either nfsv4 or nfsv3 tests
+    if grep -q '\-nfs4' /etc/portage/package.use/nfs-utils 2>/dev/null; then
```
aafeijoo-suse, line 217: isn't this check only for gentoo?

PR #2459, opened 2023-07-30
feat(sh): shell meta module

Henrik66:
Improve shell selection

## Checklist

### DIFF
--- a/modules.d/00sh/module-setup.sh
+++ b/modules.d/00sh/module-setup.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+depends() {
+    shells='dash bash mksh busybox'
+    for shell in $shells; do
+        if dracut_module_included "$shell"; then
+            echo "$shell"
+            return 0
+        fi
+    done
+
+    shell=$(realpath -e /bin/sh)
+    shell=${shell##*/}
+
+    echo "$shell"
+    return 0
+}
--- a/modules.d/99base/module-setup.sh
+++ b/modules.d/99base/module-setup.sh
@@ -7,7 +7,7 @@ check() {
 
 # called by dracut
 depends() {
-    echo udev-rules
+    echo udev-rules sh
     return 0
 }
 


Henrik66:
Current version of this PR is based on this comment on another PR https://github.com/dracutdevs/dracut/pull/2368#issuecomment-1588285469

FGrose:
Complexities arise when we seek to properly support the 4 approved command shells (_dash, bash, mksh, & busybox_) in `rd.debug` & `rd.live.debug` modes.  PR #2287 addresses a couple of bugs resulting from _mksh_ having only local xtrace.  The solution uses a feature introduced in _bash-4.4_ from 2016-09-15. This leads to the need to check the installed bash version ***after any*** native or 3rd party module might install it or link it to `/bin/sh` (as the _squash_ module does in an installpost operation). Commit 7e265ec37fce1038ff34c6bab37ccf00d1dc123f deals with the bash version check.

A supported, but less common feature of _dracut_ is the ability to build divided initramfs images where, for example, 2 or more images are entered on the GRUB initrd/initrdefi command line. In this case, or in the case of building an incremental initramfs image with `--rebuild`, one might desire to have no command shell installed.  For this, some mechanism to exclude the installation of a shell is needed.  Commit 48c7205fe073c86bb580ae8299ad5c778f73b4a9 addresses this.

The implicit installation of a command shell by dracut-install.c during the *** Resolving executable dependencies *** phase of dracut.sh means that an explicit call for a shell must occur before that phase of processing.  Here is a simple test case:

1. `ln -sf mksh /bin/sh`
2. `dracut /tmp/initrd-test.img -m selinux`
with the test PR #2459, `/usr/bin/mksh` will be installed without the dependency `/usr/bin/printf` and without reporting _mksh_ in `/usr/lib/dracut/modules.txt`.

Commit ef9faa253455bb40b2a8249737cb2834ef710b1e is designed to run after all module checks have occurred and assures that a command shell is explicitly loaded before the executable dependencies are checked.  It is also coded so that it does its work automatically without being included in `/usr/lib/dracut/modules.txt`.

PR #2462, opened 2023-07-31
docs: Make the purpose of fw_dir explicit

Zer0-One:
Updated documentation and usage text to make it explicit that `fw_dir` is only for CPU firmware.

I'm new to dracut, played around with it for a while today and got stuck trying to figure out why this configuration option wasn't working, until I read the script and realized it's working fine, it's just not named well.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -113,7 +113,7 @@ Creates initial ramdisk images for preloading modules
   -k, --kmoddir [DIR]   Specify the directory where to look for kernel
                          modules.
   --fwdir [DIR]         Specify additional colon-separated list of directories
-                         where to look for firmware files.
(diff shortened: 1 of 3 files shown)


aafeijoo-suse (commented):
No, this is also the firmware path for kernel modules.

ddiss (changes requested):
As mentioned by Antonio, `fwdir` is passed through to dracut-install via `DRACUT_FIRMWARE_PATH` and used to lookup kernel module firmware paths (e.g. `modinfo i915 | grep "^firmware:"`). The current usage looks fine.

Zer0-One:
I see. I was under the impression that it was for including firmware in the initramfs image.

PR #2476, opened 2023-08-07
fix(systemd-networkd): remove dbus hard dependency

Henrik66:
According to https://github.com/dracutdevs/dracut/issues/2378#issuecomment-1668136681 dbus is not a requirement for systemd-networkd.


aafeijoo-suse (commented):
> According to https://github.com/dracutdevs/dracut/issues/2378#issuecomment-1668136681 dbus is not a requirement for systemd-networkd.

And for `networkctl`?

PR #2479, opened 2023-08-12
feat(install.d) Support ukify for kernel-install

rahilarious:
This patch allows users to generate UKI with systemd & use dracut for initramfs. 

### How it works
Recently ([commit](https://github.com/systemd/systemd/commit/f3d1d7609d2008086f0f8369d9f842a5c09024b2)) ukify gained ability to add all initrd's placed in $KERNEL_INSTALL_STAGING_AREA to UKI file.
With this patch initrd will be placed in $KERNEL_INSTALL_STAGING_AREA for ukify users, allowing seamless automation with kernel-install.

### Checklist
- [x] I have tested it locally with latest git versions of kernel-install (systemd) & dracut

### Fixes
Need for hacks & hooks to let dracut & ukify get along well

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -3,6 +3,11 @@
 COMMAND="$1"
 KERNEL_VERSION="$2"
 BOOT_DIR_ABS="$3"
+
+if [[ ${KERNEL_INSTALL_LAYOUT} == "uki" && ${KERNEL_INSTALL_UKI_GENERATOR} == "ukify" ]]; then
+    BOOT_DIR_ABS="${KERNEL_INSTALL_STAGING_AREA}"
+fi
+
 KERNEL_IMAGE="$4"
 
 # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.


rahilarious:
> dracut-ukify seem to exists only as an AUR package for arch only. 

Well, I just used "dracut-ukify" as to indicate some kind of bridge/bond between two programs. Being a Gentoo user, I wasn't even aware that such thing exists in AUR.

> Generally distro specific changes will expect a bit more scrutiny

Indeed, this is true, if this was about AUR package.(but it's not)

This will work on any distro with systemd-boot (yes even on gentoo with openrc) with latest git version (or >=systemd 254.2)

aafeijoo-suse (commented):
Sorry, I don't agree with what this PR wants to add, so it's a no from me.

rahilarious:
> Sorry, I don't agree with what this PR wants to add, so it's a no from me.

Seems like anti-ukifyism going on here. Some grudge against ukify/systemd? Anything related to it gets denied.

aafeijoo-suse:
> > Sorry, I don't agree with what this PR wants to add, so it's a no from me.
> 
> Seems like anti-ukifyism going on here. Some grudge against ukify/systemd? Anything related to it gets denied.

It's one thing to support `kernel-install` parameters within the `kernel-install` scripts shipped with dracut (#2495), and quite another to support other random uki generators and what they want to do.

You can interpret it as you wish.

rahilarious:
> quite another to support other random uki generators and what they want to do.

Sir, this is no random generator. [Ukify](https://github.com/systemd/systemd/blob/main/man/ukify.xml) is built into systemd. 

> It's one thing to support `kernel-install`

This patch is about `kernel-install` script. So **dracut**  will generate & copy initramfs  to proper directory where it can be picked up by other plugins (install.d scripts)

PR #2481, opened 2023-08-14
fix(dracut.sh): remove microcode check based on CONFIG_MICROCODE_[AMD|INTEL]

aafeijoo-suse:
`CONFIG_MICROCODE_AMD` and `CONFIG_MICROCODE_INTEL` are hidden since https://lore.kernel.org/all/20230810160805.081212701@linutronix.de/, therefore this check is wrong and early microcode is always disabled.

[Already merged into the tip tree](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?id=029ae99c8507db15adfae24a15dc44fd568b5f93), PR created as draft meanwhile it doesn't make it to the master branch.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1564,23 +1564,20 @@ fi
 
 if [[ $early_microcode == yes ]]; then
     if [[ $hostonly ]]; then
-        if [[ $(get_cpu_vendor) == "AMD" ]]; then
-            check_kernel_config CONFIG_MICROCODE_AMD || unset early_microcode
-        elif [[ $(get_cpu_vendor) == "Intel" ]]; then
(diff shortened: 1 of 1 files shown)


lnykryn:
Do we need to wait to merge this? CONFIG_MICROCODE is set in the older kernels, so there should be no issue.
Well, there is a scenario where someone on intel compiles the kernel with CONFIG_MICROCODE_AMD=y and CONFIG_MICROCODE_INTEL=n, but ... meh.

lnykryn (approved):

aafeijoo-suse:
FTR, already merged into master (https://github.com/torvalds/linux/commit/42a7f6e3ffe06308c1ec43a7dac39a27de101574), this patch will be required for 6.6-rc1.

PR #2487, opened 2023-08-16
test: support running qemu on other architectures

bdrung:
## Changes

This merge request contains the first set of patches to support running qemu on other architectures.

## Checklist


test/run-qemu
```diff
@@ -46,4 +46,5 @@ if [[ $* == *-initrd* ]]; then
     ARGS+=(-kernel "$VMLINUZ")
 fi
 
+echo "${0##*/}: $BIN ${ARGS[*]@Q} ${*@Q}"
```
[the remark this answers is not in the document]
bdrung: This line is ugly to read, but since there are parameters that contain spaces (e.g. `-append`), this is the only way I found to do it in one line. The alternative would be a five line long for loop.

test/run-qemu
```diff
@@ -3,24 +3,31 @@
...
+[[ -c /dev/kvm && -x "/usr/bin/qemu-system-${ARCH}" ]] && BIN="/usr/bin/qemu-system-${ARCH}" && ARGS=(-enable-kvm -cpu host)
 
 [[ $BIN ]] || {
     echo "Could not find a working KVM or QEMU to test with!" >&2
     echo "Please install kvm or qemu." >&2
     exit 1
 }
 
+case "$ARCH" in
+    amd64 | i386 | x86_64)
```
aafeijoo-suse, line 24: Please replace `i386` with `i?86` for completeness.

test/run-qemu
```diff
@@ -29,14 +36,23 @@ fi
...
 if ! [ -f "$VMLINUZ" ]; then
     [[ -f /etc/machine-id ]] && read -r MACHINE_ID < /etc/machine-id
 
     if [[ $MACHINE_ID ]] && { [[ -d /boot/${MACHINE_ID} ]] || [[ -L /boot/${MACHINE_ID} ]]; }; then
         VMLINUZ="/boot/${MACHINE_ID}/$KVERSION/linux"
-    else
+    elif [ -f "/boot/vmlinuz-${KVERSION}" ]; then
         VMLINUZ="/boot/vmlinuz-${KVERSION}"
+    elif [ -f "/boot/vmlinux-${KVERSION}" ]; then
+        VMLINUZ="/boot/vmlinux-${KVERSION}"
```
aafeijoo-suse, line 51: Yes, on SUSE `ppc*` arch kernel is named `vmlinux` as well. And, also on SUSE distros, we have the following kernel names for the other architectures:
- `s390*` => `image`
- `aarch64` => `Image`
- `armv*` => `zImage`

But, I don't know if all this specific naming follows a standard convention, so it can be upstreamed.

test/run-qemu
bdrung: Thanks. Changed.

test/run-qemu
bdrung: @aafeijoo-suse Should I extend the patch to also search for those files or will you provide a follow-up merge request?

aafeijoo-suse: I'd like to know the kernel names used in Ubuntu/Debian and RH/Fedora (@lnykryn could you shed some light here?) for the different architectures, so we could decide to extend this patch based on the response.

lnykryn: We name everything vmlinuz. /boot/vmlinuz-5.14.0-358.el9.s390x, /boot/vmlinuz-5.14.0-358.el9.aarch64,....

bdrung: On Debian/Ubuntu the kernel is named `vmlinuz` (in `/etc` and has a version suffix) on most architectures, but sometimes `vmlinux` (e.g. ppc64el). Example:
```
ls /boot/vmlinuz* -l
lrwxrwxrwx 1 root root       24 Aug 12 20:16 /boot/vmlinuz -> vmlinuz-6.2.0-27-generic
-rw------- 1 root root 14482632 Jul 10 22:31 /boot/vmlinuz-6.2.0-26-generic
-rw------- 1 root root 13872968 Jul 12 23:23 /boot/vmlinuz-6.2.0-27-generic
lrwxrwxrwx 1 root root       24 Aug 12 20:16 /boot/vmlinuz.old -> vmlinuz-6.2.0-26-generic
```

aafeijoo-suse: hmm.. we cannot handle all these downstream naming differences... maybe the right approach here is to add the ability to set `VMLINUZ` using an external variable?

lnykryn: AFAIK vmlinux vs vmlinuz is an accepted standard. The first should be uncompressed second one is compressed. So at least adding vmlinux makes sense.

aafeijoo-suse: Ok, good enough for me. Thanks!

aafeijoo-suse (approved):

PR #2489, opened 2023-08-19
fix: codespell

Henrik66:
Run codespell (v2.2.5).


src/dracut-cpio/src/main.rs
```diff
@@ -112,7 +112,7 @@ struct ArchiveState {
     // offset from the start of this archive
     off: u64,
     // next mapped inode number, used instead of source file inode numbers to
-    // ensure reproducability. XXX: should track inode per mapped dev?
+    // ensure delimiters. XXX: should track inode per mapped dev?
```
aafeijoo-suse, line 115: is this correction ok? shouldn't this be `reproducibility`?

PR #2490, opened 2023-08-19
docs: set KVERSION for running test suite

bdrung:
## Changes

Running the test suite in podman will most likely require setting `KVERSION` because the running Linux kernel version will most likely be different to the kernel version in the container.

## Checklist

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -240,7 +240,7 @@ $ podman run --rm -it \
 # ./configure
 # make -j $(getconf _NPROCESSORS_ONLN)
 # cd test
-# make V=1 SKIP="16 60 61" clean check
+# make KVERSION="$(cd /lib/modules && ls -1 | tail -1)" V=1 SKIP="16 60 61" clean check
 ```
 
 with `[CONTAINER]` being one of the


bdrung:
The test-github.sh is just another layer of indirection: test-github.sh -> test/Makefile -> test/TEST-$x/Makefile -> test/Makfile.testdir -> ./test.sh. Maybe test/Makefile could set KVERSION depending on what is available.

A quick win for the documentation is useful even when you rework test-github.sh tomorrow.

docs/HACKING.md
```diff
@@ -240,7 +240,7 @@ $ podman run --rm -it \
 # ./configure
 # make -j $(getconf _NPROCESSORS_ONLN)
 # cd test
-# make V=1 SKIP="16 60 61" clean check
+# make KVERSION="$(cd /lib/modules && ls -1 | tail -1)" V=1 SKIP="16 60 61" clean check
```
aafeijoo-suse, line 243: Why someone running the testsuite on a local system should write this instead of nothing or `uname -r`? Is the kernel version used in the container always the one with the higher version? Maybe it's better to add a few words about _when_ `KVERSION` should be set and _what_ value it should contain.

bdrung: The podman container has one kernel version installed and the host is running one kernel version. The likelihood that both are identical is marginal.

aafeijoo-suse: My question is, why should I run this `cd /lib/modules..` instead of doing nothing or just `uname -r` on my local system to run the testsuite?

```
> echo $(cd /lib/modules && ls -1 | tail -1)
6.5.0-rc6-1-default
> uname -r
6.4.6-1-default
```

That's why I asked for some explanation about the container use case.

bdrung: You do not need to set it on bare metal, but this proposed change is in the "Rootless in a container with podman" section.

aafeijoo-suse: Oops... my fault, I was missing the section all the time... sorry.

aafeijoo-suse (approved):

PR #2492, opened 2023-08-19
ci: also run integration test on Ubuntu

bdrung:
## Changes

Add a Dockerfile for Ubuntu (very similar to Debian) and run the integration tests also on Ubuntu.

I built the container locally and tested that the test cases from the integration test succeed there.

## Checklist

Fixes #

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -34,6 +34,7 @@ jobs:
                     - { dockerfile: 'Dockerfile-Arch',              tag: 'arch:latest' }
                     - { dockerfile: 'Dockerfile-Debian',            tag: 'debian:latest' }
                     - { dockerfile: 'Dockerfile-Gentoo',            tag: 'gentoo:latest' }
+                    - { dockerfile: 'Dockerfile-Ubuntu',            tag: 'ubuntu:latest' }
         steps:
             -   name: Check out the repo
                 uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


aafeijoo-suse (commented):
> The Ubuntu container is almost an exact copy of the Debian container.

+1

> Adding a new test container is a good time to discuss how we can we better maintain all of these test containers.

I'd love to see Ubuntu running on the CI, but adding a new container implies having to support it, we already have 5, and upstream resources (talking about people) are veeery limited...

bdrung:
Added ubuntu to manialtest.

> I'd love to see Ubuntu running on the CI, but adding a new container implies having to support it, we already have 5, and upstream resources (talking about people) are veeery limited...

Ubuntu will use dracut-install in initramfs-tools (see [manual_add_modules is slow / consumes the most time](https://bugs.launchpad.net/ubuntu/+source/initramfs-tools/+bug/2031185)). So at least dracut-install will become part of Ubuntu main (see [[MIR] dracut](https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2031304)) and part of the 1000 packages our Ubuntu Foundations team will maintain.

aafeijoo-suse (approved):

aafeijoo-suse:
Does this PR need to be merged before any of the Ubuntu tests go green?

bdrung:
Yes, the Ubuntu tests need the Ubuntu container, which is only built from the master branch.

PR #2496, opened 2023-08-21
fix(test): use bash for jobs -r parameter

### LINKED ISSUE #1901: Test 20 - "NFSv3 root=dhcp DHCP IP:path" fails on debian
https://github.com/dracutdevs/dracut/blob/master/test/TEST-20-NFS/test.sh#L152 times out in when running on a Debian container. 4 NFS subtests pass, but the 5th times out.

The same test passes on the Fedora-latest and Arch containers, so the concern seems to be distro dependent.

bdrung:
## Changes

Test 20 - "NFSv3 root=dhcp DHCP IP:path" fails on Debian with following error message in `server.log`:

```
jobs -rp
/sbin/init: 105: jobs: Illegal option -r
```

The jobs parameter `-r` requires bash.

## Checklist

Partially fixes #1901

### DIFF
--- a/test/TEST-20-NFS/server-init.sh
+++ b/test/TEST-20-NFS/server-init.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 exec < /dev/console > /dev/console 2>&1
 set -x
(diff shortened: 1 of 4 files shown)


aafeijoo-suse (approved):
Thanks

PR #2500, opened 2023-08-24
refactor: maintain dracut version in dracut.sh

### LINKED ISSUE #2169: Release tarballs do not include dracut-version.sh
Tarballs from the GH release page lack the file `dracut-version.sh` that encodes version information. The [Makefile](https://github.com/dracutdevs/dracut/blob/master/Makefile#L3-L10) attempts to populate version information from git tags (which is impossible in an unpacked release tarball) and, failing that, reverts to whatever would be encoded in that file if it were sourced. Because the file doesn't exist, `make dracut-version.sh` just writes an empty version.

It seems the project is migrating from kernel.org to GitHub as the official site for releases. The repo lists the GH wiki as the official project link, and the ["Downloads" link on the wiki](https://github.com/dracutdevs/dracut/wiki#development) points back to the GH release page. Therefore, tagged GitHub releases ought to include a value `dracut-version.sh`.

bdrung:
## Changes

The git release tag does not contain the `dracut-version.sh` file and therefore `git archive` will lack this file as well. The Makefile attempts to populate version information from git tags (which is impossible in an unpacked release tarball) and, failing that, reverts to whatever would be encoded in that file if it were sourced.

Maintain the dracut version directly in `dracut.sh` and call `dracut.sh --version` to get the version number. The version can be changed with a simple sed command to a new version (for releases or for distributors that want to include the packaging version):

```
sed -i "s/^DRACUT_VERSION=.*$/DRACUT_VERSION=\"${NEW_VERSION}\"/" dracut.sh
```

## Checklist

Fixes #2169

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,10 @@
--include dracut-version.sh
-
-DRACUT_MAIN_VERSION ?= $(shell env GIT_CEILING_DIRECTORIES=$(CURDIR)/.. git describe --abbrev=0 --tags --always 2>/dev/null || :)
-ifeq ($(DRACUT_MAIN_VERSION),)
-DRACUT_MAIN_VERSION = $(DRACUT_VERSION)
-endif
-DRACUT_FULL_VERSION ?= $(shell env GIT_CEILING_DIRECTORIES=$(CURDIR)/.. git describe --tags --always 2>/dev/null || :)
+DRACUT_FULL_VERSION ?= $(shell ./dracut.sh --version | cut -d' ' -f 2-)
 ifeq ($(DRACUT_FULL_VERSION),)
 DRACUT_FULL_VERSION = $(DRACUT_VERSION)
+endif
+DRACUT_MAIN_VERSION ?= $(shell echo "$(DRACUT_FULL_VERSION)" | sed 's/\([0-9]\+\).*/\1/')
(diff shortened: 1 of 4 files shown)


aafeijoo-suse (commented):
I didn't want to mean this in https://github.com/dracutdevs/dracut/pull/2450#discussion_r1303857075, I was thinking of just setting `DRACUT_VERSION` at the beginning of `dracut.sh` and leave the rest of the code untouched.

If we have to `sed` any `.sh` script, I prefer #2450  and do it on the dedicated `dracut-version.sh`.

ahesford:
I don't have strong opinions on how to handle versioning in project repositories. It's a hard enough problem that, in Python, a million different packages exist to try to divine the version from various markers.

That said, the approach in this PR is pretty much how we handle versioning for ZFSBootMenu. A [releng/tag-release.sh](https://github.com/zbm-dev/zfsbootmenu/blob/master/releng/tag-release.sh) script does some git magic to extract release notes, prepare the branch for tagging, and [overwrites the version](https://github.com/zbm-dev/zfsbootmenu/blob/master/releng/tag-release.sh#L115) of the one script that requires it. It works well enough for us!

bdrung:
Changed `fix` to `refactor`.

PR #2501, opened 2023-08-24
ci: fix shellcheck warning SC2319

Henrik66:
fix shellcheck warning SC2319

### DIFF
--- a/modules.d/95iscsi/iscsiroot.sh
+++ b/modules.d/95iscsi/iscsiroot.sh
@@ -229,7 +229,8 @@ handle_netroot() {
             echo "$target"
         done
     })
-    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed with status $?" && return 1
+    ret=$?
+    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed with status $ret" && return 1
 
     found=
     for target in $targets; do
--- a/modules.d/80lvmthinpool-monitor/start-thinpool-monitor.sh
+++ b/modules.d/80lvmthinpool-monitor/start-thinpool-monitor.sh
@@ -9,7 +9,8 @@ is_lvm2_thinp_device() {
     _lvm2_thin_device=$(lvm lvs -S 'lv_layout=sparse && lv_layout=thin' \
         --nosuffix --noheadings -o vg_name,lv_name "$_device_path" 2> /dev/null)
 
-    [ -n "$_lvm2_thin_device" ] && return $?
+    ret=$?
+    [ -n "$_lvm2_thin_device" ] && return $ret
 }
 
 for LV in $LVS; do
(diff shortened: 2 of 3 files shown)


LaszloGombos (approved):
lgtm

.shellcheckrc
```diff
@@ -32,3 +32,15 @@ disable=SC3013
...
 # ... but dash supports it
 disable=SC3045
+
+# SC2086: Double quote to prevent globbing and word splitting.
+# https://github.com/koalaman/shellcheck/wiki/SC2086
+disable=SC2086
+
+# SC2317: Command appears to be unreachable. Check usage (or ignore if invoked indirectly).
+# https://github.com/koalaman/shellcheck/wiki/SC2317
+disable=SC2317
```
pvalena, line 42: I think this might be helpful... what is the reason to disable this?

.shellcheckrc
```diff
@@ -32,3 +32,15 @@ disable=SC3013
 # https://github.com/koalaman/shellcheck/wiki/SC3045
 # ... but dash supports it
 disable=SC3045
+
+# SC2086: Double quote to prevent globbing and word splitting.
+# https://github.com/koalaman/shellcheck/wiki/SC2086
+disable=SC2086
```
pvalena, line 38: I do not see why would you want not to quote `"$@"`. I'd rather solve this on case-by-case basis, as globally disabling this might lead to possibly faulty code.

modules.d/95iscsi/iscsiroot.sh
```diff
@@ -229,7 +229,8 @@ handle_netroot() {
             echo "$target"
         done
     })
-    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed with status $?" && return 1
+    ret=$?
```
pvalena, line 232: It should be also declared as `local`.

.shellcheckrc
```diff
@@ -32,3 +32,15 @@ disable=SC3013
...
+
+# SC2317: Command appears to be unreachable. Check usage (or ignore if invoked indirectly).
+# https://github.com/koalaman/shellcheck/wiki/SC2317
+disable=SC2317
+
+# SC2004: $/${} is unnecessary on arithmetic variables.
+# https://github.com/koalaman/shellcheck/wiki/SC2004
+disable=SC2004
```
pvalena, line 46: This one (`SC2004 `) seems completely reasonable to me; why would you disable this?

Henrik66:
Given the help from https://github.com/dracutdevs/dracut/pull/2530 I removed the changes related to shellcheck version changes and just kept the code changes regarding SC2319 . Thanks @pvalena for your help with the reviews.

modules.d/95iscsi/iscsiroot.sh
pvalena: Sure, you're right. I didn't know this file is supposed to be POSIX (and that posix doesn't know local). Prefixing it makes sense as well.

LaszloGombos (approved):

PR #2504, opened 2023-08-25
fix(Makefile): release is now just made out of a git commit

Henrik66:
Remove dist make target as it is no longer used or maintained.


LaszloGombos (approved):
As far as I can tell, this code is no longer used neither downstream nor here upstream.

pvalena:
LGTM.

PR #2515, opened 2023-09-12
fix(dracut-initramfs-restore.sh): do not set selinux labels if disabled

aafeijoo-suse:
Also, it's not enough to check if `SELINUX=disabled` in /etc/selinux/config, because it can be disabled via kernel command line options.

## Checklist


lnykryn (approved):
Yup, makes sense.

PR #2520, opened 2023-09-16
fix(crypt): Encrypted root FS handling with generic initrd

### LINKED ISSUE #2437: Generic initrd does not work with encrypted root FS without further configuration
Moving downstream discussion here upstream from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1029324

with an encrypted root fs, the system will end up with a configuration which
looks something like this:

```
/etc/crypttab:
	<dev>_crypt UUID=<some_uuid> none luks,discard

/etc/fstab:
	/dev/mapper/<some_LVM_LV> / ext4 <options>

/boot/grub/grub.cfg:
	linux /vmlinuz-6.0.0-6-amd64 root=/dev/mapper/<some_LVM_LV> ro quiet
```

Now, after switching to dracut, the new initrd which is built will not
contain the necessary configuration to actually setup the "<dev>_crypt"
device, since autoconfiguration (rd.auto) is disabled and a hostonly
initrd (which would include a custom /etc/crypttab in the initrd) is not
built.

The result is an initrd which will hang waiting for
/dev/mapper/<some_LVM_LV> to show up, and which will eventually time
out.

A workaround for now is to change /etc/default/grub:
	GRUB_CMDLINE_LINUX_DEFAULT="quiet rd.auto=1"

This, however, means that the name of the crypt device will change from
"<dev>_crypt" (as specified in /etc/crypttab) to "luks-<some_uuid>",
which will result in ugly error messages during boot (that's mostly
cosmetic though) since systemd will fail to setup the device
"<dev>_crypt".

That, in turn, can be fixed by changing /etc/crypttab from:
	<dev>_crypt UUID=<some_uuid> none luks,discard
so that it reads:
	luks-<some_uuid> UUID=<some_uuid> none luks,discard

A further problem with building the generic initrd is that custom crypt
addons (like the fido2 support) won't be included in the initrd even if
fido2 is defined in /usr/lib/dracut/modules.d/90crypt/module-setup.sh
("fido2-device=auto").

CC @Alphix

DanWin:
This pull request adds missing modules potentially required for disk decryption to a generic initrd. Additionally it changes the default for unlocking LUKS encrypted devices at boot from previously only doing so when rd.auto=1 was specified.

## Checklist

Fixes #2437

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -30,6 +30,12 @@ depends() {
         if grep -q "tpm2-device=" "$dracutsysrootdir"/etc/crypttab; then
             deps+=" tpm2-tss"
         fi
+    elif [[ ! $hostonly ]]; then
+        deps+=" fido2 pkcs11"
+        module_check "tpm2-tss" > /dev/null 2>&1
+        if [[ $? == 255 ]]; then
+            deps+=" tpm2-tss"
+        fi
     fi
     echo "$deps"
     return 0
--- a/modules.d/90crypt/parse-crypt.sh
+++ b/modules.d/90crypt/parse-crypt.sh
@@ -174,7 +174,7 @@ else
                 } >> "$hookdir/emergency/90-crypt.sh"
             fi
         done
-    elif getargbool 0 rd.auto; then
+    elif getargbool 1 rd.auto; then
         if [ -z "$DRACUT_SYSTEMD" ]; then
             {
                 printf -- 'ENV{ID_FS_TYPE}=="crypto_LUKS", RUN+="%s ' "$(command -v initqueue)"


paulmenzel (approved):
Thank you for these patches. No idea, how dracut development process is.

adrelanos:
Tested. Works for me. Bug fixed. Please review.

LaszloGombos:
This PR would make
`dracut --no-hostonly -a crypt` fail if the host does not have e.g. `libfido2.so.*` installed (because now fido2 is a mandatory dependency for crypt in no-hostonly mode).

I think this would be undesirable, especially where for this issue there is an obvious workaround just to manually add the missing modules.  

Also
https://bugzilla.redhat.com/show_bug.cgi?id=1018930

DanWin:
Hi @LaszloGombos , actually there are no checks for these modules that require any files to be present. The modules can currently be included as dependency, without copying any files, because their `module-setup.sh` only copies the library files with `inst_libdir_file` which in turn has a check on whether the files exist, and otherwise silently ignores them.
So the way I see it, there should be nothing against adding them as dependency, unless there are plans on adding checks whether all libraries exist? If so, we might make them optional dependencies like with the tpm2-tss module.

LaszloGombos:
> Hi @LaszloGombos , actually there are no checks for these modules that require any files to be present. 

I have not had a chance to check, just saw that in some other places (e.g. https://github.com/dracutdevs/dracut/blob/master/modules.d/99squash/module-setup.sh#L46), there is `-o`, so i just assumed that without `-o` it would fail.

> the library files with `inst_libdir_file` which in turn has a check on whether the files exist, and otherwise silently ignores them. 

Ok, I see. Thanks !

Perhaps you can upload and improved version of https://github.com/dracut-ng/dracut-ng/pull/280 to https://github.com/dracut-ng/dracut-ng/ (without the rd.auto=1 change).

Issue #2506, opened 2023-08-30
Tests failing in Arch due to segmentation fault in `run-qemu`


aafeijoo-suse:
E.g.: https://github.com/dracutdevs/dracut/actions/runs/6022358884/job/16336946730?pr=2317#step:4:87

```
2023-08-30T08:20:11.1639736Z run-qemu: /usr/bin/qemu-system-x86_64 '-cpu' 'IvyBridge-v2' '-M' 'q35' '-smp' '2' '-m' '1024' '-nodefaults' '-vga' 'none' '-display' 'none' '-no-reboot' '-device' 'virtio-rng-pci' '-serial' 'stdio' '-kernel' '/lib/modules/6.4.12-arch1-1/vmlinuz' '-drive' 'if=none,format=raw,file=/var/tmp/dracut-test.AYlYTw/marker.img,id=drive-sata0' '-device' 'ide-hd,bus=ide.0,drive=drive-sata0,id=sata0,model=disk,serial=marker' '-drive' 'if=none,format=raw,file=/var/tmp/dracut-test.AYlYTw/root.img,id=drive-sata1' '-device' 'ide-hd,bus=ide.1,drive=drive-sata1,id=sata1,model=disk,serial=root' '-append' 'root=/dev/dracut/root rw rootfstype=ext4 quiet console=ttyS0,115200n81 selinux=0' '-initrd' '/var/tmp/dracut-test.AYlYTw/initramfs.makeroot'
2023-08-30T08:20:11.1858709Z qemu-system-x86_64: warning: TCG doesn't support requested feature: CPUID.01H:ECX.x2apic [bit 21]
2023-08-30T08:20:11.1866974Z qemu-system-x86_64: warning: TCG doesn't support requested feature: CPUID.01H:ECX.tsc-deadline [bit 24]
2023-08-30T08:20:11.1867618Z qemu-system-x86_64: warning: TCG doesn't support requested feature: CPUID.07H:EDX.spec-ctrl [bit 26]
2023-08-30T08:20:11.1873851Z qemu-system-x86_64: warning: TCG doesn't support requested feature: CPUID.01H:ECX.x2apic [bit 21]
2023-08-30T08:20:11.1874469Z qemu-system-x86_64: warning: TCG doesn't support requested feature: CPUID.01H:ECX.tsc-deadline [bit 24]
[...]

PR #2521, opened 2023-09-19
fix(install.d): do not create initramfs if the supplied image is UKI

vittyvk:
When the supplied kernel image is a UKI, there's no point in creating initramfs as the UKI has it built-in already. This is the situation when dracut.install is called for a distro shipped UKI.

Note, KERNEL_INSTALL_IMAGE_TYPE == "uki" is different from KERNEL_INSTALL_LAYOUT == "uki", the later can be used to create UKI upon installing a standard kernel image.

This pull request changes...

## Checklist
- [ X ] I have tested it locally

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -11,6 +11,11 @@ if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
     exit 0
 fi
 
+# Do not attempt to create initramfs if the supplied image is already a UKI
+if [[ "$KERNEL_INSTALL_IMAGE_TYPE" = "uki" ]]; then
+    exit 0
+fi
+
 # Mismatching the install layout and the --uefi/--no-uefi opts just creates a mess.
 if [[ $KERNEL_INSTALL_LAYOUT == "uki" && -n $KERNEL_INSTALL_STAGING_AREA ]]; then
     BOOT_DIR_ABS="$KERNEL_INSTALL_STAGING_AREA"


aafeijoo-suse (approved):
LGTM, thanks!

vittyvk:
Looking at the failed integration test,

```
./test.sh: line 24:  2497 Segmentation fault      (core dumped) "$testdir"/run-qemu "${disk_args[@]}" -append "root=/dev/dracut/root rw rootfstype=ext4 quiet console=ttyS0,115200n81 selinux=0" -initrd "$TESTDIR"/initramfs.makeroot
TEST: root filesystem on a ext4 filesystem   [FAILED] 
```
I honestly don't think this has anything to do with my tiny patch but please let me know if there's anything I can take care of.

aafeijoo-suse:
> Looking at the failed integration test,
> 
> ```
> ./test.sh: line 24:  2497 Segmentation fault      (core dumped) "$testdir"/run-qemu "${disk_args[@]}" -append "root=/dev/dracut/root rw rootfstype=ext4 quiet console=ttyS0,115200n81 selinux=0" -initrd "$TESTDIR"/initramfs.makeroot
> TEST: root filesystem on a ext4 filesystem   [FAILED] 
> ```
> 
> I honestly don't think this has anything to do with my tiny patch but please let me know if there's anything I can take care of.

Unrelated to this patch. See #2506

vittyvk:
@haraldh @johannbg could you please take a look? Thanks!

aafeijoo-suse:
You should request that to @lnykryn these days. FTR, see #2274

vittyvk:
I missed that there was a discussion already. I don't fully agree with the resolution on #2274: it is perfectly valid to have a mix of Type-1 BLS entries (standard kernels which require initramfs) and Type-2 entries (which already have initramfs inside) and the patch just skips creating useless initramfs-es for the later. Such 'mixed' cases can't be resolved by not installing dracut.

johannbg:
@vittyvk care to explain those use cases in detail?

vittyvk:
@johannbg sure, let me try.

- On a system which normally uses UKIs user wants to install e.g. '-debug' kernel, customize cmdline/initramfs and do some tests/debugging/... In this case there's going to be a mix of Type-1 and Type-2 entries and the user may freely switch between them in bootloader (e.g. sd-boot).
- On a system which normally uses UKIs user wants to build kernel locally -- to try newer version, a patch,... The result can be either a standard kernel which requires initramfs or a UKI -- and dractut can be used to build these. So at the end it may still be a 'pure UKI' system (only Type-2 entries) but some of these UKIs are locally built and thus require an initramfs-making tool.
- A user of a standard (only Type-1 entries) system wants to try switching to using UKIs but when installing their first UKI they want to be safe and preserve existing, working kernels in case UKI doesn't boot. In this case dracut cannot be removed from the system (as standard kernels likely 'require' it) but we certainly don't need to generate initramfs for the newly installed UKI.

AndrewAmmerlaan:
Plus with this change distro's that currently rely on `kernel-install` to generate the initrd and install the kernel (e.g. Gentoo) won't have to change the installation logic when/if they in the future decide to distribute the pre-built kernel in UKI format instead. Without this change the UKI would be generated locally by `kernel-install` again, which is a) pointless and b) probably results in installing the locally generated unsigned UKI instead of the one that was pre-built and pre-signed. Therefore I think this change makes sense.

vittyvk:
Gentle ping)

PR #2526, opened 2023-09-26
fix(systemd-pcrphase): rename systemd-pcrphase binary to systemd-pcrextend

aafeijoo-suse:
The `systemd-pcrphase` binary has been renamed to `systemd-pcrextend` since https://github.com/systemd/systemd/commit/32295fa0

This patch will be necessary for systemd-v255.

## Checklist

### DIFF
--- a/modules.d/01systemd-pcrphase/module-setup.sh
+++ b/modules.d/01systemd-pcrphase/module-setup.sh
@@ -6,7 +6,7 @@
 check() {
 
     # If the binary(s) requirements are not fulfilled the module can't be installed.
-    require_binaries "$systemdutildir"/systemd-pcrphase || return 1
+    require_binaries "$systemdutildir"/systemd-pcrextend || return 1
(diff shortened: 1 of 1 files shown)


LaszloGombos (approved):

lnykryn:
Wouldn't this break compatibility with older systemd?

ferringb:
> Wouldn't this break compatibility with older systemd?

yep.

aafeijoo-suse:
> Wouldn't this break compatibility with older systemd?

Yes, `systemd-pcrphase` was introduced in systemd-v252. I thought that our current upstream policy was to support only the latest version of other upstream dependencies, otherwise this is unmanageable.

LaszloGombos (commented):
CI is still on systemd-v252

LaszloGombos (changes requested):
CI is still on systemd-v252

PR #2527, opened 2023-09-26
fix(resume): add new systemd-hibernate-resume.service

### LINKED ISSUE #2513: Check if the new `systemd-hibernate-resume.service` is required to resume from hibernation
Added since https://github.com/systemd/systemd/commit/a628d933

In case systemd now requires this new service to resume from hibernation, the `resume` module must install it in the initrd.

CC @YHNdnzj

### LINKED ISSUE #2587: Dracut does not generate images that support hibernation
**Describe the bug**
Images built with dracut do not allow for hibernation.

**Distribution used**
EndeavourOS Linux

**Dracut version**
dracut 060

**Init system**
systemd

**To Reproduce**
1. Build initramfs using dracut and reboot
2. Run `systemctl hibernate`
3. The command fails with `Failed to find location to hibernate to: Invalid argument`

**Expected behavior**
The command should hibernate the system, and the state of the system should be restored from disk on boot.

**Additional context**
This was not an issue before I updated systemd to version 255. This issue is not present in images built with `mkinitcpio` as they allow for hibernation. I ensured  sure that the `resume` module was included in the initramfs image using `lsinitrd /boot/initramfs-linux.img -m | grep resume`

aafeijoo-suse:
Since https://github.com/systemd/systemd/commit/a628d933, the generator only does the initial validation of the system info and then enables the new `systemd-hibernate-resume.service`.

Not tested, but this patch will be required for systemd-v255 based on https://github.com/dracutdevs/dracut/issues/2513#issuecomment-1713597890.

## Checklist

Fixes #2513

### DIFF
--- a/modules.d/95resume/module-setup.sh
+++ b/modules.d/95resume/module-setup.sh
@@ -44,6 +44,7 @@ install() {
     if dracut_module_included "systemd" && [[ -x $dracutsysrootdir$systemdutildir/systemd-hibernate-resume ]]; then
         inst_multiple -o \
             "$systemdutildir"/system-generators/systemd-hibernate-resume-generator \
+            "$systemdsystemunitdir"/systemd-hibernate-resume.service \
             "$systemdsystemunitdir"/systemd-hibernate-resume@.service \
             "$systemdutildir"/systemd-hibernate-resume
         return 0


Flowdalic:
I have tested this patch and can confirm that it is required for resume functionality with systems 255. Please consider merging it soon, since systemd 255 was released recently.

MikPisula:
I have tested this patch as well.

LaszloGombos (approved):

PR #2529, opened 2023-09-29
fix(45ifcfg): mark as deprecated and strictly opt-in

jlebon:
The `45ifcfg` module should be considered deprecated now since ifcfg files themselves are deprecated. They've been replaced by e.g. NetworkManager keyfiles or systemd-networkd files.

Currently, the `45ifcfg` module checks if the `/etc/sysconfig/network-scripts` directory exists to know if to automatically get pulled in. However, on systems with NetworkManager, this directory always exists even if the functionality to read ifcfg files is disabled by default (NM kindly ships a README in there to ease migrating to keyfiles).

Since almost everyone should've already migrated to a more modern alternative, let's make this module purely opt-in now.

## Checklist

### DIFF
--- a/modules.d/45ifcfg/module-setup.sh
+++ b/modules.d/45ifcfg/module-setup.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
+# This module is deprecated. Modern replacements are NetworkManager keyfiles and
+# systemd network files. It must now be explicitly opted in by the user to be
+# added to the initrd.
+
 # called by dracut
 check() {
-    [[ -d $dracutsysrootdir/etc/sysconfig/network-scripts ]] && return 0
     return 255
 }
 
--- a/test/TEST-60-BONDBRIDGEVLAN/test.sh
+++ b/test/TEST-60-BONDBRIDGEVLAN/test.sh
@@ -367,7 +367,7 @@ test_setup() {
     "$DRACUT" -l -i "$TESTDIR"/overlay / \
         --no-early-microcode \
         -o "plymouth" \
-        -a "debug ${USE_NETWORK}" \
+        -a "debug ${USE_NETWORK} ifcfg" \
         --no-hostonly-cmdline -N \
         -f "$TESTDIR"/initramfs.testing "$KVERSION" || return 1
 


jlebon:
An alternative is to tweak the conditional to check whether the directory is non-empty (skipping over the README) to know if ifcfg files are in use, but that creates a bootstrapping problem where you want ifcfg files generated on first boot to get propagated there in the first place. I also wanted to see if we could have a stronger stance on this and maybe eventually get rid of that module entirely.

aafeijoo-suse (commented):
`TEST-60-BONDBRIDGEVLANIFCFG` using the `network-legacy` module in Fedora needs to be addressed: https://github.com/dracutdevs/dracut/actions/runs/6353041817/job/17256980641?pr=2529#step:4:1186

The `ifcfg` module is required and not automatically included. See an old run, e.g.: https://github.com/dracutdevs/dracut/actions/runs/6309637128/job/17129970526?pr=2527#step:4:101

Apart from that, I'm ok with deprecating RH-specific code, or removing this module if it's no longer needed.

bengal:
Makes sense to me, ifcfg support is also being deprecated in NetworkManager and will be removed eventually.

jlebon:
> `TEST-60-BONDBRIDGEVLANIFCFG` using the `network-legacy` module in Fedora needs to be addressed: [dracutdevs/dracut/actions/runs/6353041817/job/17256980641?pr=2529#step:4:1186](https://github.com/dracutdevs/dracut/actions/runs/6353041817/job/17256980641?pr=2529#step:4:1186)
> 
> The `ifcfg` module is required and not automatically included. See an old run, e.g.: [dracutdevs/dracut/actions/runs/6309637128/job/17129970526?pr=2527#step:4:101](https://github.com/dracutdevs/dracut/actions/runs/6309637128/job/17129970526?pr=2527#step:4:101)

Hmm OK, I'm not sure if I'm following the testing code properly. Do I just need a `-m network-legacy` [here](https://github.com/dracutdevs/dracut/blob/2b1f338fab83d95ab46cdaecd11f28cf9308669b/test/TEST-60-BONDBRIDGEVLAN/test.sh#L367)?

jlebon:
> > Hmm OK, I'm not sure if I'm following the testing code properly. Do I just need a `-m network-legacy` [here](https://github.com/dracutdevs/dracut/blob/2b1f338fab83d95ab46cdaecd11f28cf9308669b/test/TEST-60-BONDBRIDGEVLAN/test.sh#L367)?
> 
> @jlebon can you please try to add this and see if that resolves the regression with TEST-60. Thanks !

Ahhh thanks. My mind got mixed up while looking at the test code and was thinking about `network-legacy` instead of `ifcfg`.

Looks like that fixes it, though there are other tests that failed. Seems like we'll need to tweak a few others.

LaszloGombos (approved):

PR #2530, opened 2023-10-03
Enable Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR.

It is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

## Checklist

This will resolve your issues with cleaning your codebase every time the new ShellCheck comes out: https://github.com/dracutdevs/dracut/pull/2501

Differential ShellCheck only reports defects directly related to the PR/commit. So you can focus on defects caused by your changes and deal with the existing defects later.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,11 +16,9 @@ jobs:
       - name: shfmt
         uses: luizm/action-sh-checker@v0.6.0
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SHFMT_OPTS: -s # arguments to shfmt.
         with:
           sh_checker_shellcheck_disable: false
-          sh_checker_comment: true
 
   differential-shellcheck:
     runs-on: ubuntu-latest
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,6 +22,30 @@ jobs:
           sh_checker_shellcheck_disable: false
           sh_checker_comment: true
 
+  differential-shellcheck:
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
+
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


LaszloGombos (approved):

LaszloGombos (approved):
Not sure if 3 separate commits adds value in this PR.

jamacku:
I can squash them If you prefer.

jamacku:
> @jamacku Have you had a chance to test this ? I run a brief test and at first run this will open about 300 code scanning issues based on the existing issues we have. Is this what is expected ?

You are right. There are 285 ShellCheck reports in the Security Dashboard. This is expected. It's for your awareness. You can review them and decide how to deal with them. The security dashboard allows you to set notes to each report and dismiss them.

> What do you recommend after this happens ? Just leave the existing issues alone and focus on avoiding adding more issues ? Can you perhaps point to a project where we can see this in action ?

I usually leave the existing issues open and focus on not adding more. But the choice is yours.

The security dashboard is "private" to members of the organization, so you won't see the results of other projects that use differential shellcheck unless you are a member. Reports on PR can see anyone.

For example, the [systemd](https://github.com/systemd/systemd) project is keeping shell scripts free from ShellCheck defects - https://github.com/systemd/systemd/blob/main/.github/workflows/differential-shellcheck.yml

We also use it on our RHEL8 and RHEL9 dracut distribution git:
- https://github.com/redhat-plumbers/dracut-rhel9/pull/63

It's beneficial to have differential ShellCheck scans because you can start linting your code base without dealing with current reports first (that saves time, allows you to check new incoming changes, and avoids potential regressions caused by fixing false positives).

pvalena:
LGTM.

pvalena (approved):

PR #2531, opened 2023-10-06
feat(kernel-network-modules): Add Qualcomm IPC router to enable USB

jlinton:
The x13s, and possibly other Qualcomm based devices, need the QC IPC router driver to enable USB. Without it, it's not possible to boot from USB-C attached disks.

## Checklist

Fixes #

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -95,7 +95,8 @@ installkernel() {
                 "=drivers/usb/misc" \
                 "=drivers/usb/musb" \
                 "=drivers/usb/phy" \
-                "=drivers/scsi/hisi_sas"
+                "=drivers/scsi/hisi_sas" \
+                "=net/qrtr"
         fi
 
         awk -F: '/^\// {print $1}' "$srcmods/modules.dep" 2> /dev/null | instmods


jlinton:
I don't think so, but maybe?

Despite the "networking" it is largely for the USB controller functionality. Which I guess could be a USB nic as well as USB storage. So, maybe my understanding of the difference between this file and that one surrounds whether one wants actual networking rather than the needed driver happens to be in the kernel networking directory because well its a proprietary hardware/firmware IPC protocol.

aafeijoo-suse (commented):
Hmm... it's a socket based interface, but used only for sort of "local networks". Although it may make sense to add it via `kernel-modules`, maybe the right place for a network driver would be `kernel-network-modules`.

BTW, do not prepend the module number in your commit message (`90kernel-modules` -> `kernel-modules`).

jlinton:
Alright, I will switch it to the networking modules.

aafeijoo-suse (approved):

ausil:
I strongly disagree about using the networking modules to pull in qrtr. The Fedora-Minimal disk image for instance does not install dracut-network. with the patch as is the disk image installed on a USB stick will not be bootable on the x13s as qrtr will be missing. As it is needed for working USB and block storage it needs to always be included.

aafeijoo-suse:
> I strongly disagree about using the networking modules to pull in qrtr. The Fedora-Minimal disk image for instance does not install dracut-network. with the patch as is the disk image installed on a USB stick will not be bootable on the x13s as qrtr will be missing. As it is needed for working USB and block storage it needs to always be included.

Then you can add it via conf file shipped by the distro.

ausil:
> > I strongly disagree about using the networking modules to pull in qrtr. The Fedora-Minimal disk image for instance does not install dracut-network. with the patch as is the disk image installed on a USB stick will not be bootable on the x13s as qrtr will be missing. As it is needed for working USB and block storage it needs to always be included.
> 
> Then you can add it via conf file shipped by the distro.

https://android.googlesource.com/kernel/msm/+/android-7.1.0_r0.2/Documentation/arm/msm/msm_ipc_router.txt outlines the purpose of qrtr. It does networking between Processors inside of a single system. It honestly probably needs to be moved out of net/ in the kernel tree, as I understand things, it was added there because its initial use was to configure Qualcomm modems.

jlinton:
Well (AFAIK) the stuff in net/qrtr is intended to be the userspace access point to the bus, so in that regard it makes sense to be in the upstream net directory. 

But looking closer at how dracut-network is used in fedora, I think my original assumption that it belongs in the general kernel modules is correct. This fix only works because kdump (maybe incorrectly) is pulling it in for the minority of users who might want to kdump to a network target. 

So, I agree with @ausil here that it doesn't really belong in kernel-network-modules.

aafeijoo-suse:
> As it stands, I see this as a workaround for a bug that's not in dracut. That said I am open for adding it back to the kernel-modules with some kind comment or reference or resolution to the bigger upstream issue.

I agree. Taking into consideration everything discussed, maybe it's better to add this in `kernel-modules`, but please commenting why.

jlinton:
Ok, moved back with a larger commit message about what is going on.

Conan-Kudo (approved):

Conan-Kudo:
Please downcase "add" to make the bot happy:

```
Error: commit_message:1:23: error: don't use title case in the description
feat(kernel-modules): Add Qualcomm IPC router to enable USB
                      ^~~~~~~~~~~~
```

jlinton:
Ah, right, that got lost when I reverted to the original commit. It should be fixed now.

LaszloGombos (approved):

PR #2533, opened 2023-10-13
fix(install.d): do not generate a new initrd if any INITRD_FILE is provided

aafeijoo-suse:
If `kernel-install add` is called specifying one or more INITRD-FILEs, the caller just wants to use these, so there is no point in generating a new one.

## Checklist


aafeijoo-suse:
I just noticed that this was already attempted to be fixed in #1691 -> https://github.com/dracutdevs/dracut/pull/1691/commits/56aa3e5653f6e10ed9976d82c1ed10b4058af8dd

PR #2534, opened 2023-10-13
consolidated s390 device configuration

steffen-maier:
Consolidate the persistent and dynamic configuration of s390-specific devices in Linux distributions by delegating the configuration to the existing framework `zdev` from s390-tools.

This pull request completes **distribution independent** s390 device configuration in initrds built by `dracut`.
This also **fixes** a few known and newly discovered **bugs** in addition to the changes described below. Please see the commit descriptions for more bug fix information.
Some of the commits [see their description] **depend** on certain commits from https://github.com/ibm-s390-linux/s390-tools/pull/158 and thus https://github.com/ibm-s390-linux/s390-tools/releases/tag/v2.31.0.

Zdev's job is to perform low-level configuration after which the user gets architecture-independent objects such as block devices, SCSI devices, or network interfaces. Those can and should in turn be configured with existing common code mechanisms. So there's a clear separated layering for configuration duties.

In particular, the s390-specific devices currently are: DASD (traditional disk), ZFCP (scsi), and ZNET representing channel-attached network (QETH incl. OSA and HiperSockets, LCS, CTC). Zdev has a stable command line user interface and abstracts from sysfs and from a persistent configuration representation. Zdev encapsulates configuration details. Systems management code can simply delegate configuration to zdev and thus reduce architecture-specific code.

_This improves user experience, serviceability, maintainability, and reduces test effort._

@danimo @dtardon @haraldh @hramrach @hreinecke @jstodola @mwilck @sharkcz @smitterl @teclator

## Changes

Dracut had 3 different plugin module sets to configure s390 devices. Those module sets partly competed with each other and partly contradicted each other such as activating too many devices for a memory-limited kdump environment.
- original modules from Red Hat: `dasd`, `dasd_mod`, `znet`, `zfcp`.
- additional modules from SUSE because the Red Hat modules were tightly   coupled with the Red Hat specific persistent configuration mechanism and SUSE needed their own persistent mechanism: `dasd_rules`, `qeth_rules`, `zfcp_rules`.

Along with zdev as configuration backend, IBM introduced https://github.com/ibm-s390-linux/s390-tools/tree/master/zdev/dracut
- making use of zdev as consolidated configuration mechanism: `zdev`, `zdev-kdump`.

_This change consolidates the 3 dracut module sets into one single module set. The user syntax with rd.dasd, rd.znet, and rd.zfcp remains the same; only the backend implementation changes to zdev._
Move `dasd` and `zfcp` into `zdev` of s390-tools (modules could alternatively exist in dracut but they are so tightly coupled with zdev that s390-tools seemed natural).
It supports dracut features: hostonly, hostonly-cmdline, print-cmdline.
Update `znet` and `cms` (cmsdasd & cmsconffile feature used by anaconda installer) and keep them in dracut because other code depends on these modules.

## Checklist


dtardon:
Please submit the first two commits as a separate PR (or PRs). They make sense of their own and are not tied to the rest.

steffen-maier:
> Some of the commits [see their description] **depend** on certain commits from [ibm-s390-linux/s390-tools#158](https://github.com/ibm-s390-linux/s390-tools/pull/158).

was merged, so now turning this into a regular pull request

steffen-maier:
rebased onto latest master and force pushed to resolve the merge conflicts, which meanwhile appeared due to upstream commit de8ac6300d11 ("fix(github): update format of labeler")

modules.d/80cms/module-setup.sh
```diff
@@ -4,6 +4,8 @@
 check() {
     arch=${DRACUT_ARCH:-$(uname -m)}
     [ "$arch" = "s390" -o "$arch" = "s390x" ] || return 1
+    require_binaries chzdev || return 1
```
dtardon, line 7: Just merge these two lines. `require_binaries` takes multiple arguments.

modules.d/80cms/module-setup.sh
```diff
@@ -13,13 +13,14 @@ check() {
...
     [ "$arch" = "s390" -o "$arch" = "s390x" ] || return 1
-    echo znet dasd dasd_mod bash
+    echo znet bash
     return 0
 }
 
 # called by dracut
 installkernel() {
     instmods zfcp
+    instmods dasd_mod dasd_eckd_mod dasd_fba_mod dasd_diag_mod
```
dtardon, line 23: Please add the new modules to the existing `instmods` call.

dtardon:
Not a matter for this PR, but note that NetworkManager has dropped support for the `ifcfg` format, therefore the `cms` module will have to be ported off it eventually.

dtardon:
> > Not a matter for this PR, but note that NetworkManager has dropped support for the `ifcfg` format, therefore the `cms` module will have to be ported off it eventually.
> 
> I'm struggling to understand the ifcfg vs. nm-connection in the current existing code of module 80cms. Even though I tested it with the code of this PR here, I did not rd.debug it to understand which code path it actually took. It "just worked" for me.
> 
> https://github.com/dracutdevs/dracut/blob/4980bad34775da715a2639b736cba5e65a8a2604/modules.d/80cms/cmsifup.sh#L26-L43
> 
> It does look like there is a conditional branch to already use nm-connections, if nm-initrd-generator is available, and to me it looks like it delegates higher-level network-interface config to dracut cmdline option ip= (and nameserver=) so core dracut could handle this via nm-connections instead of ifcfg?

Eh. It seems my memory has failed me: I'd have sworn that `cms` wrote `ifcfg` files to configure the network. I should have checked...

sharkcz (approved):
LGTM

I think we should be able to remove the `$(rpm -ql s390utils-base)` from `inst_multiple`, but that's a separate topic.

edit: Hmm, this was meant for the 1st patch ...

sharkcz:
Let's do it this way as I went over the patches in my review

OK bf9229cb refactor(cms): use zdev to simplify handling CMSDASD=... boot option
- remove  `$(rpm -ql s390utils-base)` from `inst_multiple` after refactoring??

OK b91b45e9 refactor(cms): use consolidated zfcp config with zdev from s390-tools

OK 5a397f95 refactor(cms): use consolidated dasd config with zdev from s390-tools

OK c5cdbfec refactor(cms): use consolidated network config with zdev from s390-tools

OK 4cee5e38 feat(zfcp_rules): remove zfcp handling consolidated in s390-tools
- the URL in the commit message could be pointing directly to the "zdev" commit instead

OK 6e91a4bb refactor(zfcp): minimize zfcp handling consolidated in s390-tools
- update 1st paragraph in the commit message, copied from 4cee5e38? missing reference to s390-tools zdev parse-zfcp.sh

OK 1c628c82 feat(dasd_rules): remove dasd handling consolidated in s390-tools
- the URL in the commit message could be pointing directly to the "zdev" commit instead

OK 49398eb8 feat(dasd_mod): remove dasd handling consolidated in s390-tools

OK f5146389 refactor(dasd): minimize dasd handling consolidated in s390-tools
- update 1st paragraph in the commit message, copied from ??? missing reference to s390-tools zdev parse-dasd.sh

OK 33d7929c docs(dracut.cmdline): generalize description of rd.znet
- perhaps replace "activates interface on s390" with "... interface on s390x platform"?

?? ae3e6923 fix(znet): append to udev rules so each rd.znet_ifname is effective
- hmm, wouldn't there be a conflict in the `ccw_ifname_end` label when multiple rules are copied into the file??

OK 6bb660f0 feat(znet): use zdev for consolidated device configuration

OK 36e2cc6d feat(qeth_rules): remove qeth handling consolidated in 95znet

OK feb09a29 refactor(ifcfg): delete code duplication using iface_get_subchannels()

OK 255094b7 feat(ifcfg): minimize s390-specific network configuration aspects

----
OK cacaf073 feat(zfcp): finally remove zfcp handling consolidated in s390-tools
OK c91ac8b6 feat(dasd): finally remove dasd handling consolidated in s390-tools
- but those 2 must be a separate PR to be merged after we implement the migration via s390utils I think.

There is no single "upgrade point" in Fedora (or ELN), so the migration needs to be done via a regular package update. And this will need some coordination and setting versioned dependencies between the involved packages (dracut, s390utils, ...)

steffen-maier:
(I do have a git range-diff invocation, if the following text is too hard to read.)

During testing I realized that I had introduced a bug in 95znet/parse-ccw.sh.
Probably when I fixed all shellcheck reports and added missing double quotes after my function tests.
The `"$*"` needs to be `"$@"`, because otherwise:
1. If there are options at the end of rd.znet, they are handed as one single argument to chzdev. This does not work if there is more than one option, as each option needs to be a separate argument to chzdev.
2. If there a no options at the end of rd.znet, it results in one empty argument to chzdev, which chzdev does not like as it interprets positional arguments as device specifiers and an empty string is not valid device specifier.
Fixed and tested.

Fixed up the commit messages (direct commit URL, and missing context) of:
- 6 was 5: feat(zfcp_rules): remove zfcp handling consolidated in s390-tools
- 7 was 6: refactor(zfcp): minimize zfcp handling consolidated in s390-tools
- 8 was 7: feat(dasd_rules): remove dasd handling consolidated in s390-tools
- 9 was 8: feat(dasd_mod): remove dasd handling consolidated in s390-tools
- 10 was 9: refactor(dasd): minimize dasd handling consolidated in s390-tools

> OK [bf9229c](https://github.com/dracutdevs/dracut/commit/bf9229cb5419809ca09d73f4761899dc412f155f) refactor(cms): use zdev to simplify handling CMSDASD=... boot option
>
> * remove  `$(rpm -ql s390utils-base)` from `inst_multiple` after refactoring??

Inserted new commit 5:
("refactor(cms): remove now unnecessary inclusion of full s390utils-base")

Tested that dasd, zfcp, and znet still work.

> OK [33d7929](https://github.com/dracutdevs/dracut/commit/33d7929c222130b62874bdfc8823f8797f10ef91) docs(dracut.cmdline): generalize description of rd.znet
>
> * perhaps replace "activates interface on s390" with "... interface on s390x platform"?

I felt more comfortable with appending "architecture" than "platform".

Additionally, I fixed up the commit description: The pre-req for the doc update is not a change in s390-tools, but the (meanwhile) preceding commit ("feat(znet): use zdev for consolidated device configuration") here in the dracut PR.
For this, I moved previous commit 12 to position 11.
So the order now is:
11: feat(znet): use zdev for consolidated device configuration
12: docs(dracut.cmdline): generalize description of rd.znet

> ?? [ae3e692](https://github.com/dracutdevs/dracut/commit/ae3e6923c0fd6b6d90177730104d7067a2950772) fix(znet): append to udev rules so each rd.znet_ifname is effective
>
> * hmm, wouldn't there be a conflict in the `ccw_ifname_end` label when multiple rules are copied into the file??

Indeed, thanks for catching this bug! This was the only commit, where I was sloppy in developing and did not really function test it.

Fixed by sanitizing the user input for `ifname` (replacing non-word characters by underscore) and then using that as LABEL string to make the labels unique.

Tested successfully with two rd.znet and one rd.znet_ifname for each.

I moved this commit from previous position 11 to now 14, which is moving the bugfix to after the sequence of feature commits consolidating znet (11..12) and removing qeth_rules (13).

steffen-maier:
> OK [cacaf07](https://github.com/dracutdevs/dracut/commit/cacaf073b46f2c76766637e0bcb9f58ba14894da) feat(zfcp): finally remove zfcp handling consolidated in s390-tools
> OK [c91ac8b](https://github.com/dracutdevs/dracut/commit/c91ac8b66f1690a496c0f3af9728a4a16c266e86) feat(dasd): finally remove dasd handling consolidated in s390-tools
>
> * but those 2 must be a separate PR to be merged after we implement the migration via s390utils I think.
>
> There is no single "upgrade point" in Fedora (or ELN), so the migration needs to be done via a regular package update. And this will need some coordination and setting versioned dependencies between the involved packages (dracut, s390utils, ...)

Understood.

Originally, [95dasd_mod/parse-dasd-mod.sh](https://github.com/dracutdevs/dracut/blob/4980bad34775da715a2639b736cba5e65a8a2604/modules.d/95dasd_mod/parse-dasd-mod.sh#L17), [95zfcp/parse-zfcp.sh](https://github.com/dracutdevs/dracut/blob/4980bad34775da715a2639b736cba5e65a8a2604/modules.d/95zfcp/parse-zfcp.sh#L16), and [95znet/parse-ccw.sh](https://github.com/dracutdevs/dracut/blob/4980bad34775da715a2639b736cba5e65a8a2604/modules.d/95znet/parse-ccw.sh#L25) perform free from cio_ignore during dracut cmdline hook (and independently of any existing corresponding cmdline option on the kernel/dracut cmdline). This triggers actual device configuration by processing any /etc/dasd.conf and /etc/zfcp.conf which were copied into initrd.

The following commits had broken above behavior. I changed them to retain the cio_ignore handling so any /etc/dasd.conf or /etc/zfcp.conf copied from the root-fs into initrd still has the same effect (and only that; the actual dracut cmdline option parsing is already part of s390-tools since https://github.com/ibm-s390-linux/s390-tools/pull/158):

* refactor(zfcp): minimize zfcp handling consolidated in s390-tools => feat(zfcp): minimize zfcp handling consolidated in s390-tools
* feat(dasd_mod): remove dasd handling consolidated in s390-tools => feat(dasd_mod): minimize dasd handling consolidated in s390-tools

And I removed the last 2 commits you mentioned from this PR here, to move them to a new PR, which will intentionally be in draft mode to defer any merging until you give the go from a Fedora packaging point of view.

BTW:
Strictly speaking, Fedora has been following a rule that anything in the (device) dependency tree to prepare for mounting the root-fs inside initrd needs to have [rd.dasd](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/performing_a_standard_rhel_9_installation/assembly_installing-on-64-bit-ibm-z_installing-rhel#dasds-that-are-part-of-the-root-file-system_configuring-a-linux-instance-on-ibm-z), [rd.zfcp](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/performing_a_standard_rhel_9_installation/assembly_installing-on-64-bit-ibm-z_installing-rhel#fcp-luns-that-are-part-of-the-root-file-system_configuring-a-linux-instance-on-ibm-z), or [rd.znet](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/performing_a_standard_rhel_9_installation/assembly_installing-on-64-bit-ibm-z_installing-rhel#configuring-an-ibm-z-network-device-for-network-root-file-system_configuring-a-linux-instance-on-ibm-z) on the kernel cmdline, if there are corresponding s390 devices among the leaf nodes of the dependency tree. The anaconda installer (with the help of python-blivet) has been taking care of that since a long time for the initial Linux installation by means of dracut_setup_args(): [rd.dasd](https://github.com/storaged-project/blivet/blob/715c1861db252fc1a1ee47ae040af9c846e3515e/blivet/devices/disk.py#L542-L585), [rd.zfcp](https://github.com/storaged-project/blivet/blob/715c1861db252fc1a1ee47ae040af9c846e3515e/blivet/devices/disk.py#L497-L504), or [rd.znet](https://github.com/rhinstaller/anaconda/blob/8b5f5920d62bcebfadc06285e9b681d4d36f45d1/pyanaconda/modules/network/nm_client.py#L1368-L1383) ([some history](https://github.com/rhinstaller/anaconda/pull/5250/commits/fc0dbe8ae95ee4874b4ec04e2e27a759c109db42)). Hence, one could challenge any functionality depending on copying of /etc/dasd.conf or /etc/zfcp.conf into initrd and those copies having an effect during early initrd.
Additionally, initrd should only activate required devices: https://github.com/dracutdevs/dracut#readme
> "The initramfs has (basically) one purpose in life -- getting the rootfs mounted so that we can transition to the real rootfs. ... This helps to keep the time required in the initramfs as little as possible so that things like a 5 second boot aren't made impossible as a result of the very existence of an initramfs."

IMO, everything else should be activated after switch root during another udev cold plug. It's easier to debug those at that point in time when we already have the full root-fs with all debug tooling.

steffen-maier:
> [LaszloGombos](https://github.com/LaszloGombos) requested a review from [lnykryn](https://github.com/lnykryn) [2023-10-30](https://github.com/dracutdevs/dracut/pull/2534#event-10808077972)

Hello @lnykryn, I'm wondering how I can help getting the required two (green) approving reviews. With you being one of the required approvers, do @sharkcz's [review comments](https://github.com/dracutdevs/dracut/pull/2534#issuecomment-1933744932) with his "grey" approval and my corresponding subsequent code updates help? [[1](https://github.com/dracutdevs/dracut/pull/2534#issuecomment-1968723358), [2](https://github.com/dracutdevs/dracut/pull/2534#issuecomment-1968730931)] Can I do anything else?

steffen-maier:
minor clarifications on `rd.zfcp` and `rd.dasd` in man/dracut.cmdline.7.asc

steffen-maier:
Put temporary files under the subdir ${DRACUT_TMPDIR} in https://github.com/dracutdevs/dracut/pull/2534/commits/5fc66498de7072a10358f1873b23eb95ff9b3acd ("feat(znet): use zdev for consolidated device configuration").
Idea from a [kdump dracut module](https://lists.fedoraproject.org/archives/list/kexec@lists.fedoraproject.org/message/WPJU2LKBNCW4S72DLHJKJZ7IFPJXQFNV/) (which additionally has its own subdir under $DRACUT_TMPDIR).
Also used in the related https://github.com/ibm-s390-linux/s390-tools/commit/849aa5b1058399447d478b0e129f23219e9dc7f1 ("zdev/dracut: put temporary files under the subdir ${DRACUT_TMPDIR}").

steffen-maier:
@johannbg moving discussion on s390 dracut modules here:

> Anyways back to the s390-tools PR's what's the current status of that and in which direction are the upstream 390 taking with regards to init systems, is it being generic or moving towards systemd only?

In case you refer to the merged https://github.com/ibm-s390-linux/s390-tools/pull/158 and the related pending https://github.com/dracutdevs/dracut/pull/2534 here, it is distro-independent and only uses udev. AFAIK, udev has always been the base for dracut so we should be good? I'm not aware of any dependencies on systemd (or any other init system) for low-level configuration of s390-specific devices.

> That said I'm not particularly fond of "partial" upstream, with that I mean we should not support some aspects of 390 here upstream while other parts of it is being maintained in 390 upstream repo which will double the load on everyone involved and quadruple it for anykind of support issues.

Do you mean the fact, that currently handling of DASD and ZFCP is moving to s390-tools, but ZNET remains in dracut? [as in the "Changes" section of the description of #2534 here] Originally, I thought I would move ZNET also into s390-tools, but refrained from doing so because other dracut modules have dependencies on `modules/95znet`, such as (these are the ones I'm aware of and were included in my function tests):
* https://github.com/dracutdevs/dracut/blob/5d2bda46f4e75e85445ee4d3bd3f68bf966287b9/modules.d/80cms/module-setup.sh#L11-L14 (note that the other module dependencies on dasd dasd_mod zfcp go away with #2534 so only znet would remain)
* as well as an out-of-tree (not in dracut upstream) kdump module such as https://src.fedoraproject.org/rpms/kexec-tools/blob/858de28447ff05b15ef3c7004f60c7babf1ab15c/f/dracut-module-setup.sh#_66.

I suppose we _could theoretically_ move znet into s390-tools upstream as well, if all affected downstream package maintainers are OK with _adapting runtime package requirements_ on (a particular minimum level of a new future version of) s390-tools, so the znet dracut module exists at initrd build time when dracut module cms or the kdump module depend on znet.
Would that be worth the effort?
In no particular order: @dtardon ? @pvalena ? (dracut packaging), @coiby (kdump-utils being factored out of downstream kexec-tools), @rvykydal (just FYI because anaconda uses znet and 80cms) @sharkcz (s390utils downsteam packaging of s390-tools), @aafeijoo-suse @tblume (dracut packaging), @ngueorguiev (s390-tools packaging), @joseivanlopez (distro installer might depend on znet), @hoeppnerj @steffen-eiden @oberpar @vneethv (s390-tools upstream and the zdev dracut module in there).

Any other alternative ideas?

steffen-maier:
rebased to resolve merge conflict with just merged #2630

dtardon:
> I suppose we _could theoretically_ move znet into s390-tools upstream as well, if all affected downstream package maintainers are OK with _adapting runtime package requirements_ on (a particular minimum level of a new future version of) s390-tools,

Doesn't this PR already require that?

> so the znet dracut module exists at initrd build time when dracut module cms or the kdump module depend on znet. Would that be worth the effort?

Yes, it would.

> Any other alternative ideas?

This would have to be tested, but I think it it'd not be a problem if both projects provided the module; they would just have to use a different leading number. (Because only the module with the lower number would be loaded, if I read the code correctly). Then, after some transitional period, the dracut `znet` module could be dropped.

tblume:
> > I suppose we _could theoretically_ move znet into s390-tools upstream as well, if all affected downstream package maintainers are OK with _adapting runtime package requirements_ on (a particular minimum level of a new future version of) s390-tools,
> 
> Doesn't this PR already require that?
> 
> > so the znet dracut module exists at initrd build time when dracut module cms or the kdump module depend on znet. Would that be worth the effort?
> 
> Yes, it would.
> 
> > Any other alternative ideas?
> 
> This would have to be tested, but I think it it'd not be a problem if both projects provided the module; they would just have to use a different leading number. (Because only the module with the lower number would be loaded, if I read the code correctly). Then, after some transitional period, the dracut `znet` module could be dropped.

I support that approach.
We are already using chzdev from the s390-tools for creating the udev rules for network device activation instead of the dracut znet module in our latest versions @SUSE.

PR #2536, opened 2023-10-17
modules: remove undefined $libdir var ref in 01-systemd-repart

berrange:
Nothing in dracut appears to define the '$libdir' environment variable when module scripts are being run. As a result when the 01-systemd-repart module script runs, it fails to copy any of the config files into the initrd. Replace it with '/usr/lib' which reflects where systemd-repart expects its global config files.


aafeijoo-suse (commented):
Thanks for spotting this, apparently no one has tested or used this module...

Minor nitpick, could you change the commit message to: `fix(systemd-repart): correct undefined $libdir`?

aafeijoo-suse (approved):

AdrianVovk:
> apparently no one has tested or used this module.

The module is used and tested. It's just that `systemd-repart` will automatically look into `/sysroot/usr/...` for its config files whenever it can, instead of using the config in the initrd

PR #2540, opened 2023-10-27
fix(pkcs11): Delete trailing dot on libcryptsetup-token-systemd-pkcs11.so

innovara:
pkcs11 module doesn't include cryptsetup/libcryptsetup-token-systemd-pkcs11.so as it's supposed to do.

## Changes

Changed the syntax of its install statement in modules.d/91pkcs11/module-setup.sh so it matches the file and it's included.

## Checklist


aafeijoo-suse (commented):
Thanks for your fix, another untested module...

Could you please change your commit message to `fix(pkcs11): delete trailing dot on libcryptsetup-token-systemd-pkcs11.so`

https://github.com/dracutdevs/dracut/actions/runs/6666390577/job/18126162850?pr=2540

innovara:
Done (I think!).

I haven't fully tested the module. I am working on getting a LUKS root partition unlocked with a smart card and now I'm troubleshooting other issues. For one, I think this module should require pcsc since I reckon a good number of pkcs11 tokens are smart cards, albeit not all of them.

But even when I add that manually, systemd reaches paths.target before pscsd.service is started and systemd-cryptsetp...  can't find the card and it asks for the password after it times out.

If I get somewhere sorting that out and, eventually, I manage to unlock LUKS with the card I'll obviously send another PR for everybody's benefit.

aafeijoo-suse (approved):

Issue #2328, opened 2023-05-08
Test 60 seems to fail on most distros


LaszloGombos:
The test runs very well on Fedora. I do not remember seeing a failure on Fedora.

Seems to fail on opensuse
https://github.com/dracutdevs/dracut/actions/runs/4932181615

Seems to fail on latest Arch
https://github.com/dracutdevs/dracut/actions/runs/4910624311

..and on latest debian
https://github.com/dracutdevs/dracut/actions/runs/4911580441

..with similar error message

grep: /run/initramfs/state/etc/sysconfig/network-scripts/ifcfg-*: No such file or directory

Related: https://github.com/dracutdevs/dracut/pull/2189

@bengal @aafeijoo-suse

bdrung:
ifcfg is RedHat/Fedora/etc specific. Debian and derivatives do not have ifcfg.

PR #2543, opened 2023-10-29
test: skip test 50,60 if ifcfg dracut module can not be installed

Henrik66:
Test 50 and 60 have a dependency on the ifcfg dracut module.
If ifcfg dracut module is not available, than the preconditions
for these tests are not met.
    
Since ifcfg dracut module is not available in many Linux enviroments
(such as Arch or Debian), it make sense to explicitelly skip these
tests instead of failing them as they are simply not applicable
to many Linux distributions in their current form.

### DIFF
--- a/test/TEST-50-MULTINIC/test.sh
+++ b/test/TEST-50-MULTINIC/test.sh
@@ -7,6 +7,11 @@ TEST_DESCRIPTION="root filesystem on NFS with multiple nics with $USE_NETWORK"
 #DEBUGFAIL="loglevel=7 rd.shell rd.break"
 #SERIAL="tcp:127.0.0.1:9999"
 
+# skip the test if ifcfg dracut module can not be installed
+test_check() {
+    test -d /etc/sysconfig/network-scripts
+}
+
 run_server() {
     # Start server first
     echo "MULTINIC TEST SETUP: Starting DHCP/NFS server"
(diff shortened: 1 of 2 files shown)


LaszloGombos (approved):
CC @bdrung

bdrung:
Please provide a reasoning in the git commit. At least these tests were failing on Debian/Ubuntu (see also bug #2328).

Henrik66:
@bdrung Thanks for the review. I added more reasoning to the git commit message.

>     test: skip test 50,60 if ifcfg dracut module can not be installed
>     
>     Test 50 and 60 have a dependency on the ifcfg dracut module.
>     If ifcfg dracut module is not available, than the preconditions
>     for these tests are not met.
>     
>     Since ifcfg dracut module is not available in many Linux enviroments
>     (such as Arch or Debian), it make sense to explicitelly skip these
>     tests instead of failing them as they are simply not applicable
>     to many Linux distributions in their current form.

As to your question, these tests are only meant to run if ifcfg dracut module is supported.
A better solution would be to reimplement these tests so that they do not require ifcfg dracut module. 
I am not able to provide such a solution at this time, so until than perhaps we can just skip the test.

bdrung:
> @bdrung Thanks for the review. I added more reasoning to the git commit message.

Thanks.

> As to your question, these tests are only meant to run if ifcfg dracut module is supported. A better solution would be to reimplement these tests so that they do not require ifcfg dracut module. I am not able to provide such a solution at this time, so until than perhaps we can just skip the test.

That information could go into the commit message as well, saying something like: Instead of disabling the test, they should be reimplement so that they do not require ifcfg dracut module in the future.

Henrik66:
> That information could go into the commit message as well, saying something like: Instead of disabling the test, they should be reimplement so that they do not require ifcfg dracut module in the future.

okay. I updated the commit message. Thanks @bdrung

PR #2544, opened 2023-10-30
fix(dracut.sh): skip README for AMD microcode generation

### LINKED ISSUE #2541: AuthenticAMD.bin includes README, breaking early loading
**Describe the bug**
Dracut builds AuthenticAMD.bin by doing more or less `cat /usr/lib/firmware/amd-ucode/* > AuthenticAMD.bin`
Since https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/commit/amd-ucode/README?id=89ec6198f13d1007563ff87aae5de209e993be07 the folder contains a README file.

**Distribution used**
Alma 8 / Alma 9 / Fedora 38

**Dracut version**
dracut-049-223.git20230119.el8.x86_64
dracut-057-21.git20230214.el9.x86_64
dracut-059-4.fc38.x86_64

**Init system**
systemd

**To Reproduce**

```
# LANG=C LC_ALL=C bash -c 'for i in/usr/lib/firmware/amd-ucode/*; do echo $i; done'
# dracut -N initramfs-test.img 6.5.6-200.fc38.x86_64
# lsinitrd --unpackearly initramfs-test.img
# binwalk kernel/x86/microcode/AuthenticAMD.bin 

DECIMAL       HEXADECIMAL     DESCRIPTION
--------------------------------------------------------------------------------
0             0x0             xz compressed data <-------- should not be there
1316          0x524           Intel x86 or x64 microcode, sig 0x01000083, pf_mask 0x00, 2008-04-30, rev 0x03c0, size 4130
...
```

**Expected behavior**
README (or README.xz or ...) is not part of AuthenticAMD.bin

**Additional context**
GenuineIntel.bin seems ok when looking at it with binwalk
I only validated that this is a problem with Alma 8 kernel, maybe newer version have some extra patches to ignore extra data, but we should still remove `README*`

aafeijoo-suse:
This file was added in https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/commit/amd-ucode/README?id=89ec6198f13d1007563ff87aae5de209e993be07 and it should be skipped.

## Checklist

Fixes #2541

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2154,6 +2154,8 @@ if [[ $early_microcode == yes ]]; then
                 done
                 for i in $_fwdir/$_fw/$_src; do
                     [[ -e $i ]] || continue
+                    # skip README{.xz,.zst,...}
+                    str_starts "$i" "$_fwdir/$_fw/README" && continue
                     # skip gpg files
                     str_ends "$i" ".asc" && continue
                     cat "$i" >> "$_dest_dir/${ucode_dest[$idx]}"


dracut.sh
```diff
@@ -2154,6 +2154,8 @@ if [[ $early_microcode == yes ]]; then
                 done
                 for i in $_fwdir/$_fw/$_src; do
                     [[ -e $i ]] || continue
+                    # skip README
```
champtar, line 2157:
```suggestion
                    # skip README*
```

aafeijoo-suse: That's why I used `str_starts` with the full path instead of `std_ends`, to handle the different possible compressions. Do you think a `*` is really necessary in the previous comment?

champtar: The `*` for me document that we are using `str_starts` because it's needed and not by accident.
Someone rewriting this code without context might think they only need to skip `README`

champtar (approved):

LaszloGombos (approved):

lnykryn (approved):

dracut.sh
```diff
@@ -2154,6 +2154,8 @@ if [[ $early_microcode == yes ]]; then
                 done
                 for i in $_fwdir/$_fw/$_src; do
                     [[ -e $i ]] || continue
+                    # skip README{.xz,.zst,...}
+                    str_starts "$i" "$_fwdir/$_fw/README" && continue
```
classabbyamp, line 2158: this just puts a bandage on a symptom of the problem, instead of fixing the actual problem: the proper way would be to only `cat` files that match the glob `*.bin` (for amd) or `??-??-??` (for intel)

PR #2545, opened 2023-10-30
fix: wait 12 hours before halt on media check fail

AdamWill:
If a modesetting driver has been loaded by the time the media check happens, halting the system (as we currently do immediately if the check fails) blanks the screen, which is confusing for the user. This adds a warning message and a 12-hour wait before the system is eventually halted, so the user can see the media check failure and (presumably) reboot and fix the medium.

It also tweaks the text of the failure message not to call it a "CD check", since it's not 1998 any more.

https://bugzilla.redhat.com/show_bug.cgi?id=2246410

## Checklist
- [-] I have reviewed and updated any documentation if relevant
- [-] I am providing new code and test(s) for it

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -76,7 +76,9 @@ if [ -n "$check" ]; then
         checkisomd5 --verbose "$check_dev"
     fi
     if [ $? -eq 1 ]; then
-        die "CD check failed!"
+        warn "Media check failed! We do not recommend using this medium. System will halt in 12 hours"
+        sleep 43200
+        die "Media check failed!"
         exit 1
     fi
     type plymouth > /dev/null 2>&1 && plymouth --show-splash


AdamWill:
Note, this fixes the problem for live images. Traditional installer images use a different codepath that lives in anaconda, I will send a PR for anaconda to fix the same problem on that path.

AdamWill:
the situation we're concerned about is that the user starts the media check and wanders off to do something else.

I considered various values from a minute upwards. 12 hours was nirik's suggestion.

AdamWill:
Also, I really doubt anyone is booting to the media check on a remote/unattended system. What would be the point? presumably, if you're in a position to check the result of the media check, you're in a position to power off the system after noticing it failed.

Conan-Kudo (approved):

AdamWill:
btw, I wasn't necessarily expecting this to be merged as-is, I was hoping someone had a better idea somehow. But I thought it was good enough for F39 at least, and wanted to submit it upstream before doing the downstream build.

PR #2547, opened 2023-11-01
fix(pcsc): add opensc load module file

innovara:
On Fedora, and possibly others, there is a file to load the module `opensc-pkcs11.so`.

In my tests, without this module, you can't unlock a LUKS encrypted volume using a smart card since `pscsd` is not able to establish any communication with it.

The file simply loads `opensc-pkcs11.so` with the following syntax:

`module: opensc-pkcs11.so`

There may be a more efficient way of doing this without having to attempt to add the file from various sources (each distro seems to have this file in different places).

## Changes

This pull request adds the installation statements needed to include the file that loads `opensc-pkcs11.so` on various distributions. I couldn't find where this file lives in Arch Linux. Maybe it's loaded in the configuration directly.

## Checklist

### DIFF
--- a/modules.d/91pcsc/module-setup.sh
+++ b/modules.d/91pcsc/module-setup.sh
@@ -29,7 +29,8 @@ install() {
     inst_simple "$moddir/pcscd.socket" "${systemdsystemunitdir}"/pcscd.socket
 
     inst_multiple -o \
-        pcscd
+        pcscd \
+        /usr/share/p11-kit/modules/opensc.module
 
     # Enable systemd type unit(s)
     for i in \


aafeijoo-suse (commented):
Thanks again for looking into this!

The current upstream rule is to follow only other upstreams, so remove all the comments and add only what the OpenSC packaging says: `/usr/share/p11-kit/modules/opensc.module`

https://github.com/OpenSC/OpenSC/blob/6d7fad101812dbcbde4a3ff069281a359fc54d1f/packaging/opensc.spec#L140

I got a smart card reader and a card with certs, so I'll test that this is working fine. One question, does your card require a pin to unlock the certs?

aafeijoo-suse:
Not looking good on my side...

```
Nov 03 10:55:02 localhost systemd[1]: Listening on pcscd.socket.
Nov 03 10:55:02 localhost systemd[1]: Started pcscd.service.
Nov 03 10:55:02 localhost pcscd[191]: 00000000 auth.c:92:IsClientAuthorized() polkit_authority_get_sync failed: Error initializing authority: Could not connect: No such file or directory
Nov 03 10:55:02 localhost pcscd[191]: 00000008 winscard_svc.c:336:ContextThread() Rejected unauthorized PC/SC client
Nov 03 10:55:02 localhost pcscd[191]: 00005539 auth.c:92:IsClientAuthorized() polkit_authority_get_sync failed: Error initializing authority: Could not connect: No such file or directory
Nov 03 10:55:02 localhost pcscd[191]: 00000003 winscard_svc.c:336:ContextThread() Rejected unauthorized PC/SC client
Nov 03 10:55:02 localhost systemd[1]: Starting dracut-cmdline-ask.service...
Nov 03 10:55:02 localhost systemd[1]: Finished dracut-cmdline-ask.service.
```

This means `pcscd` requires `polkit` and `dbus`... both are a bad idea.

innovara:
> The current upstream rule is to follow only other upstreams, so remove all the comments and add only what the OpenSC packaging says: `/usr/share/p11-kit/modules/opensc.module`

Great. I suspected there would be an upstream package path but I missed the obvious, checking their code.

> I got a smart card reader and a card with certs, so I'll test that this is working fine. One question, does your card require a pin to unlock the certs?

Yes, the slot I use requires a PIN, and when I tried with one that it doesn't, it didn't work. But be aware that you need one more change on dracut's side for it to work. I haven't sent anything yet because I need a bit of advice regarding `pcscd`

innovara:
> This means `pcscd` requires `polkit` and `dbus`... both are a bad idea.

You arrived to what it was going to be my 3rd change request but I need advice.

`pcscd` introduced `--disable-polkit` on 2.0.0 precisely to address this [1].

So the next and last change needed to make it all work is to change `pcscd.service`

` - ExecStart=/usr/sbin/pcscd --foreground --auto-exit
+  ExecStart=/usr/sbin/pcscd --foreground --auto-exit --disable-polkit`

But that change is obviously not going to be compatible with host using earlier versions of pcscd. However, dracut's pcsc module is broken anyway, so in some ways it doesn't really matter if we break it differently for pcscd < 2.0

What do you think?

I tested all this in Fedora 39 that uses pcsc-2.0.0 and it works as expected. pcscd is triggered at the right time, you get asked your PIN and it unlocks the LUKS encrypted volume.

[1] https://github.com/LudovicRousseau/PCSC/releases/tag/2.0.0

aafeijoo-suse:
> `pcscd` introduced `--disable-polkit` on 2.0.0 precisely to address this [1].
> 
> So the next and last change needed to make it all work is to change `pcscd.service`
> 
> ```
> - ExecStart=/usr/sbin/pcscd --foreground --auto-exit
> + ExecStart=/usr/sbin/pcscd --foreground --auto-exit --disable-polkit
> ```

IMO this is definitely the way to go, please update your commit or add another one with this change, so we can completely fix this module.

> But that change is obviously not going to be compatible with hosts using earlier versions of pcsc. However, dracut's pcsc module is broken anyway, so in some ways it doesn't really matter if we break it differently for pcsc < 2.0.0.
> 
> What do you think?

I'm not sure if this module has ever worked... even with polkit disabled I found some issues in systemd which I need to analyze further.

innovara:
Yeah, I spent some time troubleshooting and I got that error a few times. Usually, there is something not right with the URI, it's the wrong object, or it's not a RSA 2048 key. It also happened to me when I tried to use slot 9e which is supposed to work without PIN. Another mistake was to reuse `header.json` between a few tests. In the end I always made sure that I deleted the LUKS password slot, 1 in my case but be careful, with `sudo cryptsetup luksKillSlot /dev/xxx 1` as well as the token with `sudo cryptsetup token remove --token-id 0 /dev/xxx`. Then start again the enrol process.

One thing I did that helped me was to create a small LUKS partition and get that working with your card before getting dracut in the mix. Basically, unlocking it from the operating system with `sudo /lib/systemd/systemd-cryptsetup attach luks-UUID.... /dev/xxxx - pkcs11-uri=auto`. If you get that working then you know your token and systemd-cryptsetup can work together and move onto bootup.

innovara:
One further thought is that, if you are using a pkcs11 token, you need the change introduced by this PR for it to work at bootup: https://github.com/dracutdevs/dracut/pull/2540

LaszloGombos (approved):

innovara:
May I ask what are the next steps on a PR of this kind? Is it going to be merged on a future release? I'd love to see it on the next version of Fedora. Ironically, it's been backported to openSUSE!

aafeijoo-suse (approved):
> I'd love to see it on the next version of Fedora. Ironically, it's been backported to openSUSE!

Why is that an irony? I guess you can explicitly request that: https://src.fedoraproject.org/rpms/dracut/pull-requests

innovara:
> Why is that an irony?

`in a way that is interesting, strange, or funny because of being very different from what you would expect`

My expectation was that the fix would be applied to the upstream project first, and the downstream projects would receive it in due course at their own update cadence.

Of course, if the patches are good to go and other downstream projects can/want to use them before they are merged upstream, even better as they will reach at least a subset of the users of the project.

innovara:
> I guess you can explicitly request that: https://src.fedoraproject.org/rpms/dracut/pull-requests

Thanks for the tip. I did that and the patches will be used in future releases.

innovara:
This is still an issue in the latest release of Dracut.

PR #2551, opened 2023-11-08
fix(check_live_ram): increase /run tmpfs size if needed

### LINKED ISSUE #2550: rd.live.ram=1 may fill up /run if filesystem is too small
**Describe the bug**
Many distros now cap the total size of the `/run` filesystem at a relatively small percentage of RAM. For example, Fedora Security Live F38 on a VM with 4 GB of RAM caps `/run` at only 779.89 MB. This is not enough to contain the complete ISO when `rd.live.ram=1` is requested. When dracut writes the rootfs image to `/run/initramfs/squashed.img`, it fills up `/run` and fails the boot.

**Distribution used**
Fedora Security Live F38

**Dracut version**
059-2.fc38

**Init system**
systemd-253.2-1.fc38

**To Reproduce**
1. Obtain [`Fedora-Security-Live-x86_64-38-1.6.iso`](https://download.fedoraproject.org/pub/alt/releases/38/Labs/x86_64/iso/Fedora-Security-Live-x86_64-38-1.6.iso)
2. Configure a VM in GNOME Boxes or libvirt with 4000 MB of RAM
3. At the GRUB2 menu, choose a normal boot option like "Start Fedora-Security Live 38." Press `e` to edit the boot options. Add `rd.live.ram=1 rd.break rd.shell` to the end of the `linux` options line. Press `ctrl-x` to boot.
4. Observe that dracut fails to mount the rootfs because `mount` can't recognize the filesystem.
5. Run
   ```bash
   echo "something something something" >/run/initramfs/testfile
   ```
   and verify that a `no space left on device` error occurs. Also verify that `/run/initramfs/squashed.img` is smaller than it should be.
6. tmpfs size may be queried with
   ```bash
   cat /proc/self/mountinfo
   ```

**Expected behavior**
I believe that dracut dmsquash-live and livenet should either:
* Create a tmpfs just for `squashed.img` that is large enough to contain it; OR
* Resize the existing tmpfs for `/run` so that it is large enough to contain `squashed.img`.

Without changes to dracut, distros that want to support `rd.live.ram` will need to ensure that `/run/initramfs` is large enough.

**Additional context**
Requesting `rd.live.ram=1` on a system with not enough memory may put the system under severe memory pressure. #2344 adds the capability to check for this, and we probably don't need to second-guess this check.

The [`tmpfs(5)`](https://man7.org/linux/man-pages/man5/tmpfs.5.html) default size is 50% of memory. The default size should be large enough on most modern hardware to contain a live image of "reasonable" size. It is clear that some distos opt for a lower `size` limit for `/run`.

See also downstream bug [RH-2035641](https://bugzilla.redhat.com/show_bug.cgi?id=2035641)

FGrose:
fix(check_live_ram): increase /run tmpfs size, if needed
    
    Check the size and available space in /run to enlarge it,
      if needed.
    Introduce the check_meminfo() function to dracut-lib.sh
      to replace a less versatile sed call.
    Also, use local variables,
      use parameter expansion to assign a default value,
      use test exit logic directly, where possible,
      tweak comments,
      and move the check_live_ram() function to img-lib.sh.
### Checklist

Fixes #2550

### DIFF
--- a/modules.d/99base/dracut-lib.sh
+++ b/modules.d/99base/dracut-lib.sh
@@ -1130,6 +1130,18 @@ show_memstats() {
     esac
 }
 
+# parameter: <memory_name:>  example: MemTotal:
+# Check /proc/meminfo
+# echo the field value, if present.
+check_meminfo() {
+    local - m sz
+    set +x
+    while read -r m sz _ || [ "$m" ]; do
+        [ "$m" = "$1" ] && echo "$sz" && return 0
+    done < /proc/meminfo
+    return 1
+}
+
 remove_hostonly_files() {
     rm -fr /etc/cmdline /etc/cmdline.d/*.conf "$hookdir/initqueue/finished"
     if [ -f /lib/dracut/hostonly-files ]; then
@@ -1150,29 +1162,4 @@ load_fstype() {
         [ "${fs:-$d}" = "$_fs" ] && return 0
     done < /proc/filesystems
     modprobe "$1"
-}
-
-# parameter: size of live image
-# calls emergency shell if ram size is too small for the image
-check_live_ram() {
-    minmem=$(getarg rd.minmem)
-    minmem=${minmem:-1024}
-    imgsize=$1
-    memsize=$(($(sed -n 's/MemTotal: *\([[:digit:]]*\).*/\1/p' /proc/meminfo) / 1024))
-
-    if [ -z "$imgsize" ]; then
-        warn "Image size could not be determined"
-        return 0
-    fi
-
-    if [ $((memsize - imgsize)) -lt "$minmem" ]; then
-        sed -i "N;/and attach it to a bug report./s/echo$/echo\n\
-         echo \n\
-         echo 'Warning!!!'\n\
-         echo 'The memory size of your system is too small for this live image.'\n\
-         echo 'Expect killed processes due to out of memory conditions.'\n\
-         echo \n/" /usr/bin/dracut-emergency
-
-        emergency_shell
-    fi
 }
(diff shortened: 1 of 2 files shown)


LaszloGombos (approved):

Conan-Kudo (approved):

cbs228:
This PR is needed for #2550 and is still waiting on review.

Athwale:
Any progress on this please, can someone review this and approve it? I am hitting this problem on f39.
Thanks.

Conan-Kudo:
I approved it, but I haven't had merge permissions for a while. ☹️

Athwale:
@haraldh @johannbg @tblume Can you please have a look at this? Thank you.

Athwale:
This is a related issue also connected to the low /run size
https://github.com/dracutdevs/dracut/issues/2550
Particularly the ```rd.writable.fsimg=1``` copies rootfs.img into RAM and this image is typically even larger than the squashfs.img.

Athwale:
Still a problem.

jozzsi:
@Athwale most Linux distributions are now packaging https://github.com/dracut-ng/dracut-ng . Which Linux distributions are you using where you can still reproduce this problem ?

Athwale:
I am using fedora 41.
It looks like the dracut in repositories is: dracut.x86_64: Initramfs generator using udev 
Is the new package called dracut-ng?
I will look if F42 is using it.
Thanks.

Athwale:
Actually it looks like the fix is already in F41, in that case I will update my workarounds and this is no longer a problem.
Thanks for letting me know.

PR #2560, opened 2023-11-16
feat(resume): do not attempt to install systemd-hibernate-resume@.service

aafeijoo-suse:
This service is no longer shipped since systemd-v254 (https://github.com/systemd/systemd/commit/760e99bb).

## Checklist


Flowdalic:
Is that really required, given that `inst_multiple` is invoked with `-o` (optional, which makes the command *not* fail if a given file does not exist)? It seems that this should probably be done conditionally on the used systemd version. Or simply try to install that file for a a while.

aafeijoo-suse:
> Is that really required, given that `inst_multiple` is invoked with `-o` (optional, which makes the command _not_ fail if a given file does not exist)? It seems that this should probably be done conditionally on the used systemd version. Or simply try to install that file for a a while.

AFAIK, the current upstream policy is to support only the latest upstream version of other dependencies.

PR #2563, opened 2023-11-22
multipath cleanup

mwilck:
- feat(multipath): replace multipathd.service with a drop-in
- feat(multipath): don't add rd.driver.pre= directives to cmdline

## Changes

Simplify dracut's multipath module by using the upstream systemd
service file, and by not enforcing kernel module loading using
"rd.driver.pre".

This allows distributions to configure driver loading in the
initrd in a generic way which is consistent with the overall
configuration of the distribution, and transparent to the user.
Also, it makes it possible to avoid loading SCSI modules on 
systems where no SCSI hardware is installed.

This change may require adaptations in the way distributions
handle booting from dm-multipath. Distributions will have to
take care that the necessary modules are loaded if necessary.
Most distributions will be doing this anyway using
modules-load.d or similar mechanisms. dracut already pulls in 
modules-load.d and modprobe.d configuration files that
make this work.

## Checklist

Fixes #


mwilck:
Adding @aafeijoo-suse and @bmarzins.

bmarzins:
This looks fine to me.

mwilck:
This is currently causing regressions. I probably can't just drop the "rd.driver.pre" without making sure all test distros handle this cleanly. OTOH, this will reveal a couple of subtleties that we should fix anyway. I'm on it.

mwilck:
> Thanks @mwilck
> 
> > not enforcing kernel module loading using "rd.driver.pre".
> 
> This is a welcomed change, I also did notice that the current logic has unintended consequences.
> 
> As an example Test-01 and Test-02 tests are passing on openSUSE (and all other distributions on the CI) on master, but in the try run of this PR it seems to fail.

Right. That already revealed a multipath-tools bug (pre-0.9.7, but 0.9.7 has it for systemd versions < 245, too). `multipathd.service` is lacking a dependency on `systemd-modules-load.service`.

Good that I finally learn the dracut test suite, very useful stuff.

mwilck:
> > (multipathd) died (code=exited, status=1/FAILURE)

Same reason, probably. The `dm-multipath` kernel module is loaded after multipathd startup, which then fails. An additional dependency would fix it. For the time being, I could add that dependency here, but it must ultimately go into multipath-tools upstream.

Btw I noticed that the stdin redirection from `/dev/null` for qemu in the [test scripts](https://github.com/dracutdevs/dracut/blob/4971f443726360216a4ef3ba8baea258a1cd0f3b/test/test-functions#L163) makes it impossible to actually enter emergency mode in the test VM (when `DEBUGFAIL`) is set. For the time being I just removed this redirection. Is there a more elegant trick to do this that I'm unaware of?

PR #2576, opened 2023-11-30
Make rd.udev.debug work in systemd-based initrds

dtardon:
This makes `rd.udev.debug` work in systemd-based initrds too.

## Changes

A systemd generator is added that sets log level of `systemd-udevd.service` to debug if `rd.udev.debug` is present on the kernel command line.

## Note

`rd.udev.info` is ignored as that's the default log level of udevd. The existence of this option makes some sense in the legacy `init.sh`, as it sets the log level of udevd to `err` by default, but I don't think we want to go this way...

## Checklist


dtardon:
Oops, sorry. Apparently I wasn't fully awake yet when I tried this in the morning... The option _is_ handled (in `modules.d/98dracut-systemd/dracut-cmdline.sh`), I just didn't find it.

PR #2578, opened 2023-12-01
refactor(nfs): remove leftover code

aafeijoo-suse:
- Leftover `grep`s used when `chown` was perfomed after them.
- `$initdir/var/lib/rpcbind` already created with `mkdir -m 0770`, it is not necessary to `chmod 770` again.

Follow-up to 5ebf48d2baa7d7cbad7cfc663319f8ca80f76398

## Checklist


aafeijoo-suse:
> I am not aware of a standard that would require nfs user. Different distributions use different usernames, which is why testing the username is useful.
> 
> Better solution would be to determine the username instead of dracut hardcoding it.

This comment is unrelated to what this PR is about.

PR #2580, opened 2023-12-07
fix(configure): misleading error if C compiler is not installed

aafeijoo-suse:
While preparing a new system for development, `./configure` reaches a point where it fails with:

```
$ ./configure
dracut needs fts development files.
```

After installing the fts library, `./configure` keeps throwing the same error:

```
$ rpm -qf /usr/include/fts.h
glibc-devel-2.38-6.1.x86_64
$ ./configure
dracut needs fts development files.
```

The problem is `${CC} $CFLAGS $LDFLAGS conftest.c` can also fail if the compiler referenced by `$CC` is not installed.

## Checklist

### DIFF
--- a/configure
+++ b/configure
@@ -60,6 +60,11 @@ if ! ${PKG_CONFIG} --exists --print-errors " libkmod >= 23 "; then
     exit 1
 fi
 
+if ! command -v "${CC}" > /dev/null; then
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

PR #2586, opened 2023-12-12
fix(systemd-255): handle systemd-pcrphase -> systemd-pcrextend

### LINKED ISSUE #2583: With systemd 255, systemd-pcrphase is renamed to systemd-pcrextend
**Describe the bug**
With the systemd version 255, the systemd-pcrphase binary was renamed to systemd-pcrextend, breaking dracut initrds requiring it.

**Distribution used**
Arch linux

**Dracut version**
dracut 059

**Init system**
systemd

ferringb:
In systemd 255 systemd-pcrphase was renamed to systemd-pcrextend; there was no other change to it, just the rename.

Unit scripts are still named systemd-pcrphase, just the underlying binary got renamed.  Thus adapt the module
to check for either binary.

## Changes

Adds awareness of systemd-pcrextend binary allowing systemd-pcrphase module to work with systemd-255.

## Checklist
- []  I am providing new code and test(s) for it
// copy/rename of existing code, and existing lacks tests...

Fixes #2583

### DIFF
--- a/modules.d/01systemd-pcrphase/module-setup.sh
+++ b/modules.d/01systemd-pcrphase/module-setup.sh
@@ -6,7 +6,11 @@
 check() {
 
     # If the binary(s) requirements are not fulfilled the module can't be installed.
-    require_binaries "$systemdutildir"/systemd-pcrphase || return 1
+    # systemd-255 renamed the binary, check for old and new location.
+    if ! require_binaries "$systemdutildir"/systemd-pcrphase && \
+       ! require_binaries "$systemdutildir"/systemd-pcrextend; then
+       return 1
+    fi
 
     # Return 255 to only include the module, if another module requires it.
     return 255
@@ -28,6 +32,7 @@ install() {
 
     inst_multiple -o \
         "$systemdutildir"/systemd-pcrphase \
+        "$systemdutildir"/systemd-pcrextend \
         "$systemdsystemunitdir"/systemd-pcrphase-initrd.service \
         "$systemdsystemunitdir/systemd-pcrphase-initrd.service.d/*.conf" \
         "$systemdsystemunitdir"/initrd.target.wants/systemd-pcrphase-initrd.service


ferringb:
Pardon the noise; my testing of my previous patch was lackluster; this was verified properly.

modules.d/01systemd-pcrphase/module-setup.sh
```diff
@@ -26,8 +30,11 @@ depends() {
 # Install the required file(s) and directories for the module in the initramfs.
 install() {
 
+    binary="$systemdutildir/systemd-pcrphase"
```
dtardon, line 33: You can just put both binaries to the `inst_multiple` call. It uses `-o` (optional), hence non-existing files are ignored.

ferringb: Is there any better form for that `require_binaries` call?  as to the `inst_multiple`- and I'm probably being anal here- but that install should probably only allow optional for the `*.conf`.  The `*.service` definitely need to exist for this module to work.

Where I'm going with that is thus: how do I make this *fail* if one of those two binaries isn't found?  That problem can only occur if there is a time-of-check-time-of-use, but it's probably best to note the "I'm probably being anal here" comment from above.

dtardon: > The `*.service` definitely need to exist for this module to work.

Well, you could change this to use `inst_multiple` for stuff that *must* be present and `inst_multiple -o` for stuff that's *may* be present. But IMHO it's not strictly needed. `dracut` assumes that the stuff that's supposed to work in the initrd being created does work in the "source" system, which couldn't be true if some required files were missing.
 
> Where I'm going with that is thus: how do I make this _fail_ if one of those two binaries isn't found? 

You already do, in `check()`. Nothing else is needed.

ferringb: Pushing a rewrite w/ the requested changes; this should be good to go w/ what you've asked for.  I'm not splitting up the inst_multiple for the reasons you mentioned, and lazyness on my part. ;)

ferringb:
@dtardon what you asked for is addressed, and this is still horked in mainline.  Is there anything blocking this being merged?

dtardon:
> @dtardon what you asked for is addressed, and this is still horked in mainline. Is there anything blocking this being merged?

That doesn't depend on me... I'm just a contributor; I don't have the rights to merge PRs.

ferringb:
@dtardon pardon, my bad.

@haraldh or @johannbg , both of you are listed as code owners.  Any chance you can address this?

johannbg:
> @dtardon pardon, my bad.
> 
> @haraldh or @johannbg , both of you are listed as code owners. Any chance you can address this?

Interesting there are quite few dragons there in systemd v255 release for downstream enterprise distro's and downstream that support multiple init systems anyway there is no such thing as "just a rename" when upstream changes a executable name. 

The scope of this changes requires
A new module (  systemd-pcrextend )
Existing module being renamed and turned into a meta module to provide a clean upgrade/migration path for downstream. 

A simple optional inclusion will not be merged to address this upstream systemd change. 
It is more than it's made to appear to be as is always the case when upstream "rename" things...

dtardon:
> The scope of this changes requires A new module ( systemd-pcrextend ) Existing module being renamed and turned into a meta module to provide a clean upgrade/migration path for downstream.

Sorry, but what?! Meta modules are useful for alternative impls. of a concept, but this isn't that. It's a simple rename of an _internal_ binary. The _service_ name remains the same, which is what really matters, because it's user-visible. There will be just one of the binaries on the system at any point, hence a simple optional inclusion already provides all the migration path that's needed. Anything else is an overkill. (Even if a system--and consequently an initrd generated on it--ended up containing both binaries somehow, it wouldn't matter anyway. The old `systemd-pcrphase-initrd.service` calls `systemd-pcrphase`, while the v255 one calls `systemd-pcrextend`, hence there'd still be no problem/confusion.)

> A simple optional inclusion will not be merged to address this upstream systemd change. It is more than it's made to appear to be as is always the case when upstream "rename" things...

No, it isn't. This is an _internal_ binary, not supposed to be called by users. Hence this rename is transparent to users. It only affects `dracut` and possibly other initrd/image generators that need to poke at package internals.

lnykryn:
I agree with @dtardon here. For reference: https://github.com/systemd/systemd/commit/32295fa08f35d8c0b86c5faa03d7c818d3a9353f

lnykryn (approved):

johannbg:
> Relevant section of the dracut wiki - https://github.com/dracutdevs/dracut/wiki/Dracut-development#compatibility
> 
> > The compatibility promise is primary towards end users and not towards Linux distributions packaging dracut. Distribution packaging will need changes time-to-time between dracut releases (e.g. as files gets added and removed).

Looking at that wikipage it's quite evident that I need to start restricting who can create and edit wiki pages.

johannbg:
> I agree with @dtardon here. For reference: [systemd/systemd@32295fa](https://github.com/systemd/systemd/commit/32295fa08f35d8c0b86c5faa03d7c818d3a9353f)

Modules are based upon the upstream configuration options so correct dependency chain can be kept downstream  ( which is what this module systemd-pcrphase was based upon ) If those configuration options change upstream ( which most certainly did with this change ) those changes will be reflected here as well.

So to be perfectly clear on the matter this wont get merged in any other form than in a new module ( systemd-pcrextend ) and the existing module being renamed and turned into a meta module to provide a clean upgrade/migration path for downstream so this PR can be update to reflect that or it can be closed.

ferringb:
> > I agree with @dtardon here. For reference: [systemd/systemd@32295fa](https://github.com/systemd/systemd/commit/32295fa08f35d8c0b86c5faa03d7c818d3a9353f)
> 
> Modules are based upon the upstream configuration options so correct dependency chain can be kept downstream ( which is what this module systemd-pcrphase was based upon ) If those configuration options change upstream ( which most certainly did with this change ) those changes will be reflected here as well.
> 
> So to be perfectly clear on the matter this wont get merged in any other form than in a new module ( systemd-pcrextend ) and the existing module being renamed and turned into a meta module to provide a clean upgrade/migration path for downstream so this PR can be update to reflect that or it can be closed.

Closing this out.  Feel free to cut your own fix.

dtardon:
> > I agree with @dtardon here. For reference: [systemd/systemd@32295fa](https://github.com/systemd/systemd/commit/32295fa08f35d8c0b86c5faa03d7c818d3a9353f)
> 
> Modules are based upon the upstream configuration options so correct dependency chain can be kept downstream ( which is what this module systemd-pcrphase was based upon ) If those configuration options change upstream ( which most certainly did with this change ) those changes will be reflected here as well.
> 
> So to be perfectly clear on the matter this wont get merged in any other form than in a new module ( systemd-pcrextend ) and the existing module being renamed and turned into a meta module to provide a clean upgrade/migration path for downstream so this PR can be update to reflect that or it can be closed.

Have you even _read_ what I had written?

PR #2591, opened 2023-12-18
fix(dracut): move hooks directory from /usr/lib to /var/lib

### LINKED ISSUE #2588: hookdir can't be in /usr if used for writing
https://github.com/dracutdevs/dracut/blob/master/modules.d/99base/dracut-lib.sh#L402 sets `hookdir=/lib/dracut/hooks`. Later the code writes there. With https://github.com/systemd/systemd/pull/30255 systemd makes `/usr` read-only even in initrd though. So a dracut generated initrd fails to boot with errors like

```
Dec 14 15:38:23 localhost dracut-cmdline[257]: //lib/dracut/hooks/cmdline/00-parse-root.sh: line 28: /lib/dracut/hooks/initqueue/finished/devexists-\x2fdev\x2fdisk\x2fby-uuid\x2f8b3c35f1-610b-48d8-ae73-aecdaa924731.sh: Read-only file system
...
Dec 14 15:38:24 localhost dracut-initqueue[374]: rm: cannot remove '/lib/dracut/hooks/initqueue/finished/devexists-\x2fdev\x2fdisk\x2fby-path\x2fpci-0000:00:03.0-part.sh': Read-only file system
```

Seen in openSUSE tumbleweed with systemd from git

aafeijoo-suse:
Since https://github.com/systemd/systemd/commit/ffc1ec73, `/usr` is mounted as read-only in the initramfs by default.

## Checklist

Fixes #2588

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -183,42 +183,42 @@ You are encouraged to provide a README that describes what the module is for.
 
 init has the following hook points to inject scripts:
 
-`/lib/dracut/hooks/cmdline/*.sh`
+`/var/lib/dracut/hooks/cmdline/*.sh`
    scripts for command line parsing
 
-`/lib/dracut/hooks/pre-udev/*.sh`
+`/var/lib/dracut/hooks/pre-udev/*.sh`
    scripts to run before udev is started
 
-`/lib/dracut/hooks/pre-trigger/*.sh`
+`/var/lib/dracut/hooks/pre-trigger/*.sh`
    scripts to run before the main udev trigger is pulled
 
-`/lib/dracut/hooks/initqueue/*.sh`
+`/var/lib/dracut/hooks/initqueue/*.sh`
    runs in parallel to the udev trigger
    Udev events can add scripts here with /sbin/initqueue.
    If /sbin/initqueue is called with the "--onetime" option, the script
    will be removed after it was run.
-   If /lib/dracut/hooks/initqueue/work is created and udev >= 143 then
+   If /var/lib/dracut/hooks/initqueue/work is created and udev >= 143 then
    this loop can process the jobs in parallel to the udevtrigger.
    If the udev queue is empty and no root device is found or no root
    filesystem was mounted, the user will be dropped to a shell after
    a timeout.
    Scripts can remove themselves from the initqueue by "rm $job".
 
-`/lib/dracut/hooks/pre-mount/*.sh`
+`/var/lib/dracut/hooks/pre-mount/*.sh`
    scripts to run before the root filesystem is mounted
    Network filesystems like NFS that do not use device files are an
    exception. Root can be mounted already at this point.
 
-`/lib/dracut/hooks/mount/*.sh`
+`/var/lib/dracut/hooks/mount/*.sh`
    scripts to mount the root filesystem
    If the udev queue is empty and no root device is found or no root
    filesystem was mounted, the user will be dropped to a shell after
    a timeout.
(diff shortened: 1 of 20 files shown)


aafeijoo-suse:
@lnussel I'm not sure if `/var` is a good long-term option, see https://github.com/systemd/systemd/blob/5a4631bd8fde4122bf6f25872df07229a6b3ed74/TODO#L966-L970

A specific directory outside the FHS (such as `/.dracut`) would not suffer these types of changes.

lnussel:
That's talking about `/sysroot/var` IMO

aafeijoo-suse:
> That's talking about `/sysroot/var` IMO

Right, I was wrong. Thanks.

johannbg:
@aafeijoo-suse did you look into how this change would affect non systemd based distributions and which locations those users/distro's might want to use instead since the scope of this change is not limited to systemd based distributions.

aafeijoo-suse:
> @aafeijoo-suse did you look into how this change would affect non systemd based distributions and which locations those users/distro's might want to use instead since the scope of this change is not limited to systemd based distributions.

No, I didn't. That's why upstream PRs are public, so people from other environments can comment whether this change suits their systems. What I know is that initrds generated by dracut don't work with the upcoming systemd-v256, but we are not in a hurry, the new place to save the hooks is something to discuss here.

aafeijoo-suse:
> > @aafeijoo-suse did you look into how this change would affect non systemd based distributions and which locations those users/distro's might want to use instead since the scope of this change is not limited to systemd based distributions.
> 
> No, I didn't. That's why upstream PRs are public, so people from other environments can comment whether this change suits their systems. What I know is that initrds generated by dracut don't work with the upcoming systemd-v256, but we are not in a hurry, the new place to save the hooks is something to discuss here.

Bumping this topic again, systemd will tag v256-rc1 soon and dracut is broken without this fix.

@thesamesam @floppym @AndrewAmmerlaan @ahesford ping, I don't know if `/var` is a good idea on non-systemd systems, or an out-of-tree directory like `/.dracut` is better.

ahesford:
I'll tag @zdykstra for further comment as the maintainer of Void's dracut package and fellow ZFSBootMenu author.

On the surface, this is probably irrelevant to ZFSBootMenu, because we should be using dracut's hook installation functions to put our stuff where it needs to go. We don't work with systemd in the initramfs and just exec into our own event loop anyway.

In the meantime, I'll grumble quietly, but loudly enough to say that this is yet another example of a questionable decision by systemd that produces no discernible benefit (why is `/usr` even a separate filesystem in the initramfs?!?!?) and makes the quest to incorporate systemd into initramfs images seem even more quixotic.

LaszloGombos:
> Regardless of the new writable location, can dracut maintain a link from /lib/dracut/hooks to the new directory location for improved compatibility ?

I do not see how we can maintain [Compatibility](https://github.com/dracutdevs/dracut/wiki/Dracut-development#compatibility) without having a link from /lib/dracut/hooks.

Random (perhaps not the best) example that I think would break by this PR (from openshift) - https://github.com/openshift/installer/blob/master/pkg/asset/agent/image/agentartifacts.go#L164

zdykstra:
There's nothing I can see on the Void side that will break because of this change.  I seriously question the need for making `/usr` read-only in an initramfs, but that ship seems to have already sailed.

LaszloGombos (approved):
LGTM

Thanks for adding a link from /lib/dracut/hooks.

PR #2596, opened 2023-12-20
feat(dracut-initramfs-restore.sh): get compression method from image

### LINKED ISSUE #1995: Prefer XBOOTLDR to ESP when installing unified kernel images.
When installing unified kernel images images (uefi=yes) dracut currently searches `/efi`, `/boot/efi` and `/boot` to determine where the ESP is mounted. I suggest that the install location be instead queried by `bootctl -x` in order to also support a possible XBOOTLDR partition.

aafeijoo-suse:
Instead of trying every possible compression method until the right one is found, parse the header of the initramfs image to get the right one and use it without hiding possible decompression errors, using the same logic that is implemented in `lsinitrd.sh`.

For that, refactor `lsinitrd.sh` moving common code to `dracut-functions.sh`. In particular:
- `get_machine_id`
- `get_default_initramfs_image [<kernel_version>]`
- `has_early_microcode <initramfs_image>`
- `get_decompression_command <initramfs_image_header>`

Also, added a new `get_dollar_boot` function to get the `$BOOT` partition placeholder, as defined in the [Boot Loader Specification](https://uapi-group.org/specifications/specs/boot_loader_specification). If `bootctl` is available, use it to support a possible XBOOTLDR partition, otherwise, check only the ESP, as before.

Also refactored `dracut.sh` using these new functions to get the `outfile` path.

## Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1082,11 +1082,56 @@ pe_get_image_base() {
     echo "$((16#$base_image))"
 }
 
-# get_machine_id
+# get_dollar_boot
+# $BOOT is the primary place to put boot menu entry resources into
+# see https://uapi-group.org/specifications/specs/boot_loader_specification
+get_dollar_boot() {
+    local _root_arg
+    local _esp
+    local _xbootldr
+    local _dollar_boot
+
+    if type -P bootctl &> /dev/null; then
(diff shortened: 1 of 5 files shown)


aafeijoo-suse:
@tblume JFYI, this is what we discussed internally this week.

PR #2600, opened 2024-01-02
fix(github): update format of labeler

aafeijoo-suse:
Since we are using `actions/labeler@main`, its format changed, displaying the following error with every new PR:

```
Run actions/labeler@main
The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
Error: Error: found unexpected type for label 'repository' (should be array of config options)
Error: found unexpected type for label 'repository' (should be array of config options)
```

## Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,292 +1,543 @@
 repository:
-  - ./*
+  - changed-files:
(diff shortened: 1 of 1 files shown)


johannbg (approved):
LGTM

PR #2604, opened 2024-01-09
fix(dmsquash-live): use the overlay size with thin provisioning

gregory-lee-bartholomew:
Using the overlay size in the device mapper table allows the filesystem in the base image to be extended at runtime if the overlay is larger than the base image (i.e. this makes `resize2fs /dev/mapper/live-rw` work when `rd.live.overlay.thin` is used).

This pull request changes...

## Changes

This patch changes the upper-bound of the `live-rw` device when the user requests to create a thin-provisioned overlay by specifying the `rd.live.overlay.thin` kernel parameter. The size of the overlay (typically 32GiB, but it can be specified by the user with a kernel parameter) is used rather than the size of the base image.

## Checklist

Fixes #

Users of Fedora Linux have been reporting this bug here: https://discussion.fedoraproject.org/t/fedora-liveos-root-system-and-available-ram/82531/1

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -296,7 +296,7 @@ do_live_overlay() {
         dmsetup message /dev/mapper/live-overlay-pool 0 "create_thin 0"
 
         # Create a snapshot of the base image
-        echo 0 "$sz" thin /dev/mapper/live-overlay-pool 0 "$base" | dmsetup create live-rw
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

gregory-lee-bartholomew:
Thanks @Conan-Kudo. I'll close this MR since the change has been merged in another MR. 🙂

gregory-lee-bartholomew:
I think I may have closed this prematurely. I see now that what was merged earlier was to "dracut-ng", not "dracut". I'll reopen this against "dracut" and I'll fix the commit message to be consistent with the one that was merged earlier. Sorry about the confusion.

PR #2605, opened 2024-01-13
Prevent assembly of unwanted MD RAID arrays

chriscray:
The mdraid module modifies the 65-md-incremental-imsm.rules  during the pre-trigger phase.  Before modification, this rule will start any RAID; after modification, this rule will start only the RAID passed in with rd.md.uuid.  Because this modification occurs while udevd is running, unwanted MD arrays may be assembled if any of their component drives are enumerated before modification.

To prevent unwanted MD devices from being assembled, the modification of 65-md-incremental-imsm.rules should occur before udev is started.

This pull request changes 90mdraid/module-setup.sh so that 90mdraid/parse-md.sh is run in the pre-udev stage instead of the pre-trigger phase.

## Checklist

Fixes #

### DIFF
--- a/modules.d/90mdraid/module-setup.sh
+++ b/modules.d/90mdraid/module-setup.sh
@@ -114,7 +114,7 @@ install() {
     fi
 
     inst_hook pre-udev 30 "$moddir/mdmon-pre-udev.sh"
-    inst_hook pre-trigger 30 "$moddir/parse-md.sh"
+    inst_hook pre-udev 40 "$moddir/parse-md.sh"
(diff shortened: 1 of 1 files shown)


chriscray:
I'm closing this.  The solution I suggested breaks all parameters that depend on udev being up when parse-md.sh is run since script sets global udev property.  I've worked around that locally by modifying the 65-md-incremental-imsm.rules udev rule so that it's configured pre-udev in all cases, but this entails more changes than I'm comfortable pushing....

PR #2607, opened 2024-01-16
fix(dracut): correct regression with multiple `rd.break=` options

aafeijoo-suse:
Since the introduction of the `dracut-util` binary to parse kernel command line arguments, if the user inputs multiple `rd.break=` options, dracut only stops the boot process at the last one.

```
[    0.985362] localhost dracut-cmdline[245]: /bin/dracut-cmdline@18(): info 'Using kernel command line parameters:' ' ... rd.debug rd.break=cmdline rd.break=pre-pivot'
...
[    1.044979] localhost dracut-cmdline[245]: /bin/dracut-cmdline@48(): getarg rd.break=cmdline -d rdbreak=cmdline
[    1.044979] localhost dracut-cmdline[245]: /lib/dracut-lib.sh@155(getarg): debug_off
[    1.044979] localhost dracut-cmdline[245]: /lib/dracut-lib.sh@23(debug_off): set +x
[    1.044979] localhost dracut-cmdline[245]: /lib/dracut-lib.sh@218(getarg): return 1
```

For options that can be specified multiple times, `getargs` should be used instead.

```
master> export CMDLINE="rd.break=cmdline rd.break=pre-udev rd.break=pre-pivot"
master> ./dracut-getarg rd.break=cmdline
master> echo $?
1
master> ./dracut-getarg rd.break=pre-udev
master> echo $?
1
master> ./dracut-getarg rd.break=pre-pivot
master> echo $?
0
master> ./dracut-getargs rd.break=cmdline
cmdline
master> echo $?
0
master> ./dracut-getargs rd.break=pre-udev
pre-udev
master> echo $?
0
master> ./dracut-getargs rd.break=pre-pivot
pre-pivot
master> echo $?
0
```

## Checklist

Fixes 501d82f79675a6bf9b37e8250152515863a80236

### DIFF
--- a/modules.d/98dracut-systemd/dracut-cmdline.sh
+++ b/modules.d/98dracut-systemd/dracut-cmdline.sh
@@ -45,7 +45,7 @@ export fstype
 
 make_trace_mem "hook cmdline" '1+:mem' '1+:iomem' '3+:slab'
 # run scriptlets to parse the command line
-getarg 'rd.break=cmdline' -d 'rdbreak=cmdline' && emergency_shell -n cmdline "Break before cmdline"
+getargs 'rd.break=cmdline' -d 'rdbreak=cmdline' && emergency_shell -n cmdline "Break before cmdline"
 source_hook cmdline
 
 [ -f /lib/dracut/parse-resume.sh ] && . /lib/dracut/parse-resume.sh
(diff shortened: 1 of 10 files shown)


aafeijoo-suse:
> > if the user inputs multiple rd.break= options, dracut only stops the boot process at the last one.
> 
> The dracut documentation seems to suggest that dracut works as intended and this is not a bug.

Then why the following excerpt of the documentation states that we can stop at `pre-shutdown` **and** `shutdown`?

https://github.com/dracutdevs/dracut/blob/4980bad34775da715a2639b736cba5e65a8a2604/man/dracut.usage.asc?plain=1#L513

This used to work at least until `dracut-049` (SLE15-SP3), so it's a clear regression.

PR #2610, opened 2024-01-18
fix(systemd-255): handle systemd-pcr{phase -> extend} rename

### LINKED ISSUE #2583: With systemd 255, systemd-pcrphase is renamed to systemd-pcrextend

LaszloGombos:
The binary systemd-pcrphase was renamed to systemd-pcrextend in systemd 255, but the backing units were all left named systemd-pcrphase.

Fixes: #2583

### DIFF
--- a/modules.d/01systemd-pcrphase/module-setup.sh
+++ b/modules.d/01systemd-pcrphase/module-setup.sh
@@ -6,7 +6,11 @@
 check() {
 
     # If the binary(s) requirements are not fulfilled the module can't be installed.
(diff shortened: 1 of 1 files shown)


freswa:
This is still an issue

PR #2621, opened 2024-02-10
fix(network-legacy): make sure "$hookdir"/initqueue/finished exists

### LINKED ISSUE #2620: network-legacy gen-rules.sh failes due to missing initqueue/finished directory
**Describe the bug**

Attached are the init.log and build logs.

[init-sanitized.log](https://github.com/dracutdevs/dracut/files/14224195/init-sanitized.log)

[dracut.log](https://github.com/dracutdevs/dracut/files/14224193/dracut.log)

I've been trying to build [zfsbootmenu with remote ssh access](https://docs.zfsbootmenu.org/en/v2.2.x/guides/general/remote-access.html) in a reproducible way using a debian:bookworm docker based workflow.

zfsbootmenu prohibits systemd dracut modules, so I've enabled network-legacy instead. hostonly is disabled/ set to `no`. zfsbootmenu calls dracut (see the logs), and finally produces a UEFI stub image which I test using qemu. I boot the image with ip=dhcp and rd.neednet=1, as well as some debug flags to increase dracut logging.

The image builds fine, however when booting the `eth0` interface stays DOWN and no IP is assigned via dhcp.

I can then manually bring up the interface just fine by calling `dhclient eth0`.

Looking `/run/initramfs/init.log` I see 

```
//lib/dracut/hooks/pre-udev/60-net-genrules.sh: line 120: /lib/dracut/hooks/initqueue/finished/wait-network.sh: No such file or directory
```

**Fix :question: **

If I add `mkdir -p "$hookdir"/initqueue/finished` just before [that line](https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-legacy/net-genrules.sh#L120), similar to [network-manager](https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/nm-lib.sh#L23) and [connman](https://github.com/dracutdevs/dracut/blob/master/modules.d/35connman/cm-lib.sh#L7) modules and rebuild the image, the interface comes up automatically at boot, as expected.

**Distribution used**

Docker version of debian bookworm.

**Dracut version**

059-4

**Init system**

? systemd when generating dracut, but systemd is omitted in dracut's conf files.

**To Reproduce**

I can share my dockerfile is necessary, but, without being a dracut expert or experienced user *at all*, I wonder how network-legacy can work if neither connman or network-manager modules are used and the finished dir is never created?

**Expected behavior**

I'd have expected the eth0 interface to be brought up properly by network-legacy.

LaszloGombos:
Just like in network-manager and connman modules, make sure "$hookdir"/initqueue/finished exists in network-legacy as well.

Fixes #2620

### DIFF
--- a/modules.d/35network-legacy/net-genrules.sh
+++ b/modules.d/35network-legacy/net-genrules.sh
@@ -116,6 +116,7 @@ command -v fix_bootif > /dev/null || . /lib/net-lib.sh
         # if you change the name of "91-default-net.rules", also change modules.d/80cms/cmssetup.sh
         echo "$cond, $runcmd" > /etc/udev/rules.d/91-default-net.rules
         if [ "$NEEDNET" = "1" ]; then
+            mkdir -p "$hookdir"/initqueue/finished
             # shellcheck disable=SC2016
             echo 'for i in /tmp/net.*.did-setup; do [ -f "$i" ]  && exit 0; done; exit 1' > "$hookdir"/initqueue/finished/wait-network.sh
         fi


ahesford:
This seems entirely appropriate to me.

PR #2622, opened 2024-02-10
fix(i18n): silence spurious setfont stderr warning

LaszloGombos:
This error has been here for years, but unreported by setfont. Not actionable.

CC @zdykstra


zdykstra:
Thanks - I'll drop Void's patch once this lands in a release.

zdykstra:
This isn't stale. It's just not being reviewed and merged.

PR #2626, opened 2024-02-18
fix(dracut.sh): recognize kernel file in /boot named vmlinux too

Icenowy:
This pull request adds /boot/vmlinux-${kernelver} to recognized kernel file when trying to detect the output path of the initramfs file, which makes it broken on AOSC OS loongarch64 (but coincidentally worked before /efi mount point detection is implemented).

## Checklist


LaszloGombos:
@Icenowy  Seem loongarch64 is still experimental https://wiki.aosc.io/aosc-os/information/arch-support/ ?

Would this impact other distributions that you know of ?

Icenowy:
The English document is out of sync, sorry... It's already primary architecture at AOSC OS now.

I am not aware of any other distro affected by this, but GRUB recognizes vmlinux (and also kernel), see https://git.savannah.gnu.org/cgit/grub.git/tree/util/grub.d/10_linux.in#n174 .

PR #2630, opened 2024-02-28
fix(zfcp_rules): correct shellcheck regression when parsing ccw args

aafeijoo-suse:
## Checklist

Fixes 032ecd95c94b77f3f08237e0f765b355dacb9573


Vogtinator (approved):
Might be worth checking other places whether they got similarly broken.

aafeijoo-suse:
> Might be worth checking other places whether they got similarly broken.

I agree

aafeijoo-suse:
BTW this file will be removed by #2534, so I don't know if this fix makes sense upstream at this point.

tblume (approved):
LGTM

johannbg (approved):
LGTM

PR #2634, opened 2024-03-06
Fixed some issues found by static analyzers

lnykryn:


src/util/util.c
```diff
@@ -180,12 +180,12 @@ static int getarg(int argc, char **argv)
         char *end_value = NULL;
         bool bool_value = false;
         char *cmdline = NULL;
+        char *cmdline_iter = NULL;
```
dtardon, line 183: Merge this with the previous line?

src/util/util.c
```diff
@@ -204,9 +204,16 @@ static int getarg(int argc, char **argv)
         if (strlen(search_key) == 0)
                 usage(GETARG, EXIT_FAILURE, "search key undefined");
 
+        cmdline = strdup(p);
```
dtardon, line 207: Hmm, I see no reason to copy `p` in the first place... The function could just use it directly, i.e., initialize `cmdline` (the iterator) simply as `cmdline = p`. Or have I overlooked something?

dtardon: I have... `next_arg()` modifies the string, so it _must_ be copied. So scratch this.

src/util/util.c
dtardon, line 183: Merge this into the preceding line?

src/install/dracut-install.c
```diff
@@ -1140,6 +1140,7 @@ static int parse_argv(int argc, char *argv[])
                         kerneldir = optarg;
                         break;
                 case ARG_FIRMWAREDIRS:
+                        strv_free(firmwaredirs);
                         firmwaredirs = strv_split(optarg, ":");
```
dtardon, line 1144: Not related to the fix, but maybe add a check that `strv_split()` succeeded?

dtardon (changes requested):

PR #2663, opened 2024-07-09
WIP: Add support for erofs + overlayfs live rootfs

bcl:
This PR adds support for erofs compressed live rootfs images, used in conjunction with https://github.com/weldr/lorax/pull/1403 to enable creating a boot.iso using erofs instead of squashfs for the install.img

NOTE: This is WIP and has only been tested by hand editing /sbin/dmsquash-live-root with an erofs boot.iso

## Changes
Recognizes erofs as a supported filesystem in dmsquash-live-root

## Checklist

Fixes #


bcl:
Looks like this is the old project location.

PR #2665, opened 2024-07-19
support kernels without initrd

teknoraver:
If the installed kernel has no initrd support, skip the image creation.

This pull request changes...

## Checklist

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1858516


LaszloGombos:
@teknoraver Fedora is now using https://github.com/dracut-ng/dracut-ng . Perhaps you can upload this PR there. Should you also do the same for rescue initrd ?

teknoraver:
Will do!

teknoraver:
Done in https://github.com/dracut-ng/dracut-ng/pull/520

This one needs to be discarded then?

PR #2672, opened 2024-08-27
fix(lvm): clean up whitespace in messages

teigland:
The lvm module has been printing messages with whitespace damage:

  Scanning devices sda sdb  for LVM logical volumes vg/root
  vg/swap

Remove the trailing space after the final device name, and remove the newlines between LV names.  The message above becomes:

  Scanning devices sda sdb for LVM logical volumes vg/root vg/swap

## Checklist
- [ x] I have tested it locally

### DIFF
--- a/modules.d/90lvm/lvm_scan.sh
+++ b/modules.d/90lvm/lvm_scan.sh
@@ -134,7 +134,8 @@ check_lvm_ver 2 3 14 "$maj" "$min" "$sub" \
     && scan_args="$scan_args --nohints"
 
 if [ -n "$LVS" ]; then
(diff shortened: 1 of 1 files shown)


jozzsi:
@teigland  Fedora is using https://github.com/dracut-ng/dracut-ng . Would you consider uploading it there. Thanks !

aafeijoo-suse:
> @teigland Fedora is using https://github.com/dracut-ng/dracut-ng . Would you consider uploading it there. Thanks !

This has to be a bot, there is no other explanation.

aafeijoo-suse:
> This has to be a bot, there is no other explanation.

Oh, now we know what it is, and it's worse than a bot... though I'm not surprised (facepalm)

PR #2681, opened 2025-01-23
Extend description of `inst` command in dracut.modules man page

jougs:
This pull request extends the description of the `inst` command to mention that the filename has to be present in the `<dst>` given and that directories in `<dst>` will be created if they don't exist yet.

Background: When omitting the filename from `<dst>`, only the first call to `inst` seems to succeed, while subsequent calls are skipped without an error. Feel free to accept this PR in case this is the intended behavior. If the current behavior is considered a bug, it might be better to fix it and adapt the manpage accordinly.

## Checklist

### DIFF
--- a/man/dracut.modules.7.asc
+++ b/man/dracut.modules.7.asc
@@ -261,9 +261,11 @@ not lead to an error.
 ==== inst <src> [<dst>]
 
 installs _one_ file <src> either to the same place in the initramfs or to an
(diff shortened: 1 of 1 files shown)


jougs:
Sorry, I seem to have missed the comment by stalebot above. Anyway: I think this deserves to be re-opened. Or shall I create a new PR from the same branch instead? Thanks!

PR #2696, opened 2025-04-23
Fix kernel module name normalization in drivers lists

sndirsch:
## Changes
In drivers/add_drivers/force_drivers/omit_drivers kernel modules lists replace all occurrences of "-" with "_", not only the first one!

## Checklist
 [x ] I have tested it locally
 [x ] I have reviewed and updated any documentation if relevant

## Fixes 
The issue that kernel modules in drivers/add_drivers/force_drivers/omit_drivers modules lists with "-" in their name are ignored except the first module with only one "-" in its name.


sndirsch:
Wrong project. Submitted to dracut-ng instead.
https://github.com/dracut-ng/dracut-ng/pull/1287

PR #2703, opened 2025-07-04
fix(dracut.sh): Ensure hardlink deduplication is reproducible

cjriches:
By default, `hardlink` will only deduplicate files with identical mtimes, down to one-second granularity. If a dracut module rapidly generates multiple identical files, it is completely up to chance as to whether their mtimes cross a second boundary or not, and thus whether they get deduplicated or not. This results in non-reproducible output.

Perform timestamp clamping before invoking `hardlink` if a reproducible build is requested.

## Checklist


cjriches:
There's a few open questions / alternative approaches here:

* Doing this mtime clamping will also result in more files getting more consistently deduplicated, so it potentially has benefits for everyone - perhaps we should do it unconditionally regardless of `--reproducible`?
* We could also avoid touching mtimes and just run `hardlink -t` to deduplicate regardless of mtime match, but there's a small chance this breaks someone who was relying on identical files having different mtimes. The `-t` option also isn't present in all distros (e.g. it's in RHEL9 but not RHEL8).

cjriches:
Linked RHEL bug ticket: https://issues.redhat.com/browse/RHEL-101901

aafeijoo-suse (approved):
Thanks for fixing this hidden issue.

This repo is dead, could you please resubmit this to https://github.com/dracut-ng/dracut-ng?

cjriches:
> Thanks for fixing this hidden issue.
> 
> This repo is dead, could you please resubmit this to https://github.com/dracut-ng/dracut-ng?

Thanks, have done: https://github.com/dracut-ng/dracut-ng/pull/1429

If you have the power to do so, I would suggest updating the README of this repo and marking it as archived, so that future contributors don't make the same mistake.

PR #2705, opened 2025-07-18
When --entry-type=type2 is used (for UKI), will not remove normal kernel IMAGE.

xxiong2021:
When --entry-type=type2 is used (for UKI), will not remove normal kernel IMAGE.
according to the changes of https://github.com/systemd/systemd/pull/37897

This pull request changes...

## Changes
add check if the entry-type=type2, not remove the normal kernel IMAGE

## Checklist

Fixes #
https://issues.redhat.com/browse/RHEL-83932


jozzsi:

PR #2713, opened 2025-10-10
Fix compiler check in configure

snickl:
Quoting breaks the check in situations like
CC="/home/user/buildroot/output/host/bin/ccache /usr/bin/gcc"


jozzsi:

snickl:
Thanks for letting me know. This repo is still referenced by buildroot head.
Would be good if the admins marked this repo as archived then.

PR #2722, opened 2026-01-04
fix(kernel-modules-extra): remove redundant escape for '/' in re_escape

panzhe-0322:
The re_escape function in 90kernel-modules-extra/module-setup.sh incorrectly added a backslash escape for the '/' character , which is not a regular expression special character. This redundant escape triggers a 'stray \ before /' warning in grep when processing paths containing '/'.

This pull request changes...

## Checklist

Fixes #


jozzsi:

PR #2740, opened 2026-08-03
fix(dracut-systemd): pass through dissect and gpt-auto-force root= va…

supakeen:
…lues

systemd's gpt-auto-generator supports root=dissect, root=dissect-force, and root=gpt-auto-force in addition to root=gpt-auto, but dracut's cmdline parser did not recognize these values. This caused a fatal error during boot: "Don't know how to handle 'root=dissect'".

Add all three values to the passthrough case in dracut-cmdline.sh so they are accepted and left for systemd to handle, just like gpt-auto and tmpfs already are.

mount.usr=dissect is not affected since dracut does not parse mount.usr= at all. The 98usrmount module skips itself when systemd is present, and the parameter is consumed entirely by systemd's gpt-auto-generator.


LaszloGombos:
Development moved to https://github.com/dracut-ng/dracut

